PR #1, opened 2020-04-16
chore: create .gitlab-ci.yml pipeline

fampinheiro:


jeysal:
Force pushed a config I'm happy with that does all the checks for now :+1: 
You can take over again @fampinheiro :sweat_smile:

.gitlab-ci.yml
```diff
@@ -0,0 +1,46 @@
+image: node:12
+
+stages:
+  - setup
+  - build
+  - test
+
+setup:packages:
```
jeysal, line 8: this added on purpose? we already have the `check:install` step

package.json
```diff
@@ -72,5 +73,9 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
```
jeysal, line 77: If we start distinguishing between `deps` and `devDeps` here in the workspace root, all the existing `deps` should also be `devDeps` instead (as the workspace root ofc will never have actual 'production dependencies')

fampinheiro: It was a consequence of using `--dev`, I can change the others.

packages/example-lib/src/index.ts
```diff
@@ -1 +1 @@
-export const LINK_TEXT = 'React is cool.';
+export const LINK_TEXT = 'React is cool :)';
```
jeysal, line 1: This is the key change :D

PR #3, opened 2020-04-20
replace React brand with ASAP brand

jeysal:


apps/frontend/package.json
```diff
@@ -6,7 +6,7 @@
     "node": "12.x"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "TSC_COMPILE_ON_ERROR=true react-scripts start",
```
fampinheiro, line 9: The TSC_COMPILE_ON_ERROR is mandatory everytime? If that is the case can we use something like https://www.npmjs.com/package/cross-env?

apps/frontend/package.json
```diff
@@ -6,13 +6,18 @@
...
     "test": "echo Run the workspace root script instead",
     "eject": "react-scripts eject"
   },
   "dependencies": {
     "@asap-hub/example-lib": "workspace:packages/example-lib",
+    "react": "^16.13.1",
+    "react-app-polyfill": "^1.0.6",
+    "react-dom": "^16.13.1"
+  },
+  "devDependencies": {
```
jeysal, line 20: Split up into deps and devDeps, just for you @fampinheiro :heart: ;D

apps/frontend/package.json
jeysal: fixed, thanks! :+1:

jeysal:
Improved the lint config. PR now does a bit more than it says because I also ended up updating Jest and stuff for no reason, but :shrug: :sweat_smile:

PR #4, opened 2020-04-20
chore: build master and PRs on CI

fampinheiro:


.gitlab-ci.yml
```diff
@@ -58,37 +100,28 @@ build:native-deps:
...
 
 deploy:sls:branch_remove:
+  <<: *tmpl_branch_remove_rules
+  dependencies:
+    - build:native-deps
   environment:
     name: review/$CI_COMMIT_REF_SLUG
     action: stop
   script:
+    - git checkout $CI_COMMIT_SHA # the branch may not exist
```
jeysal, line 126: Isn't it gonna fail even before this in that case? Because checking out the branch is like a pre script step right?

.gitlab-ci.yml
```diff
@@ -1,5 +1,47 @@
...
+    SLS_STAGE: $CI_COMMIT_REF_SLUG
+
+.tmpl:deploy: &tmpl_deploy
+  artifacts:
+    reports:
+      dotenv: deploy.env
+  script:
+    - yarn pnpify serverless deploy --verbose
+    - echo "APP_URL=$(yarn pnpify serverless info --verbose | grep -w "CloudFrontDistributionDomain" | cut -d ' ' -f2)" >> deploy.env
+  dependencies:
```
jeysal, line 34: We need to distinguish between master and branch here. master also needs to depend on all checks.

.gitlab-ci.yml
fampinheiro: This job should execute after a pull request is merged and branch removed. 
The remove job doesn't rely on the branch name due to that, having a detached state with the git SHA is a better approach to avoid failure caused by removing the branch.

fampinheiro: "Dependencies" refer to the artifacts copied to the working directory, "needs" allow for "parallelization". For deploy, we "need" every previous stage to finish.

.gitlab-ci.yml
fampinheiro: "Dependencies" refer to the artifacts copied to the working directory, "needs" allow for "parallelization". For deploy, we "need" every previous stage to finish.

.gitlab-ci.yml
jeysal: I understand what you're trying to do, I'm just not convinced it works - if it usually tries to check out the branch before running the scripts, doesn't it already fail then before even getting to this script?

.gitlab-ci.yml
jeysal: IMO it should be `needs` for branches and `dependencies` for master?
Why should we prevent failing lints, tests etc. from deploying a feature env - e.g. if I want to show someone my changes despite not having fixed all tests yet.

fampinheiro: We can add needs to branches to avoid waiting for checks to speed up the process but the pipeline will still fail. Is this what you have in mind?

jeysal: Yes, the pipeline should def fail if checks or tests or anything are red - but the deploy can go ahead regardless (on branches only, not master obviously)

.gitlab-ci.yml
jeysal: I guess we can also just see if it works with this branch after merge? :sweat_smile:

.gitlab-ci.yml
fampinheiro: https://github.com/yldio/asap-hub/pull/4/files#diff-96edf7a6f008de9e928d04e1ae5e12a5R117

.gitlab-ci.yml
fampinheiro: We can iterate on this later then 🙂

PR #10, opened 2020-05-07
Auth0 integration

jeysal:
Login / Register / Logout
No
* fancy design
* proper ASAP production Auth0 account


jest-base.config.js
```diff
@@ -3,7 +3,16 @@ module.exports = {
   testEnvironment: 'node',
 
   modulePathIgnorePatterns: ['<rootDir>/build/', '<rootDir>/coverage/'],
+
   collectCoverageFrom: ['<rootDir>/src/**/*.{js,jsx,ts,tsx}'],
+  coverageThreshold: {
```
jeysal, line 8: This is obviously a bit ... ambitious, I haven't seen many projects do this.
But I figured, it's where we're at now, and while building things like a component library, it may actually be relatively easy to keep up and makes you aware of potentially missing tests.
When it becomes too much, we can lower the thresholds just as much as we need to, or even better just throw in a couple of `istanbul-ignore` comments.

jeysal:
This is now ready for review @fampinheiro 
Obviously this is *a lot*, so feel free to review however large parts you deem necessary, I'm fairly confident in this after some iteration. A lot of it is copied from the Auth0 docs.

PR #13, opened 2020-05-14
Document Auth0 / ORCID configuration

jeysal:
Some things in here can be split out into preceding PRs


docs/config/auth0.md
```diff
@@ -0,0 +1,74 @@
...
+**Settings**
+
+- `Company Logo` - URL to the ASAP logo
+- `Primary Color` - pick any matching branding
+- `Background Color` - pick any matching branding
+
+**Login**
+
+- `Customize Login Page` - `true`
+- `HTML` - [here](auth0/login.html)
```
fampinheiro, line 36: Is this the HTML you mentioned we could generate with the same components used on the website?

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build:babel": "yarn workspaces foreach -pt run build:babel",
-    "build:typecheck": "tsc -b -v packages/example-lib",
+    "build:typecheck": "tsc -b -v packages/example-lib apps/orcid-user-profile-script",
```
fampinheiro, line 13: Can we leverage `yarn workspace foreach` to have each package we need to typecheck?
Not an action for now I'm just wondering how to chain this operation.

docs/config/auth0.md
jeysal: Yes, this is basically the Auth0 default HTML with some slight changes made for ORCID

package.json
jeysal: Yeah so fortunately we only need to specify the final things we want to build in here, which would be `apps/*`, and TSC will take care of building all the dependencies from `packages/*` that they require. However, unfortunately, the 'special' CRA frontend cannot be typechecked in this same simple way. We probably just want to specify 'all packages and apps except the frontend' in here, but I guess that requires writing a small JS script.

jeysal: I will note this down and think of a solution

PR #14, opened 2020-05-15
chore: lambda setup

fampinheiro:
This PR adds two lambdas:
 - one that returns an object `{ hello: "world" }`
- one that throws an error

The main purpose of it is to demonstrate how to integrate a lambda function on the current repository.

**Not implemented**
- Sourcemaps on the deployed function
- Simplified boilerplate to add a new function


serverless.js
```diff
@@ -1,23 +1,13 @@
-const { mkdirSync, existsSync, unlinkSync, symlinkSync } = require('fs');
-const { join, dirname } = require('path');
 const { paramCase } = require('param-case');
 const pkg = require('./package.json');
 
-const service = paramCase(pkg.name);
-const plugins = ['serverless-s3-sync'];
+const { NODE_ENV = 'production' } = process.env;
```
jeysal, line 4: I would argue it's safer to have nothing default to production and only CI master builds set things to prod

package.json
```diff
@@ -14,11 +14,13 @@
     "build:frontend": "yarn workspace @asap-hub/frontend build",
     "build:frontend:typecheck": "yarn workspace @asap-hub/frontend typecheck",
     "build": "yarn build:typecheck && yarn build:babel && yarn build:frontend:typecheck && yarn build:frontend",
+    "watch:babel": "yarn workspaces foreach -pi -j 1000 run watch:babel",
```
jeysal, line 17: Reordered some things so that similar things are grouped together

package.json
```diff
@@ -14,11 +14,13 @@
...
+    "watch:typecheck": "yarn build:typecheck -w",
+    "dev": "cross-env NODE_ENV=development yarn sls offline",
+    "sls": "node serverless/link-plugins.js && yarn pnpify serverless",
```
jeysal, line 20: This is awesome I think, now CI and `dev` care neither about `serverless` being `pnpify`ed nor about the initialization step, they just let `yarn sls` handle it :ok_hand:

apps/hello-world/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "watch:babel": "yarn build:babel -w --verbose"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "8.10.51",
+    "@types/jest": "^25.2.2",
+    "@types/source-map-support": "0.5.1"
+  },
+  "dependencies": {
+    "aws-lambda": "^1.0.6",
+    "aws-sdk": "^2.678.0",
```
jeysal, line 21: What is this for?

package.json
```diff
@@ -14,11 +14,13 @@
     "build:frontend": "yarn workspace @asap-hub/frontend build",
     "build:frontend:typecheck": "yarn workspace @asap-hub/frontend typecheck",
     "build": "yarn build:typecheck && yarn build:babel && yarn build:frontend:typecheck && yarn build:frontend",
+    "watch:babel": "yarn workspaces foreach -pi -j 1000 run watch:babel",
+    "watch:typecheck": "yarn build:typecheck -w",
+    "dev": "yarn sls offline",
```
jeysal, line 19:
```suggestion
    "start:backend": "yarn sls offline",
```
my suggestion - `start` would be consistent with CRA, and backend clarifies which part it starts

package.json
```diff
@@ -70,8 +81,10 @@
     "@typescript-eslint/eslint-plugin": "^2.28.0",
     "@typescript-eslint/parser": "^2.28.0",
     "@yarnpkg/pnpify": "^2.0.0-rc.20",
+    "@zeit/ncc": "^0.22.1",
     "babel-eslint": "^10.1.0",
     "babel-preset-react-app": "^9.1.2",
+    "cross-env": "^7.0.2",
```
jeysal, line 87: unused I think, remove

serverless.js
```diff
@@ -30,6 +22,11 @@ module.exports = {
     region: `\${env:AWS_REGION, "us-east-1"}`,
     stage: `\${env:SLS_STAGE, "development"}`,
   },
+  package: {
+    individually: true,
+    // we don't need this because of ncc
```
jeysal, line 27: I see, ncc only includes what is needed anyway. But do we need to explicitly exclude things here then?

apps/hello-world/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "@asap-hub/hello-world",
+  "private": true,
+  "version": "0.1.0",
+  "engines": {
+    "node": "12.x"
+  },
+  "main": "build/handler.js",
+  "types": "build/handler.d.ts",
+  "scripts": {
+    "build:babel": "babel src --source-maps true -Dd build -x .js,.jsx,.ts,.tsx --root-mode upward",
```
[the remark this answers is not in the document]
fampinheiro: I ended up removing source-maps. Let's test them and use it in a separate PR.

apps/hello-world/package.json
fampinheiro: To remove.

serverless.js
fampinheiro: This configuration is for serverless. Since `ncc` is packaging the source code serverless doesn't need to remove `devDependencies`.

PR #15, opened 2020-05-18
build(ts): generate project list automatically

jeysal:


package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build:babel": "yarn workspaces foreach -pt run build:babel",
-    "build:typecheck": "tsc -b -v packages/example-lib apps/orcid-user-profile-script",
+    "build:typecheck": "tsc -b -v $(node scripts/get-composite-ts-projects.js)",
```
jeysal, line 13: I don't have a Windows machine and we don't have a Windows CI to check, but apparently this should work cross platform with Yarn 2 https://yarnpkg.com/api/modules/yarnpkg_shell.html

PR #17, opened 2020-05-19
ci: build with NODE_ENV=production

jeysal:
CRA will already set this when running `react-scripts build`, but our packages have just one build script no matter if the final apps are being watch-run during development or built for production, so we should make sure everything is built for production with any optimizations applied on CI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,30 +3,29 @@
 .tmpl:master_deploy_rules: &tmpl_master_deploy_rules
   rules:
     - if: $CI_COMMIT_BRANCH == 'master'
-  variables:
-    SLS_STAGE: production
+.tmpl:master_deploy_variables: &tmpl_master_deploy_variables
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -25,8 +25,9 @@
   artifacts:
     reports:
       dotenv: deploy.env
+  variables:
+    NODE_ENV: production
```
fampinheiro, line 29: `gitlab-ci` doesn't merge variables in different templates. 

```
deploy:sls:branch:
  <<: *tmpl_branch_deploy_rules
  <<: *tmpl_deploy
```

The result of this template usage is 
```
deploy:sls:branch:
  variables:
    SLS_STAGE: $CI_COMMIT_REF_SLUG
  variables:
    NODE_ENV: production
```

To use variables from multiple templates https://docs.gitlab.com/ee/ci/yaml/#yaml-anchors-for-variables needs to be used.

PR #22, opened 2020-05-20
React component library

jeysal:
and usage in the frontend


jeysal:
Oops, just noticed the build on this didn't pass :raised_eyebrow: I'll look into it

jeysal:
https://github.com/yldio/asap-hub/pull/23 will help make the build deterministic again, then it should pass on CI as well as any local environment.

jeysal:
rebased, should work now :pray:

packages/react-context/tsconfig.json
```diff
@@ -4,5 +4,6 @@
     "rootDir": "src",
     "outDir": "build"
   },
-  "include": ["src"]
+  "include": ["src"],
+  "references": [{ "path": "../auth" }]
```
fampinheiro, line 8: Can we infer this from dependencies?

jeysal: #23 actually kind of implemented that, but only checks that they are there. TypeScript inferring them automatically is unfortunately still not possible https://github.com/microsoft/TypeScript/issues/25376

PR #28, opened 2020-05-22
feat: Storybook

jeysal:


apps/storybook/.eslintrc.js
```diff
@@ -0,0 +1,19 @@
+const { resolve } = require('path');
+
+module.exports = {
```
jeysal, line 3: Made a note for a follow-up to do some refactoring to extract the ESLint config shared between React packages/apps into the root somewhere.

PR #30, opened 2020-05-25
feat: users service

fampinheiro:


apps/users-service/src/controllers/users.ts
```diff
@@ -0,0 +1,84 @@
+import Boom from '@hapi/boom';
+import aws from 'aws-sdk';
+import { generate } from 'shortid';
```
jeysal, line 3: > ShortId does not generate cryptographically secure ids, so don't rely on it to make IDs which are impossible to guess.

Given that these could be used to impersonate a user, we should probably use e.g. UUID v4 from the `uuid` package.

apps/users-service/types/apr-intercept.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'apr-intercept';
```
jeysal, line 1: We should create proper definitions for these so that we don't have `any`s floating around in our code. No need to worry about that now though, I can see if I can come up with good definitions for these packages once it's merged into master.

PR #31, opened 2020-05-26
chore: use terraform and aws cert to create certificates

fampinheiro:


.gitlab-ci.yml
```diff
@@ -147,11 +174,12 @@ deploy:sls:master:
     <<: *tmpl_master_deploy_variables
   environment:
     name: production
-    url: https://$APP_URL
-  needs:
+    url: https://$BASE_URL
+  dependencies:
```
jeysal, line 178: Can we leave this as `needs`? Otherwise it'll wait for `test` before even starting the deploy, and the deploy is already quite slow :|

serverless.js
```diff
@@ -1,7 +1,21 @@
...
-const { NODE_ENV = 'development' } = process.env;
+const {
+  AWS_REGION = 'us-east-1',
+  SLS_STAGE = 'development',
+  BASE_URL,
+  AWS_ACM_CERTIFICATE_ARN,
+  NODE_ENV = 'development',
+} = process.env;
+
+assert.ok(BASE_URL, 'process.env.BASE_URL not defined');
```
jeysal, line 14: I think these checks should be in a conditional `NODE_ENV==='production'`? Otherwise `sls package` etc locally will fail

serverless.js
```diff
@@ -96,6 +77,48 @@ module.exports = {
   },
   resources: {
     Resources: {
+      HttpApiDomain: {
+        Type: 'AWS::ApiGatewayV2::DomainName',
+        Properties: {
+          DomainName: `api.\${self:custom.origin}`,
```
jeysal, line 83: This means it's no longer on the same domain as the frontend? That seems non-ideal as it will require us to take care of the corresponding CORS settings, right?

serverless.js
fampinheiro: We have two options
* Validate when production
* Create the custom code only when the variables are present

serverless.js
fampinheiro: Life is made of decisions 🙂. We can choose to have a record set pointing to CloudFront or the HttpApiGateway.

.gitlab-ci.yml
jeysal: Updated in the other job as well

serverless.js
jeysal: How does it work for the branch deploys? `api.my-branch.asap.yld.io`?

jeysal: Yeah, apparently

jeysal: We can ship this as is for now I guess?
Two questions:
This means things are no longer deployed to a cloudfront domain, ONLY the `*.asap.yld.io`, correct?
And once the time comes to connect FE and API, would you rather configure CORS rules or try to get them hosted on one origin again?

serverless.js
jeysal: The first is much easier and good enough I guess

serverless.js
fampinheiro: CORS rules is my bet.

PR #33, opened 2020-05-26
build(lint): extract shareable ESLint configs

jeysal:
This already works when running something like `y eslint --max-warnings=0 --report-unused-disable-directives --resolve-plugins-relative-to=./packages/eslint-config-asap-hub '{packages,apps}/*/src/**/*.{js,ts,jsx,tsx}'`. But the `resolve-plugins-relative-to` option is needed because of ESLint's weird config mechanisms, and to set that when running through `jest-runner-eslint`, we'll need https://github.com/jest-community/jest-runner-eslint/issues/94.


jeysal:
There is now https://github.com/jest-community/jest-runner-eslint/pull/95 which will allow us to ship this when merged and released.

jeysal:
updated `jest-runner-eslint`, should work now

PR #34, opened 2020-05-26
ci(storybook): serve index.html at /.storybook/

jeysal:


jeysal:
I am using S3's `IndexDocument` to serve `index.html` on every subdirectory. According to [SO](https://stackoverflow.com/a/15528757/7618180), CloudFront should respect that, if the Origin Domain Name is set. And we do set the `DomainName` for the origin in the CloudFront distribution.
But https://d3lxhi2lc27sxb.cloudfront.net/.storybook/ returns the main frontend still.
Don't want to spend a lot more time on this minor thing that maybe is easier to solve once we have a proper domain name anyway, but @fampinheiro do you by any chance have a clue what is going wrong here?

jeysal:
It does seem like the S3 bucket is the problem though - https://asap-hub-storybook-index-storybook.s3.amazonaws.com/.storybook/ does not work either

jeysal:
Oh, there's a separate domain for 'S3 Website Hosting': http://asap-hub-storybook-index-storybook.s3-website-us-east-1.amazonaws.com/.storybook/ works

jeysal:
If we manage to link the CloudFront distribution to that, it should work.

jeysal:
I made it work with a manual edit on AWS CloudFront, let's see if my CloudFormation code also works

jeysal:
This is now doing exactly [this](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serve-static-website/) from the AWS docs

jeysal:
Works :tada: The splitting the URL string by `http://`, and the `http-only` policy custom origin seem sketchy, but it's exactly the setup you get if you manually follow the official guide :shrug:

serverless.js
```diff
@@ -221,21 +224,23 @@ module.exports = {
                 },
               },
               {
-                DomainName: {
-                  'Fn::GetAtt': ['StorybookBucket', 'RegionalDomainName'],
+                CustomOriginConfig: {
+                  OriginProtocolPolicy: 'http-only',
```
fampinheiro, line 228: Is the `http-only` intended?

serverless.js
```diff
@@ -221,21 +224,23 @@ module.exports = {
...
-                        'origin-access-identity/cloudfront',
-                        { Ref: 'CloudFrontOriginAccessIdentityStorybook' },
+                DomainName: {
+                  'Fn::Select': [
+                    '1',
+                    {
+                      'Fn::Split': [
+                        'http://',
+                        {
+                          'Fn::GetAtt': ['StorybookBucket', 'WebsiteURL'],
```
fampinheiro, line 237: We are accessing the bucket directly 😐
That explains having the http configuration.

jeysal: Yes, exactly. As stated in my comments this is the same setup as https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serve-static-website/

PR #36, opened 2020-05-28
feat(components): Button

jeysal:


jeysal:
Oops, should have probably left this as a draft until design review went through. Whatever, it's not like the components are used anywhere yet, can do amends in a follow up :smile:

fampinheiro:
We are in a state of the project we can merge and proceed :) Nothing a new pull request can't solve.

PR #39, opened 2020-05-29
feat(buttons): icons

jeysal:
Will follow up with a PR adapting the Login/Logout Button and creating the Google / ORCID SSO buttons. As you said though, might as well merge these things already and let James review designs on master Storybook at this stage.

This interface (combined icons and text in `children` in any order) is a bit magic compared to `<Button icon={<svg/>}>Text</Button>` and I was unsure with what route to go with, but it's well encapsulated and tested so I think this more dynamic and flexible one is fine, even if we may not need it.


jeysal:
https://buttons-icons.asap.yld.io/.storybook/?path=/story/atoms-button--icon-and-text

PR #40, opened 2020-05-29
SSO buttons

jeysal:
do not squash


jeysal:
https://sso-buttons.asap.yld.io/.storybook/?path=/story/organisms-auth-google-signin-button--normal

PR #42, opened 2020-05-29
build: update Jest, Prettier, related deps

jeysal:


jeysal:
passed, and https://dep-updates.asap.yld.io/ plus storybook looking good, merging

PR #44, opened 2020-06-01
feat(buttons): pointer cursor when hovering

fampinheiro:


jeysal:
Afaik this is pretty controversial, some argue that all clickable elements should have a pointer cursor, some that only link-type elements should. See e.g. https://ux.stackexchange.com/questions/105024/why-dont-button-html-elements-have-a-css-cursor-pointer-by-default
I would personally prefer the latter (that's why I implemented it this way) given that our buttons already have another visible hover effect, but I think James should be the one to make that call?

fampinheiro:
I'm a simple person. I see a link I expect a small glove instead of an arrow. 
But I agree let's bring James to the discussion.

PR #46, opened 2020-06-03
chore: use wildcard certificate

fampinheiro:


jeysal:
Well, I better not name my branches `api-*` then :D
Is this ready to ship as a workaround for now @fampinheiro?

PR #47, opened 2020-06-04
feat: connect user

fampinheiro:


apps/users-service/src/config.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const {
+  APP_BASE_URL,
+  AUTH0_BASE_URL,
+  MONGODB_CONNECTION_STRING,
+  NODE_ENV,
+} = process.env;
+
+export const auth0BaseUrl = AUTH0_BASE_URL || 'https://jeysal.eu.auth0.com';
+export const globalToken = 'change_me_when_we_have_admins';
```
jeysal, line 11: We should also read this from the env and define something random on CI that we can share with each other privately, so that not everyone who looks at the source code can use the form to spam people with emais. I would do something like
```js
if (NODE_ENV==='production' && !GLOBAL_TOKEN) {
  throw new Error('GLOBAL_TOKEN variable must be set in production')
}
export const globalToken = GLOBAL_TOKEN || 'local_dev_token';
```
to also ensure that a production build cannot fall back to the guessable default token

PR #48, opened 2020-06-05
feat(admin): user invitation form

jeysal:


apps/frontend/src/config.ts
```diff
@@ -0,0 +1,15 @@
+let baseUrl: string;
+switch (window.location.host.split('.').length) {
```
jeysal, line 2: This is ... less than ideal, ideas @fampinheiro?
I also don't really wanna pass the base URL via the environment when building though, that'd mean the same commit builds a different frontend bundle on `master` vs a branch, which is also weird

fampinheiro: Apart from an environment variable I'm not seeing many options.

serverless.js
```diff
@@ -82,6 +82,9 @@ module.exports = {
     webpack: {
       config: 'serverless/webpack.config.js',
     },
+    'serverless-offline': {
+      useChildProcesses: true, // needed for hot reloading to work https://github.com/dherault/serverless-offline/issues/931
```
fampinheiro, line 86: Nice one.

jeysal:
Two issues left before this is good to go:
* Two test TODOs left, will take care of them on Monday @jeysal 
* API requests fail due to missing CORS headers @fampinheiro

jeysal:
removing any CORS experiments from this PR, can land those separately to make it *actually* work

PR #49, opened 2020-06-08
chore: add cors headers

fampinheiro:


apps/users-service/src/framework/lambda.ts
```diff
@@ -28,6 +28,17 @@ export interface RequestContext {
   connection: MongoClient;
 }
 
+export const response = (res: APIGatewayProxyResult): APIGatewayProxyResult => {
+  return {
+    ...res,
+    headers: {
+      'Access-Control-Allow-Origin': '*',
```
jeysal, line 35: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) the wildcard does not work with credentials (and it'd also be insecure), so I think we'll need to pass in the `asdfghjkl.asap.yld.io` domain via the environment.

jeysal:
This LGTM, but for the invitation form to work we'll also need the headers to be present on the `OPTIONS /users` request that is not handled by our backend but by the API gateway directly I think

PR #50, opened 2020-06-08
chore: cors in the provider

fampinheiro:


jeysal:
Let's see - but I'm worried that even if it sets the headers, it'll be `allowcredentials false` and thus not help :|

serverless.js
```diff
@@ -56,6 +41,11 @@ module.exports = {
     memorySize: 512,
     region: AWS_REGION,
     stage: SLS_STAGE,
+    httpApi: {
```
fampinheiro, line 44: When creating `AWS::ApiGatewayV2::Api` cors configuration is at the provider level.

PR #51, opened 2020-06-08
refactor(ts): precise and global typings

jeysal:
* Proper definitions instead of just `declare module` so that we have fewer `any`s floating around in our code
* Moved the type definitions out of one specific package into workspace-local `@types/*` packages that can be used by multiple packages


apps/users-service/src/framework/lambda.ts
```diff
@@ -90,7 +90,9 @@ export const http = <T>(
     query: event.queryStringParameters,
   } as Request;
 
-  const [err, res] = await Intercept(fn(request));
+  const [err, res] = await Intercept(
+    (async (): Promise<Response> => fn(request))(),
```
fampinheiro, line 94: Any particular reason for wrapping promises in a new `async` function?

jeysal: The definition of `fn` above returns `Response | Promise<Response>`. This is to prevent the `Response` case from breaking (`Intercept` can't deal with something that doesn't have a `.then`). If we don't actually need that we can change `fn` to return only `Promise<Response>` (probably better in fact)

jeysal: pushed

PR #54, opened 2020-06-09
fix(security): use cryptograhically secure random id generator

jeysal:


jeysal:
Tested by making an API call and receiving an email with a UUID in the link

PR #59, opened 2020-06-09
chore: return 409 on duplicate email

fampinheiro:


apps/users-service/test/handlers/users.create.test.ts
```diff
@@ -134,7 +134,7 @@ describe('POST /users', () => {
       null,
     )) as APIGatewayProxyResult;
 
-    expect(res.statusCode).toStrictEqual(403);
+    expect(res.statusCode).toStrictEqual(409);
```
jeysal, line 137: make sure to update the test name as well

PR #68, opened 2020-06-22
feat: custom Auth0 universal login page

jeysal:


jeysal:
This should be ready to merge as an improvement over the current login page (bearing in mind that signing up a new email-password requires `&screen_hint=signup` which will later be set by the referring frontend) without any email invite code connection (and with open design issues that I haven't looked at yet)

PR #70, opened 2020-06-25
feat: render template and upload to aws

fampinheiro:


jeysal:
I've glanced over it and it makes sense to me, good job :+1: 

Only thing I'd like to see changed is IMO `react-templates` should go into the `apps` folder, not the `packages` folder, because its build setup is very much unlike the normal packages, and it'd fit the distinction of packages and apps in README/RepoStructure better.

Optional suggestion if you're not tired of this yet and want to do something cool:
Because the build setup of this is so complex, we could extract that into a package `something something react email compiler` to be able to use proper `imports`, webpack plugin types, etc. Should *just work* :tm: if it is added to the app as a `devDependency`, the compiler package will be built first and then used to compile the templates app.
But could also leave a todo and do it later.

Also idk if you've thought about how to do integration testing for this?

fampinheiro:
There are three main improvements needed on this pull request.
* Create a mechanism to have feature branches to test templates
* Create integration tests, possibly mocking the request done to aws
* Export types to avoid trying to send messages with the wrong arguments (we can create notification system to detect this)

PR #77, opened 2020-06-30
feat: add fetch user handler and extends user schema

johnytiago:


apps/users-service/src/cms/users.ts
```diff
@@ -38,6 +50,16 @@ export default class Users extends Base<User> {
     return items[0] as User;
   }
 
+  async fetchById(id: string): Promise<User | null> {
+    const { items } = await this.client
+      .get('users', {
```
fampinheiro, line 55: We can access the user using the id, the id is unique so we don't need a search operation.
https://cloud.squidex.io/api/content/asap-hub/docs#operation/GetUsersContent

johnytiago: Fixed 👍

PR #81, opened 2020-07-02
chore: app using api from env vars

fampinheiro:
To enable independent frontend and backend development the api url can be set using an environment variable. [Here](https://github.com/yldio/asap-hub/pull/81/files#diff-1eb6c386a33702c4811031f80affa3c0).

To do this in the CI a [script](https://github.com/yldio/asap-hub/pull/81/files#diff-2550986331811c521c3d9c78191f209f) is responsible for setting the `APP_HOSTNAME` and `API_HOSTNAME` explicitly.


ci/env-create.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env sh
```
johnytiago, line 1: Are we sure this file is well-formed?

I was looking into the docs https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html#artifactsreportsdotenv.
> The .env file can’t have empty lines or comments (starting with #).

apps/react-templates/.gitlab-ci.yml
```diff
@@ -1,18 +1,15 @@
...
+    - setup
+    - build
   needs:
+    - setup
     - build
   variables:
-    <<: *tmpl_master_deploy_variables
-    ORIGIN: https://$BASE_URL
+    SLS_STAGE: production
+    APP_ORIGIN: https://$APP_ORIGIN
```
jeysal, line 15: This is very confusing, either it's with a protocol, then it's an `origin`, or it's without a protocol, then it's `host`. We shouldn't have `origin` sometimes include protocol and sometimes not. In this case, it seems like the reference should be `$APP_HOST`.

apps/frontend/src/config.ts
```diff
@@ -1,20 +1,4 @@
...
-  case 4: {
-    baseUrl = `${window.location.protocol}//api-${window.location.host}`;
-    break;
-  }
-  default: {
-    baseUrl = 'http://localhost:3333/development';
-  }
-}
-export const API_BASE_URL = baseUrl.toString();
+export const API_BASE_URL =
```
jeysal, line 3: Should it not be `'http://localhost:3333/development'`?

apps/frontend/README.md
```diff
@@ -12,6 +12,12 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 The page will reload if you make edits.<br />
 You will also see any lint errors in the console.
 
+#### Environment Variables
+
+The frontend application uses the production API by default, to change this behavior you can set the `REACT_APP_API_BASE_URL` environment variable.
```
jeysal, line 17: Does it? It seems like it defaults to `localhost`

apps/react-templates/.gitlab-ci.yml
fampinheiro: The script exports just hostnames and they are used as is or with a protocol when used as origins.

apps/frontend/README.md
fampinheiro: Last minute change that wasn't properly updated. Thanks

PR #91, opened 2020-07-07
fix(text-field): avoid showing valid indicator below custom indicator

jeysal:
aka
fix(password-field): do not show valid indicator below show/hide button


jeysal:
failure is just the deploy because of too long branch name :joy:

PR #92, opened 2020-07-09
feat: get /me endpoint

johnytiago:
Depends on #84

### DIFF
--- a/apps/users-service/src/handlers/fetch-me.ts
+++ b/apps/users-service/src/handlers/fetch-me.ts
@@ -1,51 +1,8 @@
-import Boom from '@hapi/boom';
-import got from 'got';
-import Intercept from 'apr-intercept';
-import { config as authConfig } from '@asap-hub/auth';
 import { framework as lambda } from '@asap-hub/services-common';
 import { APIGatewayProxyHandler } from 'aws-lambda';
 
+import validateUser from '../utils/validate-user';
 import Users from '../controllers/users';
-import * as auth0 from '../entities/auth0';
-
-const validateUser = async (
-  request: lambda.Request,
-): Promise<auth0.UserInfo> => {
-  const headers = request.headers as {
-    authorization: string;
-  };
-
-  if (!headers.authorization) {
-    throw Boom.unauthorized();
-  }
-
-  const [type, token] = headers.authorization.split(' ');
-
-  if (type.toLowerCase() !== 'bearer') {
-    throw Boom.unauthorized();
-  }
-
-  const [err, res] = await Intercept(
-    got(`https://${authConfig.domain}/userinfo`, {
-      headers: {
-        Authorization: `Bearer ${token}`,
-        'Content-Type': 'application/json',
-      },
-    }).json<auth0.UserInfo>(),
-  );
-
-  if (err) {
-    if (err.response.statusCode < 500) {
-      throw Boom.forbidden('Forbidden', {
-        error: err,
-      });
-    }
-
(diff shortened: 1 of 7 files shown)


apps/users-service/src/handlers/fetch-me.ts
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  if (!headers.authorization) {
+    throw Boom.unauthorized();
+  }
+
+  const [type, token] = headers.authorization.split(' ');
+
+  if (type.toLowerCase() !== 'bearer') {
+    throw Boom.forbidden();
```
jeysal, line 25: Afaik usually Unauthorized is for when you haven't provided a valid token, and Forbidden for when you have provided a valid token, but do not have permissions for the operation. So I think this should be unauthorized instead?

apps/users-service/src/handlers/fetch-me.ts
```diff
@@ -0,0 +1,57 @@
...
+    got(`https://${authConfig.domain}/userinfo`, {
+      headers: {
+        Authorization: `Bearer ${token}`,
+        'Content-Type': 'application/json',
+      },
+    }).json<auth0.UserInfo>(),
+  );
+
+  if (err) {
+    throw Boom.forbidden('Forbidden', {
```
jeysal, line 38: Can (and do we want to) forward the status code from auth0 userinfo? I think probably yes - if they give a 4xx it's probably something about the token the client provided, we want to pass that through, if it's 5xx it's probably some Auth0 outage or whatever and we don't want the client to think their authentication is just invalid.

apps/users-service/src/handlers/fetch-me.ts
johnytiago: Just revisited the HTTP codes spec. You're totally right! 👍

apps/users-service/src/handlers/fetch-me.ts
johnytiago: That's right, let me add a test to cover that scenario.

apps/users-service/test/handlers/users.fetch-me.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import { apiGatewayEvent } from '../helpers/events';
+import { APIGatewayProxyResult } from 'aws-lambda';
+import { createRandomUser } from '../helpers/create-user';
+import { config as authConfig } from '@asap-hub/auth';
+
+jest.mock('@asap-hub/auth');
+
+const chance = new Chance();
+
+describe('GET /users/invites/{code}', () => {
```
fampinheiro, line 13:
```suggestion
describe('GET /me', () => {
```

apps/users-service/test/handlers/users.fetch-me.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  test('return 401 when Authentication is not set', async () => {
```
fampinheiro, line 14:
```suggestion
  test('return 401 when authorization header is not set', async () => {
```

apps/users-service/src/handlers/fetch-me.ts
johnytiago: The fix is done. But I'll need someones help on fixing the typescript error.
image

jeysal: I can TAL (not sure when exactly because of upcoming flight)

johnytiago: I think I managed to work around it 😉 For now it's ok!

PR #95, opened 2020-07-15
feat: add create ROMP endpoint

johnytiago:


apps/users-service/src/cms/romps.ts
```diff
@@ -0,0 +1,43 @@
+import { Base, BaseOptions } from '@asap-hub/services-common';
+import { Romp } from '../entities/romps';
+
+export interface CreateRomp {
```
jeysal, line 4: We can throw this into `packages/model` already so it's accessible to FE+BE :upside_down_face:

apps/users-service/src/cms/index.ts
```diff
@@ -1,11 +1,15 @@
 import Users from './users';
+import Romps from './romps';
```
jeysal, line 2: The team's outputs overall are called the ROMP, so I think for the individual 'posts' we should find a different terminology. Maybe `ResearchOutput(s)` or just `Output(s)`? @fampinheiro

apps/users-service/test/handlers/users.create-romp.test.ts
```diff
@@ -0,0 +1,216 @@
...
+import { config as authConfig } from '@asap-hub/auth';
+
+jest.mock('@asap-hub/auth');
+
+const chance = new Chance();
+
+describe('GET /users/{id}/romp', () => {
+  let id, code;
+
+  const romp = {
```
jeysal, line 16: Should be typed as `const romp: CreateRomp = ...` to make sure properties are updated whenever we change the interface

apps/users-service/src/cms/index.ts
fampinheiro: I've been referring to them as `RompItems`, ReseachOutputs` is an option. Let me raise this to see the terminology product and design is using.

apps/users-service/test/handlers/users.create-research-output.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    code = user.connections[0].code;
+  });
+
+  test('return 401 when Authentication header is not set', async () => {
+    const result = (await handler(
+      apiGatewayEvent({
+        httpMethod: 'post',
+        body: output,
+        pathParameters: {
+          id,
```
fampinheiro, line 50: I'm wondering if we need the `id` in the header. We can extract it from the authenticated user.

PR #100, opened 2020-07-19
feat: record output form

jeysal:


jeysal:
can do multiline text fields in a follow up and merge this one

PR #102, opened 2020-07-22
feat: fetch team by id

johnytiago:


apps/teams-service/test/handlers/teams.fetch-by-id.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import { APIGatewayProxyResult } from 'aws-lambda';
+import { createRandomUser, createRandomOutput } from '../helpers/create-user';
+import { config as authConfig } from '@asap-hub/auth';
+
+jest.mock('@asap-hub/auth');
+
+const chance = new Chance();
+
+describe('GET /teams/{id}', () => {
+  const id = '53411d87-5cfe-40c9-946e-1eda0086a4cb';
```
johnytiago, line 14: Do you guys think I should store this in a better way? This will fail locally btw

jeysal: I thought we should have a `beforeAll` and `afterAll` that create and delete a team by speaking directly with the CMS

PR #103, opened 2020-07-22
refactor(frontend): authenticate requests

jeysal:


packages/model/src/user.ts
```diff
@@ -8,3 +8,8 @@ export interface Invitee {
   orcid?: string;
   institution?: string;
 }
+
+export interface UserResponse extends Invitee {
+  id: string;
+  connections?: [{ code: string }];
```
fampinheiro, line 14:
```suggestion

```

We don't want to expose the connection until we have a better understanding on how we'll map multiple accounts.

jeysal: But that's not how TypeScript works, we need to declare it or the backend e.g. in `create-user.ts` will error because it's trying to set the property.

fampinheiro: In that case `create-user.ts` is exposing a property it shouldn't.

apps/frontend/src/members/Members.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Profile from './Profile';
+import CheckAuth from '../auth/CheckAuth';
+
+const Members: React.FC<{}> = () => {
+  const { path } = useRouteMatch();
+
+  return (
+    <CheckAuth>
+      <Switch>
+        <Route exact path={path} component={ProfileList} />
```
fampinheiro, line 14: Interesting approach :+1:

PR #108, opened 2020-07-25
feat(frontend): create research outputs

jeysal:


jeysal:
@fampinheiro not up to usual quality standards (UI not showing request progress, and esp. no test coverage), but maybe you can get this over the line next week

apps/frontend/src/api/__tests__/util.test.tsx
```diff
@@ -20,7 +20,7 @@ it('throws if the user is not authenticated', async () => {
 });
 
 const wrapper: React.FC<{}> = ({ children }) => (
-  <authTestUtils.LoggedIn user={{}}>{children}</authTestUtils.LoggedIn>
+  <authTestUtils.LoggedIn user={undefined}>{children}</authTestUtils.LoggedIn>
```
fampinheiro, line 23: What is the advantage of setting the value to undefined over not providing the property?

jeysal: In the props type definitions for `LoggedIn` I've forbidden omitting it (`: User | undefined` instead of `?: User`) with a comment that it is a weird state when Auth0 has a token but the user info was not fetched yet. It should only be used when you know that the user info is not relevant, only the fact that the user is logged in, that's why you must not omit `user` by simply forgetting it.

jeysal:
rebased, looking at design QA comments

jeysal:
Design QA comments implemented.
Tested the whole flow on the environment again, appears to work :tada:

PR #120, opened 2020-07-29
feat: return one item when fetching by slug

fampinheiro:


apps/asap-server/src/cms/content.ts
```diff
@@ -20,6 +21,11 @@ export default class ContentCMS extends Base {
         },
       })
       .json();
-    return items;
+
+    if (items.length === 0) {
+      throw Boom.notFound();
```
johnytiago, line 26: This condition handling should be done at the `controller` level.

apps/asap-server/src/controllers/content.ts
```diff
@@ -4,10 +4,14 @@ import { CMSContent } from '../entities/content';
 
 interface ReplyContent {
```
johnytiago, line 5: Maybe we should move this interface to `asap-hub/model`

PR #122, opened 2020-07-30
feat: extend user with more props and team information

fampinheiro:


packages/model/src/user.ts
```diff
@@ -4,11 +4,18 @@ export interface Invitee {
...
+  jobTitle?: string;
   orcid?: string;
+  department?: string;
   institution?: string;
+  location?: string;
 }
 
 export interface UserResponse extends Invitee {
   id: string;
+  teams: {
```
johnytiago, line 16: What do you think of re-using the TeamMember type https://github.com/yldio/asap-hub/blob/master/packages/model/src/team.ts#L1

fampinheiro: I understand the reasoning, but this is just an happy coincidence. I don't expect them to evolve in the same way.

johnytiago: Do you expect to return different data if you fetch team members or if you fetch team affiliations?

fampinheiro: I'm expecting the team information when you get `/users/:id` to be the short form (`id`, `displayName`, `role`) but when fetching teams `/teams/:id` the user may have more information.

PR #127, opened 2020-08-03
feat: welcome page

fampinheiro:


jeysal:
Oh, was gonna review this last thing today, sorry for taking too long

fampinheiro:
You can still give it a go.

packages/react-components/src/organisms/Layout.tsx
```diff
@@ -11,8 +12,8 @@ import {
 
 const containerStyles = css({
   position: 'relative',
-  width: '100%',
-  height: '100%',
+  minWidth: '100vw',
```
jeysal, line 15: Glanced over it, this is the only thing I'm concerned about - `100` is a value that should rarely be used with vw/vh. In this case, it already broke the layout on small sizes, because as soon as a scroll bar appears, `100vw` does not make sense anymore.
I'll open a PR changing it back - if there was a reason to make this change feel free to comment there.

PR #133, opened 2020-08-04
feat: add user create & update webhook

johnytiago:

### DIFF
--- a/apps/asap-server/src/cms/users.ts
+++ b/apps/asap-server/src/cms/users.ts
@@ -3,7 +3,7 @@ import { Base, BaseOptions } from '@asap-hub/services-common';
 import { Invitee } from '@asap-hub/model';
 import get from 'lodash.get';
 
-import { CMSUser } from '../entities/user';
+import { CMSUser, CMSOrcidWork } from '../entities/user';
 import { CMSTeam } from '../entities/team';
 
 export interface Connection {
@@ -110,6 +110,22 @@ export default class Users extends Base {
         json: {
           email: { iv: user.data.email.iv },
           connections: { iv: connections },
+        },
+      })
+      .json();
+  }
+
+  updateOrcidWorks(
+    user: CMSUser,
+    lastModifiedDate: string,
+    works: CMSOrcidWork[],
+  ): Promise<CMSUser> {
+    return this.client
+      .patch<CMSUser>(`users/${user.id}`, {
+        json: {
+          email: { iv: user.data.email.iv },
+          orcidLastModifiedDate: { iv: lastModifiedDate },
+          orcidWorks: { iv: works },
         },
       })
       .json();
--- a/apps/asap-server/src/entities/user.ts
+++ b/apps/asap-server/src/entities/user.ts
@@ -6,16 +6,16 @@ export interface CMSUser {
   data: {
     displayName: { iv: string };
     email: { iv: string };
-    firstName: { iv: string };
-    middleName: { iv: string };
-    lastName: { iv: string };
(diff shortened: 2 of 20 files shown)


apps/asap-server/src/cms/users.ts
```diff
@@ -114,4 +114,20 @@ export default class Users extends Base {
...
+  updateOrcidWorks(
+    user: CMSUser,
+    lastModifiedDate: string,
+    works: CMSOrcidWork[],
+  ): Promise<CMSUser> {
+    return this.client
+      .patch<CMSUser>(`users/${user.id}`, {
+        json: {
+          email: { iv: user.data.email.iv },
+          orcidLastModifiedDate: { iv: lastModifiedDate },
```
fampinheiro, line 127: Is the last modified date from the orcid system or defined by us?

apps/asap-server/src/handlers/users/sync-orcid.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (newOrcid && newOrcid !== payload.dataOld?.orcid?.iv) {
+        return {
+          statusCode: 200,
+          payload: await users.syncOrcidProfile(id),
+        };
+      }
+      return { statusCode: 204 };
+    }
+
+    throw Boom.notImplemented();
```
fampinheiro, line 58: Can you check if we need to return 200? I'm assuming a non-success code will make a new request to the lambda.

johnytiago: To be safe, I'm going to return a 204 instead 👍

apps/asap-server/src/cms/users.ts
johnytiago: This one come from the orcid system

apps/asap-server/src/entities/user.ts
```diff
@@ -6,17 +6,32 @@ export interface CMSUser {
   data: {
     displayName: { iv: string };
     email: { iv: string };
-    firstName: { iv: string };
-    middleName: { iv: string };
-    lastName: { iv: string };
-    jobTitle: { iv: string };
-    orcid: { iv: string };
-    institution: { iv: string };
+    firstName?: { iv: string };
```
fampinheiro, line 9: Is there any particular reason to make this optional? We should consider redefining the type for the particular use-case with need this flexibility.

apps/asap-server/test/helpers/create-user.ts
```diff
@@ -26,12 +26,12 @@ function transform(user: CMSUser): TestUserResponse {
...
     orcid: user.data.orcid && user.data.orcid.iv,
     institution: user.data.institution && user.data.institution.iv,
-    teams: user.data.teams,
+    teams: user.data.teams?.iv,
     connections: user.data.connections.iv,
   };
 }
 
-export const createUser = (): Promise<CMSUser> => {
+export const createUser = (overwrites: Invitee | {}): Promise<CMSUser> => {
```
fampinheiro, line 34:
```suggestion
export const createUser = (overwrites: Invitee | object = {}): Promise<CMSUser> => {
```

apps/asap-server/src/entities/user.ts
johnytiago: I just updated it to reflect what is mandatory in the createSchema.

apps/asap-server/test/helpers/create-user.ts
johnytiago: 🤦

PR #139, opened 2020-08-06
feat(profile): tabs stub

jeysal:


apps/frontend/src/users/Profile.tsx
```diff
@@ -28,7 +38,19 @@ const Page: React.FC<{}> = () => {
...
+          outputsHref={join(url, 'outputs')}
         />
+        <Switch>
+          <Route path={`${path}/about`} render={() => 'About'} />
+          <Route
+            path={`${path}/research-interests`}
+            render={() => 'Research interests'}
+          />
+          <Route path={`${path}/outputs`} render={() => 'Outputs'} />
+          <Redirect to={join(url, 'about')} />
```
fampinheiro, line 52: I was imagining we would keep about as the default route (render it on `/`). This is also an option but do we need the extra redirect?

jeysal: I don't have a strong opinion on this, only reason I did it this way was symmetry / consistency

jeysal: Relative links can be a bit confusing when tabs are at different levels, but with the `join` structure I used it should be fine

jeysal:
I also considered building a more sophisticated tab manager that takes care of all the routing and the outlet for the tab content etc., but I think it's overkill at this point

apps/frontend/src/users/Profile.tsx
jeysal: I guess a third option would be having the default tab as a "fallback" non-exact third route with path `/`, so that both `/` and `/about` (and `/whatever`) match it.
Not something I prefer, just naming it for completeness

jeysal: Let me know whether you want this changed or you're equally unpassionate about it :sweat_smile:

PR #143, opened 2020-08-11
refactor(components): rework layouting

jeysal:


jeysal:
This is still a draft because it does not include the changes to page grid/padding handling that it needs, I didn't manage to finish those today. But it is a preview of the layout with fixed header, navigation, etc. working

jeysal:
This is now ready I think. Make sure to go through the Storybook and Frontend and double check if everything looks okay. The JSON renderer pages I have not paid any attention to layouting nicely / giving any padding at all because it doesn't seem worth it to invest any time there.
I've decided to not have a `Container`-ish component, because not only the side padding, but also the top/bottom padding seems to vary greatly between pages, and we would either start nesting padding components, or such a component would essentially be a function translating lots of React props to CSS styles on a div. Also, this way every page can use whichever semantic tags etc. it likes and just compose the styles onto it.

apps/frontend/src/teams/TeamList.tsx
```diff
@@ -22,18 +22,18 @@ const Page: React.FC = () => {
 
   if (teams) {
     return (
-      <Container>
+      <>
         {teams.map((team) => {
           const { id } = team;
           return (
-            <div key="id">
+            <div key={id}>
```
fampinheiro, line 29: 🤦

PR #147, opened 2020-08-12
feat: custom title and subject on emails

fampinheiro:


apps/react-templates/lib/template-plugin.js
```diff
@@ -39,7 +40,11 @@ const apply = async (stats, origin) => {
     .map(async (f) => {
       const cache = createCache();
       const { extractCritical } = createEmotionServer(cache);
-      const { default: Component } = require(path.resolve(outputDir, f));
+      const pkg = require(path.resolve(outputDir, f));
+      const Component = pkg.default;
+
+      const { name } = path.parse(f);
+      let subject = pkg.subject ? pkg.subject : titleCase(name);
```
johnytiago, line 47: `const`?

johnytiago: Btw, the title will be a path name if no path is supplied? 🤔

fampinheiro: Regarding the title I didn't change the previous behavior. The title is subject of the email.

PR #150, opened 2020-08-13
feat: create tag label component

fampinheiro:


packages/react-components/src/atoms/TagLabel.tsx
```diff
@@ -0,0 +1,45 @@
...
+  transition: '200ms',
+  paddingLeft: '6px',
+  paddingRight: '6px',
+
+  color: lead.rgb,
+  fontSize: '0.8em',
+});
+
+type TagLabelProps = {
+  readonly highlight?: boolean;
```
jeysal, line 35: remove

packages/react-components/src/atoms/TagLabel.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import css from '@emotion/css';
+import { steel, lead } from '../colors';
+
+const borderWidth = 1;
+const containerStyles = css({
+  display: 'flex',
+  cursor: 'default',
```
jeysal, line 8: ?

packages/react-components/src/atoms/TagLabel.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import css from '@emotion/css';
+import { steel, lead } from '../colors';
+
+const borderWidth = 1;
+const containerStyles = css({
```
jeysal, line 6: Isn't this container already the molecule that contains multiple tag labels next to each other? It seems to me like this component should just have the inner element.
Also, the best tags are probably `ul > li` - I think a screen reader would otherwise read them as a sentence "Publication Preprint Paper Neurology"

jeysal:
Oh also, try zooming in using the Storybook buttons - that should be a good explanation of why I do things like `${6 / perRem}em` in so many places, in case that's something you wondered about :D

packages/react-components/src/atoms/TagLabel.tsx
fampinheiro: Can remove but I'm following https://3.basecamp.com/4011604/buckets/16601866/uploads/2894021447.

jeysal: I know and it can be reintroduced, but it seems dead / not implemented atm and users of the component might expect  something to happen when settings the prop

packages/react-components/src/atoms/TagLabel.tsx
fampinheiro: The reason to have the wrapper is  to ensure the tag always matches the content, otherwise we can have a tag stretched and it's the parent responsibility to ensure the border fits to content.

fampinheiro: Screenshot to represent what I'm saying.

image
image

jeysal: It's probably not a problem anymore when we have an `ul` tab label list molecule? Then we can even stop exporting `TagLabel` from the component library to make sure it cannot be used outside of an `ul`. For now can leave it as is or try messing around with `display: inline-block` or something

fampinheiro:
Thanks for the tip, no magic there :). Just forgot about the em approach.

PR #151, opened 2020-08-13
feat: create recent works component

fampinheiro:


packages/react-components/src/organisms/ProfileRecentWorks.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <div css={elementStyle}>
+              <TagLabel>{typeMap[type] || typeMap.UNDEFINED}</TagLabel>
+              <Headline3>{title}</Headline3>
+              <p>
+                Originally Published:{' '}
+                {format(
+                  date,
+                  `${publicationDate.day ? 'Do ' : ''}${
+                    publicationDate.month ? 'MMMM ' : ''
+                  }${publicationDate.year ? 'yyyy' : ''}`,
```
johnytiago, line 93: Does it make a difference that this is not capitalised, `YYYY`?

PR #152, opened 2020-08-13
fix: sync CMS not watching sub directories

johnytiago:


.gitlab-ci.yml
```diff
@@ -180,4 +180,4 @@ deploy:sync-cms:
   rules:
     - if: $CI_COMMIT_BRANCH == 'master'
       changes:
-        - dev/squidex
+        - dev/squidex/**/*
```
fampinheiro, line 183: Should we add dev/squidex/*` also? Or does the glob matches files on root?

johnytiago: `      - dockerfiles/**/*`
> Any of the files and subdirectories inside the dockerfiles directory.

johnytiago: I think it will track everything. 😉

PR #156, opened 2020-08-13
feat: avatarURL on User and TeamMember

johnytiago:
Changed the merge base to make it more obvious what I'm changing. Once #156 is merged, we can point this back to master

### DIFF
--- a/apps/asap-server/src/controllers/users.ts
+++ b/apps/asap-server/src/controllers/users.ts
@@ -15,23 +15,25 @@ import { origin } from '../config';
 import { fetchOrcidProfile, ORCIDWorksResponse } from '../utils/fetch-orcid';
 
 export const transform = (user: CMSUser): UserResponse => {
-  return {
-    id: user.id,
-    displayName: user.data.displayName.iv,
-    email: user.data.email.iv,
-    firstName: user.data.firstName?.iv,
-    middleName: user.data.middleName?.iv,
-    lastName: user.data.lastName?.iv,
-    biography: user.data.biography?.iv,
-    jobTitle: user.data.jobTitle?.iv,
-    institution: user.data.institution?.iv,
-    teams: user.data.teams?.iv || [],
-    orcid: user.data.orcid?.iv,
-    orcidLastModifiedDate: user.data.orcidLastModifiedDate?.iv,
-    orcidWorks: user.data.orcidWorks?.iv,
-    skills: user.data.skills?.iv || [],
-    avatarURL: user.data.avatarURL?.iv,
-  };
+  return JSON.parse(
+    JSON.stringify({
+      id: user.id,
+      displayName: user.data.displayName.iv,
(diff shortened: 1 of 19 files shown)


johnytiago:
Will have to add some tests before merging this ⏳

jeysal:
Thanks for pushing it nonetheless, even without tests I can base on this and start using the property. Will it be
1. an absolute URL, or
2. relative to some CMS URL, or
3. will we proxy avatar requests through the backend?

johnytiago:
It will be most likely an absolute URL to S3.

apps/asap-server/src/controllers/users.ts
```diff
@@ -15,22 +15,25 @@ import { origin } from '../config';
...
-    biography: user.data.biography?.iv,
-    jobTitle: user.data.jobTitle?.iv,
-    institution: user.data.institution?.iv,
-    teams: user.data.teams?.iv || [],
-    orcid: user.data.orcid?.iv,
-    orcidLastModifiedDate: user.data.orcidLastModifiedDate?.iv,
-    orcidWorks: user.data.orcidWorks?.iv,
-    skills: user.data.skills?.iv || [],
-  };
+  return JSON.parse(
```
johnytiago, line 18: Tell me what you guys think about my approach on removing undefined values for now

apps/asap-server/test/handlers/teams/fetch-by-id.test.ts
```diff
@@ -116,7 +116,9 @@ describe('GET /teams/{id}', () => {
         },
       }),
     )) as APIGatewayProxyResult;
+    const body = JSON.parse(result.body);
 
+    console.log(body);
```
fampinheiro, line 121: :(

packages/model/src/user.ts
```diff
@@ -10,6 +10,8 @@ export interface Invitee {
   department?: string;
   institution?: string;
   location?: string;
+  avatarURL?: string;
+  skills?: string[];
```
fampinheiro, line 14: We don't expect skills when adding a new user. The data is changed on the CMS.

dev/squidex/schemas/teams.json
```diff
@@ -108,6 +108,21 @@
           "placeholder": "",
           "isRequired": false
         }
+      },
```
fampinheiro, line 111: Do we need this?

johnytiago: Either the `fix:format` or squidex CLI's changes 🤷

packages/model/src/user.ts
johnytiago: I wanted to add all user props to the create user method so that I could test all of them return. I will address this on another PR. 👍

jeysal:
@fampinheiro is this ready?

jeysal:
@jt ready 2 merge?

PR #160, opened 2020-08-17
fix(profile): crash due to incorrect date format

jeysal:


jeysal:
works on storybook now, merging hotfix

PR #161, opened 2020-08-17
feat(profile): design QA amends

jeysal:


packages/react-components/src/atoms/Tag.tsx
```diff
@@ -10,29 +10,23 @@ const containerStyles = css({
   justifyContent: 'center',
   alignItems: 'center',
 
-  marginTop: '6px',
-  marginBottom: '6px',
-
-  [`@media (min-width: ${tabletScreen.width}px)`]: {
```
fampinheiro, line 16: Are we sure we want to remove this?

jeysal: I checked the design guide and it just says 12 top 6 bottom, no conditions

fampinheiro: image

It's based on this.

jeysal: Yeah, it matches that now, expect I made the padding (see Measure>Right side) 7 top 5 bottom because James asked me to move the text down by 1px in Design QA

PR #167, opened 2020-08-18
feat: adds connect by code for rules webhook

johnytiago:


fampinheiro (approved):

packages/auth0-functions/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@asap-hub/auth0-functions",
+  "private": true,
+  "version": "0.1.0",
+  "engines": {
+    "node": "12.x"
+  },
+  "main": "build/index.js",
```
jeysal, line 8: These do not exist - we can remove the fields or (this is my preference) reduce te app to just this function in an `index.ts`, and when we create another function, just create a different app for it

jeysal (changes requested):

jeysal (approved):
will need to extend the auth0.md docs but that can be done later once the new rules are in place and tested :+1:

PR #171, opened 2020-08-19
feat: rename profile skills to skills section

fampinheiro:


jeysal:
:+1: although if we end up sharing a lot between users and teams, we might need terminology to encompass both - maybe profile is actually such a term? "user profile" and "team profile"

PR #172, opened 2020-08-19
feat: members section

fampinheiro:
Rough implementation of members section.


packages/react-components/src/atoms/__tests__/Link.test.tsx
```diff
@@ -37,6 +37,17 @@ describe.each`
...
 
+  it('respects the underline prop', () => {
+    const { getByRole } = render(
+      <Link underline={false} href={href}>
+        text
+      </Link>,
+      { wrapper },
+    );
+    const { textDecoration } = getComputedStyle(getByRole('link'));
+    expect(textDecoration).toBe('none');
```
jeysal, line 48: First render without the prop and `expect(textDecoration).toMatchInlineSnapshot()` or so, then rerender with the prop set and to this :)

apps/storybook/src/MembersSection.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { MembersSection } from '@asap-hub/react-components';
+
+export default {
+  title: 'Organisms / Team / Members',
+};
+
+export const Normal = () => (
```
jeysal, line 8: Can we add an empty variant on this?

packages/react-components/src/organisms/__tests__/MembersSections.test.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import MembersSection from '../MembersSection';
+
+it('renders an an header with number of members', () => {
```
jeysal, line 5: tests appear to be off (don't test what their titles say)

packages/react-components/src/organisms/MembersSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const avatarContainer = css({
+  flexBasis: 1,
+});
+
+const detailsContainer = css({
+  paddingLeft: `${12 / perRem}em`,
+});
+
+const memberContainerStyles = css({
```
jeysal, line 35: We can merge this / not block the PR on this for now, but the layouting needs a bit of work, it has weird states atm
image
The media query should probably be removed in favor of automatic wrapping since the text sizes are unknown, and this should probably be an `auto-fill` grid or so, not a flex, since otherwise there can be holes
image
Let me know if you want me to take over part of this

jeysal (changes requested):
Mostly good work esp. with using list tags :raised_hands: 
I have a couple of comments, the big one in particular could be addressed post merge

packages/react-components/src/organisms/MembersSection.tsx
fampinheiro: We can discuss this when you're back :)

packages/react-components/src/atoms/__tests__/Link.test.tsx
fampinheiro: It looks like we can't have `toMatchInlineSnapshot` twice on the same test.

packages/react-components/src/organisms/MembersSection.tsx
jeysal: I might also be able to fit something in tomorrow after finishing the current ticket

packages/react-components/src/atoms/__tests__/Link.test.tsx
jeysal: Yes you can o.O any error or so?

fampinheiro: `Jest: Multiple inline snapshots for the same call are not supported.`

PR #173, opened 2020-08-20
feat: adds connect and user-metadata rules

johnytiago:
Changed the merge base to reduce noise. Once #167 is merged, we can point this to master


babel.config.js
```diff
@@ -7,4 +7,5 @@ module.exports = {
     [require.resolve('@babel/plugin-transform-runtime'), { corejs: 3 }],
   ],
   babelrcRoots: ['.', 'apps/*', 'packages/*'],
+  ignore: ['apps/auth0-add-user-metadata/*', 'apps/auth0-connect-user/*'],
```
jeysal, line 10: Can we solve this like in `orcid-user-profile-script`? It has a separate `babel.config` (thus a separate Babel root) that is completely independent of this root. It can thus still use all newest features and TS though

fampinheiro (approved):

babel.config.js
fampinheiro: Can you do that change @jeysal?

jeysal: I think @johnytiago has already done it, this line just needs to be removed (is dead now)

PR #174, opened 2020-08-20
feat: use last update date instead of orcid date

fampinheiro:


packages/react-components/src/templates/ProfileHeader.tsx
```diff
@@ -67,16 +67,17 @@ const iconStyles = css({
...
   | 'displayName'
-  | 'institution'
-  | 'orcidLastModifiedDate'
   | 'firstName'
+  | 'institution'
+  | 'jobTitle'
+  | 'lastModifiedDate'
   | 'lastName'
   | 'location'
+  | 'orcidLastModifiedDate'
```
jeysal, line 79: probably unused now? But I guess the red CI likely already complained :sweat_smile:

packages/react-components/src/templates/__tests__/ProfileHeader.test.tsx
```diff
@@ -68,11 +68,6 @@ it('does not show team information if the user is not on a team', async () => {
 });
 
 it('generates the last updated text', () => {
-  const { container } = render(
-    <ProfileHeader
-      {...boilerplateProps}
-      orcidLastModifiedDate={formatISO(subYears(new Date(), 2))}
-    />,
-  );
+  const { container } = render(<ProfileHeader {...boilerplateProps} />);
```
jeysal, line 71: sorry to be a pain but can I ask you to revert moving the prop to the `boilerplateProps`? The `expect`ation is confusing if you can't see straight in the test why it should be "2 years ago". All the other tests pass those props that are relevant to them themselves.

fampinheiro: Ok, I can override it. `lastModifiedDate` is required now.

PR #176, opened 2020-08-20
refactor: eliminate pages indirection layer

jeysal:
This technically fixes all of it, but I'll probably want to restructure the profile page to use nested routing for tabs as well, because that working without rerenders is kind of a conincidence atm.
Anyway I'll think about it with a fresh mind tomorrow.

Also I love that (and wouldn't have expected that) this is net lines deleted :muscle:


fampinheiro:
Nice one :+1: 
The only points I would discuss further (future work):
- Don't have a NotFound on every nested switch (I understand it's related to nesting routes)
- Extract the loading component to a common place

jeysal:
Def agree on loading and error pages.
On nested Switch I couldn't think of an easy solution for now unfortunately, I also don't like it :/

fampinheiro (approved):

PR #186, opened 2020-08-25
fix: connect webhook payload

johnytiago:


johnytiago:
Just tested and the rule is now working fine and connecting users based on `invitation_code` ✅

fampinheiro (approved):

PR #187, opened 2020-08-25
fix: prevent multiple connects with same code

johnytiago:


apps/asap-server/src/cms/users.ts
```diff
@@ -118,7 +118,11 @@ export default class Users extends Base {
       .json();
   }
 
-  connectByCode(user: CMSUser, id: string): Promise<CMSUser> {
+  connectByCode(user: CMSUser, id: string): Promise<CMSUser> | CMSUser {
+    if (user.data.connections.iv.find(({ code }) => code === id)) {
+      return user;
```
fampinheiro, line 123: can we return `Promise.resolve(user)` and don't change the method signature?

johnytiago: Sure! 👍

fampinheiro (approved):

PR #190, opened 2020-08-26
feat: rich-text first iteration

fampinheiro:


apps/frontend/src/api/util.ts
```diff
@@ -16,15 +16,18 @@ export const useFetchOptions = ({
     options: { headers, ...options },
     ...args
   }) => {
-    if (!isAuthenticated) {
-      throw new Error('No authorization bearer token available');
-    }
+    const authHeaders = isAuthenticated
```
jeysal, line 19: I think we should hide this behind an option. Reason why I originally made it throw is if you accidentally call it without waiting for user authorization and expect it to set the header, it'll be a very confusing hard to debug failure.

packages/react-components/src/atoms/Display.tsx
```diff
@@ -6,8 +6,14 @@ interface DisplayProps {
   readonly children: TextChildren;
   readonly styleAsHeading?: keyof typeof headlineStyles;
 }
-const Display: React.FC<DisplayProps> = ({ children, styleAsHeading = 1 }) => (
-  <h1 css={[layoutStyles, headlineStyles[styleAsHeading]]}>{children}</h1>
+const Display: React.FC<DisplayProps> = ({
```
jeysal, line 9: I want to keep allowed component usage as low as possible because it my experience otherwise it leads to a hard to understand component structure like on the YLD website. In particular, I've avoided passing arbitrary props to HTML elements, or passing arbitrary styles to components. Can you tell me what is the use case for the extra props on the typography element here, so that we can maybe look for a less invasive solution? What does rehype need to set other than children?

apps/frontend/src/pages/content.css
```diff
@@ -0,0 +1,7 @@
+.toc-level {
```
jeysal, line 1: Can this be set as [Nested Selectors](https://emotion.sh/docs/nested) in the component lib / rich text?

jeysal (changes requested):
Nice, just some comments on changes to existing code that I'd consider too invasive

packages/react-components/src/atoms/Display.tsx
fampinheiro: It needs to set the id for jumping into the multiple areas. I changed it to be more intentional.
But we need to revisit this :)

apps/frontend/src/api/util.ts
fampinheiro: We don't need to authenticate the request that lead me to do this change, so I'm reverting it.

apps/frontend/src/pages/content.css
fampinheiro: Sure :+1:

packages/react-components/src/atoms/Display.tsx
jeysal: @fampinheiro do you think I should write a `react-components/README.md` with some principles that the library follows?

jeysal (commented):
LGTM except I think you missed one component when implementing review feedback.
Also, I can look into completing the typings for `rehype-slug` later, and see if I have any ideas if we can avoid the super weird typings in `RichText` :sweat_smile:

fampinheiro:
There is a lot to do on rehype and the components :+1: 
This is a good starting point.

jeysal (approved):
LGTM, will look into TS improvements later :+1:

PR #193, opened 2020-08-28
feat: create lambda to fetch pages

fampinheiro:


johnytiago (changes requested):

apps/asap-server/test/handlers/pages/fetch.test.ts
```diff
@@ -0,0 +1,33 @@
+import { APIGatewayProxyResult } from 'aws-lambda';
+import { handler } from '../../../src/handlers/pages/fetch';
+import { apiGatewayEvent } from '../../helpers/events';
+
+describe('Get /pages/{path+}', () => {
```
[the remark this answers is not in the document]
fampinheiro: It's a catch all, everything from the path is captured in that parameter.

apps/asap-server/test/handlers/pages/fetch.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  afterEach(() => {
```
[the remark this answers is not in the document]
fampinheiro:
```suggestion
```

apps/asap-server/test/handlers/pages/fetch.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    jest.clearAllMocks();
+  });
```
fampinheiro, line 8:
```suggestion
```

packages/services-common/src/squidex/client.ts
```diff
@@ -0,0 +1,57 @@
...
+      form: {
+        grant_type: 'client_credentials',
+        scope: 'squidex-api',
+        client_id: squidex.clientId,
+        client_secret: squidex.clientSecret,
+      },
+    }).json();
+
+    token = res.then((r) => (r as { access_token: string }).access_token);
+    /* eslint-enable @typescript-eslint/camelcase */
```
johnytiago, line 24:
```suggestion
    /* eslint-enable @typescript-eslint/camelcase */
    token = res.then((r) => (r as { access_token: string }).access_token);
```

johnytiago (changes requested):

packages/services-common/src/squidex/client.ts
fampinheiro: I need the disable on the promise on line 23.

johnytiago: You should disable the rule for the line after. Not the line before

PR #203, opened 2020-09-02
fix(members): dense grid in all circumstances

jeysal:
includes a bunch of refactors around links / theme


packages/react-components/src/text.ts
```diff
@@ -78,19 +78,18 @@ export const headlineStyles = {
 } as const;
 
 export type AccentColorName =
-  | 'berry'
```
jeysal, line 81: Reverted somewhat recent change adding `charcoal` by @fampinheiro - it's not an accent color, it's the default on light theme. I've handled the case it was added for with an IMO better structure

packages/react-components/src/organisms/MembersSection.tsx
```diff
@@ -1,72 +1,72 @@
...
-  flexBasis: '50%',
-  [`@media (max-width: ${mobileScreen.width}px)`]: {
-    flexBasis: '100%',
-    flexDirection: 'column',
+  fontWeight: 'bold',
+});
+const roleStyles = css({
+  gridColumn: 2,
+  [`@media (min-width: ${tabletScreen.min}px)`]: {
+    'li:nth-of-type(2n) &': {
```
jeysal, line 37: I didn't realize `emotion` supported this (referring to parent elements by putting `&` at the end, this opens up amazing possibilities...)

packages/react-components/src/atoms/Link.tsx
```diff
@@ -1,50 +1,49 @@
 import React, { ReactNode } from 'react';
-import css from '@emotion/css';
+import css, { CSSObject, SerializedStyles } from '@emotion/css';
 import { Link as ReactRouterLink } from 'react-router-dom';
 
-import { fern, pine, paper } from '../colors';
+import { fern, paper, pine } from '../colors';
```
fampinheiro, line 5: Thanks for sorting them :)

jeysal: lol that was more of a remove one and add it back later, could have gone the other way as well :see_no_evil:

fampinheiro (approved):

jeysal:
coverage degradation is just because of the deletions

PR #204, opened 2020-09-03
build: update TS and CRA

jeysal:
typescript@4 cannot be used yet due to lack of CRA support, but this at least gets us closer and thus gives us assurance that the update will be as easy as possible


fampinheiro (approved):

jeysal:
passed typecheck and CRA built a working frontend that was deployed, so looks good

PR #206, opened 2020-09-03
chore: custom auth0 variables

fampinheiro:


jeysal:
Hmm, I think TypeScript has a point here saying `process` is not defined, *a lot* of things depend on `@asap-hub/auth` and I wouldn't rely on them inlining env variables as needed.
If we want to keep this centralized, I think perhaps we should go for https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables? (I can write typings to go with that change in minutes if we decide for it)

packages/auth/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "@asap-hub/eslint-config-asap-hub": "workspace:packages/eslint-config-asap-hub",
-    "@auth0/auth0-spa-js": "^1.9.0"
+    "@auth0/auth0-spa-js": "^1.9.0",
```
jeysal, line 19: tbc back to 12 when rebasing on #211

jeysal:
It seems the Auth0 dev environment isn't configured according to docs yet, can we do that before merge so that signin page etc. already match prod closely? @fampinheiro

packages/auth/package.json
jeysal: Can rebase on master now and use its something something 12.12 version everywhere :sweat_smile:

jeysal (approved):
LGTM, as I said I think we should configure the Auth0 dev env as well before merging

fampinheiro:
The only thing missing is the authentication page, the rules are already in place and working :+1:

jeysal:
Okay. Any questions on `auth0.md` lmk

PR #208, opened 2020-09-03
Add team members to team endpoint

kmaid:


fampinheiro (approved):

apps/asap-server/src/entities/team.ts
```diff
@@ -1,5 +1,7 @@
 export interface CMSTeam {
   id: string;
+  created: string;
```
johnytiago, line 3: Are we using this attribute atm or planning on doing so soon?

johnytiago (approved):

apps/asap-server/src/entities/team.ts
kmaid: We aren't using created or updated date currently. The new figma designs have updated date in them. Can't hurt as it is provided by Squidex? Ideally this type would be complete.

PR #209, opened 2020-09-03
feat(rich-text): improve (type) safety

jeysal:


jeysal:
@fampinheiro this is the rough direction of my attempt to fix the safety issues that the types are pointing at for your eyes. obviously not final either, this will have to grow as we support more tags etc.

jeysal:
Won't focus on this for now but will probably occasionally push something here when I need a break from my main work ;D

fampinheiro:
It looks like an improvements, do we want to break hard on lack of `id`? The jump it's a nice to have, don't want the page to break on slug misbehaving.

jeysal:
> It looks like an improvements, do we want to break hard on lack of `id`? The jump it's a nice to have, don't want the page to break on slug misbehaving.

My understanding is that `rehype-slug` is used to generate an `id` for *every* `<h*>` based on its text content, so the `id` not being a `string` (but e.g. `undefined`) would be a hypothetical case that should never occur at runtime, thus warranting an assertion.

jeysal:
I see, only with `toc` prop

jeysal:
The typings for `rehype-react` are indeed garbage. This will need a cast, but I think what I've done now  is what we need: Converting from props that rehype can normally pass to a `p / h1 / a / ...` to props that we can pass to our `Paragraph / Headline2 / Link / ...`. It's tested, and because I think I did the *right* cast the tests now manage to cover all cases that could happen with rehype in reality.

fampinheiro (approved):

jeysal:
when deployed should check that terms page still works, cannot test locally atm due to a CORS config problem

jeysal:
seems to work https://209.hub.asap.science/privacy-policy

PR #213, opened 2020-09-04
Feature: Team card

kmaid:


packages/react-components/src/atoms/Link.tsx
```diff
@@ -6,7 +6,35 @@ import { fern, paper, pine } from '../colors';
...
+const BaseLink: React.FC<BaseLinkProps> = ({ children, href, styles }) => {
+  if (useHasRouter()) {
+    return (
+      <ReactRouterLink css={styles} to={href}>
+        {children}
+      </ReactRouterLink>
+    );
+  }
+
+  const internal =
```
kmaid, line 23: I think this is the problem for the bug mentioned in standup. I think we are using the react router link because there is a router. We might need to test for both

jeysal: Hmm yeah. Tests should have caught that - I have an idea how to improve the tests so that they catch this kind of thing, will pick the bug up at some point. Thanks!

packages/react-components/src/molecules/TeamCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+  skills,
+  members,
+}) => {
+  const shownSkills = skills.slice(0, 2);
+  const hiddenSkillsCount = skills.slice(2).length;
+
+  return (
+    <Link theme={null} href={`/teams/${id}`}>
+      <Card>
+        <Headline4>{displayName}</Headline4>
```
jeysal, line 41: This should probably be a `Headline2 styleAsHeading={4}` (headings numbers must be in order in HTML mostly for accessibility reasons, even when we want to be more flexible from a styling view)

packages/react-components/src/molecules/__tests__/TeamCard.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    'Adult Neurology',
+    'Neuroimaging',
+    'Neurologic Examination',
+    'Neuroprotection',
+  ],
+  members: [member],
+  lastModifiedDate: '2020-07-31T11:45:14Z',
+};
+
+it('renders the content', () => {
```
jeysal, line 33: You can make this just about `renders the team name as the headline`, check that `getByRole('heading')` has the expected text content and has `tagName H2`.
Project summary check is not needed as that pretty declarative - don't need to assert on bits of 'static markup'.
1 Team Member check can go into a test case at the bottom colocated with the other team members test so that there is one for singular and one for plural down there.
Neurological Diseases check is redundant, there are enough tests for the skills.

jeysal (changes requested):
Good work, all of my comments are minor details! :rocket: 
(except around extracting a component, but that's for a follow up PR)

packages/react-components/src/molecules/TeamCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  '> li:not(:last-of-type)': {
+    paddingRight: `${12 / perRem}em`,
+  },
+});
+
+interface TeamCardProps {
+  id: string;
+  displayName: string;
+  projectSummary?: string; // Should this be optional?
```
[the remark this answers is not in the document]
fampinheiro: We can discuss it, I would prefer typings to have as little optionals as possible and use the api to ensure they are empty string on most cases.

Up for discussion :point_up:

apps/storybook/src/NetworkCards.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  title: 'Organisms / Network Cards',
+};
+const member = {
+  id: 'ff0e04ac-4769-44ed-8d3b-245c1bfe17b3',
+  firstName: 'Mason',
+  lastName: 'Carpenter',
+  displayName: 'Birdie Romeo',
+  role: 'VrrPdl',
+};
+const teamCardProps = {
```
fampinheiro, line 14: While you are at it can we have more meaningful example data?

apps/storybook/src/NetworkCards.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+      'Project Summary',
+      'Molecular actions of PD-associated pathological proteins using in vitro human pluripotent stem cell-derived brain organoids',
+    ),
+    skills: [...Array(numberOfSkills).fill(skill, 0, numberOfSkills)],
+    members: [...Array(numberOfMembers).fill(member, 0, numberOfMembers)],
+    lastModifiedDate: '2020-07-31T11:45:14Z',
+  };
+};
+
+export const TeamCardNormal = () => <TeamCard {...teamCardProps()} />;
```
jeysal, line 39:
```suggestion
export const Normal = () => <TeamCard {...teamCardProps()} />;
```
(Team Card it already says in the overall Story title)

apps/storybook/src/NetworkCards.stories.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { number, text, withKnobs } from '@storybook/addon-knobs';
+
+import { TeamCard } from '@asap-hub/react-components';
+
+export default {
+  title: 'Organisms / Team / Team Card',
+  decorators: [withKnobs],
```
jeysal, line 8:
```suggestion
```
knobs are already enabled on all stories

jeysal (approved):
LGTM, nice work!
(please click apply suggestions for the two more minor things I found unless you have concerns)
@fampinheiro do you want to make the model change you suggested? (since I think it'd be in another PR probably)

PR #214, opened 2020-09-04
fix(link): identify external links even in a router context

jeysal:
Fixes https://trello.com/c/kgrugxNf/288-website-link-redirects-the-user-to-an-internal-page

First I thought I'd need to write a test that mocks the `location`, actually clicks the link, and then checks where it *really* redirected to (which would still be the safest way to test, but make the test even more complex), but this also catches such an error nicely:
image


kmaid (approved):
LGTM. Time to rebase my pr :D

fampinheiro (approved):

PR #215, opened 2020-09-04
build: only use @types/node where appropriate

jeysal:
Unfortunately, TS is quite lax on importing transitive type definitions, so `import 'fs'` is still possible in many locations (e.g. `frontend`) where it shouldn't. But this is at least somewhat closer to reality and e.g. doesn't allow that in `auth`...


jeysal:
```
➤ YN0000: [@asap-hub/frontend]: TypeScript error in undefined(undefined,undefined):
➤ YN0000: [@asap-hub/frontend]: Cannot find type definition file for 'apr-intercept'.  TS2688
```
wtf, apr-intercept is not used by the frontend.
no idea what is going on and I'll pass on dealing with that, thanks.

PR #216, opened 2020-09-04
fix: cronjob schedule

johnytiago:


johnytiago:
image
It is working now. Good for mergex! ✅

PR #221, opened 2020-09-07
feat(layout): basic mobile and desktop structures

jeysal:


packages/react-components/src/templates/BasicLayout.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
jeysal, line 1: This now has essentially zero overlap with the main layout and would only make it even more complex if in the same file, so I extracted it.

packages/react-components/src/templates/WelcomePage.tsx
```diff
@@ -50,7 +50,7 @@ const containerStyles = css({
 });
 
 const headerStyles = css({
-  flexGrow: 9999,
+  flexGrow: 1,
```
fampinheiro, line 53: Thank you

packages/react-components/src/templates/Layout.tsx
```diff
@@ -1,65 +1,179 @@
...
+import { tabletScreen } from '../pixels';
+import { MenuHeader, MainNavigation, UserNavigation } from '../organisms';
+import { UserMenuButton } from '../molecules';
+import { Overlay } from '../atoms';
 
-const showNavigationQuery = `@media (min-width: ${smallDesktopScreen.width}px)`;
+const drawerQuery = `@media (max-width: ${tabletScreen.width - 1}px)`;
+const crossQuery = `@media (min-width: ${tabletScreen.width}px)`;
 
 const styles = css({
```
jeysal, line 14: This is about as small as this component can get unfortunately, I've already extracted out everything that isn't tightly coupled to the grid layout

packages/react-components/src/templates/WelcomePage.tsx
jeysal: lol, I kinda just moved the hack

jeysal: but it's a fairly common CSS hack :shrug:

jeysal:
This looks like *a lot* to review, but really `Layout` and `Overlay` are new, everything else is basically stub components.

fampinheiro (approved):

kmaid (approved):

PR #227, opened 2020-09-08
Serverless dotend setup

kmaid:


serverless.js
```diff
@@ -27,6 +27,7 @@ const {
 
 const service = paramCase(pkg.name);
 const plugins = [
+  'serverless-dotenv-plugin',
```
fampinheiro, line 30: Can we load this just when it's not production, just like `serverless-offline`?

fampinheiro (changes requested):

fampinheiro (approved):

PR #231, opened 2020-09-09
Feat: People Card

kmaid:
Refactored profile text to be in seperate reusable component


apps/asap-server/src/controllers/users.ts
```diff
@@ -17,6 +17,7 @@ export const transform = (user: CMSUser): UserResponse => {
   return JSON.parse(
     JSON.stringify({
       id: user.id,
+      createdDate: user.created,
       lastModifiedDate: user.data.lastModifiedDate?.iv,
```
kmaid, line 21: Should this be lastModified? This data isn't shown on the frontend

packages/react-components/src/molecules/PeopleCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  createdDate,
+  firstName,
+  lastName,
+  location,
+  teams,
+  jobTitle,
+  avatarURL,
+}) => {
+  return (
+    <Link theme={null} href={`/users/${id}`}>
```
jeysal, line 68: I may have missed this in other reviews, but we should pass these hrefs in as props. Route structure is a frontend concern, its definition in `Route`s should not be separated from its usage in `Link`s

packages/react-components/src/index.ts
```diff
@@ -30,17 +30,19 @@ export {
...
+  LabeledDropdown,
+  LabeledPasswordField,
   LabeledTextArea,
   LabeledTextField,
-  LabeledPasswordField,
-  LabeledDropdown,
+  MenuButton,
+  OrcidSigninButton,
+  PeopleCard,
+  ProfilePersonalText,
```
jeysal, line 43: I believe we should not export this, it is only to be used as part of other components and does not make much sense standalone. Basically everything that we want to show standalone in storybook needs to be exported, nothing else. This is not technically enforced and possibly not consistent at the moment, but seems like the best guiding principle to me.

packages/react-components/src/molecules/PeopleCard.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
```
jeysal, line 1: Can we move this to `organisms`? Unless you've thought about it and disagree that it fits in there

packages/react-components/src/molecules/ProfilePersonalText.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
```
jeysal, line 1: Great to have this extracted!

packages/react-components/src/molecules/PeopleCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  alignItems: 'start',
+});
+
+const textContainerStyles = css({
+  flexGrow: 1,
+  display: 'flex',
+  flexWrap: 'wrap',
+  justifyContent: 'space-between',
+  alignItems: 'start',
+  '> *': {
```
jeysal, line 21: This invades `ProfilePersonalText`'s styling and should IMO be avoided. We can wrap it a container that ensures it is big enough and then gives that space to `ProfilePersonalText`.

jeysal (changes requested):

packages/react-components/src/molecules/PeopleCard.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import css from '@emotion/css';
+import format from 'date-fns/format';
+import { UserResponse } from '@asap-hub/model';
+
+import { Card, Link, Headline2, Avatar, Caption } from '../atoms';
+import ProfilePersonalText from './ProfilePersonalText';
+import { tabletScreen } from '../pixels';
+
+const containerStyles = css({
```
jeysal, line 10: Heh, was about to comment on this in my review since it seemed overly complex. Is there a case where even no styling (`display: block`) would not suffice?

jeysal: Oh seems the behavior changed with your recent push

jeysal (changes requested):
found a few more small problems but overall looking good!

jeysal (approved):
Test failure appeared now unfortunately, but other than that LGTM!

PR #234, opened 2020-09-10
refactor(team-card): use CSS to cut off skill tags overflow

jeysal:
This CSS is a bit ... _wild_, but I think it's a good solution for the issue we chatted about @kmaid.
Also don't forget the list of tags should be extracted into a `TagList` molecule that can take a prop for whether it should cut off. That molecule can then be used in `ProfileSkills` as well.


kmaid (approved):

jeysal:
Being able to properly test the overflow indicator would be nice, maybe if I get to configuring Playwright in-browser testing or so that'll work out.

PR #235, opened 2020-09-10
fix(auth): session preservation without 3rd party cookies

jeysal:
This seems to fix it for me on Brave Browser with 3rd party cookie blocking.
Throw your browsers at it, folks! ^^


kmaid (approved):
Tested on latest firefox and chrome

PR #238, opened 2020-09-11
refactor(auth0): use redirect_uri to determine user claim key

jeysal:
This means feature environments now get a key with their PR#
from the Auth0 dev environment, instead of always `dev`.
This way, all environments know which key to expect and
can read the token easily.


jeysal:
Will merge this as it's tested and required for me to proceed, but would appreciate your reviews when you're back from holiday!

PR #240, opened 2020-09-14
Design Amends

kmaid:


packages/react-components/src/organisms/PeopleCard.tsx
```diff
@@ -32,6 +32,7 @@ const moveStyles = css({
   [`@media (min-width: ${tabletScreen.min}px)`]: {
     flexBasis: 'auto',
     order: -1,
+    lineHeight: `${lineHeight}px`,
```
jeysal, line 35: This should be inherited from the  `html` element already o.O

packages/react-components/src/organisms/TeamCard.tsx
```diff
@@ -53,6 +55,11 @@ const overflowContentStyles = css({
 const teamMemberStyles = css({
   display: 'flex',
   alignItems: 'center',
+  color: lead.rgb,
+});
+
+const summaryStyles = css({
```
jeysal, line 61: `Paragraph` has an `accent` prop because changing color is very common and explicit in the design system

packages/react-components/src/templates/NetworkPageHeader.tsx
```diff
@@ -31,15 +31,22 @@ const buttonTextStyles = css({
   },
 });
 
+const textStyles = css({
+  color: lead.rgb,
```
jeysal, line 35: see other comment about accent

packages/react-components/src/organisms/PeopleCard.tsx
kmaid: Caption seems to have a line height specified within it making it hard to override.

packages/react-components/src/organisms/TeamCard.tsx
kmaid: Ah! Whoops. Thanks 👍

packages/react-components/src/organisms/PeopleCard.tsx
jeysal: Oh, I didn't see there was a caption - this looks like it shouldn't be a caption, but normal text with styling adjustments. This likely generates illegal HTML, a `<figcaption>` without a `<figure>`.

jeysal: We may have to rethink the `Caption` component, but either way given that this deviates even from the `Caption` in styles, this probably isn't a usage of the Caption in design system terms

jeysal (approved):

PR #242, opened 2020-09-14
Chore: Refactor skills list to taglist

kmaid:


packages/react-components/src/molecules/__tests__/TagList.test.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import TagList from '../TagList';
+
+it('shows all skills when there are few', () => {
```
jeysal, line 6:
```suggestion
it('shows all tags when there are few', () => {
```

jeysal: can you also move this into a describe block `when summarized` or so, together with the other test it applies to

packages/react-components/src/molecules/TagList.tsx
```diff
@@ -0,0 +1,80 @@
...
+  '::after': {
+    content: '"+" counter(tags)',
+  },
+});
+
+interface TagListProps {
+  tags: string[];
+  summarize?: boolean;
+}
+const TagList: React.FC<TagListProps> = ({ tags, summarize = false }) =>
```
jeysal, line 57: It seems you haven't used this in `TeamCard` yet

jeysal (changes requested):

fampinheiro (approved):

PR #246, opened 2020-09-15
feat(layout): user menu

jeysal:


apps/frontend/src/App.tsx
```diff
@@ -18,11 +18,20 @@ import Page from './pages/Content';
...
   const { isAuthenticated } = useAuth0();
-  return isAuthenticated ? (
+  const user = useCurrentUser();
+  return isAuthenticated && user ? (
     <Layout
       libraryHref="/library"
       networkHref="/users"
       newsAndEventsHref="/news-and-events"
+      profileHref={`/users/${user.id}`}
+      teams={[]} // TODO
```
jeysal, line 28: This requires further Auth0 changes that would make the PR too big, doing it in a follow up

apps/frontend/src/App.tsx
```diff
@@ -18,11 +18,20 @@ import Page from './pages/Content';
...
+      settingsHref="/settings"
+      feedbackHref="/feedback"
+      logoutHref="/logout" // TODO
```
jeysal, line 31: This adds (frontend) and removes (LoginLogoutButton) a lot of code that would make the PR too big, doing it in a follow up

packages/react-components/src/organisms/MainNavigation.tsx
```diff
@@ -6,28 +6,31 @@ import {
...
-  mobileScreen,
-  312,
-  largeDesktopScreen,
-  258,
-  'px',
-)})`;
-
 const containerStyles = css({
-  width,
+  minWidth: `max(${vminLinearCalc(
```
jeysal, line 15: James requested the change on small desktop, this is now essentially a V-shaped function, with the bottom point at small desktop

kmaid: I recently added a vminLinearCalcCapped function. Maybe it would make sense to extend this to support a minimum value in addition to a maximum value?

fampinheiro (approved):

packages/react-components/src/organisms/MainNavigation.tsx
jeysal: TBH I'm not convinced that should exist. It's just a composition of `vminLinearCalc` and CSS `min`. Sometimes you want `min`, sometimes you want `max`, sometimes you want both, sometimes, like here, you want a `max` of *two* calcs. I would prefer removing that function as it's overly specific and its goal can be achieved by composition of functions

jeysal: @kmaid  further thoughts?

jeysal: Merging for now so that I submit a follow up for review. As I said I'd like to remove `vminLinearCalcPadded` entirely, maybe I'll open a PR+discussion for that at some point.

PR #247, opened 2020-09-15
Feat:  Toggle

kmaid:


packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ['&:not(:checked) ~ .left']: activeStyles,
+  ['&:checked ~ .right']: activeStyles,
+});
+const buttonRight = css({ marginLeft: '-20px', padding: '0 10px' });
+const buttonLeft = css({ marginRight: '-20px', padding: '0 10px' });
+
+interface ToggleProps {
+  defaultChecked?: boolean;
+  leftButtonText: string;
+  leftButtonIcon: React.ReactElement<SVGElement>;
```
jeysal, line 50: This isn't actually enforced by TypeScript (yet) FYI

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+const checkboxStyles = css({
+  display: 'none',
+  '&:not(:checked) ~ .left': activeStyles,
+  '&:checked ~ .right': activeStyles,
+});
+const buttonRight = css({ marginLeft: '-20px', padding: '0 10px' });
+const buttonLeft = css({ marginRight: '-20px', padding: '0 10px' });
+
+interface ToggleProps {
+  defaultChecked?: boolean;
```
jeysal, line 48: I think this interface is confusing. Is this just a `checked` prop, what does the `default` mean? And if it is `checked`, does that mean `left` or `right` is enabled? I would assume `right`, but representing the selection using `'left'` or `'right'` may be more intuitive.

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+  g: {
+    stroke: charcoal.rgb,
+  },
+});
+const checkboxStyles = css({
+  display: 'none',
+  '&:not(:checked) ~ .left': activeStyles,
+  '&:checked ~ .right': activeStyles,
+});
+const buttonRight = css({ marginLeft: '-20px', padding: '0 10px' });
```
jeysal, line 44: I think a grid with overlap could help eliminate these negative margins?

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import css from '@emotion/css';
+import { lineHeight, perRem } from '../pixels';
+import { steel, paper, tin, charcoal } from '../colors';
+const checkboxLabelStyles = css({
+  display: 'flex',
+  height: '54px',
```
jeysal, line 7: Can we make these relative as well, like other values further down?

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+const buttonRight = css({ marginLeft: '-20px', padding: '0 10px' });
+const buttonLeft = css({ marginRight: '-20px', padding: '0 10px' });
+
+interface ToggleProps {
+  defaultChecked?: boolean;
+  leftButtonText: string;
+  leftButtonIcon: React.ReactElement<SVGElement>;
+  rightButtonText: string;
+  rightButtonIcon: React.ReactElement<SVGElement>;
+  onChange: (e: React.FormEvent<HTMLInputElement>) => void;
```
jeysal, line 53: Related to the interface definition, can we not expose the whole event? The component that receives this should not care about `target` etc., and giving them just the new state is the simplest in usage as well

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { steel, paper, tin, charcoal } from '../colors';
+const checkboxLabelStyles = css({
+  display: 'flex',
+  height: '54px',
+  borderRadius: '27px',
+  border: `1px solid ${steel.rgb}`,
+  overflow: 'hidden',
+  backgroundColor: steel.rgb,
+  color: tin.rgb,
+  g: {
```
jeysal, line 13: Hmm, this seems prone to break with changes of the `svg` as it's not colocated with the actual definition of the icon? Should we just have two icons?

jeysal (changes requested):

packages/react-components/src/atoms/Toggle.tsx
kmaid: Maybe we could pass in props to the icon?

jeysal: Or does normal `fill: #color` do the job here?

kmaid: No it doesn't 👎

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -0,0 +1,85 @@
...
+  leftButtonIcon,
+  leftButtonText,
+  rightButtonIcon,
+  rightButtonText,
+  onChange,
+}) => (
+  <label css={checkboxLabelStyles}>
+    <input
+      type="checkbox"
+      defaultChecked={defaultChecked === 'right'}
```
jeysal, line 70: I still think this should be `checked`, not `defaultChecked`. Otherwise the component becomes uncontrolled, not (necessarily) in sync with the state of the app

packages/react-components/src/icons/profile-icon.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+/* istanbul ignore file */
+
+const profileIcon = (
```
jeysal, line 5: Is this the same as the `user` icon?

jeysal (commented):
Nice, great improvements to the (styling) structure IMO

packages/react-components/src/atoms/Toggle.tsx
jeysal: You can actually already see this happen in Storybook

jeysal:
Merging, I'll push some a11y fixes and tests afterwards.

PR #249, opened 2020-09-15
feat(user-menu): logout functionality

jeysal:


packages/react-components/src/index.ts
```diff
@@ -1,7 +1,6 @@
 import * as messages from './messages';
 import * as pixels from './pixels';
 import * as text from './text';
-// TODO when LoginLogoutButton is removed, move this out of components into frontend
```
jeysal, line 4: I ended up not doing this, and it will not happen, because the auth test utils now have other legitimate use cases inside of `react-components`

jeysal:
+60 -190, I love it

apps/frontend/src/auth/Logout.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { useAuth0 } from '@asap-hub/react-context';
+
+const Logout: React.FC<{}> = () => {
+  const { logout } = useAuth0();
+
+  useEffect(() => {
+    logout({ returnTo: globalThis.location.origin });
+  });
+
+  return 'Logging you out ...';
```
fampinheiro, line 11: Do we need a design for this?

jeysal: Not super urgently I guess, but for a 100% smooth experience yeah

jeysal: I'd be very interested to see the analytics for this, I always struggle to understand why logging out is such a common use case :sweat_smile:

jeysal:
Hi @kmaid can I haz review? :D

jeysal:
@fampinheiro @kmaid can I haz review? :D

kmaid (approved):

jeysal:
ty ^^

PR #250, opened 2020-09-15
build: use yarn-plugin-dotenv

jeysal:
This will initialize *all* scripts with the dotenv variables


serverless.js
```diff
@@ -91,9 +91,6 @@ module.exports = {
     'serverless-offline': {
       useChildProcesses: true, // needed for hot reloading to work https://github.com/dherault/serverless-offline/issues/931
     },
-    dotenv: {
-      path: 'apps/asap-server/.env',
```
kmaid, line 95: There is also a 'apps/asap-server/.env.example'

kmaid: Where are we loading the .env file? the one in the root directory is aimed at Docker and public. Could probably load a .env.local but it likely pickup all the docker variables and vice versa

jeysal: Oh, I thought I had moved the `.env.example` to the root

jeysal: Oh, the existing `.env` is a problem. `dotenv` recommends against having multiple `.env` files, and against committing `.env`. I don't really have an understanding of the backend configuration, @fampinheiro @johnytiago can you advise how we could eliminate the existing `/.env`? Can they just be defaults somewhere in code?

jeysal: I've solved it, hopefully. Can you take another look?

jeysal:
Dependent on you backend folks to review this, I don't really understand enough about the docker config to test if everything works properly this way, I only know the user-defined `.env` variables are now present in *every* script being run, as discussed with @kmaid

docker-compose.yaml
```diff
@@ -21,14 +21,15 @@ services:
...
-      - URLS__ENFORCEHTTPS=${SQUIDEX_FORCE_HTTPS}
+      - SQUIDEX_PROTOCOL=${SQUIDEX_PROTOCOL:-http}
+      - URLS__BASEURL=${SQUIDEX_URL:-http://localhost:4004}
+      - URLS__ENFORCEHTTPS=${SQUIDEX_FORCE_HTTPS:-False}
       - EVENTSTORE__MONGODB__CONFIGURATION=mongodb://mongo
       - STORE__MONGODB__CONFIGURATION=mongodb://mongo
-      - IDENTITY__ADMINEMAIL=${SQUIDEX_ADMINEMAIL}
-      - IDENTITY__ADMINPASSWORD=${SQUIDEX_ADMINPASSWORD}
+      - IDENTITY__ADMINEMAIL=${SQUIDEX_ADMINEMAIL:-admin@squidex.io}
+      - IDENTITY__ADMINPASSWORD=${SQUIDEX_ADMINPASSWORD:-SuperS3cure123./.?sdf}
```
kmaid, line 32: There is still a password

PR #252, opened 2020-09-16
feat(user-menu): show user's teams

jeysal:


packages/auth/src/index.ts
```diff
@@ -8,7 +8,9 @@ export { config };
 export type User = Pick<
   UserResponse,
   'id' | 'displayName' | 'email' | 'firstName' | 'lastName' | 'avatarURL'
->;
+> & {
+  teams: ReadonlyArray<Pick<UserResponse['teams'][0], 'id' | 'displayName'>>;
```
kmaid, line 12: Should this be name or display name?

jeysal: `displayName` - `name` is a property defined by `UserNavigation`. We could consider renaming *that* to `displayName` to bring it closer to the model, but not sure there's a point since its other property is `href` which isn't present in the model at all

kmaid (approved):

jeysal:
Rebased

jeysal:
Tested with two teams, works :ok_hand:

PR #253, opened 2020-09-16
Feat: Search Field

kmaid:


packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { TextField } from '../atoms';
+import { perRem } from '../pixels';
+import { searchIcon } from '../icons';
+
+const containerStyles = css({
+  paddingBottom: `${18 / perRem}em`,
+});
+
+type SearchFieldProps = {} & Exclude<ComponentProps<typeof TextField>, 'id'>;
```
jeysal, line 12: This should probably whitelist (Pick, not Exclude) only what we need for our use case, most of the `TextField` props don't make sense on the search

packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -0,0 +1,23 @@
...
+const SearchField: React.FC<SearchFieldProps> = (textFieldProps) => (
+  <div css={containerStyles}>
+    <TextField
```
jeysal, line 15: `type` should be `'search'`

jeysal (approved):

PR #257, opened 2020-09-17
feat: add text and type to outputs

fampinheiro:


packages/react-components/src/templates/ResearchOutputPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+  alignSelf: 'stretch',
+  background: pearl.rgb,
+  padding: `${36 / perRem}em ${contentSidePaddingWithNavigation(8)}`,
+});
+
+const articleStyles = css({
+  alignSelf: 'stretch',
+});
+
+const ResearchOutputPage: React.FC<ResearchOutputResponse> = ({
```
jeysal, line 20: Quick test to check that the individual parts are rendered?

packages/react-components/src/templates/ResearchOutputPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+  title,
+  text,
+}) => {
+  return (
+    <div css={articleStyles}>
+      <main css={[mainStyles]}>
```
jeysal, line 26: `<main>` is already done by the `Layout`, you can collapse the two elements into just the div

packages/model/src/research-output.ts
```diff
@@ -6,7 +8,7 @@ export type ResearchOutputAuthor =
   | { readonly id: string };
 
 export type ResearchOutputFormData = {
-  readonly url: string;
+  readonly url?: string;
```
jeysal, line 11: Why this change? In the form design it was a mandatory field

jeysal (changes requested):

packages/model/src/research-output.ts
fampinheiro: The proposal doesn't have a URL. But it's also not created by users.
Is keeping the URL required for the form and change it on the ResearchOutputResponse an option?

jeysal: Yeah so this type was specifically for the input form data, I don't think it even reaches the backend before it is re-mapped

fampinheiro:
Merging this and addressing the extraneous comments in my next PR.
@jeysal let me know if I left something out.

PR #258, opened 2020-09-17
Feat: Network page header

kmaid:


apps/frontend/src/network/ProfileList.tsx
```diff
@@ -22,18 +16,10 @@ const Page: React.FC<{}> = () => {
   if (usersData) {
     const users = usersData.map((user: UserResponse) => ({
       ...user,
-      href: join(url, user.id),
+      href: join('/users', user.id),
```
jeysal, line 19: Are you changing the path of individual entries later in a follow up?

apps/frontend/src/network/Routes.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const searchOnChange = (newQuery: string) => {
+    currentUrlParams.set('query', newQuery);
+    history.replace({ search: currentUrlParams.toString() });
+  };
+  return (
+    <Switch>
+      <Route
+        exact
+        path={`${path}/users`}
+        render={() => (
```
jeysal, line 32: `react-router-dom` now (idk since when exactly, actually) supports `children`! It's a much more readable substitute for `render` :tada:

apps/frontend/src/network/Routes.tsx
```diff
@@ -0,0 +1,62 @@
...
+            page="teams"
+            toggleOnChange={toggleOnChange('users')}
+            searchOnChange={searchOnChange}
+            query={currentUrlParams.get('query') || ''}
+          >
+            <TeamList />
+          </NetworkPage>
+        )}
+      />
+      <Route>Not Found</Route>
```
jeysal, line 57: Should we have a redirect to `users`? Could also change e.g. the `networkHref` to just `/network` then and have the Network define it's internal structure itself :thinking:

apps/frontend/src/network/__tests__/ProfileList.test.tsx
```diff
@@ -36,7 +36,7 @@ const users: ReadonlyArray<UserResponse> = [
     skills: [],
   },
   {
-    id: '55724942-3408-4ad6-9a73-14b92226ffb7',
+    id: '55724942-3408-4ad6-9a73-14b92226ffb77',
```
jeysal, line 39: ?
It's already one higher than the previous id :D

apps/frontend/src/network/__tests__/Routes.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    .reply(200, UsersResponse);
+});
+
+describe('Test toggle to team', () => {
+  it('Changes the placeholder', async () => {
+    const { getByText, queryByText, getByRole } = await renderNetworkPage(
+      '/network/teams',
+    );
+    const toggle = getByText('People');
+    const searchBox = getByRole('textbox') as HTMLInputElement;
```
jeysal, line 58: I *think* there's a `search` `role` that we should probably apply, if it's not already applied due to the input type, and test here using `getByRole('search')`

packages/react-components/src/atoms/Toggle.tsx
```diff
@@ -43,9 +43,9 @@ const buttonLeft = css({
   gridColumn: '1 / span 2',
   gridRow: 1,
 });
-
+export type Position = 'left' | 'right';
```
jeysal, line 46: export not needed

packages/react-components/src/templates/NetworkPage.tsx
```diff
@@ -19,10 +19,27 @@ const articleStyles = css({
   padding: `0 0 ${vminLinearCalc(mobileScreen, 36, tabletScreen, 72, 'px')}`,
 });
 
-const NetworkPage: React.FC = ({ children }) => {
+type NetworkPageProps = {
+  searchOnChange: (newQuery: string) => void;
```
jeysal, line 23: change handlers in the component lib are always optional (`?:`), initialized to `= noop`. This aids with testing, Storybook, etc., where fewer props need to be provided as a consequence.
Also, naming is usually `onChange*`, so `onChangeToggle`, `onChangeSearch`, ...

packages/react-components/src/templates/NetworkPageHeader.tsx
```diff
@@ -16,9 +18,27 @@ const containerStyles = css({
   }em `,
 });
 
+const textStyles = css({
+  maxWidth: `${610 / perRem}em`,
+});
+
 const controlsStyles = css({
```
jeysal, line 25: This IMO breaks way too soon (e.g. 768px width, small desktop rotated). Not sure what design specs say about this

packages/react-components/src/templates/NetworkPageHeader.tsx
```diff
@@ -16,9 +18,27 @@ const containerStyles = css({
...
-  display: 'flex',
-  justifyContent: 'flex-end',
+  display: 'grid',
+  gridRowGap: `${8 / perRem}em`,
+  gridColumnGap: `${18 / perRem}em`,
+  alignItems: 'center',
+  paddingTop: `${18 / perRem}em`,
+  [`@media (min-width: ${smallDesktopScreen.min}px)`]: {
+    paddingTop: `${2 / perRem}em`,
+    gridTemplateColumns: '230px auto',
```
jeysal, line 33: Can the toggle define this on its own based on its intrisic size, so that we only have to do `min-content` here?

jeysal (changes requested):
The structure LGTM :ok_hand: all comments are all fairly local things

apps/frontend/src/network/__tests__/Routes.test.tsx
kmaid: Ill address it in the search integration ticket if thats cool?

apps/frontend/src/fixtures/users.json
```diff
@@ -0,0 +1,65 @@
+[
+  {
```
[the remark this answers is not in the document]
kmaid: Yeah, makes sense!

apps/frontend/src/network/__tests__/Routes.test.tsx
jeysal: ya

apps/frontend/src/network/__tests__/Routes.test.tsx
```diff
@@ -0,0 +1,143 @@
...
```
jeysal, line 58:
```suggestion
    const searchBox = getByRole('textbox') as HTMLInputElement; // TODO 'search' role
```

apps/frontend/src/network/__tests__/ProfileList.test.tsx
jeysal, line 39:
```suggestion
    id: '55724942-3408-4ad6-9a73-14b92226ffb7',
```

apps/frontend/src/network/Routes.tsx
```diff
@@ -0,0 +1,59 @@
...
+          page="teams"
+          onChangeToggle={onChangeToggle('users')}
+          onChangeSearch={onChangeSearch}
+          query={currentUrlParams.get('query') || ''}
+        >
+          <TeamList />
+        </NetworkPage>
+      </Route>
+      <Route path={`${path}/teams/:id`} component={Team} />
+      <Redirect to={'/network/users'} />
```
jeysal, line 54:
```suggestion
      <Redirect to={`${path}/users`} />
```

apps/frontend/src/App.tsx
```diff
@@ -21,12 +20,12 @@ const ConfiguredLayout: React.FC = ({ children }) => {
   return isAuthenticated && user ? (
     <Layout
       libraryHref="/library"
-      networkHref="/users"
+      networkHref="/network/users"
```
jeysal, line 23:
```suggestion
      networkHref="/network"
```

jeysal (approved):
just applying a couple of suggestions for the unresolved comments

PR #265, opened 2020-09-21
Fix: People design amends

kmaid:


packages/react-components/src/molecules/ProfilePersonalText.tsx
```diff
@@ -5,21 +5,33 @@ import { UserResponse, UserTeam } from '@asap-hub/model';
...
   display: 'inline-block',
   width: `${lineHeight / perRem}em`,
   height: `${lineHeight / perRem}em`,
   paddingRight: `${6 / perRem}em`,
 });
 
+const paragraphStyles = css({
+  marginTop: 0,
+  marginBottom: `${18 / perRem}em`,
+  fontSize: `${17 / perRem}em`,
```
jeysal, line 29: The font size and line height should be inherited properly so we could just omit those, but the letter spacing is only defined on `Paragraph` because other texts should not have it as specified by design. Repeating it on every custom `<p>` seems like it will be (probably has been) forgotten a lot.
What do you think about adding `p { letter-spacing: ... }` to the global styles? That'd be the only solution that comes to my mind atm.

jeysal (approved):

PR #266, opened 2020-09-21
Misc styling changes

kmaid:


packages/react-components/src/form.ts
```diff
@@ -40,6 +40,12 @@ export const styles = {
   ':focus': {
     borderColor: fern.rgb,
   },
+  ' ~ div svg': {
```
jeysal, line 43: This seems a bit spooky, how can `form.ts` know about the `div` structure of the various form elements

packages/react-components/src/atoms/__tests__/TextField.test.tsx
```diff
@@ -78,7 +80,7 @@ describe('with a custom indicator positioned right', () => {
 
     expect(customIndicatorPaddingRight).toBeCloseTo(
       // times 2 because there is now padding on both sides of the indicator
```
jeysal, line 82:
```suggestion
```

packages/react-components/src/atoms/__tests__/TextField.test.tsx
```diff
@@ -123,7 +125,7 @@ describe('with a custom indicator positioned left', () => {
 
     expect(customIndicatorPaddingLeft).toBeCloseTo(
       // times 2 because there is now padding on both sides of the indicator
```
jeysal, line 127:
```suggestion
```

packages/react-components/src/templates/NetworkPageHeader.tsx
```diff
@@ -29,7 +29,7 @@ const controlsStyles = css({
   gridColumnGap: `${18 / perRem}em`,
   alignItems: 'center',
   paddingTop: `${18 / perRem}em`,
-  [`@media (min-width: ${smallDesktopScreen.min}px)`]: {
+  [`@media (min-width: ${smallDesktopScreen.min + 1}px)`]: {
```
jeysal, line 32: `min-width` being one greater than `smallDesktop`s `min` doesn't seem to make sense to me - previously it applied to small desktop regardless of orientation, now it only just misses portrait small desktop but hits landscape small desktop. Perhaps the intention was to make it miss tablets regardless of orientation? If so,
```suggestion
  [`@media (min-width: ${tabletScreen.max + 1}px)`]: {
```
(needs import)

jeysal (approved):

PR #268, opened 2020-09-22
draft: send mails flow

johnytiago:
queue

### DIFF
--- a/apps/asap-server/src/controllers/users.ts
+++ b/apps/asap-server/src/controllers/users.ts
@@ -1,5 +1,4 @@
 import Boom from '@hapi/boom';
-import Debug from 'debug';
 import path from 'path';
 import url from 'url';
 import Intercept from 'apr-intercept';
@@ -11,7 +10,7 @@ import { Invitee, UserResponse, ListUserResponse } from '@asap-hub/model';
 
 import { CMSUser, CMSOrcidWork } from '../entities/user';
 import { sendEmail } from '../utils/send-mail';
-import { origin } from '../config';
+import { origin, environment } from '../config';
 import { fetchOrcidProfile, ORCIDWorksResponse } from '../utils/fetch-orcid';
 import { createURL } from '../utils/squidex';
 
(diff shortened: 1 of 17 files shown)


apps/asap-server/src/controllers/users.ts
```diff
@@ -176,4 +158,41 @@ export default class Users {
 
     return transform(user);
   }
+
+  async sendWelcomeEmail(user: CMSUser) {
+    const code = `ASAP-${uuidV4()}`;
```
fampinheiro, line 163: Does this changes the URL sent to the user? I would prefer to keep the link as `/welcome/${uuid}`.
We can keep the prefix but just for internal use. Similar to what Auth0 does with `google-oauth2|${id}`, `autg0|${id}`.

johnytiago: Didn't thought about that. Let me fix it 👍🏽

apps/asap-server/test/handlers/webhooks/webhook-send-mails.test.ts
```diff
@@ -0,0 +1,259 @@
...
+      },
+      Template: 'Welcome',
+      TemplateData: JSON.stringify({
+        firstName: 'TestUser',
+        link: `${origin}/welcome/uuid`,
+      }),
+    });
+  });
+
+  //test('returns 200 when sends users emails - dev environments', async () => {
```
johnytiago, line 95: This test is commented because I'm not sure how to test this without getting a typing error.
`config.environment` is a `ReadOnly` prop imported from `process.env` and I can't change it mid tests. Do you know a way to address this?

apps/asap-server/src/handlers/webhooks/webhook-connect-by-code.ts
```diff
@@ -24,7 +26,20 @@ export const handler: Handler = lambda.http(
...
 
     const users = new Users();
-    await users.connectByCode(code, userId);
+    const squidex: Squidex<CMSUser> = new Squidex('users');
+    const user = await squidex
+      .fetchOne({
+        filter: {
+          path: 'data.connections.iv.code',
+          op: 'eq',
+          value: `ASAP|${code}`,
```
johnytiago, line 35: This will not be backwards compatible with the codes created prior to this unless we try to match both new and old welcomeCode structures

johnytiago:
Send mails lambda needs to assume a role that allows for SES usage

PR #269, opened 2020-09-22
feat(profile): subtitle-styled headings

jeysal:


jeysal:
This is the first time I've seen James request a 'Subtitle' font style (regular size but bold) being used and it seems to be used where it's semantically a heading, so I'll add it as a heading level for now. As usual in the designs we get, optical heading level is totally unrelated to semantic heading level, whatever.

packages/react-components/src/atoms/__tests__/Subtitle.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  expect(getByText('text').tagName).toBe('H5');
+});
+
+it('applies the text margin', () => {
+  const { getByText } = render(<Subtitle>text</Subtitle>);
+  const { marginTop } = getComputedStyle(getByText('text'));
+  expect(marginTop).toMatchInlineSnapshot(`"12px"`);
+});
+
+it('renders the text in regular size bold', () => {
```
kmaid, line 17: There is no assertion around bold

kmaid (approved):

packages/react-components/src/atoms/__tests__/Subtitle.test.tsx
jeysal: True dat. I noticed the other heading tests were only checking font size, not weight, and skipped it

PR #271, opened 2020-09-22
feat: change roms schema and adapt endpoints

fampinheiro:


apps/asap-server/test/handlers/research-outputs/fetch.test.ts
```diff
@@ -0,0 +1,146 @@
...
+import { apiGatewayEvent } from '../../helpers/events';
+import { ResearchOutputResponse } from '@asap-hub/model';
+
+const chance = new Chance();
+describe('GET /research-outputs', () => {
+  beforeEach(() => {
+    nock.cleanAll();
+  });
+
+  afterEach(() => {
```
kmaid, line 19: Shouldn't need to do this twice

kmaid (approved):

apps/asap-server/test/handlers/research-outputs/fetch.test.ts
fampinheiro: I agree with you :( but we have some misbehaving test suits. It's on my todo list to address this :+1:

PR #273, opened 2020-09-23
chore: change favicon

fampinheiro:


jeysal (approved):
:+1: don't forget to update the HTML in prod and dev Auth0

PR #280, opened 2020-09-24
chore: use asset as avatar

fampinheiro:
* Use asset to store avatar (profile picture)

### DIFF
--- a/apps/asap-server/src/utils/assets.ts
+++ b/apps/asap-server/src/utils/assets.ts
@@ -6,5 +6,5 @@ export const createURL = (assets: string[]): string | undefined => {
   }
 
   const asset = assets[0];
-  return `${cms.baseUrl}/api/assets/${cms.appName}/${asset}`;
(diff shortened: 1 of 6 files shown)


apps/asap-server/src/utils/assets.ts
```diff
@@ -0,0 +1,10 @@
+import { cms } from '../config';
+
+export const createURL = (assets: string[]): string | undefined => {
+  if (assets.length === 0) {
+    return undefined;
+  }
+
+  const asset = assets[0];
+  return `${cms.baseUrl}/api/assets/${cms.appName}/${asset}`;
```
jeysal, line 9: idk what `cms.baseUrl` contains, if it's something like `https://my.squidex.com/` then this might be better to deal slash / no slash
```suggestion
  return new URL(`/api/assets/${cms.appName}/${asset}`, cms.baseUrl);
```

jeysal (commented):
`createURL` missing tests, should be easy enough?

jeysal (approved):

PR #281, opened 2020-09-24
Feat: Rich text error handling

kmaid:


fampinheiro:
image

It looks like that in mobile the padding bottom is smaller than padding top.

jeysal (approved):
Cool, obviously error messages can always be better but for now not showing a blank page was the goal I think.
If you'd like to make the tests a bit less brittle for when we improve error messages in the future, I'd recommend using assertions like `toMatch(/invalid.+h1/i)`, but not a deal breaker for me :+1:

PR #282, opened 2020-09-24
Fix: Move SVG stroke default to top level of SVG component

kmaid:


packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -1,18 +1,31 @@
 import React, { ComponentProps } from 'react';
+import css from '@emotion/css';
 
 import { TextField } from '../atoms';
 import { searchIcon } from '../icons';
+import { lead, pine } from '../colors';
+
+const containerStyles = css({
```
jeysal, line 8: I think this was better placed in `TextField` than `SearchField`

* would need to know `TextField`s internal structure, and
* any `TextField` consumer can set a custom indicator, and the indicator will need to respond to `:focus`

jeysal (commented):
Nice, this seems like a good option :ok_hand:

packages/react-components/src/molecules/SearchField.tsx
kmaid: I put it here because some of the indicators (Show and Hide) need to not respond to :focus

jeysal: Right, in fact regardless on focus they must not have `stroke`. Hmm ...

jeysal: revert this and add `stroke="none"` to the show and hide `<text>` elements? :thinking: not super sleek...

PR #291, opened 2020-09-25
feat: first iteration on user pagination

fampinheiro:


jeysal:
I was gonna say isn't it a bad idea to introduce this without the frontend having a way to show the second page, but did I read correctly that this was kind of always the case, since we never returned more than 30 items in the first place?

fampinheiro:
You are correct, we were already returning partial results.

packages/services-common/test/squidex/squidex.test.ts
```diff
@@ -90,7 +91,7 @@ describe('squidex wrapper', () => {
 
     const client = new Squidex<Content>(collection);
     const result = await client.fetch();
-    expect(result).toEqual({ items: [] });
+    expect(result).toEqual({ total: 0, items: [] });
     expect(nock.isDone()).toBeTruthy();
```
johnytiago, line 95: `.toBe(true)` ? 😅

johnytiago (approved):

PR #295, opened 2020-09-28
feat: adds teams pagination, refactors tests

johnytiago:
This PR makes `cms/teams` obsolete for now.
Also, tests may fail while older versions of the codebase run simultaneous on the CI.
Once merged, no `teams` or `users` will be left on the CMS after testing.


apps/asap-server/src/controllers/teams.ts
```diff
@@ -37,28 +37,68 @@ const transformUser = (users: CMSUser[], teamId: string): TeamMember[] =>
...
+      skip: (page - 1) * pageSize,
+      sort: [{ path: 'data.displayName.iv' }],
+    });
+
     const teamUsers = await Promise.all(
-      teams.map((team) => this.cms.users.fetchByTeam(team.id)),
+      teams.map((team) =>
+        this.users
+          .fetch({
+            take: undefined,
```
fampinheiro, line 66: Do we know what is the default value for this?

apps/asap-server/src/controllers/teams.ts
```diff
@@ -37,28 +37,68 @@ const transformUser = (users: CMSUser[], teamId: string): TeamMember[] =>
...
+
+    return {
+      total,
+      items: teamItems,
+    };
   }
 
   async fetchById(teamId: string): Promise<TeamResponse> {
-    const [notFound, team] = await Intercept(this.cms.teams.fetchById(teamId));
+    const [notFound, team] = await Intercept(this.teams.fetchById(teamId));
```
fampinheiro, line 88: Not found is handled by the squidex wrapper.

```suggestion
    const team = await this.teams.fetchById(teamId);
```

apps/asap-server/src/controllers/teams.ts
```diff
@@ -37,28 +37,68 @@ const transformUser = (users: CMSUser[], teamId: string): TeamMember[] =>
...
     if (notFound) {
       throw Boom.notFound();
     }
```
fampinheiro, line 91:
```suggestion
```

apps/asap-server/src/controllers/teams.ts
johnytiago: It's 8, but in this case I don't want to limit it. I could potentially leave a big enough team size if we prefer

fampinheiro: Does this snippet does the same then?

```suggestion
```

johnytiago:
You can consider that the tests are passing at this point. 
Ready for merge

fampinheiro (approved):

PR #300, opened 2020-09-29
ci: preserve Jest cache

jeysal:


johnytiago (approved):

jeysal:
We'll see how long the master test run will take

jeysal:
Ideally we would use `policy: push` on master and `policy: pull` on branches, but that seems to require two jobs sharing things with a template, which would be very verbose. Let me know if you have any ideas @fampinheiro

PR #307, opened 2020-09-30
Feat: Mailto links 

kmaid:


packages/react-components/src/molecules/CtaCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  )}`,
+  paddingBottom: `${vminLinearCalcClamped(
+    mobileScreen,
+    18,
+    tabletScreen,
+    24,
+    'px',
+  )}`,
+  [`@media (min-width: ${smallDesktopScreen.min}px)`]: {
+    gridTemplateColumns: 'auto min-content',
```
jeysal, line 40:
```suggestion
    gridTemplateColumns: 'auto max-content',
```
I think our button probably has the same min-content as max-content because of `white-space: pre`, but if it ever doesn't because we need wrapping text in buttons, we should probably give this the space it wants

packages/react-components/src/molecules/CtaCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  flexGrow: 1,
+  display: 'flex',
+  justifyContent: 'center',
+  [`@media (min-width: ${tabletScreen.min}px)`]: {
+    display: 'block',
+  },
+});
+
+type CtaCardProps = {
+  href: string;
```
jeysal, line 54: `children: TextChildren` to restrict and because children should probably be mandatory for this component

packages/react-components/src/templates/ProfileHeader.tsx
```diff
@@ -102,9 +104,9 @@ const ProfileHeader: React.FC<ProfileProps> = ({
       </section>
       <section css={actionsStyles}>
         <div css={contactStyles}>
-          <Button small primary>
+          <Link small buttonStyle primary href={`mailto:${email}`}>
```
jeysal, line 107: Does the email need to be URL encoded? Asking, not sure myself

packages/react-components/src/molecules/CtaCard.tsx
jeysal: oh nvm, not text children since we need more in it, but still mandatory

jeysal (changes requested):

packages/react-components/src/templates/ProfileHeader.tsx
jeysal: Looked up, `?=&` need to be escaped apparently, and email addresses can in theory contain those, so yeah we probably should. Can make a helper that builds a `mailto` link or so

fampinheiro:
image

I wild padding appears.

jeysal (approved):
I've pushed a commit to improve the `createMailTo` test so that it doesn't rely on 'magic strings' in `toEqual` expected values. You can TAL and if you agree with those tests merge.
(Also pushed a fix for a place where you forgot to use `createMailTo`)

PR #308, opened 2020-09-30
feat: adds team email

johnytiago:


packages/model/src/team.ts
```diff
@@ -13,6 +13,7 @@ export interface TeamMember {
   firstName?: string;
   lastName?: string;
   displayName: string;
+  email: string;
```
fampinheiro, line 16: Do we need to export users emails?

johnytiago: Not mandatory. Yet it simplifies a lot the logic on the controller. It is a field that we might need there sooner or later.

fampinheiro (approved):

PR #312, opened 2020-10-01
fix: amend orcid works

fampinheiro:


packages/react-components/src/organisms/ProfileRecentWorks.tsx
```diff
@@ -55,49 +59,65 @@ const typeMap: { [key in OrcidWork['type']]: string } = {
 };
 
 const containerStyles = css({
-  display: 'flex',
-  flexDirection: 'row',
-  flexWrap: 'wrap',
-  marginLeft: '-8px',
```
jeysal, line 61: You're lucky I never came across this :axe:

packages/react-components/src/organisms/ProfileRecentWorks.tsx
```diff
@@ -55,49 +59,65 @@ const typeMap: { [key in OrcidWork['type']]: string } = {
...
-  marginRight: '-8px',
+  display: 'grid',
+  gridRowGap: `${12 / perRem}em`,
 });
 
-const elementStyle = css({
-  paddingLeft: '8px',
-  paddingRight: '8px',
+const elementStyle = css({});
```
jeysal, line 66: ..

jeysal (approved):
coverage increase <3 :D

packages/react-components/src/organisms/ProfileRecentWorks.tsx
fampinheiro: If it works it ain't stupid.

packages/react-components/src/organisms/ProfileRecentWorks.tsx
fampinheiro: Where does that come from :thinking:

PR #314, opened 2020-10-01
feat(api): change page to take skip

fampinheiro:


johnytiago:
👏🏽 For the very descriptive branch name. We should do this more often and completely dish out the PR description. What do you think?

johnytiago (approved):

PR #315, opened 2020-10-01
feat: search by team name

johnytiago:


apps/asap-server/src/controllers/teams.ts
```diff
@@ -39,40 +39,48 @@ const transformUser = (users: CMSUser[], teamId: string): TeamMember[] =>
...
+    const { search, ...opts } = options;
     const { total, items: teams } = await this.teams.fetch({
-      ...options,
+      ...opts,
+      ...(search
+        ? {
+            fullText: search,
+            filter: {
+              or: [
+                { path: 'data.displayName.iv', op: 'contains', value: search },
```
fampinheiro, line 78: To make the search more helpful we should tokenize `search` and filter on multiple strings.

johnytiago: You mean, splitting by whitespace and applying the same filters on all of those tokens?

PR #318, opened 2020-10-01
feat: redirect user to previous page on login

fampinheiro:


kmaid (approved):

apps/frontend/src/auth/Signin.tsx
```diff
@@ -6,9 +6,12 @@ const Home: React.FC<{}> = () => {
   const { loginWithRedirect } = useAuth0();
 
   const signin = () => {
-    // TODO pass a redirect_uri with a query param containing the current page;
-    // and send the user back to that original page in handleRedirectCallback
-    return loginWithRedirect({ prompt: 'login' });
+    return loginWithRedirect({
+      prompt: 'login',
+      appState: {
+        targetUrl: window.location.pathname + window.location.search,
```
jeysal, line 12: Any reason why this includes only `pathname` and `search`?

jeysal (changes requested):
Test please :upside_down_face: `prompt` is already being tested, so the fundamental test setup is there

apps/frontend/src/auth/Signin.tsx
fampinheiro: The redirect URL expects a pathname on the same origin.
https://github.com/yldio/asap-hub/blob/master/apps/frontend/src/auth/AuthProvider.tsx#L16

jeysal: Yeah it seems like an absolute URL would not be supported, but there seems to be no reason why `hash` should be left out

PR #319, opened 2020-10-02
feat: Generate URLS for Auth pages

kmaid:


apps/auth-frontend/src/login/Login.tsx
```diff
@@ -7,6 +7,16 @@ import {
...
 import { extractErrorMessage, WebAuthError } from '../auth0/errors';
 
+const getAppUrl = () => {
+  const redirectUri = new URL(
+    new URLSearchParams(window.location.search).get('redirect_uri') ||
+      'https://hub.asap.science/',
+  );
+  return `${redirectUri.protocol}//${redirectUri.hostname}${
+    redirectUri.port ? `:${redirectUri.port}` : ''
+  }`;
```
fampinheiro, line 17:
```suggestion
  return redirectUri.origin;
```
I believe `origin` already returns this.

apps/auth-frontend/src/login/Login.tsx
```diff
@@ -7,6 +7,16 @@ import {
 } from '../auth0/web-auth';
 import { extractErrorMessage, WebAuthError } from '../auth0/errors';
 
+const getAppUrl = () => {
```
fampinheiro, line 10: We can move this to `config`. 
We can treat it as a constant, no need to keep invoking a function.

jeysal: I don't think so, it reads the current `window.location.search`

apps/auth-frontend/src/login/Login.tsx
jeysal: This will also allow removing some test cases that 'test that specific parts of the URL are there' :+1:

apps/auth-frontend/src/login/Login.tsx
```diff
@@ -7,6 +7,16 @@ import {
...
+  const redirectUri = new URL(
+    new URLSearchParams(window.location.search).get('redirect_uri') ||
```
jeysal, line 12: I think we should throw if `redirect_uri` is not there, to avoid having to hardcode any URL here.
While it is optional in OAuth2, doesn't mean that our flow needs to make it optional, and we already know without a `redirect_uri` the flow will fail because our `add-user-metadata` Auth0 hook throws if it's missing.

apps/auth-frontend/src/login/Login.tsx
fampinheiro: I'm assuming the `auth-frontend` "entrypoint" needs the `redirect_uri` when loaded.
Haven't blocked the review because this isn't something i have strong opinions about :+1:

jeysal: Yeah sorry I noticed that after writing, so yeah it should work as a `constant`

jeysal: The problem is that it's very annoying to test. The only way is dynamically loading modules so that you have the opportunity to change `location` before loading the module. This is why I would pref reverting it

fampinheiro: Let's park this idea then :slightly_smiling_face:

apps/auth-frontend/src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  Object.defineProperty(globalThis, 'location', {
+    configurable: true,
+    enumerable: true,
+    value: originalLocation,
+  });
+});
+it('to get origin from redirect URI', () => {
+  mockLocation.mockReturnValue(
+    new URL(
+      `?redirect_uri=${encodeURIComponent('http://google.com')}`,
```
jeysal, line 24:
```suggestion
      `?redirect_uri=${encodeURIComponent('http://google.com/page')}`,
```
to test that it uses `origin`, not `href`

jeysal (approved):

fampinheiro (approved):

PR #321, opened 2020-10-02
feat: news and events page

fampinheiro:


packages/react-components/src/atoms/Paragraph.tsx
```diff
@@ -14,23 +14,31 @@ const primaryStyles = css({
   lineHeight: `${24 / 18}em`,
 });
 
+const smallStyles = css({
```
jeysal, line 17: This is not in the design spec o.O are you sure it's not `Caption` (13.6)

packages/react-components/src/atoms/Datetime.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { format } from 'date-fns';
+
+interface DatetimeProps {
+  readonly date: Date;
+}
+
+const Datetime: React.FC<DatetimeProps> = ({ date }) => {
```
jeysal, line 8: This should probably be just a function? I could see this causing problems with components that expect text as children

packages/react-components/src/templates/NewsAndEventsPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+} from '../pixels';
+import { pearl } from '../colors';
+import NewsAndEventsPageHeader from './NewsAndEventsPageHeader';
+import { contentSidePaddingWithNavigation } from '../layout';
+
+const mainStyles = css({
+  alignSelf: 'stretch',
+  background: pearl.rgb,
+  display: 'grid',
+  gridRowGap: `${vminLinearCalcClamped(
```
jeysal, line 18: I wouldn't overdo it with `Clamped`, only use if necessary. I was hesitant to add it because of this, it's more for ensuring things don't go into negative range. Extremely large screen (well beyond large desktop) warrants extreme numbers.

packages/react-components/src/atoms/Paragraph.tsx
fampinheiro: Maybe it is :) checking.

packages/react-components/src/templates/NewsAndEventsPage.tsx
fampinheiro: We probably want to address this in a separate PR since this is being used on every listing?

kmaid: This was picked up by James in an earlier PR. The issue is that the jump is quite large from mobile to tablet so it was very large at large desktop size.

jeysal: Ohh I'm sorry, thought it was a mobile - large desktop scale

apps/frontend/src/news/__tests__/NewsAndEvents.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    nock.cleanAll();
+    nockInterceptor = nock(API_BASE_URL, {
+      reqheaders: {
+        authorization: 'Bearer token',
+        'content-type': 'application/json',
+      },
+    }).get('/news-and-events');
+  });
+
+  // ('renders page error', async () => {
```
jeysal, line 66: leftovers

babel.config.js
```diff
@@ -5,7 +5,10 @@ module.exports = {
   ],
   plugins: [
     [require.resolve('@babel/plugin-transform-runtime'), { corejs: 3 }],
-    [require.resolve('babel-plugin-transform-inline-environment-variables')],
+    [
+      require.resolve('babel-plugin-transform-inline-environment-variables'),
+      { include: ['AUTH0_DOMAIN', 'AUTH0_CLIENT_ID'] },
```
jeysal, line 10: Curious what problem this caused? :upside_down_face:

packages/react-components/src/images/events-placeholder.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
```
jeysal, line 1: SVG belongs in `icons` (raw files in `images`)

jeysal: (other one as well)

apps/frontend/src/news/__tests__/NewsAndEvents.test.tsx
fampinheiro: This is something I want to discuss. Can't have the commented test and the other running. 
I know there is an `it` missing before the test name, CI was complaining :)

babel.config.js
fampinheiro: I was testing dates and one approach was to set process.env.TZ to UTC. 
`babel-plugin-transform-inline-environment-variables` was throwing an error. Not needed anymore.

packages/react-components/src/images/events-placeholder.tsx
fampinheiro: It was my first approach, but it's not an icon 😛.

jeysal: A distinction based on visual content is pretty arbitrary though, what the `Icon`/`Image` suffix is trying to achieve is that you know what you've got when you `import { *Image, *Icon } from '@asap-hub/react-components'`

fampinheiro: Was just messing with you 🙂

apps/frontend/src/news/__tests__/NewsAndEvents.test.tsx
jeysal: Is it because of the weird nock behavior?

packages/react-components/src/templates/NewsAndEventsPageBody.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface LibraryPageBodyProps {
+  readonly newsAndEvents: ReadonlyArray<
+    ComponentProps<typeof NewsAndEventsCard>
+  >;
+}
+
+const NewsAndEventsPageBody: React.FC<LibraryPageBodyProps> = ({
+  newsAndEvents,
+}) => (
+  <>
```
jeysal, line 14: This will need to change slightly due to the `CardList` changes, sorry :/
Check e.g. `LibraryPageBody` - basically the cards are now wrapped in a `CardList`, and paddings/gridGaps that used to be in `NewsAndEventsPage` can also go away because they are now handled by `CardList`

jeysal (commented):
@kmaid you should probably also take a look at this just so you're aware what changes are happening and whether there's any conflicts with your work.
Should @johnytiago look at the backend changes?

apps/asap-server/src/utils/squidex.ts
```diff
@@ -0,0 +1,17 @@
...
+import { parse, format } from 'date-fns';
+import { cms } from '../config';
+
+export const createURL = (assets: string[]): string[] => {
+  return assets.map((asset) =>
+    new URL(`/api/assets/${cms.appName}/${asset}`, cms.baseUrl).toString(),
+  );
+};
+
+export const parseDate = (date: string): Date => {
```
kmaid, line 11: Isn't this just ISO date?

apps/frontend/src/news/__tests__/NewsAndEvents.test.tsx
fampinheiro: Yes it is

apps/asap-server/src/utils/squidex.ts
fampinheiro: Squidex doesn't store miliseconds https://support.squidex.io/t/postponed-support-iso-8601-missing-miliseconds/2191

packages/react-components/src/templates/NewsAndEventsPageBody.tsx
fampinheiro: No worries, thanks for pointing it out.

apps/asap-server/src/utils/squidex.ts
jeysal: milliseconds are optional in ISO, so when *parsing* it should work? Only when serializing we should not pretend we have more information (`.000`) than we have

apps/frontend/src/news/__tests__/NewsAndEvents.test.tsx
jeysal: I guess skip the error testing for now? :/

jeysal: idk if you had any ideas how to make it work

apps/asap-server/src/utils/squidex.ts
fampinheiro: Our problem is more on sending data to squidex, we get a 400.

packages/react-components/src/templates/NewsAndEventsPageBody.tsx
```diff
@@ -0,0 +1,31 @@
...
+const NewsAndEventsPageBody: React.FC<LibraryPageBodyProps> = ({
+  newsAndEvents,
+}) => (
+  <CardList
+    numberOfPages={1}
+    numberOfItems={newsAndEvents.length}
+    currentPageIndex={0}
+    renderPageHref={() => ''}
+  >
+    {newsAndEvents.map((newsOrEvent, idx) => {
```
jeysal, line 21: This should use the actual `id`, better key than index

jeysal (commented):
From a frontend perspective LGTM except for one minor change (could have noticed that in a previous review, sorry for round trip)

johnytiago (approved):
Let's address this chores on another PR 😉

PR #322, opened 2020-10-02
feat(react-components): mostly standardize card paddings

jeysal:
and remove naming / story inconsistencies in the profile cards


apps/storybook/src/ProfileSkills.stories.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
 import { array } from '@storybook/addon-knobs';
-import { SkillsSection } from '@asap-hub/react-components';
+import { ProfileSkills } from '@asap-hub/react-components';
```
fampinheiro, line 3: Not sure about this name change. This component is used on Profile and Teams.

jeysal: Hmm, I wasn't aware of that.
What do you think of wording everything that is specific to users `UserProfile`, everything that is specific to teams `TeamProfile` and everything that is for both `Profile`?

jeysal: (in another PR ofc)

jeysal:
rebased

fampinheiro (approved):

jeysal:
patch diff check complains about a line in `TeamAbout` where only a variable name change happened, ignoring

PR #326, opened 2020-10-02
refactor: serverless-webpack => serverless-plugin-tree-shake

fampinheiro:


apps/asap-server/tsconfig.json
```diff
@@ -3,7 +3,9 @@
   "compilerOptions": {
     "rootDir": ".",
     "resolveJsonModule": true,
-    "outDir": "build"
+    "outDir": "build",
+    "module": "commonjs",
```
jeysal, line 7: Does the bundler use actual TSC to output files? o.O

jeysal: Wouldn't it be better and faster to still bundle from `build`? We've already done the TS=>JS transpilation (and downleveling to current Node version instead of hardcoded es2019)?

jeysal (approved):
seems to still work
@fampinheiro good to merge if you are

PR #327, opened 2020-10-02
feat: pearl background

fampinheiro:


jeysal (changes requested):
This is interesting. I've thought about it a bit and disagree with the approach - I think the main content area should get `pearl` and the default should remain `paper`. This is because
1. `theme.ts` is supposed to be imported by small components that want to render e.g. a `grey` part of the page. While this is happening in very few places so far, if you want `light`, you want `charcoal` on `paper`, there is almost never actual `charcoal` text on `pearl`
2. Pages that do not conform to the typical layout, say the `SigninPage`, seem to use `paper` as a background - really the main content area on the typical layout is the only thing that uses `pearl`

fampinheiro:
The main problem this pull request targets is to ensure main pearl areas shorter than browser height leak body background color. 

We can move the pearl from the theme to global styles to avoid the use case you describe in 1. I reckon there may be other ways to do this and I'm open for suggestions :)

jeysal:
Hmm, I can't quite imagine the problem with applying it to main content background that you're describing. I'll try to reproduce

packages/react-components/src/atoms/Card.tsx
```diff
@@ -9,7 +9,7 @@ import { paddingStyles } from '../card';
 export type AccentVariant = 'default' | 'red' | 'green';
 
 export const accents: Record<AccentVariant, string> = {
-  default: `border-color: ${colors.silver.rgb};`,
+  default: `background-color: ${colors.paper.rgb}; border-color: ${colors.silver.rgb};`,
```
jeysal, line 12: Slightly unfortunate that `Card` loses the ability to be used in a `grey` or `dark` context. Maybe I'll come up with a solution at some point.
Either way:
```js
import {light} from '../theme'
default: `${light} border-color: ...`
```
so that both background-color and color from the light theme are applied, otherwise things get ugly in a dark context (paper on paper)

packages/react-components/src/templates/Layout.tsx
```diff
@@ -27,7 +27,7 @@ const styles = css({
 
 const headerStyles = css({
   gridArea: 'header',
-
+  backgroundColor: paper.rgb,
```
jeysal, line 30: Shouldn't be needed

packages/react-components/src/templates/Layout.tsx
```diff
@@ -38,6 +38,8 @@ const headerMenuShownStyles = css({
 });
 
 const contentStyles = css({
+  backgroundColor: pearl.rgb,
```
jeysal, line 41: Please move below the positioning styles to inner styles. I'm not usually this pedantic, but this layout would be extremely hard to grasp without
order

packages/react-components/src/templates/Layout.tsx
```diff
@@ -69,6 +71,8 @@ const userButtonStyles = css({
     display: 'none',
   },
 
+  backgroundColor: paper.rgb,
```
jeysal, line 74: All areas on the left and top in the layout seem to be `paper` so this is not needed
(At least in the original designs like [this](https://projects.invisionapp.com/d/main#/console/20054351/425905082/preview), some other page designs since then have various inconcistencies

fampinheiro: This is the changes I had when the body was pearl, need to remove/adjust.

jeysal (approved):

PR #328, opened 2020-10-02
fix: link members on the header to valid page

fampinheiro:


jeysal (approved):
We should start doing some higher level testing that actually navigates around the app. Could be e2e (probably post MVP), or could be in JSDOM with request mocking

PR #330, opened 2020-10-05
Feat: Search Controls & Labeled Checkbox

kmaid:


apps/storybook/src/CheckboxGroup.stories.tsx
```diff
@@ -18,20 +18,23 @@ export const Normal = () => (
...
+          Luton: 'LTN',
+          City: 'LCY',
+          Southend: 'SEN',
+        },
+        ['LHR'],
+        {
+          display: 'inline-check',
+        },
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ) as any
```
kmaid, line 37: optionsKnob returns a const type when it should be an array of const. Taking suggestions or ill take another look at the end of the ticket

apps/storybook/src/SearchControls.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  title: 'Organisms / Search Controls',
+  component: SearchControls,
+};
+
+export const Normal = () => (
+  <SearchControls
+    placeholder={text('Placeholder', 'Search for a protein, a method…')}
+    onChangeSearch={() => action('Search')}
```
jeysal, line 15: Can we go with either `search` or `query` as the terminology, or even `searchQuery` I guess?

packages/react-components/src/organisms/CheckboxGroup.tsx
```diff
@@ -0,0 +1,40 @@
...
+export interface Option<V extends string> {
+  value: V;
+  label: string;
+  disabled?: boolean;
+}
+
+interface CheckboxGroupProps<V extends string> {
+  readonly options: ReadonlyArray<Option<V>>;
+
+  readonly values?: V[];
```
jeysal, line 16:
```suggestion
  readonly values?: ReadonlyArray<V>;
```

packages/react-components/src/atoms/Checkbox.tsx
```diff
@@ -0,0 +1,75 @@
...
+      borderColor: pine.rgb,
+      backgroundColor: pine.rgb,
+    },
+  },
+});
+
+interface CheckboxProps {
+  readonly id?: string;
+  readonly groupName: string;
+  readonly disabled?: boolean;
```
jeysal, line 50: can we please invert this to `enabled` (default `true`) for consistency with `Button` and `TextField`

packages/react-components/src/organisms/__tests__/SearchControls.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  expect(getByRole('textbox')).toHaveValue('test123');
+});
+
+it('shows and hides the dropdown menu', () => {
+  const { getByRole, getByText } = render(
+    <SearchControls {...props} query="test123" filterTitle="Filter Dropdown" />,
+  );
+  const filterButton = getByRole('button');
+  userEvent.click(filterButton);
+  expect(getByText('Filter Dropdown')).toBeVisible();
```
jeysal, line 32: hide missing at the end, only testing show so far.

packages/react-components/src/organisms/__tests__/SearchControls.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+});
```
jeysal, line 33: Menu auto-close on filter change would be good to test.
Slightly odd behavior though - maybe first check if this is actually a requirement, it seems like it could be considered annoying rather than helpful

packages/react-components/src/templates/NetworkPageHeader.tsx
```diff
@@ -81,19 +66,28 @@ const NetworkPageHeader: React.FC<NetworkPageHeaderProps> = ({
...
+          onChangeSearch={onChangeSearch}
+          placeholder={
+            page === 'users' ? 'Search for someone…' : 'Search for a team…'
+          }
+          query={query}
+          filterEnabled={page === 'users'}
+          filterOptions={
+            page === 'users'
+              ? [
+                  { label: 'Lead PI', value: '' },
```
jeysal, line 79: @fampinheiro just tagging so you are aware since this looks like it's gonna be an interesting architecture challenge :D

jeysal (changes requested):
Wow this was a lot, but haven't found much compared to the sheer size of the diff :+1: :sweat_smile:

packages/react-components/src/templates/NetworkPageHeader.tsx
kmaid: I could pass this in from Frontend?

jeysal: Not sure that improves anything - it's okay to rely on the model here, it's just that we have a hardcoded role list in many places

apps/storybook/src/SearchControls.stories.tsx
kmaid: This changes allot of files. Mind if i do it in my next PR?

packages/react-components/src/organisms/__tests__/SearchControls.test.tsx
kmaid: This does not hide automatically. I am just testing the show hide dropdown button.

packages/react-components/src/atoms/Checkbox.tsx
```diff
@@ -0,0 +1,75 @@
...
+  marginBottom: `${12 / perRem}em`,
+
+  appearance: 'none',
+  outline: 'none',
+  borderRadius: 0,
+  borderStyle: 'solid',
+  borderWidth: `${1 / perRem}em`,
+  borderColor: steel.rgb,
+
+  ':enabled:hover, :focus': {
```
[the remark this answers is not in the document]
kmaid: Ive gone back to James' design document and it works as intended as far as i can see

packages/react-components/src/organisms/__tests__/SearchControls.test.tsx
jeysal: Then what is `filterOptions` doing in the `useEffect` hook dependencies?

packages/react-components/src/atoms/Checkbox.tsx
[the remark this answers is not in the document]
jeysal: I don't doubt that it works as per spec, I'm complaining about the `:focus` here being dead because the `borderColor` on `:focus` is being overwritten further down anyway

apps/storybook/src/SearchControls.stories.tsx
jeysal: ok

packages/react-components/src/atoms/Checkbox.tsx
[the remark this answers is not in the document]
jeysal: I would suspect that the intention is
```suggestion
  ':enabled:hover, :enabled:focus': {
```
though?

PR #332, opened 2020-10-06
feat: default to research tab on profile

fampinheiro:


apps/frontend/src/network/__tests__/Profile.test.tsx
```diff
@@ -74,9 +74,9 @@ describe('without team', () => {
     expect((await findByText(user.displayName)).tagName).toBe('H1');
   });
 
-  it('renders the about member content', async () => {
+  it('renders the research member content', async () => {
```
jeysal, line 77: :joy:

packages/react-components/src/atoms/Avatar.tsx
```diff
@@ -63,7 +63,7 @@ const circleStyle = css({
   borderRadius: '50%',
 });
 const imageStyle = css({
-  objectFit: 'contain',
+  objectFit: 'cover',
```
jeysal, line 66: I get that borders on the pic aren't nice, but showing only half of a face seems *much* worse to me :/

fampinheiro: Both of them aren't perfect. I spoke with James and we rather have a perfect circle than a perfect avatar, and we can always upload a better picture.

jeysal (approved):

PR #335, opened 2020-10-07
feat(react-components): page controls

jeysal:


packages/react-components/src/icons/PreviousPage_Inactive.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fampinheiro, line 1: Is it intentional to keep this as an svg?

jeysal: oops, will delete

babel.config.js
```diff
@@ -5,7 +5,11 @@ module.exports = {
   ],
   plugins: [
     [require.resolve('@babel/plugin-transform-runtime'), { corejs: 3 }],
-    [require.resolve('babel-plugin-transform-inline-environment-variables')],
+
+    require.resolve('babel-plugin-transform-inline-environment-variables'),
+
+    require.resolve('babel-plugin-lodash'),
```
kmaid, line 11: Is this still required?

jeysal: I included it because why not, even the Serverless deployment doesn't need to be bigger than it has to :sweat:

kmaid (approved):

jeysal:
restarting Jest in hopes that codecov check will update

jeysal:
yaaaay it did

jeysal:
although it has no clue about diff with master
but whatever, the component has a trillion branches and 100% of it (at least the component itself) is covered, so this is definitely an overall increase

PR #344, opened 2020-10-09
Feat: Search integration

kmaid:


apps/frontend/src/api/util.ts
```diff
@@ -43,3 +47,26 @@ export const useFetchOptions = ({
...
+  searchQuery?: string;
+  filters?: string[];
+}
+
+export const useApiGet = <T>(
+  endpoint: string,
+  parameters: Record<string, string | string[] | undefined> = {},
+  options?: Parameters<typeof useFetchOptions>[0],
+) => {
+  const url = new URL(`${API_BASE_URL}${endpoint}`);
```
jeysal, line 61:
```suggestion
  const url = new URL(endpoint, API_BASE_URL);
```

apps/frontend/src/api/util.ts
```diff
@@ -43,3 +47,26 @@ export const useFetchOptions = ({
...
+
+export interface BasicOptions {
+  searchQuery?: string;
+  filters?: string[];
+}
+
+export const useApiGet = <T>(
+  endpoint: string,
+  parameters: Record<string, string | string[] | undefined> = {},
+  options?: Parameters<typeof useFetchOptions>[0],
```
jeysal, line 59: So this is `UseFetchOptions & RequestInit`? :sweat_smile:

apps/frontend/src/api/util.ts
```diff
@@ -43,3 +47,26 @@ export const useFetchOptions = ({
...
+) => {
+  const url = new URL(`${API_BASE_URL}${endpoint}`);
+  Object.entries(parameters).forEach(([key, value]) => {
+    if (Array.isArray(value)) {
+      value.forEach((item) => url.searchParams.append(key, item));
+    } else if (value) {
```
jeysal, line 65:
```suggestion
    } else if (value !== undefined) {
```
should we ever want `?x=` :D

apps/frontend/src/library/Routes.tsx
```diff
@@ -16,9 +17,22 @@ const Library: React.FC<{}> = () => {
...
+  const searchQuery = currentUrlParams.get('searchQuery') || '';
+  const [searchQueryDebounce] = useDebounce(searchQuery, 500);
 
   const onChangeSearch = (newQuery: string) => {
-    currentUrlParams.set('query', newQuery);
+    currentUrlParams.set('searchQuery', newQuery);
+    history.replace({ search: currentUrlParams.toString() });
+  };
+  const onChangeFilter = (filter: string) => {
+    currentUrlParams.delete('filter');
```
jeysal, line 29: should `filter` be a `Set`?

jeysal: looks like it could make this handling here easier (condition-less)

apps/frontend/src/network/Routes.tsx
```diff
@@ -19,12 +20,25 @@ const Network: React.FC<{}> = () => {
...
   const currentUrlParams = new URLSearchParams(search);
-
+  let filters = currentUrlParams.getAll('filter');
+  const searchQuery = currentUrlParams.get('searchQuery') || '';
+  const [searchQueryDebounce] = useDebounce(searchQuery, 400);
   const onChangeToggle = (pathname: string) => () => {
+    currentUrlParams.delete('filter');
     history.push({ pathname, search: currentUrlParams.toString() });
   };
   const onChangeSearch = (newQuery: string) => {
```
jeysal, line 30: These things may also want to move into a hook. But can look at that later when pagination is also in the mix

apps/frontend/src/api/util.ts
kmaid: I did this initially but there is /development on the API_BASE_URL

apps/frontend/src/api/util.ts
kmaid: Yes

apps/frontend/src/api/util.ts
jeysal: Okay, but then the problem is really that `endpoint` is absolute.
Can we change the paths so that my solution effectively does `new URL('users', 'https://host/development/')`? (which works)

apps/frontend/src/api/util.ts
jeysal: Or wait, this behavior seems to be used by the components atm. This util should probably still not erase empty string values, seems cleaner if the components just don't pass empty string values in the first place

apps/frontend/src/library/Routes.tsx
kmaid: Set doesn't work with arrays 👎

jeysal: No I mean `filters` being a Set

jeysal: convert when retrieving from SearchParams, and convert back when writing to SearchParams

apps/frontend/src/network/Routes.tsx
kmaid: Yeah, I agree but maybe not this PR 😅

apps/frontend/src/network/Routes.tsx
```diff
@@ -19,12 +20,23 @@ const Network: React.FC<{}> = () => {
   const { search } = useLocation();
   const history = useHistory();
   const currentUrlParams = new URLSearchParams(search);
-
+  const filters = new Set(currentUrlParams.getAll('filter'));
+  const searchQuery = currentUrlParams.get('searchQuery') || undefined;
+  const [searchQueryDebounce] = useDebounce(searchQuery, 400);
```
jeysal, line 25:
```suggestion
  const [searchQueryDebounce] = useDebounce(searchQuery, 500);
```
lol @ two different numbers :D

apps/frontend/src/api/__tests__/hooks.test.tsx
```diff
@@ -0,0 +1,54 @@
+import useFetch from 'use-http';
+import { useFetchOptions } from '../util';
+
+import { useApiGet } from '../hooks';
+
+jest.mock('use-http');
```
jeysal, line 6: not a big fan of this mock-heavy testing :/ this e.g. misses deleting the `return ` from the hook :D
I was hoping for a `@testing-library/react-hooks` test.
I think this may be a good topic to pair on on Monday, we haven't actually done before! :sweat_smile:

jeysal (approved):

PR #345, opened 2020-10-09
fix: allow filters to be a string & add tests

johnytiago:


apps/asap-server/test/handlers/users/fetch.test.ts
```diff
@@ -99,6 +99,7 @@ describe('GET /users', () => {
         httpMethod: 'get',
         queryStringParameters: {
           search: 'first last',
+          filter: 'Lead PI',
```
kmaid, line 102: Should you test for multiple and none?

apps/asap-server/src/handlers/users/fetch.ts
```diff
@@ -19,8 +19,8 @@ export const handler: Handler = lambda.http(
     const query = lambda.validate('query', request.query, querySchema) as {
       take: number;
       skip: number;
-      search: string;
-      filter: string[];
+      search?: string;
+      filter?: string[] | string;
```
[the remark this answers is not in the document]
fampinheiro: You can use stripUnkown option on the schema to remove filter before implementing it.
We can discuss how we can upgrade serverless to payload v2.0 next week.

apps/asap-server/test/handlers/users/fetch.test.ts
fampinheiro: I'm assuming implementation comes later, this is to ensure a `filter` on `qs` doesn't give `400`.

fampinheiro (approved):

PR #346, opened 2020-10-09
perf(frontend): lazy load (almost) all the things

jeysal:


jeysal:
This will be way better once we have a proper loading animation :D

jeysal:
[profile.zip](https://github.com/yldio/asap-hub/files/5357114/profile.zip)

Chromium perf profiles before and after.
Loading now happens very incrementally, e.g. Layout paint moved from 12.6s in to 7.2s in.
I've built in prefetching so that the key navigations shouldn't take longer either.

kmaid (approved):

jeysal:
rebased

PR #347, opened 2020-10-09
build: parallelize more build steps

jeysal:


jeysal:
brings absolutely zero benefit

PR #348, opened 2020-10-10
fix: squidex query for auth0 integration

fampinheiro:


.gitlab-ci.yml
```diff
@@ -100,10 +100,10 @@ test:format:
   script:
     - yarn run lint:format
 
-test:jest:
+test:unit:
   <<: *tmpl_test
   artifacts:
-    expire_in: 1 day
+    expire_in: 5 day
```
fampinheiro, line 106: Artifact expire date needs to be greater than `auto_stop`

PR #357, opened 2020-10-15
feat: adds user filters

johnytiago:


apps/asap-server/test/handlers/users/fetch.test.ts
```diff
@@ -52,44 +50,10 @@ describe('GET /users', () => {
...
-                value: 'last',
-              },
-            ],
-          },
-          sort: [{ path: 'data.displayName.iv' }],
-        }),
+        $top: 8,
+        $orderby: 'data/displayName/iv',
+        $filter:
+          "data/teams/iv/role eq 'Lead PI' and (contains(data/displayName/iv, 'first') or contains(data/firstName/iv, 'first') or contains(data/displayName/iv, 'last') or contains(data/firstName/iv, 'last'))",
```
fampinheiro, line 56: Can we keep this under 80 char long?

apps/asap-server/test/handlers/users/fetch.test.ts
```diff
@@ -98,7 +62,9 @@ describe('GET /users', () => {
         httpMethod: 'get',
         queryStringParameters: {
           search: 'first last',
-          filter: 'Lead PI',
```
fampinheiro, line 101: We can keep this to test if multiValue overrides singular value

packages/services-common/src/framework/lambda.ts
```diff
@@ -79,10 +79,14 @@ export const http = <T>(fn: (request: Request) => Promise<Response>) => async (
...
-    headers,
+    headers: { ...headers, ...event.multiValueHeaders },
     params: event.pathParameters,
     payload: body,
-    query: event.queryStringParameters || {},
+    query:
+      {
+        ...event.queryStringParameters,
+        ...event.multiValueQueryStringParameters,
+      } || {},
```
fampinheiro, line 89:
```suggestion
      },
```

Since we are already creating an object no need to fallback.

apps/asap-server/test/handlers/users/fetch.test.ts
johnytiago: Not really, but why is that? I thought we had seen that the max query string length was 1024 chars. This on right here has ~200 chars

fampinheiro: 80 char long line, to be more readable 😛

apps/asap-server/src/controllers/users.ts
```diff
@@ -153,29 +153,44 @@ export default class Users {
...
+      .split(' ')
+      .filter(Boolean) // removes whitespaces
+      .reduce(
+        (acc: string, word: string) =>
+          acc.concat(
+            `contains(data/displayName/iv, '${word}') or contains(data/firstName/iv, '${word}') or `,
+          ),
+        '',
+      )
+      .slice(0, -4); // remove last ' or '
```
fampinheiro, line 170: We could concat to an array and use `.join(' or ')`.

apps/asap-server/src/controllers/users.ts
```diff
@@ -153,29 +153,44 @@ export default class Users {
...
+
+    const filterQ = (filter || [])
+      .reduce(
+        (acc: string, word: string) =>
+          acc.concat(`data/teams/iv/role eq '${word}' and `),
+        '',
+      )
+      .slice(0, -5); // remove last ' and '
```
fampinheiro, line 178: We could concat to an array and use `.join(' and ')`.

apps/asap-server/src/controllers/users.ts
johnytiago: Hell yeah! Sound even better! 😏

fampinheiro (approved):

PR #364, opened 2020-10-16
feat: remove middle name and add degree

fampinheiro:


jeysal (approved):
How does something like this work with Squidex btw? Does it automatically delete the removed field on all the data? What if the new field is mandatory?

fampinheiro:
If you remove, it gets removed, no questions asked.
If it's mandatory, we have two options:

1. Have a default value
2. Create a script to change the property onevery entity. (this probably needs to be a two step deploy)

PR #365, opened 2020-10-19
feat: discover asap

fampinheiro:


apps/frontend/src/api/discover.ts
```diff
@@ -0,0 +1,8 @@
+import useFetch from 'use-http';
+import { DiscoverResponse } from '@asap-hub/model';
+
+import { API_BASE_URL } from '../config';
+import { useFetchOptions } from './util';
+
+export const useDiscover = () =>
+  useFetch<DiscoverResponse>(`${API_BASE_URL}/discover`, useFetchOptions(), []);
```
jeysal, line 8: Please do not `useFetch` directly anymore, instead `useGetList` or (I assume in this case) `useGetOne`. See structure of one of the other components that `useGetOne` for how to consume.
Side note: I love using the `use` that is part of a hook's name as the verb for a sentence ^^

apps/frontend/src/api/dashboard.ts
```diff
@@ -1,12 +1,5 @@
...
-import { useFetchOptions } from './util';
+import { useGetOne } from './get-one';
 
-export const useDashboard = () =>
-  useFetch<DashboardResponse>(
-    `${API_BASE_URL}/dashboard`,
-    useFetchOptions(),
-    [],
-  );
+export const useDashboard = () => useGetOne<DashboardResponse>(`dashboard`);
```
jeysal, line 5: Thanks for changing it here as well! We should also adapt the FE `Dashboard` component to use the new structure of `GetOne`-type components. But doesn't need to be in this PR and doesn't need to be you who does it

apps/frontend/src/discover/__tests__/Discover.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+});
+
+test('renders dicover with guidance, about and members', async () => {
+  nock(API_BASE_URL, {
+    reqheaders: { authorization: 'Bearer token' },
+  })
+    .get('/discover')
+    .once()
+    .reply(200, {
+      aboutUs: '<h1>About us</h1>',
```
jeysal, line 54: Now that Tom has started with `@asap-hub/fixtures`, would it be possible to source this example data there (while of course object spreading and inserting the data points relevant to the test in here)?

jeysal: They would also add typing (whether in fixtures or not), which is helpful for future schema changes

apps/frontend/src/api/dashboard.ts
jeysal: Nvm, you did that as well! Thanks :tada:

apps/frontend/src/discover/__tests__/Discover.test.tsx
jeysal: We already have a lot of data spread over many tests and stories that will take a long time to centralize, so would appreciate not adding more data types with example data all over the place, that's why I'm asking to do this straight away from now on :sweat_smile:

jeysal: Looks like the Discover story has much more example data for this data type, we should probably use that for the fixture :)

packages/react-components/src/organisms/MainNavigation.tsx
```diff
@@ -43,13 +43,16 @@ const listStyles = css({
 });
 
 export interface MainNavigationProps {
-  networkHref: string;
+  discoverASAPHref: string;
```
jeysal, line 46: Should we name this
```suggestion
  discoverAsapHref: string;
```
? That's how I (and I thought most people?) have been treating acronyms in camel case, so that it's easy to identify as one word inside the identifier. @kmaid how have you been doing this?

packages/react-components/src/icons/index.ts
```diff
@@ -2,6 +2,7 @@ export { default as chevronCircleLeftIcon } from './chevron-circle-left';
 export { default as chevronDownIcon } from './chevron-down';
 export { default as chevronUpIcon } from './chevron-up';
 export { default as crossIcon } from './cross';
+export { default as discoverIcon } from './discover';
```
jeysal, line 5: Please add to icon story

packages/react-components/src/organisms/MembersSection.tsx
```diff
@@ -45,12 +45,15 @@ const roleStyles = css({
 });
 
 type MembersSectionProps = {
-  readonly members: TeamMember[];
+  readonly title: string;
+  readonly members: (Omit<TeamMember, 'role'> & {
+    role: string;
```
jeysal, line 50: What is happening here, this doesn't look right - is the component not able to deal with the specific subtype of string for some reason?

packages/react-components/src/templates/DiscoverPageBody.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import { PagesSection, RichText, MembersSection } from '../organisms';
+import { perRem } from '../pixels';
+import { Display, Card, Paragraph, Link } from '../atoms';
+import { CtaCard } from '../molecules';
+
+const styles = css({
+  display: 'grid',
+  gridRowGap: `${72 / perRem}em`,
+  marginBottom: `${24 / perRem}em`,
```
jeysal, line 12: I think we usually use `padding` for those (margins are usually not what you want regarding collapsing, background, ...)

jeysal: idk if there's a specific reason here, otherwise
```suggestion
  paddingBottom: `${24 / perRem}em`,
```

packages/react-components/src/templates/DiscoverPageBody.tsx
```diff
@@ -0,0 +1,67 @@
...
+});
+
+const containerStyles = css({
+  marginTop: `${24 / perRem}em`,
```
jeysal, line 16:
```suggestion
  paddingTop: `${24 / perRem}em`,
```

packages/react-components/src/templates/DiscoverPageBody.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Display, Card, Paragraph, Link } from '../atoms';
+import { CtaCard } from '../molecules';
+
+const styles = css({
+  display: 'grid',
+  gridRowGap: `${72 / perRem}em`,
+  marginBottom: `${24 / perRem}em`,
+});
+
+const containerStyles = css({
```
jeysal, line 15: this is `aboutUsStyles` from what I understand?

packages/react-components/src/templates/__tests__/DiscoverPageBody.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    {
+      path: '/',
+      title: 'Page 2 Title',
+      text: 'Page 2 Text',
+    },
+  ],
+  members: [],
+};
+
+it('renders multiple page cards', () => {
```
jeysal, line 23: These test cases confuse me a bit, could you go through them and make sure the assertions are exactly what the tests titles say and the other way around :sweat_smile:

jeysal (changes requested):
I was very chatty lol, I think I need to go through and resolve comments that aren't actionable :joy:

packages/react-components/src/organisms/MembersSection.tsx
fampinheiro: The role on user.teams limits to a closed list. The role on the discover page isn't on that list.
This property change removes the constraints on the presentation part (on members component) to accommodate the hardcoded role on discover.

packages/react-components/src/organisms/MembersSection.tsx
```diff
@@ -45,12 +45,15 @@ const roleStyles = css({
 });
 
 type MembersSectionProps = {
-  readonly members: TeamMember[];
+  readonly title: string;
```
jeysal, line 48: Hmm, this duplicates the `Team Members` in other places. How would you feel about having this optional, defaulting to `Team Members (members.length)`?

apps/frontend/src/discover/__tests__/Discover.test.tsx
fampinheiro: I'm changing the page card in a follow up PR. I'll do this work on that PR :+1: ?

jeysal: Okay, yeah. It might actually reduce the size of this PR by avoiding duplication, but not sure. Make sure to catch all places where you've added this type of data though :sweat_smile: 
And can you add the Response type annotation here (`{...} as BlaResponse`) so it's safe at least

packages/react-components/src/organisms/MembersSection.tsx
fampinheiro: I thought about it, no strong feelings about it.

packages/react-components/src/organisms/MainNavigation.tsx
kmaid: Same

apps/frontend/src/discover/__tests__/Discover.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      members: [],
+      pages: [],
+      aboutUs: '',
+    });
+
+  const { getByText } = await renderHome();
+  expect(getByText(/discover/i, { selector: 'h1' })).toBeVisible();
+});
+
+test('renders dicover with guidance, about and members', async () => {
```
johnytiago, line 47:
```suggestion
test('renders discover with guidance, about and members', async () => {
```

packages/model/src/discover.ts
```diff
@@ -0,0 +1,8 @@
+import { PageResponse } from './page';
+import { UserResponse } from './user';
+
+export interface DiscoverResponse {
+  pages: ReadonlyArray<PageResponse>;
```
johnytiago, line 5: Why do we call this pages for the FE but guidance for the BE?

apps/asap-server/test/handlers/discover/fetch.test.ts
```diff
@@ -0,0 +1,128 @@
...
+describe('GET /discover', () => {
+  beforeAll(() => {
+    identity();
+  });
+
+  afterEach(() => {
+    expect(nock.isDone()).toBe(true);
+  });
+
+  test('returns 200 when no inforamtion exists', async () => {
```
johnytiago, line 21:
```suggestion
  test('returns 200 when no information exists', async () => {
```

johnytiago (approved):

packages/model/src/discover.ts
fampinheiro: I created guidance first on the CMS and on the frontend the component is expecting pages. I'm changing the CMS property name.

PR #367, opened 2020-10-19
Feat: Asap Support Emails

kmaid:


packages/react-components/src/atoms/__tests__/NavigationLink.test.tsx
```diff
@@ -74,4 +74,15 @@ describe('with a router', () => {
     );
     expect(fireEvent.click(getByRole('link'))).toBe(false);
   });
+
+  it('triggers a full page navigation on click of an external link', () => {
+    const { getByRole } = render(
+      <StaticRouter>
+        <NavigationLink href="http://test.com/" icon={<svg />}>
```
jeysal, line 81: FYI `example.com` (and some others) are the [official example domains](https://en.wikipedia.org/wiki/Example.com). Let's prefer those over domains that somebody actually uses (already are in most places).
```suggestion
        <NavigationLink href="http://example.com/" icon={<svg />}>
```

packages/react-components/src/utils.ts
```diff
@@ -22,7 +22,20 @@ export const getSvgAspectRatio = (element: React.ReactElement): number => {
...
-  `mailto:${email.split('@').map(encodeURIComponent).join('@')}`;
+export const createMailTo = (
+  email: string,
+  { subject, body }: { subject?: string; body?: string } = {},
+): string => {
+  const mailtoParams = new URLSearchParams();
+  if (subject) mailtoParams.set('subject', subject);
+  if (body) mailtoParams.set('body', body);
+
+  return `mailto:${email.split('@').map(encodeURIComponent).join('@')}${
```
jeysal, line 33: Can we add a test for specifying both `body` and `subject` and preferably use the `URL` class instead of string concatenation to build this link now that it's becoming more complex (I can't tell with certainty that this works for all cases anymore)

jeysal (changes requested):
Mostly LGTM, small comments, other than that I would like to see `createMailTo` more tested and preferably avoiding string concatenation

apps/frontend/src/home/__tests__/Home.test.tsx
```diff
@@ -100,7 +100,7 @@ test('renders dashboard with correct links', async () => {
     '/news-and-events',
```
jeysal, line 100: need to work on the quality of frontend tests, this array is bullshit :see_no_evil: but not for now

packages/react-components/src/utils.ts
```diff
@@ -22,7 +22,21 @@ export const getSvgAspectRatio = (element: React.ReactElement): number => {
...
+  if (subject) mailTo.searchParams.set('subject', subject);
+  if (body) mailTo.searchParams.set('body', body);
+
+  return mailTo.toString();
+};
 
 export const formatDate = (date: Date): string => format(date, 'do MMMM yyyy');
+
+export const isInternalLink = (href: string): boolean =>
+  new URL(href ?? '', window.location.href).origin === window.location.origin;
```
jeysal, line 42:
```suggestion
  new URL(href, window.location.href).origin === window.location.origin;
```
string is mandatory

jeysal (approved):

PR #369, opened 2020-10-20
feat: updating payload v2

johnytiago:


apps/asap-server/test/handlers/users/fetch.test.ts
```diff
@@ -100,8 +98,8 @@ describe('GET /users', () => {
           Authorization: `Bearer token`,
         },
         queryStringParameters: {
-          take: 8,
-          skip: 8,
+          take: '8',
```
fampinheiro, line 101: Why a string?

johnytiago: QueryStringParameters will never be a string https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L107

fampinheiro: If Joi parses it as number I don't see a problem :+1:

johnytiago: These are not the queryStringParameters the FE sends us. This are the ones that are presented to the lambda framework after AWS processes them. These will always be strings and it's up to Joi to then convert them to numbers then. Them not being strings is actually wrong.

fampinheiro (approved):
Assuming the patch works with serverless-offline :)

PR #370, opened 2020-10-20
feat: create page cards

fampinheiro:
On this PR:
- Normalize the usage of `link`, `linkText` and `shortText` on pages and news.
- Make page card closer to the designs

**Note**: The drop on coverage is tied to the reuse of avatar parser on REST and GraphQL queries. This will be addressed in a future change.


jeysal:
Could you spread `reply({...page, someText: 'something'})` in the tests, otherwise it's pretty magic where the `/text/i` comes from. Other than that LGTM.
Oh and maybe check what is not covered by tests, looks like `asap-server` changes

PR #376, opened 2020-10-21
feat: Rich text improvements

kmaid:


fampinheiro (approved):

packages/react-components/src/text.ts
```diff
@@ -31,6 +39,28 @@ export function assertIsTextChildren(
   }
 }
 
+const isAllowedElement = (child: unknown): child is AllowedChild => {
```
jeysal, line 42: These *definitely* need tests :grimacing:

packages/react-components/src/text.ts
```diff
@@ -10,19 +11,26 @@ import {
 
 export type TextChild = React.ReactText | boolean | null | undefined;
 export type TextChildren = TextChild | ReadonlyArray<TextChild>;
+export type AllowedChild = React.ReactElement<HTMLStyleElement>;
```
jeysal, line 14: `TextChild` should be included in here already, not only in `AllowedChildren` (otherwise it's not simply the plural but actually different)

jeysal (changes requested):

packages/react-components/src/organisms/__tests__/RichText.test.tsx
```diff
@@ -52,6 +60,24 @@ it('Displays error when styling applied to h2', () => {
...
 
+it('passes through image props', () => {
+  const { getByRole } = render(
+    <RichText text={'<img data-test="test123" alt="alt text"></img>'} />,
+  );
+  expect(getByRole('img')).toHaveAttribute('data-test', 'test123');
+  expect(getByRole('img')).toHaveAttribute('alt', 'alt text');
+});
+
+it('sets blank alt when omitted', () => {
```
jeysal, line 71: Does this actually happen? :scream: 
I hoped the rich text editor would mandate alt text like MD does.
`alt=""` signals that an image is purely decorative, which we can't really guarantee

kmaid: This was to keep ES Lint happy

jeysal: Yeah I figured, but doesn't answer the question. If it's just to keep ESLint happy and not supposed to happen (i.e. Squidex only ever allows images with alt text), we could throw if it is missing.

jeysal: [we discussed this in call]

packages/react-components/src/__tests__/text.test.ts
```diff
@@ -46,3 +50,40 @@ describe.each([
...
+
+describe.each([
+  [{}],
+  [Symbol('Forbidden primitive type')],
+  [createElement('div')],
+  [createElement('b')],
+  [createElement('i', { children: [createElement('div')] })],
+  [createElement(React.Fragment, { children: {} })],
+])('for the children %p', (children) => {
+  describe('isTextChildren', () => {
```
jeysal, line 84:
```suggestion
  describe('isAllowedChildren', () => {
```

jeysal (approved):

PR #378, opened 2020-10-21
feat: add avatar to discover

fampinheiro:


jeysal (approved):
:+1: esp. for the 99% of the PR that is renaming :sweat_smile:

PR #380, opened 2020-10-21
ci: try bundlewatch instead of bundlesize

jeysal:
bundlesize is dead and has bugs like 'diff against master' not working anymore at all


jeysal:
looks ... better, so far

PR #383, opened 2020-10-21
fix: search field losing focus on key press

jeysal:
I noticed that when using `react-error-boundary@2`
(without https://github.com/bvaughn/react-error-boundary/pull/68),
and fixing my oversights from the error boundary PR,
they work properly with the intended `resetKeys` prop instead of `key`,
and we thus avoid the tree unmount issue that `key` introduced.


fampinheiro (approved):

jeysal:
Will file a bug report to `react-error-boundary`

PR #385, opened 2020-10-21
perf(frontend): prefetch more sections

jeysal:
And prefetch sequentially. This means that on slower networks (where
prefetching matters most, and the bandwidth is so constrained that
fetching more bundles at once does not increase overall throughput), the
first sections will complete prefetching earlier because they do not
share bandwidth with other bundles being prefetched. The order should
reflect how likely we think users are to navigate to each section.

### DIFF
--- a/apps/frontend/src/App.tsx
+++ b/apps/frontend/src/App.tsx
@@ -37,9 +37,11 @@ const Discover = React.lazy(loadDiscover);
 
 const Prefetch: React.FC<{}> = () => {
   useEffect(() => {
-    loadNetwork();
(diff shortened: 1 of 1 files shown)


fampinheiro (approved):

fampinheiro (approved):

jeysal:
Wow @fampinheiro you must really like this to double approve it :joy: 
I'm thinking about whether this is a good opportunity to split up `App` and get some tests for our fundamental app structure in now that I'm reminded by codecov...

fampinheiro (approved):

PR #386, opened 2020-10-22
feat: adjust header to design

fampinheiro:


jeysal:
Check https://386.hub.asap.science/.storybook/?path=/story/pages-team--about-tab on mobile landscape and tablet landscape - `Last updated` wraps which is not ideal. Other than that LGTM :sweat_smile:

PR #389, opened 2020-10-23
fix(network-users): Co-Investigator filter

kmaid:


apps/asap-server/src/controllers/teams.ts
```diff
@@ -19,6 +19,9 @@ const priorities: Record<TeamRole, number> = {
   'Project Manager': 3,
   Collaborator: 4,
   'Key Personnel': 5,
+  Advisor: 6,
+  Guest: 7,
+  Staff: 8,
```
kmaid, line 24: Took a guess at this

packages/model/src/research-output.ts
```diff
@@ -1,7 +1,15 @@
 import { TeamResponse } from './team';
 import { ListResponse } from './common';
 
-export type ResearchOutputType = 'proposal';
+export type ResearchOutputType =
+  | 'proposal'
```
kmaid, line 5: @fampinheiro You said this should match the values shown in squidex which i assume to have proper casing

fampinheiro: On squidex we have proper case :). Can you change it to Proposal?

apps/asap-server/src/controllers/teams.ts
fampinheiro: I works, it's not 100% accurate but we can adjust in the future 🙂.

jeysal (approved):

PR #390, opened 2020-10-23
fix: search results race condition

jeysal:


kmaid (approved):

jeysal:
added tests

kmaid (approved):

jeysal:
Seems no one thinks it doesn't work, and I can't manage to break it, so merging :sweat_smile:

PR #392, opened 2020-10-23
feat: revert react-components bundle and keep things simple

fampinheiro:


jeysal:
What's the problem this solves? :sweat_smile: because it does not seem simpler ^^

fampinheiro:
The rendering automation of emails wasn't working. I spent a day trying to come up with an automated solution and couldn't find one. So this pull request was born, there is no automation creating the templates, just automation pushing them to aws.

jeysal:
How did we generate the email previously?

fampinheiro:
It worked on a subset of the components we had. Somewhere during the iterations of the components we introduced a rule that makes the bundled code throw an exception when loading it to inline the css styles.

PR #393, opened 2020-10-24
feat: training hidden news-and-events

fampinheiro:


jeysal:
@fampinheiro conflicts a low in case you hadn't noticed. Will leave review up to @johnytiago didn't see any problems myself

apps/asap-server/src/controllers/news-and-events.ts
```diff
@@ -16,6 +16,11 @@ export default class ResearchOutputs {
   }): Promise<ListNewsAndEventsResponse> {
     const { total, items } = await this.newsAndEvents.fetch({
       ...options,
+      filter: {
+        path: 'data.type.iv',
+        op: 'ne',
+        value: 'Training',
+      },
```
johnytiago, line 23: Why don't we make an `or` here with the data types we want to fetch?

johnytiago (approved):

apps/asap-server/src/controllers/news-and-events.ts
fampinheiro: News and events doesn't filter by type for now. This filter just ensures we don't see Training entries.

PR #398, opened 2020-10-26
feat: filters should be or not and

fampinheiro:


johnytiago:
If the filter is a checkbox, I would assume you want all the filters to match. That is how must online shops work at least

jeysal:
Sooner or later we will need more advanced filtering anyway, but for roles, this definitely makes more sense. Being a Project Manager AND Researcher is a very rare case that requires being on two teams in the first place, which is already going to be rare, so filtering for AND does not make much sense for roles at least.

kmaid (approved):

PR #399, opened 2020-10-26
feat: use job and institution on asap member

fampinheiro:


jeysal (commented):
Add typings `reply(200, {} as Something)` please, otherwise schema modifications cause hard to track down problems. Other than that LGTM

PR #400, opened 2020-10-26
fix: team header Avatar scaling

jeysal:
I've changed the team page to have enough space available.

Avatars can now scale according to the available space.

I hadn't considered the difficulty scaling the initials,
but with a simple SVG it's reasonably easy,
and the autoscaling has e.g. the advantage of being able to
fill exactly one row with Avatars, as is done in the new team header.

I've also refactored the team header's extra users bubble
into the Avatar component to share styling with the real Avatar bubbles.

### DIFF
--- a/apps/storybook/src/Avatar.stories.tsx
+++ b/apps/storybook/src/Avatar.stories.tsx
@@ -9,7 +9,6 @@ export default {
 export const Normal = () => (
   <Avatar
     border={boolean('Border', false)}
-    small={boolean('Small', false)}
     imageUrl={text(
       'Image URL',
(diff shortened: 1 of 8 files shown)


fampinheiro:
It works. James will most certainly complain about the size of the avatars.
We can address that later.

jeysal:
> James will most certainly complain about the size of the avatars.

Yeah the scaling itself (and thus exact numbers deviating) I'll just point out that it's better as it can fill the row exactly. The max width we're flexible on.

jeysal:
and actually I'll change the maximum from 64 to 60 real quick, that's at least a multiple of our 6 design constant :D

PR #401, opened 2020-10-26
add graphQL fetch user

johnytiago:


fampinheiro (approved):
So much to improve but we are getting there :+1:.

PR #402, opened 2020-10-27
fix: team href on discover page

fampinheiro:


jeysal:
Any chance we can make this handle team href arrays? This would be the first place to re-introduce lack of multi-team support in the frontend afaik. The text could include team name only if there are multiple, so that for the standard case we retain the current copy. (If merging is urgent can defer for a follow-up)

fampinheiro:
Nice observation, this PR is to fix a broken link.
We can create a follow-up PR to address multiple teams (need to check, how we would differentiate them).

jeysal:
Yeah, differentiating would probably work like in `UserNavigation`:
```ts
  readonly teams: ReadonlyArray<{ name: string; href: string }>;
```

PR #404, opened 2020-10-27
fix: change case on avatar url

fampinheiro:


apps/asap-cli/package.json
```diff
@@ -5,6 +5,8 @@
   "bin": {
     "asap-import": "build/cli.js"
   },
+  "module": "build/cli.js",
```
jeysal, line 8: The module doesn't actually export anything so I think it's correct without?

PR #405, opened 2020-10-27
fix(avatar): overflow on small avatars

jeysal:
Background images are just so much easier to deal with,
and now images are pretty much guaranteed to
behave like the fallback initials, since it's the same element...


jeysal:
idk if images now having a background color while loading would be considered a problem, I'll change it to be sure for 100% behavior as was before :shrug:

jeysal:
@fampinheiro lmk if you find any problems

fampinheiro (approved):

PR #407, opened 2020-10-27
feat: auth0 id_token authentication

johnytiago:


apps/asap-server/src/utils/validate-user.ts
```diff
@@ -23,18 +19,7 @@ export default async function validateUser(
...
-      },
-    }).json<auth0.UserInfo>(),
-  );
-
-  if (error) {
-    throw Boom.forbidden('Forbidden', { error });
-  }
-
-  return res;
+  return decodeToken(token).catch(() => {
```
fampinheiro, line 22: Can we use the same pattern as above? 
`const [error, res] = await Intercept(`

apps/asap-server/jest.setup.ts
```diff
@@ -0,0 +1,28 @@
+jest.mock('@asap-hub/auth', () => ({
+  decodeToken: jest.fn().mockResolvedValue({
```
jeysal, line 2: We already have a default mock for `@asap-hub/auth` with some things defined in `packages/auth/src/__mocks__/index.ts`. Can we add this in there instead? Then we also probably don't need to configure a setup file anymore because using it becomes a matter of `jest.mock('@asap-hub/auth')`

apps/asap-server/src/utils/validate-user.ts
```diff
@@ -23,18 +19,7 @@ export default async function validateUser(
...
+    throw Boom.forbidden();
```
jeysal, line 23: Unauthorized is correct if the token is invalid - forbidden would be the token is valid but has insufficient privileges for the operation

apps/asap-server/test/handlers/research-outputs/fetch-by-id.test.ts
```diff
@@ -42,25 +41,9 @@ describe('GET /research-outputs/{id} - validations', () => {
...
-      }),
-    )) as APIGatewayProxyResult;
-
-    expect(result.statusCode).toStrictEqual(403);
-  });
-
-  test('returns 403 when Auth0 is unavailable', async () => {
-    nock(`https://${authConfig.domain}`).get('/userinfo').reply(500);
+    jest
+      .requireMock('@asap-hub/auth')
```
jeysal, line 45: I'd recommend against using this API - it'll return the mock no matter if it is actually active or not - if it's not then it's extremely confusing to track down why this is not doing anything.
I'd import `decodeToken` normally instead and then cast `const mockDecodeToken = decodeToken as jest.MockedFunction<typeof decodeToken>`

packages/auth/src/decode-token.ts
```diff
@@ -0,0 +1,20 @@
+import { default as jwt, JwtHeader } from 'jsonwebtoken';
+import { promisify } from 'util';
+import { Auth0User } from '.';
+
+import pubKeys from './pubKeys';
+
+const getPublicKey = (header: JwtHeader, cb: Function): string => {
+  const key = pubKeys.find(({ kid }) => header.kid);
+  return key ? cb(null, key) : cb(new Error());
```
jeysal, line 9: Error message please

packages/auth/src/pubKeys/index.ts
```diff
@@ -0,0 +1,17 @@
+import { keys as prodPubKeys } from './auth0-pubkeys.prod.json';
+import { keys as devPubKeys } from './auth0-pubkeys.dev.json';
+
+interface JWK {
```
jeysal, line 4: Just define them as TS instead of JSON I'd say, should save you from manually maintaining typings

jeysal (changes requested):
Would be happier to see the Auth0-recommended API approach but if we go for this here's so comments :sweat_smile:

packages/auth/src/pubKeys/index.ts
johnytiago: I would rather just have to `wget` the file if it needs to be updated. I want to keep it in the default format

jeysal: Hmm, but it's literally prepending `export default` isn't it, the formatter does the rest

johnytiago: Yes, I would rather not have to prepend anything. Plus, the type is simple enough that I don't mind defining it.

fampinheiro (approved):

packages/react-components/src/auth-test-utils.tsx
```diff
@@ -40,7 +40,9 @@ export const Auth0Provider: React.FC<{
         loading: !auth0Client,
         popupOpen: false,
         isAuthenticated: false,
-        getIdTokenClaims: notImplemented('getIdTokenClaims'),
+        getIdTokenClaims: auth0Client
+          ? (...args) => auth0Client.getIdTokenClaims(...args).then()
```
jeysal, line 44: `.then()`?

apps/asap-server/__mocks__/@asap-hub/auth/index.ts
```diff
@@ -0,0 +1,26 @@
+export const decodeToken = jest.fn().mockResolvedValue({
```
jeysal, line 1: any reason why this is not in the standard mock in `packages/auth` itself?

jeysal:
Phew, good that we have the bundlewatch check :see_no_evil: that would have probably been missed.
I assume this is from the `jsonwebtoken` package that is now included in the FE - adding `"sideEffects": false` (see `react-components/package.json`) should probably allow the bundler to kick this out

packages/react-components/src/auth-test-utils.tsx
johnytiago: I would suggest you hit `cmd + g` next time 😉

jeysal:
Code (without having looked at backend in detail ofc) LGTM, but Dev Auth0 is not working for me (and thu not for this PR either ofc), know what's going on?

jeysal (approved):

PR #410, opened 2020-10-28
feat: Team Internal page and Link Card

kmaid:
Team internal


fampinheiro (approved):

apps/frontend/src/network/team/__tests__/Routes.test.tsx
```diff
@@ -114,3 +105,29 @@ it('renders the not found page for a 404', async () => {
   const { getByText } = await renderTeam();
   expect(getByText(/sorry.+page/i)).toBeVisible();
 });
+
+describe('the workspace', () => {
+  it('does not render a link to workspaces when tools omitted', async () => {
+    interceptor.reply(200, createTeamResponse());
```
jeysal, line 111:
```suggestion
    interceptor.reply(200, createTeamResponse({ tools: undefined }));
```

packages/react-components/src/organisms/LinkCard.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import { Card, Headline3, Paragraph, Button } from '../atoms';
+import { noop } from '../utils';
+
+type LinkCardProps = {
+  readonly name: string;
+  readonly description: string;
+  readonly onClick?: () => void;
```
jeysal, line 9: I suspect that this would have an `href` instead? At least for modals I want to use routes for reason already stated. But can change it in the future

packages/react-components/src/templates/TeamHeader.tsx
```diff
@@ -135,8 +137,26 @@ const TeamHeader: React.FC<TeamProps> = ({
         )}
       </section>
       <TabNav>
-        <TabLink href={aboutHref}>About</TabLink>
-        <TabLink href={outputsHref}>Outputs</TabLink>
+        {tools
```
jeysal, line 140: Isn't this way simpler and doesn't need keys if we use `{tools && ...TabLink...}` just on the middle one?

packages/react-components/src/templates/__tests__/TeamWorkspace.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+it('renders the team workspace page', () => {
+  const team = createTeamResponse({ teamMembers: 1, tools: 0 });
+
+  const { getByRole } = render(<TeamWorkspace {...team} />);
+
+  expect(getByRole('heading').textContent).toEqual('Team Collaboration Tools');
+});
+
+const team = createTeamResponse({ teamMembers: 1, tools: 0 });
```
jeysal, line 15: Probably meant to be declared above and used in the first test as well

packages/react-components/src/utils.ts
```diff
@@ -22,21 +22,9 @@ export const getSvgAspectRatio = (element: React.ReactElement): number => {
...
-  );
-
-  if (subject) mailTo.searchParams.set('subject', subject);
-  if (body) mailTo.searchParams.set('body', body);
-
-  return mailTo.toString();
-};
-
 export const formatDate = (date: Date): string => format(date, 'do MMMM yyyy');
+export const formatDateAndTime = (date: Date): string =>
```
jeysal, line 26: sneaky ^^

jeysal (changes requested):
Nice work, couple small things, and pushing some probably uncontroversial things in a sec :sweat_smile:

packages/react-components/src/templates/TeamHeader.tsx
kmaid: Tab Nav will render a li for undefined :(

jeysal: Can we fix that in `TabNav` using
```
  ':empty': {
    display: 'none',
  },
```

packages/react-components/src/organisms/LinkCard.tsx
kmaid: Yeah, I'll do it in the modal PR 👍

packages/react-components/src/templates/TeamHeader.tsx
fampinheiro: I'd rather have the element not present than hidden.

jeysal: But this is IMO a bug in `TabNav` and should not be worked around here. Inserting `{false}`, `{null}`, `{<></>}` etc. at arbitrary positions in JSX should not change behavior. Alternative fixes would be more complicated, like iterating over children checking whether they are 'empty'. Also note that `display: none` elements are ignored by screen readers.

kmaid: I already filtered it in tabnav in my amends

jeysal: Did `:empty` not work? The reason why I didn't suggest filtering is exactly because it's a huge PITA, there's infinite cases due to arrays and fragments alone

kmaid: It works in this case and most cases I can think of. Can't we just iterate if it becomes a problem? To test this worked with the CSS solution we would have to do a browser test

jeysal (approved):

PR #413, opened 2020-10-29
feat: create staff profile

fampinheiro:


apps/frontend/src/network/profile/__tests__/Routes.test.tsx
```diff
@@ -144,4 +145,40 @@ describe('with team', () => {
...
+          ],
+          role: 'Staff',
+        } as UserResponse);
+
+      const { queryAllByRole, getByText } = await renderProfile('43');
+      const loadingIndicator = getByText(/loading/i);
+      await waitForElementToBeRemoved(loadingIndicator);
+      const links = (await queryAllByRole('link')) as HTMLAnchorElement[];
+      expect(links.map(({ href }) => href)).toMatchInlineSnapshot(`
+        Array [
```
jeysal, line 176: These kinds of tests are getting a bit out of hand in the FE, the test above you copied this from is quite possibly the worst test in the frontend.
I would suggest having a case for the render condition (that looks for some `'from ASAP'` rendered by `Staff`) and then a test that gets one of the links and checks its href

apps/storybook/src/ProfilePage.stories.tsx
```diff
@@ -75,8 +76,10 @@ const commonProps = () => ({
         year: '2015',
         month: '05',
       },
+      lastModifiedDate: '1478865224685',
```
jeysal, line 79: Is this not an ISO 8601 string?

jeysal (changes requested):

packages/react-components/src/templates/ProfileHeader.tsx
```diff
@@ -84,9 +92,24 @@ const ProfileHeader: React.FC<ProfileProps> = ({
   aboutHref,
   researchHref,
   outputsHref,
+  discoverHref,
+  role,
 }) => {
+  const teamsProps: ProfileProps['teams'] =
```
jeysal, line 98: Based on the conversation yesterday, I thought we would solve this by passing the `role` into `ProfilePersonalText` as well and letting it decide what to show for user with role `Staff`

apps/frontend/src/network/profile/Routes.tsx
```diff
@@ -16,9 +16,12 @@ const loadAbout = () =>
   import(/* webpackChunkName: "network-profile-about" */ './About');
 const loadOutputs = () =>
   import(/* webpackChunkName: "network-profile-outputs" */ './Outputs');
+const loadStaff = () =>
+  import(/* webpackChunkName: "network-profile-staff" */ './Staff');
 const Research = React.lazy(loadResearch);
 const About = React.lazy(loadAbout);
 const Outputs = React.lazy(loadOutputs);
+const Staff = React.lazy(loadStaff);
 loadResearch();
```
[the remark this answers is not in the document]
jeysal: @fampinheiro ^ what about this?
Doesn't need testing either, just perf optimization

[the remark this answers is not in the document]
fampinheiro: I'll create a separate PR for this and improve testing.

apps/asap-server/src/entities/user.ts
```diff
@@ -91,62 +65,98 @@ export const createSchema = Joi.object({
   connections: Joi.string(),
 });
 
-export const parseGraphQLUserTeamConnection = (
-  item: CMSGraphQLUserTeamConnection,
+export const parseGraphqlUserTeamConnection = (
```
johnytiago, line 68: Keep it consistent across functions and variables
```suggestion
export const parseGraphQLUserTeamConnection = (
```

johnytiago: Or go the other way around, always lowercase, but we should try to keep it consistent

apps/asap-server/src/entities/user.ts
```diff
@@ -92,61 +66,97 @@ export const createSchema = Joi.object({
...
-    orcidLastModifiedDate: user.data.orcidLastModifiedDate?.iv,
-    orcidWorks: user.data.orcidWorks?.iv,
-    skills: user.data.skills?.iv || [],
-    questions: user.data.questions?.iv.map(({ question }) => question) || [],
-    avatarUrl: user.data.avatar && createURL(user.data.avatar.iv)[0],
-  };
+  const teams: UserTeam[] =
+    user.data.teams?.iv.map(({ id, ...t }) => ({
+      id: id[0],
+      displayName: 'Unknown',
```
johnytiago, line 131: `undefined`?

fampinheiro: Not valid.

johnytiago (approved):

PR #418, opened 2020-11-02
build: use wildcard versions for workspace packages

jeysal:


jeysal:
Apparently this is a thing, learned from a random PR on Jest. Much simpler, and while totally not relevant here also learned that it's better in some wild sub-workspaces setups.

fampinheiro (approved):
What is the reasoning behind this change? Why is a wildcard better than specifying the path to the package.

jeysal:
For us it's literally just removing the overhead, and makes it easier to change directory structure because it now just looks at package names.
Not relevant for us: It helps when embedding a workspace in another, because the path would no longer be correct from the workspace root https://github.com/facebook/jest/issues/10761

fampinheiro:
Thanks. I was more interested in the second part of the explanation :+1:

PR #423, opened 2020-11-04
feat(profile-header): edit buttons

jeysal:
This is behind a 'feature flag' in that the props aren't passed yet


jeysal:
Breaks on small devices. I'll TAL
423 hub asap science_network_users_364b93d8-c153-4513-bad0-ebf6e57f711e_research(Galaxy S5)

jeysal:
This does change the width of the profile header slightly so that it does not match up with the width of the profile page body, but it doesn't look unintentional or weird or anything and the next edit buttons PR will make the body follow suit so IMO not a blocker

fampinheiro (approved):

PR #427, opened 2020-11-05
feat: change team roles

fampinheiro:


packages/squidex/schema/schemas/users.json
```diff
@@ -483,8 +495,11 @@
               "allowedValues": [
                 "Lead PI",
                 "Co-Investigator",
-                "Project Manager",
                 "Collaborator",
+                "Lead PI (Core Leadership)",
+                "Co-PI (Core Leadership)",
+                "Project Manager",
+                "Collaborating PI",
```
fampinheiro, line 502: Since we need to migrate the roles to the new designation I am keeping the previous ones on this PR.

jeysal (approved):
Could have just made Squidex contain technical `name`s and the frontend translates it to the `label`s I guess, that may avoid migrations and stuff

fampinheiro:
Thought about it @jeysal.
Although the names are also used be the people managing the CMS.

PR #432, opened 2020-11-06
feat: edit user profile

johnytiago:


apps/asap-server/src/entities/team.ts
```diff
@@ -11,7 +11,9 @@ export const updateSchema = Joi.object({
       }),
     )
     .required(),
-});
+})
+  .min(1)
```
fampinheiro, line 15: Users can remove every tool on the profile, the min should be 0.

apps/asap-server/src/entities/user.ts
```diff
@@ -51,9 +53,28 @@ export interface CMSUser {
...
+  orcid?: string;
+  department?: string;
+  institution?: string;
+  degree?: UserDegree;
+  location?: string;
+  skills?: string[];
+  skillsDescription?: string;
+  questions?: { question: string }[];
+  responsibilities?: string;
+  reachOut?: string;
```
fampinheiro, line 72:
```suggestion
```

Responsibilities and reach out aren't editable via the API

apps/asap-server/src/entities/team.ts
johnytiago: That would mean that tools is an empty array, no?

PR #433, opened 2020-11-06
Fix: Change link

kmaid:
This is a fix for Network not being highlighted. I spoke with James and their ideal would be network highlighted for all pages under network except team internal and user internal pages. This sounds like it needs specced out to understand how hilighting of navigation links will work on all pages post MVP so I will bring this up in standup on Monday


fampinheiro (approved):
We can have a better control with `isActive` property. https://reactrouter.com/web/api/NavLink/isactive-func

Let's merge this and evaluate if we need the `isActive` property after MVP.

PR #437, opened 2020-11-09
filter hidden profiles

fampinheiro:


johnytiago (approved):

apps/asap-server/src/controllers/users.ts
```diff
@@ -218,10 +220,12 @@ export default class Users {
       )
       .join(' or ');
 
-    const and = filter && search ? ['and (', ')'] : ['', ''];
+    const and = filterQ && searchQ ? ['and (', ')'] : ['', ''];
```
johnytiago, line 223: `searchQ` will always be true at this point

PR #440, opened 2020-11-09
Tools Persist

kmaid:


fampinheiro (approved):

apps/frontend/src/api/__tests__/util.test.tsx
```diff
@@ -24,17 +24,6 @@ describe('useFetchOptions', () => {
     <authTestUtils.LoggedIn user={undefined}>{children}</authTestUtils.LoggedIn>
   );
 
-  it('sets the default headers', async () => {
```
jeysal, line 27: Might still want to leave it in since I don't assume the test breaks now? I don't really trust `use-http` not to break this logic in an update

apps/frontend/src/network/team/Routes.tsx
```diff
@@ -98,6 +101,11 @@ const Team: React.FC<{}> = () => {
                         {...tool}
                         title="Edit Link"
                         backHref={join(url, 'workspace')}
+                        onSave={(data: TeamTool) =>
+                          patch({
+                            tools: Object.assign([], team.tools, { [i]: data }),
```
jeysal, line 106: Sick technique lol

packages/react-components/src/molecules/ModalEditHeader.tsx
```diff
@@ -36,7 +36,7 @@ const ModalEditHeader: React.FC<ModalEditHeaderProps> = ({
 }) => (
   <div css={headerStyles}>
     <div css={controlsContainerStyles}>
-      <Button onClick={onSave} primary small>
+      <Button onClick={onSave} primary small submit>
```
jeysal, line 39: This should be default for `primary` buttons already. Not saying it hurts to have it, just so you're aware

apps/frontend/src/network/team/Routes.tsx
```diff
@@ -87,6 +87,9 @@ const Team: React.FC<{}> = () => {
                     <ToolModal
                       title="Add Link"
                       backHref={join(url, 'workspace')}
+                      onSave={(data: TeamTool) =>
+                        patch({ tools: [...(team.tools ?? []), data] })
```
jeysal, line 91: In this case I suppose it doesn't make a difference since the patch is idempotent, but should we check if a req is already pending before sending one? Or will this be solved in UI with a loading state that disables the button or so?

jeysal (commented):
Nice :ok_hand: good work

packages/react-components/src/molecules/ModalEditHeader.tsx
kmaid: Removed

apps/frontend/src/network/team/Routes.tsx
kmaid: I will sort this out when i get to loading :)

PR #458, opened 2020-11-13
refactor: edit saving

jeysal:
* move navigation out of component library into frontend
* type the requests

### DIFF
--- a/packages/model/src/user.ts
+++ b/packages/model/src/user.ts
@@ -100,4 +100,21 @@ export interface UserResponse extends Invitee {
   role: Role;
 }
 
+export interface UserPatchRequest {
+  displayName?: string;
+  email?: string;
+  firstName?: string;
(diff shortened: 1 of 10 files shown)


packages/model/src/user.ts
```diff
@@ -100,4 +100,21 @@ export interface UserResponse extends Invitee {
   role: Role;
 }
 
+export interface UserPatchRequest {
```
kmaid, line 103: Couldn't this just be a partial?

jeysal: I wanted to, but it's for some reason very different that the other types, some fields aren't even of the same type, so I didn't dare and just copied it over

kmaid (approved):

jeysal:
Missing branch cov is `tools===undefined` which isn't really a use case afaik, idk why the typings even have it as optional

PR #463, opened 2020-11-17
feat: add contact email to user

fampinheiro:


jeysal:
(just rebased it)

apps/asap-server/src/handlers/users/update.ts
```diff
@@ -20,16 +20,16 @@ export const handler: Handler = lambda.http(
       id: string;
     };
 
+    if (user.id !== params.id) {
```
johnytiago, line 23: I don't really agree with changing the ordering here. We need to validate that the body schema is valid before doing the more advanced authorisations, like the one on #465

jeysal:
This needs a rebase, and is this otherwise good to go no @johnytiago? I need this in for my PR

PR #464, opened 2020-11-17
feat(textarea): higher by default on mobile

jeysal:


apps/storybook/src/ProfilePage.stories.tsx
```diff
@@ -150,7 +150,7 @@ export const AboutTabEditBiography = () => (
   </ProfilePage>
 );
 
-export const AboutTabPersonalInfoModal = () => (
+export const AboutTabEditPersonalInfo = () => (
```
jeysal, line 153: naming consistency with the other profile modal, which in turn was named after the first modal (team tools)

fampinheiro (approved):

jeysal:
pushed another thing after some clarifying design chat, unrelated, just to save time, will try to not squash when merging

PR #465, opened 2020-11-18
feat: add edit user teams

johnytiago:


packages/model/src/user.ts
```diff
@@ -115,6 +115,7 @@ export interface UserPatchRequest {
   skills?: string[];
   skillsDescription?: string;
   questions?: string[];
+  teams?: Omit<UserTeam, 'displayName' | 'role' | 'proposal'>[];
```
kmaid, line 118: Shouldn't this be a Pick to match the Joi Schema?

apps/asap-server/src/entities/user.ts
```diff
@@ -75,6 +75,15 @@ export const userUpdateSchema = Joi.object({
...
   skillsDescription: Joi.string().allow(''),
   questions: Joi.array().items(Joi.object({ question: Joi.string() })),
+  teams: Joi.array().items(
+    Joi.object({
+      id: Joi.string().required(),
+      responsibilities: Joi.string(),
+      approach: Joi.string(),
+    })
+      .min(2)
+      .required(),
```
kmaid, line 85: Will this mean we always have to provide teams in a patch request?

kmaid (approved):

apps/asap-server/src/entities/user.ts
johnytiago: No, just that the teams array has to have an object in it

packages/model/src/user.ts
johnytiago: Was having a TS error using Pick, I will paste it here

kmaid: `teams?: Pick<UserTeam, 'id' | 'approach' | 'responsibilities'>[];`
seemed to work

PR #469, opened 2020-11-19
refactor(frontend): split profile component

jeysal:
This should probably be reviewed commit by commit 😅

- refactor(frontend): rename network 'Routes' files
- refactor(frontend): split profile component
- refactor(frontend): replace useless tests and start safer routing


jeysal:
image
lol, conincidence...almost

apps/frontend/src/network/routes.ts
```diff
@@ -0,0 +1 @@
+export const TEAMS_PATH = 'teams';
```
jeysal, line 1: BTW I've only *started* with these, we can gradually use for more and more routes...

fampinheiro (approved):

PR #474, opened 2020-11-23
feat(user-profile): edit open questions

kmaid:


jeysal (approved):
LGTM! Only thing I would ask as a general note is mind the test specificity. `/question.+1/i` e.g. is perfectly sufficient usually, and matching texts too exactly in many places will leave us having to go into more tests to change this than necessary

PR #476, opened 2020-11-23
feat: increase tags and remove hover effect

fampinheiro:


packages/react-components/src/atoms/Tag.tsx
```diff
@@ -27,9 +27,9 @@ const styles = css({
   borderRadius: `${18 / perRem}em`,
 
   transition: '200ms',
-  ':hover': {
-    borderColor: charcoal.rgb,
-  },
+  // ':hover': {
```
jeysal, line 30: yaaaaay. I hated those, made it look like you're supposed to click ^^ why not remove it entirely though instead of commenting, incl. transition?

fampinheiro: I can remove it. I wanted to save time to my future self. This will come back.

jeysal (approved):

PR #484, opened 2020-11-24
feat: allow for more types on research outputs

fampinheiro:


packages/react-components/src/templates/ResearchOutputPage.tsx
```diff
@@ -50,7 +51,7 @@ const postedStyles = css({
 
 type ResearchOutputPageProps = Pick<
   ResearchOutputResponse,
-  'publishDate' | 'title' | 'type' | 'created' | 'team' | 'text'
+  'team' | 'text' | 'type' | 'created' | 'publishDate' | 'title' | 'type'
```
jeysal, line 54: Do we need `type` twice? :D

fampinheiro: It's just to make sure it's there :P

johnytiago (approved):

PR #493, opened 2020-11-27
feat: upload user avatar

johnytiago:


apps/asap-server/src/controllers/users.ts
```diff
@@ -284,6 +290,28 @@ export default class Users {
     return parseGraphQLUser(items[0]);
   }
 
+  async updateAvatar(id: string, avatar: Buffer): Promise<UserResponse> {
+    const form = new FormData();
+    form.append('file', avatar, {
+      filename: `${id}.jpg`,
+      contentType: 'image/jpeg',
```
kmaid, line 297: Are we enforcing only jpeg? I have allowed image/* in the frontend

apps/asap-server/test/handlers/users/update-avatar.test.ts
```diff
@@ -0,0 +1,266 @@
...
+import { handler } from '../../../src/handlers/users/update-avatar';
+import { apiGatewayEvent } from '../../helpers/events';
+import { identity } from '../../helpers/squidex';
+import { body } from './update-avatar.fixtures';
+import { patchResponse } from './update.fixtures';
+import decodeToken from '../../../src/utils/validate-token';
+
+jest.mock('../../../src/utils/validate-token');
+
+describe('PATCH /users/{id}/avatar - validations', () => {
```
kmaid, line 14: In the frontend we are using post as you have required in the handler

kmaid (changes requested):

apps/asap-server/test/handlers/users/update-avatar.test.ts
johnytiago: Yeah, this is just a copy & paste error. We don't look into the request method during tests.

apps/asap-server/src/controllers/users.ts
johnytiago: Can you pass the content type as an header? I don't want to be looking into the buffer to figure that out 😅

johnytiago: As discussed on zoom: the backend will implement a dataURI parser and extract the type from that.

PR #494, opened 2020-11-27
feat(textarea): disabled state

jeysal:


jeysal:
Note that this is not actually specced, but I need it to disable all fields in a modal. A specced version might also make the character count grey or something, but can do that if we ever get it specced.

kmaid (approved):

PR #495, opened 2020-11-27
feat(user-profile): modal saving states

jeysal:
Would have liked to manage the loading and disabling fields centrally in
`Modal`, but that would result in a weird structure like render props or
something and not really help a lot, because each individual Modal is
still responsible for passing the enabled prop to every single one of its fields,
unless we go completely crazy and make a context that
automatically disables all input fields below it or something.

The cancel state is actually not quite as unproblematic as we thought.
Yes, it's true that there is no instant at which a user can cancel the
edit saving and be sure that the edit was not yet performed,
and based on this, it seems like we do not have to care about
cancellation at all. However, the finished request triggers a navigation
to the corresponding profile tab, even if the user has meanwhile
navigated away from it.

This is huge, and can theoretically be split into react-components
first, frontend second. But the react-components changes are the same
thing over and over again, so maybe it's fine and I'll just do a
walkthrough.


apps/frontend/src/auth/state.ts
```diff
@@ -0,0 +1,20 @@
+import { atom, selector } from 'recoil';
+import { Auth0 } from '@asap-hub/auth';
+
+export const auth0State = atom<Auth0 | undefined>({
+  key: 'getIdToken',
```
jeysal, line 5: outdated name

apps/frontend/src/auth/state.ts
```diff
@@ -0,0 +1,20 @@
...
+  default: undefined,
+  dangerouslyAllowMutability: true,
```
jeysal, line 7: outdated

kmaid (approved):

jeysal:
We've decided that the inconvenient behavior in the navigation edge case isn't important enough to block merging this - aborting the request kind of goes into error handling territory (promise rejects) so that'll be tackled up next

PR #498, opened 2020-12-02
feat: add link to tags

fampinheiro:


jeysal:
Just a quick hint how this can perhaps become cleaner by allowing us to centralize the details of how the link is constructed in the FE:
Look at `PageControls` `renderPageHref` prop - maybe a function prop like this which generates the href for a tag would help here?

PR #499, opened 2020-12-02
feat: multiple improvements to looks

fampinheiro:


packages/react-components/src/templates/__tests__/TeamProfileHeader.test.tsx
```diff
@@ -92,7 +92,7 @@ it('renders tabs', () => {
   const { getAllByRole } = render(<TeamProfileHeader {...boilerplateProps} />);
   expect(getAllByRole('link').map(({ textContent }) => textContent)).toEqual([
     'About',
-    'Outputs',
+    'Team Outputs',
```
jeysal, line 95: ewwwww brittle tests
if you don't want these needing to change again, change to `expect.stringMatching(/output/i)` etc. :sweat_smile:

jeysal (approved):

packages/react-components/src/templates/__tests__/TeamProfileHeader.test.tsx
fampinheiro: :fire:

PR #501, opened 2020-12-03
Feat: Frontend Image Upload

kmaid:


packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -123,6 +123,14 @@ const lastModifiedNoContactStyles = css({
 const tabNavStyles = css({
   gridArea: 'tab-nav',
 });
+const avatarContainer = css({ display: 'grid', width: 90, height: 90 });
+const imageContainer = css({ gridRow: 1, gridColumn: '1 / span 2' });
```
jeysal, line 127: Does this actually need more than 1 row/column at all? it seems overlaying both in row1 col1 and then aligning and justifying should do the trick?

packages/react-components/src/templates/__tests__/UserProfileHeader.test.tsx
```diff
@@ -111,6 +111,29 @@ describe('an edit button', () => {
...
+      {...boilerplateProps}
+      onImageSelect={(file: File) => {}}
+    />,
+  );
+  expect(getByLabelText(/edit.+avatar/i)).toBeVisible();
+  expect(getByLabelText(/upload.+avatar/i)).not.toHaveAttribute('disabled');
+});
+
+it('is disabled for avatar (REGRESSION)', () => {
+  disable('EDIT_PROFILE_REST');
```
jeysal, line 126: We probably want a separate flag for this? `REST` is already active on prod and this shouldn't go live without error handling

packages/react-components/src/atoms/Button.tsx
```diff
@@ -46,6 +46,7 @@ interface LinkStyleButtonProps {
 }
 type ButtonProps = (NormalButtonProps | LinkStyleButtonProps) & {
   readonly submit?: boolean;
+  readonly label?: string;
```
jeysal, line 49: unused?

apps/frontend/src/network/users/UserProfile.tsx
```diff
@@ -85,6 +87,22 @@ const User: React.FC<{}> = () => {
         isOwnProfile && tab
           ? join(url, tab, EDIT_CONTACT_INFO_PATH)
           : undefined,
+      onImageSelect:
+        currentUser?.id === id && tab
+          ? (file: File) =>
```
jeysal, line 92: Can we extract the ugly bits of this into some sort of `fileToDataUrl` util please

apps/frontend/src/network/users/UserProfile.tsx
```diff
@@ -85,6 +87,22 @@ const User: React.FC<{}> = () => {
...
+                setAvatarSaving(true);
+                const reader = new FileReader();
+                reader.readAsDataURL(file);
+                reader.onload = async () => {
+                  if (reader.result) {
+                    resolve(await patchUserAvatar(reader.result.toString()));
+                  }
+                };
+                reader.onerror = (error) => reject(error);
+              }).finally(() => setAvatarSaving(false))
```
jeysal, line 103: The reset to enabled appears to not work on error

jeysal (changes requested):

packages/react-components/src/atoms/Button.tsx
kmaid: Whoops, Yeah. I made it work better with a label.

apps/frontend/src/network/users/__mocks__/api.ts
```diff
@@ -17,3 +21,13 @@ export const patchUser = jest.fn(
     };
   },
 );
+
+export const postUserAvatar = jest.fn(
+  async (id: string, post: UserAvatarPostRequest): Promise<UserResponse> => {
+    const user = await getUser(id);
+    return {
+      ...user,
+      avatarUrl: `url: ${post.avatar}`,
```
jeysal, line 30: Assuming `post.avatar` is a URL already?
```suggestion
      avatarUrl: post.avatar,
```

apps/frontend/src/network/users/__tests__/api.test.ts
```diff
@@ -80,3 +84,44 @@ describe('patchUser', () => {
     );
   });
 });
+
+describe('postUserAvatar', () => {
+  it('makes an authorized PATCH request for the user id', async () => {
```
jeysal, line 89:
```suggestion
  it('makes an authorized POST request for the user id', async () => {
```

apps/frontend/src/network/users/__tests__/api.test.ts
```diff
@@ -80,3 +84,44 @@ describe('patchUser', () => {
...
+    const post: UserAvatarPostRequest = { avatar: '123' };
+    nock(API_BASE_URL, { reqheaders: { authorization: 'Bearer x' } })
+      .post('/users/42/avatar')
+      .reply(200, {});
+
+    await postUserAvatar('42', post, 'Bearer x');
+    expect(nock.isDone()).toBe(true);
+  });
+
+  it('passes the patch object in the body', async () => {
```
jeysal, line 99:
```suggestion
  it('passes the post object in the body', async () => {
```

apps/frontend/src/network/users/api.ts
```diff
@@ -36,3 +44,24 @@ export const patchUser = async (
...
+    method: 'POST',
+    headers: {
+      authorization,
+      'content-type': 'application/json',
+    },
+    body: JSON.stringify(post),
+  });
+  if (!resp.ok) {
+    throw new Error(
+      `Failed to update user with id ${id}. Expected status 2xx. Received status ${`${resp.status} ${resp.statusText}`.trim()}.`,
```
jeysal, line 63:
```suggestion
      `Failed to update avatar for user with id ${id}. Expected status 2xx. Received status ${`${resp.status} ${resp.statusText}`.trim()}.`,
```

jeysal: (causes snapshot updates :grimacing: )

jeysal (commented):
Cool, this is much cleaner now. I left some nits then good to go.
Will need to remember to extract a file input component at the latest when we get another one. 
Unfortunately does not save us from good error handling incl. specialized messages for file too big etc. - my ~8MiB profile pic original e.g. is compressed only to 2.x MB and fails the request

apps/frontend/src/network/users/__mocks__/api.ts
kmaid: We submit a base64 string and squidex gives us a random url back. We can't exactly do that so ive returned whats sent with a prefix of url:

jeysal: Yeah that's true of course, it doesn't have to be the same (a data URL) but can be any URL. Just thought we'd make it a bit more realistic that way (could of course also be `https://example.com/someimage.jpg`). But I guess `url: data:base64;bla` is also technically a valid URL with protocol `url` so whatever :joy:

jeysal (approved):

PR #510, opened 2020-12-09
feat: makes description optional

johnytiago:


kmaid (changes requested):
I have modified this to accept '' like other endpoints but '' rather than undefined is coming back in the response. It displays correctly in the frontend but it be better to work like other endpoints

kmaid (approved):

PR #514, opened 2020-12-10
Initial commit of search errors

kmaid:


apps/frontend/src/network/Network.tsx
```diff
@@ -65,7 +65,10 @@ const Network: React.FC<{}> = () => {
           filters={filters}
           searchQuery={searchQuery}
         >
-          <ErrorBoundary>
+          <ErrorBoundary
```
jeysal, line 68: Are we expecting the config for this component to vary a lot? Can these be default values?

packages/react-components/src/molecules/__tests__/ErrorCard.test.tsx
```diff
@@ -6,7 +6,7 @@ import ErrorCard from '../ErrorCard';
 jest.useFakeTimers('modern');
 
 it('renders a plain message', () => {
-  const { getByText } = render(<ErrorCard>oops</ErrorCard>);
+  const { getByText } = render(<ErrorCard title="oops" description="" />);
```
jeysal, line 9: This can't be all the test changes required surely? You added new props and stuff

apps/frontend/src/network/Network.tsx
kmaid: I was hoping we would put in more error boundaries with specific error messages where possible

jeysal: Yeah but as long as that doesn't happen? :sweat_smile: 
Or even then, if we have 5 with different custom descriptions and 5 with this generic one, seems good to have the component define a standard appearance with description and refresh link instead of duplicating the generic one

jeysal: I'm still unhappy about this :( maybe we can ask product if they intend to have this error message on all unexpected errors? I would kind of assume so though, because if we set a different error message then it's really an expected error isn't it

kmaid: Ive refactored it to have a more specific search results error.

apps/frontend/src/structure/Frame.tsx
```diff
@@ -1,12 +1,28 @@
...
+  boundaryProps?: Omit<ComponentProps<typeof ErrorBoundary>, 'children'>;
+};
+
+const Frame: React.FC<FrameProps> = ({ children, boundaryProps }) => (
+  <ErrorBoundary {...boundaryProps}>
     <Suspense fallback={<Loading />}>{children}</Suspense>
   </ErrorBoundary>
 );
 
+export const SearchFrame: React.FC<{}> = ({ children }) => (
```
jeysal, line 16: Not sure where this belongs yet, but we'll figure it out sooner or later

jeysal (approved):

PR #518, opened 2020-12-11
Feat: Refresh JWT on profile updates

kmaid:


fampinheiro (approved):

jeysal (changes requested):
test. :see_no_evil: 
Oh and manual verification as well of course, how will we do that? Some other branch checked out locally?

jeysal:
Have you deployed this on the dev auth0 tenant?

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -14,7 +14,9 @@ const addUserMetadata: Rule<{ invitationCode: string }> = async (
 
   const redirect_uri = new URLSearchParams(context.request.query).get(
     'redirect_uri',
-  );
+  )
+    ? new URLSearchParams(context.request.query).get('redirect_uri')
```
jeysal, line 18: Umm... `??`? :D

apps/frontend/src/auth/test-utils.tsx
```diff
@@ -66,6 +66,7 @@ const createAuth0 = (
       logout: auth0Client
         ? (...args) => auth0Client.logout(...args)
         : notReady('logout'),
+      refreshUser: auth0Client ? async () => {} : notReady('refreshUser'),
```
jeysal, line 69: dead

apps/frontend/src/network/users/state.ts
```diff
@@ -35,17 +36,31 @@ const userState = selectorFamily<UserResponse | undefined, string>({
 
 export const useUserById = (id: string) => useRecoilValue(userState(id));
 export const usePatchUserById = (id: string) => {
+  const { getTokenSilently } = useAuth0();
   const authorization = useRecoilValue(authorizationState);
   const setPatchedUser = useSetRecoilState(patchedUserState(id));
   return async (patch: UserPatchRequest) => {
     setPatchedUser(await patchUser(id, patch, authorization));
+    await getTokenSilently({
```
jeysal, line 44: Sorry to start another round of this, but ... test? :D
E.g. do a `Provider` that re-provides the auth0 context with `getTokenSilently` overriden with a mock. Then a test case similar to the one checking the patch happening, except it checks that this is called

docs/config/auth0.md
```diff
@@ -122,4 +122,4 @@ function() {
 
 Where `$ADD_USER_METADATA` is the [build output](../../apps/auth0-rules/build/add-user-metadata.js) of `@asap-hub/auth0-rules`.
 
-**You'll need to force the import of a specific `got` version. On the Auth0 editor, where you find `require('got')` replace by `require('got@$VERSION')`, where \$VERSION is the version that `yarn why got` shows for `@asap-hub/auth0-rules`.**
+**You'll need to force the import of a specific imports for example `got` version. On the Auth0 editor, where you find `require('got')` replace by `require('got@$VERSION')`, where \$VERSION is the version that `yarn why got` shows for `@asap-hub/auth0-rules`.**
```
jeysal, line 125: idk what happened here but this is not a sentence :joy: 
Do we have another external dependency other than `got`? Was that the intention behind this?

jeysal (changes requested):

apps/frontend/src/network/users/state.ts
```diff
@@ -35,17 +36,31 @@ const userState = selectorFamily<UserResponse | undefined, string>({
 
 export const useUserById = (id: string) => useRecoilValue(userState(id));
 export const usePatchUserById = (id: string) => {
+  const { getTokenSilently } = useAuth0();
   const authorization = useRecoilValue(authorizationState);
   const setPatchedUser = useSetRecoilState(patchedUserState(id));
   return async (patch: UserPatchRequest) => {
     setPatchedUser(await patchUser(id, patch, authorization));
```
jeysal, line 43: I would think that once the `setPatchedUser` completes, the profile should be updated, no matter what happens with the token, but on the environment it seems that the profile also waits for the token request to finish, and if that errors no update happens?

apps/auth0-rules/src/add-user-metadata.ts
kmaid: I did this initially but it compiles weirdly making the manual adjustments harder

jeysal: we need to do manual adjustments here? o.O
thought only wrapping the whole thing in a fn, and the dep version lockdown

docs/config/auth0.md
kmaid: We got additional dependencies using ??

jeysal: oh. some kind of babel runtime. can probably tweak babel config to inline all of that, then we can use `??` as normal again

kmaid: Is it really worth it? I would prefer if the code generated didn't require manual changes to use map instead of some imported one.

PR #520, opened 2020-12-11
feat(edit-modals): prompt when dirty and navigating away

jeysal:


packages/react-components/src/organisms/__tests__/EditModal.test.tsx
```diff
@@ -38,78 +77,144 @@ describe('when saving', () => {
...
+          const { getByTitle, getByText } = result;
+          userEvent.click(getByTitle(/close/i));
+          expect(getUserConfirmation).not.toHaveBeenCalled();
+
+          await waitFor(() =>
+            expect(getByText(/^save/i).closest('button')).toBeEnabled(),
+          );
+        });
+
+        it('prompts when trying to leave after making edits again', async () => {
```
jeysal, line 186: This be deeeeeeeeeep...as in user action depth.

* modal initially created
* user changes things
* user saves
* modal still open (can't guarantee in this component that saving will always unmount the modal)
* user changes things again
* user closes
* expectation: gets a prompt

Crucially, when the modal becomes pristine (`!dirty`) again, it needs to reset from `hasSaved` to `initial` so that when it then becomes `dirty` again it'll prompt...

kmaid (approved):

jeysal:
Maybe I should explain the missing coverage: Exploring every path of the dirty equality checks does not seem worth it. It's just as easy to forget checking it for one field in the test as it is in the prod code. The main risk I see is forgetting to check for dirtiness as a whole, which I've tried to prevent by making the `dirty` prop mandatory

PR #521, opened 2020-12-14
Enable avatar

kmaid:


jeysal:
Shall we instruct everyone on the team (incl. us) to test this on their dev profiles with a couple of images?

PR #527, opened 2020-12-16
feat: update import cli to import social links

fampinheiro:


apps/asap-cli/src/import/parse-data.ts
```diff
@@ -77,5 +98,15 @@ export default (data: string[]): Data => {
     asapRole: ['Staff', 'Grantee', 'Guest'].includes(asapRole)
       ? (asapRole as Role)
       : 'Guest',
+    social: {
+      website1: website1 ? website1 : undefined,
```
johnytiago, line 102: what other values can `website1` have? Could it be `null`? If not, we can remove the ternary ternary operator

fampinheiro: It can be an empty string.

PR #528, opened 2020-12-16
fix(text-field): rules for showing valid indicator

kmaid:


packages/react-components/src/atoms/TextField.tsx
```diff
@@ -107,15 +115,10 @@ const customIndicatorStyles = (aspectRatio: number, position: Position) =>
     width: `${(indicatorSize * aspectRatio) / perRem}em`,
   });
 
+const showValidationType: FieldType[] = ['date', 'email', 'tel', 'url'];
```
jeysal, line 118: `Set` makes sense I think, e.g. duplicates would not be meaningful here

packages/react-components/src/atoms/TextField.tsx
```diff
@@ -152,9 +160,7 @@ const TextField: React.FC<TextFieldProps> = ({
   customIndicatorPosition = 'right',
   loading = false,
   indicateValid = customIndicator === undefined &&
-    (required !== undefined ||
-      maxLength !== undefined ||
-      pattern !== undefined),
+    (pattern !== undefined || showValidationType.has(type)),
```
jeysal, line 163: The `has` check here needs a test, one example each i.e. `email` shows indicator even without `pattern` or `indicateValid` set, `text` does not show indicator without `pattern` or `indicateValid` set

jeysal (changes requested):

packages/react-components/src/atoms/__tests__/TextField.test.tsx
```diff
@@ -141,9 +142,35 @@ describe('when valid', () => {
     expect(getComputedStyle(getByRole('textbox')).backgroundImage).toBe('');
   });
 
+  describe('when a type prop is set', () => {
```
jeysal, line 145:
```suggestion
  describe('when the type is set to a validated type', () => {
```
or so

jeysal (approved):

PR #529, opened 2020-12-16
feat(user-profile): show picture upload error message

jeysal:


jeysal:
I gave up on making the welcome page also use this mechanism for the signin errors.
It's not worth the effort, that one can stay as is, it's more of a `Route` that shows the toast to be honest, except based on query params not patchname, and closing the toast is a `history` change. It's too far from how these toasts here work (imperatively push one, and let the `ToastStack` take over from them and manage dismissal).

kmaid (approved):

PR #530, opened 2020-12-18
Fix: Global Changes

kmaid:


fampinheiro (approved):

packages/react-components/src/button.tsx
```diff
@@ -53,19 +53,19 @@ const largeStyles = css({
...
-  marginTop: `${16 / perRem}em`,
-  marginBottom: `${16 / perRem}em`,
+  marginTop: `${18 / perRem}em`,
+  marginBottom: `${18 / perRem}em`,
 
-  paddingTop: `${(15 - borderWidth) / perRem}em`,
-  paddingBottom: `${(15 - borderWidth) / perRem}em`,
-  paddingLeft: `${(18 - borderWidth) / perRem}em`,
-  paddingRight: `${(18 - borderWidth) / perRem}em`,
+  paddingTop: `${15 / perRem}em`,
```
jeysal, line 65: How is this no longer dependent on border width? If the border comes on top of this value, it'll be a weird total height won't it

PR #532, opened 2020-12-21
feat(research-output): dynamic back button destination

kmaid:


fampinheiro (approved):

jeysal (changes requested):
What are the reasons for doing it this way instead of with explicit back hrefs? What if the user is coming from `about:blank` or so? I think we should discuss the approach for this and similar future cases first

jeysal:
I've added a comment to reflect that we shouldn't do it like this elsewhere. I'm aware we already have it in the auth frontend, but there it's kind of a different situation because you can get to that page from an infinite number of previous URLs

jeysal (approved):
I've been overruled on doing this properly, so I guess I approve :D

PR #533, opened 2020-12-22
Feat: Change Rich Text mappings

kmaid:


jeysal (changes requested):
Not reviewing for now since I believe @fampinheiro already (rightfully) raised objections to changing this at all

packages/react-components/src/organisms/RichText.tsx
```diff
@@ -59,13 +59,22 @@ const components = {
...
+      <Headline5 id={id}>{children}</Headline5>
     ) : (
       <ErrorCard
         title="Styling Error"
         description="Invalid h2 heading styling"
       />
     ),
+  h3: ({ children, id }: HTMLAttributes<HTMLHeadingElement>) =>
+    isAllowedChildren(children) ? (
+      <Paragraph id={id}>{children}</Paragraph>
```
jeysal, line 71: Can we extract a `Headline6` for this please. I know it currently has no extra same styling, but that's a design coincidence, and it should still be `<h6>` and the `id` that is needed for headlines need not leak into `Paragraph`

jeysal (commented):
Been overruled on preserving correct heading structure as well :see_no_evil:

jeysal (approved):
LGTM aside from the heading order issue that was decided will not be tackled

PR #535, opened 2021-01-05
build: update to CRA 4, React 17, ESLint 7

jeysal:
I've given up trying to make it work without warnings. React/CRA are what I consider the most important updates. ESLint update is required by CRA 4, and obviously required a lot of changes, but never to actual code behavior. We should verify thoroughly that all parts of the hub work, esp. since we do not have comprehensive automated e2e tests.
Also, React Fast Refresh is happening :tada: 
image


jeysal:
No one found any issues, so #yolo

PR #541, opened 2021-01-08
refactor(model): restrict calendar colors

jeysal:


johnytiago (approved):

packages/squidex/schema/schemas/calendars.json
```diff
@@ -34,9 +34,53 @@
...
+            "#42104A",
+            "#875509",
+            "#8D6F47",
+            "#6B3304",
+            "#333333"
+          ],
           "isUnique": false,
           "inlineEditable": false,
-          "editor": "Color",
+          "editor": "Radio",
```
jeysal, line 83: This is gonna be a nice editing experience :joy: 
Can we do a different label and value?
If so, could use color 'names' at least: https://webmastermind.net/knowledgebase.php?action=displayarticle&id=15

fampinheiro: No way to have a different value and label. We can create a mapping on the backend :shrug:

fampinheiro (approved):

PR #544, opened 2021-01-11
Feat: Calendar List

kmaid:


apps/storybook/src/CalendarList.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import { CalendarList } from '@asap-hub/react-components';
+import { color, text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'Organisims / Calendar List',
+  component: CalendarList,
+};
+// number('Items', 3, { min :0, max: 10})
```
jeysal, line 10: ?

apps/storybook/src/CalendarList.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import { CalendarList } from '@asap-hub/react-components';
+import { color, text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'Organisims / Calendar List',
```
jeysal, line 7:
```suggestion
  title: 'Organisms / Calendar List',
```

packages/react-components/src/molecules/ExternalLink.tsx
```diff
@@ -20,6 +20,8 @@ const styles = css({
   margin: `${12 / perRem}em 0`,
   svg: {
     stroke: fern.rgb,
+    width: '17.8px',
```
jeysal, line 23: This and above height should also use `em` otherwise this would scale weirdly

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -0,0 +1,105 @@
...
+  li: {
+    width: '100%',
+    display: 'inline-flex',
+  },
+  '& li + li': {
+    borderTop: `1px solid ${steel.rgb}`,
+  },
+});
+interface CalendarListProps {
+  calendars: {
```
jeysal, line 58: source from model type please (`ReadonlyArray<CalendarResponse>`)

packages/react-components/src/organisms/__tests__/CalendarList.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+          colour: 'red',
+          name: 'Test Event',
+          id: '1',
+        },
+      ]}
+    />,
+  );
+  expect(getByRole('link')).toHaveAttribute(
+    'href',
+    'https://calendar.google.com/calendar/u/1/r?cid=1',
```
jeysal, line 41: `/u/1/`? o.O

jeysal (commented):
Wow, didn't someone say external link looks very different here than in existing uses and would require quite some adaptation? This icon override seems quite like the opposite :sweat_smile:

packages/react-components/src/organisms/__tests__/CalendarList.test.tsx
kmaid: I have no idea. I just used the example from the trello card. (https://trello.com/c/GvMZ2DV4/832-create-a-component-with-the-calendar-list). If we had some example data from somewhere that would be helpful

jeysal: Yeah I'm sure that wasn't there on purpose. Let's not assume people want to add it to the second google account. In fact, it is allowed to just omit that part entirely: `.../calendar/r?cid=...`, that would be preferable I think

fampinheiro: The `/1` is an hint for google to know what logged in user has access to the calendar. `/0` is the first user, `/1` the second, ...

jeysal: Yeah I know what it does, but the second user seems arbitrary, and omitting it just leaves it up to Google to decide, presumably always meaning the first user

fampinheiro (approved):

jeysal (approved):

PR #550, opened 2021-01-12
fix: infinite loading when session so old that refresh token expired

kmaid:
This is a bit of a weird one. 

See: https://github.com/auth0/auth0-spa-js/issues/449

I have been able to test this with my own expired refresh token and @fampinheiro's. If you have an expired refresh token you will be sent to the login page and signin as usual.

What i have been unable to test is when the access token has expired (after 24 hours) and the refresh token is still valid because it is signed.

I have saved my current session to test this tomorrow but I think this should go in as this will allow people who are currently locked out to access the hub.


jeysal (approved):
I'll just trust you that this fixes it - it definitely doesn't seem to break anything :D
Edited PR title to say what is actually fixed by this (as it was it was more of a refactor type of commit)

fampinheiro (approved):

PR #556, opened 2021-01-14
Fix: pagination first page no query parameters

kmaid:


jeysal:
But `'.'` should already remove all query params? o.O

fampinheiro (approved):

PR #568, opened 2021-01-20
feat: add webcal button

fampinheiro:


packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const CalendarLink: React.FC<CalendarLinkProps> = ({ id }) => {
+  const reference = useRef<HTMLDivElement>(null);
+  const handleClick = () => setMenuShown(!menuShown);
+  const [menuShown, setMenuShown] = useState(false);
+  useEffect(() => {
+    setMenuShown(false);
+  }, []);
+
+  useEffect(() => {
```
jeysal, line 92: Wow, with all other things, like the user menu, we just got away without implementing this ^^

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -0,0 +1,171 @@
...
+                </div>
+              </Anchor>
+            </li>
+            <li>
+              <Anchor href={webcal.toString()}>
+                <div css={calendarLinkStyles}>{outlookIcon} Add to Outlook</div>
+              </Anchor>
+            </li>
+            <li>
+              <Anchor
```
jeysal, line 152: I think this should be a `<button>` if it does not actually cause a navigation?

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -0,0 +1,171 @@
...
+  }, []);
+
+  useEffect(() => {
+    const handleClickOutside = (event: Event) => {
+      if (
+        reference &&
+        reference.current &&
+        !reference?.current?.contains(event.target as Node)
+      ) {
+        setMenuShown(false);
```
jeysal, line 99: This should be easily testable with `react-testing-library` - render `<><Component /><something-else /></>`, open the menu and click that something else to close it :sweat_smile:

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+type CalendarLinkProps = {
+  readonly id: string;
+};
+
+const CalendarLink: React.FC<CalendarLinkProps> = ({ id }) => {
+  const reference = useRef<HTMLDivElement>(null);
+  const handleClick = () => setMenuShown(!menuShown);
+  const [menuShown, setMenuShown] = useState(false);
+  useEffect(() => {
```
jeysal, line 88: what is this for?

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -0,0 +1,171 @@
...
+});
+
+const calendarLinkListStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  listStyle: 'none',
+  margin: 0,
+  padding: 0,
+
+  '& li': {
```
jeysal, line 60:
```suggestion
  '& > li': {
```
just to be safe, who knows what else we'll put somewhere deep down in there

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -100,19 +93,26 @@ const CalendarList: React.FC<CalendarListProps> = ({
...
-    )}
-  </Card>
+          ))}
+        </ul>
+      )}
+    </Card>
+    <Paragraph accent="lead">
+      Having issues? Set up your calendar manually with these instructions for{' '}
+      <Link href="https://support.apple.com/en-us/guide/calendar/icl1022/mac">
+        iCal
```
jeysal, line 107: Umm, did apple seriously call their calendar `iCal`? I don't think so actually, isn't it `Apple Calendar` or so

jeysal (changes requested):
I absolutely love the scammy buttons that all do the same thing :joy:

packages/react-components/src/organisms/CalendarList.tsx
fampinheiro: Yeah, they just call it Calendar.

packages/react-components/src/molecules/CalendarLink.tsx
fampinheiro: :man_shrugging:

packages/react-components/src/organisms/CalendarList.tsx
jeysal: But...should we not ask design whether the copy should be changed then? :D

jeysal (approved):
LGTM if CI and patch cov becomes happy :sweat_smile:

packages/react-components/src/organisms/CalendarList.tsx
fampinheiro: Will they notice?

PR #572, opened 2021-01-21
feat(react-components): group tools

kmaid:


apps/asap-server/src/entities/group.ts
```diff
@@ -28,14 +30,22 @@ export const parseGraphQLGroup = (item: GraphqlGroup): GroupResponse => {
     }),
   );
 
+  const tools: GroupTools = item.flatData?.tools ? item.flatData?.tools : {};
+
+  if (item.flatData?.calendars?.length) {
+    const url = new URL('https://calendar.google.com/calendar/r');
```
jeysal, line 36: Bit awkward that we do this in the backend here, but in the frontend for the calendar section

apps/storybook/src/GroupTools.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ComponentProps } from 'react';
+import { GroupTools } from '@asap-hub/react-components';
+import { text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'Molecules / Group / Tools',
```
jeysal, line 6:
```suggestion
  title: 'Molecules / Group Profile / Tools',
```

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -51,3 +52,24 @@ describe('isInternalLink', () => {
...
 });
+
+describe('getIconFromUrl', () => {
+  describe.each`
+    title                | href
+    ${'Google Drive'}    | ${`https://drive.google.com/wrong`}
+    ${'Protocols'}       | ${`https://protocols.io/wrong`}
+    ${'Slack'}           | ${`https://asap.slack.com/wrong`}
+    ${'Google Calendar'} | ${`http://calendar.google.com/r/calendar?12w3`}
+  `('for an $href to render svg with title $title', ({ href, title }) => {
```
jeysal, line 63: I don't think this is a sentence :D

jeysal: Also it's strange that this is a describe that synchronously tests things, how does this not cause any lint error?

packages/react-components/src/molecules/GroupTools.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
jeysal, line 1: I think the sections of profiles are usually organism level?

packages/react-components/src/button.tsx
```diff
@@ -81,6 +80,7 @@ const smallStyles = css({
   '> span + svg': {
     marginLeft: `${6 / perRem}em`,
   },
+  alignItems: 'center',
```
jeysal, line 83: It seems strange that large and small should differ in this way

packages/react-components/src/molecules/GroupTools.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { getIconFromUrl } from '../utils';
+import { perRem } from '../pixels';
+
+const BUTTON_SPACING = 28 / perRem;
+const buttons = css({
+  display: 'flex',
+  flexFlow: 'wrap',
+  marginTop: `${18 / perRem}em`,
+  width: `calc(100% + ${BUTTON_SPACING}em)`,
+  marginRight: `-${BUTTON_SPACING}em`,
```
jeysal, line 15: y tho?

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -82,12 +82,12 @@ const CalendarList: React.FC<CalendarListProps> = ({ calendars }) => (
               <div css={dataGrid}>
                 <div css={gridText}>
                   <Paragraph accent="charcoal">
-                    <div css={{ display: 'flex' }}>
+                    <span css={{ display: 'flex' }}>
```
jeysal, line 85: valid HTML :+2:

packages/react-components/src/organisms/GoogleCalendar.tsx
```diff
@@ -21,7 +21,11 @@ const GoogleCalendar: React.FC<GoogleCalendarProps> = ({ calendars }) => {
   });
 
   return (
-    <iframe src={iframeSrc.toString()} title="Calendar" css={iframeStyles} />
+    <iframe
+      src={iframeSrc.toString()}
+      title="Google Calendar Iframe"
```
jeysal, line 26: I prefer the previous title, this doesn't seem user focused (user doesn't necessarily know what 'iframe' means)

jeysal (changes requested):

apps/asap-server/src/entities/group.ts
fampinheiro: Agree that it may be awkward however we want to have a clean interface on tools. I don't want to revisit this component if we decide that we the google calendar isn't based on an id :|

packages/react-components/src/organisms/GoogleCalendar.tsx
kmaid: The problem is we have an iframe called Google Calendar and an SVG which is also a Google Calendar. Would you prefer embed?

packages/react-components/src/button.tsx
kmaid: This keeps the same behaviour. I can set the line height lower and remove the height restriction for it to wrap the same way as the line style if you prefer

packages/react-components/src/molecules/GroupTools.tsx
kmaid: I don't see another way to get the spacing in the design.

apps/asap-server/src/entities/group.ts
kmaid: Happy to change @fampinheiro ?

packages/react-components/src/organisms/GoogleCalendar.tsx
jeysal: If it's just about finding it in a test surely it can also be achieved with a more detailed query using `selector` or so?

apps/asap-server/src/entities/group.ts
fampinheiro: The comment was more in the "keep it as it is" direction.

packages/react-components/src/organisms/GroupTools.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { perRem } from '../pixels';
+
+const BUTTON_SPACING = 28 / perRem;
+const buttons = css({
+  display: 'flex',
+  flexFlow: 'wrap',
+  marginTop: `${18 / perRem}em`,
+  width: `calc(100% + ${BUTTON_SPACING}em)`,
+  marginRight: `-${BUTTON_SPACING}em`,
+  '> a': {
```
jeysal, line 16: Can these children be `li`s with 'sandboxed' styles instead of reaching into `Link`? (and the `div` of course `ul`)

jeysal: (I realize we can't reset `marginTop` then, but seems we can work with `marginTop` and reduce our own)

packages/react-components/src/molecules/GroupTools.tsx
jeysal: ok I found out

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -51,3 +52,24 @@ describe('isInternalLink', () => {
     });
   });
 });
+
+describe('getIconFromUrl', () => {
+  describe.each`
```
jeysal, line 57:
```suggestion
  it.each`
```

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -51,3 +52,24 @@ describe('isInternalLink', () => {
...
+    ${'Google Drive'}    | ${`https://drive.google.com/wrong`}
+    ${'Protocols'}       | ${`https://protocols.io/wrong`}
+    ${'Slack'}           | ${`https://asap.slack.com/wrong`}
+    ${'Google Calendar'} | ${`http://calendar.google.com/r/calendar?12w3`}
+  `('$href to render a svg with title $title', ({ href, title }) => {
+    const { getByTitle } = render(<>{getIconFromUrl(href)}</>);
+    expect(getByTitle(title)).toBeInTheDocument();
+  });
+
+  it('to return undefined for unknown urls', () => {
```
jeysal, line 68:
```suggestion
  it('returns undefined for unknown urls', () => {
```

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -51,3 +52,24 @@ describe('isInternalLink', () => {
...
+    expect(getIconFromUrl('http://example.com')).toBeUndefined();
+  });
+
+  it('to return undefined for invalid urls', () => {
```
jeysal, line 72:
```suggestion
  it('returns undefined for invalid urls', () => {
```

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -51,3 +52,24 @@ describe('isInternalLink', () => {
...
 });
+
+describe('getIconFromUrl', () => {
+  describe.each`
+    title                | href
+    ${'Google Drive'}    | ${`https://drive.google.com/wrong`}
+    ${'Protocols'}       | ${`https://protocols.io/wrong`}
+    ${'Slack'}           | ${`https://asap.slack.com/wrong`}
+    ${'Google Calendar'} | ${`http://calendar.google.com/r/calendar?12w3`}
+  `('$href to render a svg with title $title', ({ href, title }) => {
```
jeysal, line 63:
```suggestion
  `('$href renders a svg with title "$title"', ({ href, title }) => {
```

jeysal (approved):

apps/asap-server/src/controllers/groups.ts
```diff
@@ -15,9 +15,8 @@ flatData {
   summary
```
johnytiago, line 15: This should be deleted from the query if we are deleting summary also in this PR

johnytiago (changes requested):

apps/asap-server/src/controllers/groups.ts
johnytiago: This needs to be reflected on the entity and fixture files

PR #575, opened 2021-01-22
feat(react-components): GroupMembersSection with leaders

jeysal:
teams not yet included but submitting this as a first PR because all at once would be quite big


packages/react-components/src/molecules/MembersList.tsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
```
jeysal, line 1: This is extracted from `TeamMembersSection`, but then enhanced by functionality so may deserve full re-review

apps/storybook/src/MembersList.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  />
+);
+
+export const WithTeam = () => (
+  <MembersList
+    singleColumn={boolean('Force Single Column', false)}
+    members={createListUserResponse(number('Number of members', 6)).items.map(
+      (user) => ({
+        ...user,
+        team: { name: "Tom's Team", href: '#team' },
```
kmaid, line 24: 👍

kmaid (approved):

PR #576, opened 2021-01-22
feat: add groups to express handler

peterstarling:
see: https://trello.com/c/RZp9YZg5/893-update-express-router-to-reply-to-groups-resources-and-endpoints


apps/asap-server/test/routes/events.test.ts
```diff
@@ -1,12 +1,14 @@
 import supertest from 'supertest';
 import { appFactory } from '../../src/app';
 
+jest.mock('../../src/utils/validate-token');
```
peterstarling, line 4: @johnytiago I think we should get rid of this and try to use some sort of DI here: https://github.com/yldio/asap-hub/pull/576/commits/3680d9973a2969ee0aeeb98ca453d09f275f8446#diff-22b746f2e428e4960dd3cebd09513b984e202dc47ee52548a213cb56843dbb4cR7, what do you think?

johnytiago: Agreed 👍🏽

apps/asap-server/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  rules: {
+    'no-use-before-define': 'off',
+    'no-unused-vars': 'off',
```
johnytiago, line 4: Do you need to disable the rule, before overwriting it?

peterstarling: yeah i think so

apps/asap-server/src/controllers/groups.ts
```diff
@@ -58,13 +58,17 @@ export interface ResponseFetchGroups {
   };
 }
 
-type FetchOptions = {
-  take: number;
-  skip: number;
+export type FetchOptions = {
```
johnytiago, line 61: For now we can keep this here, but eventually we need to move it into a shared file since it will be required by all other controllers

johnytiago (approved):

PR #578, opened 2021-01-22
feat(react-components): group information

kmaid:


packages/react-components/src/organisms/__tests__/GroupInformation.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+it('does not show ellipses or show more button on short descriptions', () => {
+  const { queryByText } = render(
+    <GroupInformation {...props} description={'The best group'} />,
+  );
+  expect(queryByText(/…/)).toBeNull();
+  expect(queryByText('more')).toBeNull();
+});
+
+it('shows the ellipses and show more button on long descriptions', () => {
```
jeysal, line 27: nice 1

packages/react-components/src/organisms/__tests__/GroupInformation.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+  );
+  expect(getByText('The best group')).toBeVisible();
+});
+
+it('does not show ellipses or show more button on short descriptions', () => {
+  const { queryByText } = render(
+    <GroupInformation {...props} description={'The best group'} />,
+  );
+  expect(queryByText(/…/)).toBeNull();
+  expect(queryByText('more')).toBeNull();
```
jeysal, line 24:
```suggestion
  expect(queryByText(/more/i)).toBeNull();
```

jeysal (approved):
There is one kinda ugly edge case that comes to my mind, due to space collapsing, which is:
```
TRIM_CONTENT=2
description='x  '
```
I believe this would show `x …\nShow more`, then `x  \nShow less`. But I wouldn't consider this important enough to handle unless anyone complains.

PR #581, opened 2021-01-26
feat: Move groups-by-team-id handler to express

peterstarling:
see: https://trello.com/c/RZp9YZg5/893-update-express-router-to-reply-to-groups-resources-and-endpoints

### DIFF
--- a/apps/asap-server/src/routes/teams.route.ts
+++ b/apps/asap-server/src/routes/teams.route.ts
@@ -1,0 +1,36 @@
+import { Router } from 'express';
+import { framework } from '@asap-hub/services-common';
+import Joi from '@hapi/joi';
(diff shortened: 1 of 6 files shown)


apps/asap-server/src/routes/teams.route.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const result = await groupsController.fetchByTeamId(teamId, options);
+
+    res.json(result);
+  });
+
+  return groupRoutes;
+};
+
+const querySchema = Joi.object({
```
johnytiago, line 28: Eventually we should move this to a shared file. This is the joy schema for `FetchOptions` which is also shared across endpoints

johnytiago (approved):

PR #584, opened 2021-01-26
feat: create user profile group component

fampinheiro:


packages/react-components/src/organisms/UserProfileGroups.tsx
```diff
@@ -0,0 +1,96 @@
...
+    <Card>
+      <div
+        css={{
+          display: 'flex',
+          flexDirection: 'column',
+        }}
+      >
+        <Headline2 styleAsHeading={3}>{firstName}’ Groups</Headline2>
+        <Paragraph accent="lead">
+          {firstName}’ team is collaborating with other teams via groups, which
```
jeysal, line 87:
```suggestion
          {firstName}’s team is collaborating with other teams via groups, which
```
I assume?

packages/react-components/src/organisms/UserProfileGroups.tsx
```diff
@@ -0,0 +1,96 @@
...
+    [`&:not(:first-of-type) > :nth-child(odd)`]: {
+      display: 'none',
+    },
+  },
+});
+
+type UserProfileGroupsProps = Pick<UserResponse, 'firstName'> & {
+  groups: ReadonlyArray<{
+    name: string;
+    role: 'Leader' | 'Member';
```
jeysal, line 52: It feels like this should be defined somewhere in model?

packages/react-components/src/organisms/UserProfileGroups.tsx
```diff
@@ -0,0 +1,96 @@
...
+  gridColumnGap: `${18 / perRem}em`,
+
+  margin: 0,
+  marginTop: `${24 / perRem}em`,
+
+  padding: 0,
+  listStyle: 'none',
+});
+
+const listItemStyle = css({
```
jeysal, line 28: nice grid m8

jeysal (commented):
mostly lgtm after I understood (umm, somewhat) the grid ^^

packages/react-components/src/organisms/UserProfileGroups.tsx
fampinheiro: I'm clarifying this with James. The designs don't use the same roles as the group page, that's why they differ.
Leader is someone that is a Chair or Project Manager in a group. Member is someone that is part of a team that is part of a group.

jeysal (approved):

packages/react-components/src/organisms/UserProfileGroups.tsx
fampinheiro: Thanks for pointing this out. The requirement was to have the group roles extended with Member.
The "member" role only exists on the frontend, and it's inferred from the group response. Do you still believe the type definition should be elsewhere?

PR #585, opened 2021-01-26
fix: dont throw for tokens from other envs

johnytiago:


fampinheiro (approved):

jeysal:
What is the reason we want to accept them? (I know the code was not expected to throw in this place, but why do we not want to check this?)

johnytiago:
@jeysal We can and in fact do throw in the new express app. But the only possibility a user can reach here is by getting a token from on of the dev environments. Still, I'll update this to reflect the new handler settings.

PR #587, opened 2021-01-26
feat: ensure publish date is set

fampinheiro:


packages/react-components/src/organisms/__tests__/UserProfileRecentWorks.test.tsx
```diff
@@ -122,3 +122,42 @@ it('renders date with day, month and year', async () => {
     `"Originally Published: 12th May 2020"`,
   );
 });
+
+it('renders date when values are set', async () => {
```
jeysal, line 126: A bit confused, you probably either meant to remove the test above in favor of this one (because it is completely copied into this one), or have this contain just the bottom part?

jeysal (commented):
other than this lgtm

packages/react-components/src/organisms/__tests__/UserProfileRecentWorks.test.tsx
fampinheiro: I rerender just to ensure we had the two states. I changed the approach to have the first test render the component without any date, since the following test cover the multiple date combinations.

jeysal (approved):

PR #588, opened 2021-01-27
feat: Move /teams endpoints to express

peterstarling:
See: https://trello.com/c/JwDWDpdp/899-migrate-teams-resources-to-the-new-express-endpoint


apps/asap-server/src/controllers/teams.ts
```diff
@@ -122,7 +122,21 @@ const fetchUsers = async (id: string, client: Got): Promise<RestUser[]> => {
   return res ? res.items : [];
 };
 
-export default class Teams {
+export interface TeamController {
+  update: (id: string, tools: TeamTool[], user: User) => Promise<TeamResponse>;
+  fetch: (
+    options: {
```
johnytiago, line 128: Now would be a good time to expose those `FetchOptions` to the other controllers and add reuse them for this PR. What do you think?

PR #594, opened 2021-01-29
feat: migrates user controller tests and fix search

johnytiago:
Because this PR will become massive if I move all the endpoints at once, I'm only moving the controller tests for now. 
This PR also fixes a bug with the user search, bug I caught while migrating the endpoint.

### DIFF
--- a/apps/asap-server/test/controllers/users.test.ts
+++ b/apps/asap-server/test/controllers/users.test.ts
@@ -182,8 +182,8 @@ describe('Users controller', () => {
     });
   });
 
-  describe.only('connectByCode', () => {
-    test('Throws forbidden when doesn find connection code', async () => {
+  describe('connectByCode', () => {
+    test('Should throw forbidden when doesn find connection code', async () => {
       nock(config.baseUrl)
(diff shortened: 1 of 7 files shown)


apps/asap-server/test/controllers/users.test.ts
```diff
@@ -0,0 +1,532 @@
...
+      nock(config.baseUrl)
+        .post(`/api/apps/${config.appName}/assets`)
+        .reply(500);
+
+      await expect(
+        users.updateAvatar('user-id', Buffer.from('avatar'), 'image/jpeg'),
+      ).rejects.toThrow();
+    });
+
+    test('should return 500 when fails to update user', async () => {
```
peterstarling, line 358: probably not a 500 as this is http specific and our controller should really just throw a more descriptive error. 

But since you're not really asserting the error message in these tests anyway i think just a generic 'error' should do

peterstarling (changes requested):
aside from the invalid error assertion there isn't really anything super important so feel free to address the rest of them or not

fampinheiro (approved):

PR #599, opened 2021-02-01
refactor(react-components): do not use Link where we really just want the HashLink/anchor behavior

jeysal:
As discussed. I've tried extracting it into just a hook but returning an element type and props there is just stupid, so it's a component after all, but one that with its props interface and everything is kind of a transparent anchor and can be passed all anchor attributes.


kmaid (approved):

jeysal:
599.hub.asap.science looks good from what I can tell :rocket:

PR #600, opened 2021-02-01
fix: express tracing middleware

johnytiago:


johnytiago:
This is not working properly yet

PR #606, opened 2021-02-03
refactor(storybook): one react-component per story

jeysal:
Some stories still aren't very high quality of course (some page stories have 20 knobs, and some smaller stories have none), but with no more huge story files we should now be in a better position to make/keep the quality high


jeysal:
Not gonna ask you to review this code @fampinheiro @kmaid , it's impossible, but maybe check if everything in the deployed SB looks alright

kmaid (approved):

PR #610, opened 2021-02-04
feat: Add local server for development

peterstarling:
Adds a local express server for development by using ts-node-dev watcher. This means that the server does not require to run the build and comes with hot reloading.

## Notable changes:
* moved declarations to a standalone file 
* modified tsconfig for asap-server to get ts-node to work
* changed the `yarn start:backend` command definition so it only runs the web server (instead of the whole docker-compose)

## Important
* the express migration does not (at least for now) involve migrating the /webhook endpoints meaning that we will still need to run serverless-offline to test these endpoints locally
* by default the API points to the local squidex instance, but the `yarn start:backend` will no longer provide one, it is therefore required to either point to the DEV instance manually by changing the env vars or run the docker-compose


fampinheiro (approved):

peterstarling:
Parking this until all the endpoints are migrated to express

PR #614, opened 2021-02-05
feat: add endpoint /groups/id

fampinheiro:


apps/asap-server/test/controllers/groups.test.ts
```diff
@@ -59,10 +60,39 @@ describe('Group controller', () => {
...
+        .reply(200, fixtures.queryGroupsResponse);
 
       const result = await groups.fetch(fetchOptions);
-      expect(result).toEqual(fixtures.expectation);
+      expect(result).toEqual(fixtures.queryGroupsExpectation);
+    });
+  });
+
+  describe('Fetch by id method', () => {
+    test('Should return an empty result', async () => {
```
peterstarling, line 71: it doesn't actually return the empty result, it throw a 'Not found' error

peterstarling (approved):
I approved it but with a comment, you just need to rephrase the test case and you're good to go

PR #619, opened 2021-02-08
feat(frontend): Group Profile

jeysal:


apps/storybook/src/GroupProfileAbout.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+              teams: user.teams.map((team) => ({ ...team, href: '#' })),
+            },
+            role,
+            href: '#',
+            teams: user.teams.map((team) => ({ ...team, href: '#' })),
+          }))
+        : []
+    }
+    teams={
+      boolean('Has Teams', true)
```
kmaid, line 28: can

kmaid (approved):

apps/storybook/src/GroupProfileAbout.stories.tsx
jeysal: Oh, yeah, I remember actually having thought about that too (who knows if the fixture doesn't just have `[]` for that), but then didn't do anything about it. Adding `teamsCount` etc...

PR #620, opened 2021-02-08
fix: fetch user first on groups by userId

johnytiago:


apps/asap-server/test/routes/user.test.ts
```diff
@@ -124,8 +124,19 @@ describe('/users/ route', () => {
     });
   });
 
-  describe('GET /users/{user_id}/groups', () => {
+  describe.only('GET /users/{user_id}/groups', () => {
```
fampinheiro, line 127:
```suggestion
  describe('GET /users/{user_id}/groups', () => {
```

fampinheiro (changes requested):

peterstarling (approved):

PR #621, opened 2021-02-09
feat: import protocols from protocols.io csv

fampinheiro:


johnytiago:
I think you need to rebase this branch again. ff8ba68 is already on master

apps/asap-cli/src/cli.ts
```diff
@@ -6,20 +6,27 @@
...
+    command: 'import <entity> <path>',
+    describe: 'import entities data to squidex from csv',
     builder: (cli) =>
-      cli.positional('path', {
-        describe: 'path to csv file',
-        type: 'string',
-      }),
-    handler: async ({ path }) => importUsers(path as string),
+      cli
+        .positional('entity', {
```
johnytiago, line 19: Is entity required? Maybe we should use something like this http://yargs.js.org/docs/#demand

apps/asap-cli/src/import/protocols/insert.ts
```diff
@@ -0,0 +1,157 @@
...
+    },
+    tags: {
+      iv: data.keywords,
+    },
+  };
+
+  const [e1, teamId] = await Intercept(fetchTeamByName(data.team, promises));
+  const [e2, current] = await Intercept(fetchProtocolByLink(data.link));
+
+  const err1 = e1 as HTTPError;
```
johnytiago, line 112: Isn't this just a got error?

apps/asap-cli/src/cli.ts
fampinheiro: The `<>` wrap ensures the property is required.

johnytiago (approved):

PR #622, opened 2021-02-09
feat: e2e/smoke tests

peterstarling:
See: https://trello.com/c/FMU3iWrH/933-create-e2e-smoke-tests


apps/e2e-tests/config.ts
```diff
@@ -0,0 +1,4 @@
+export const config = {
+  appOrigin: process.env.APP_ORIGIN || 'https://api-dev.hub.asap.science',
```
jeysal, line 2: I think app origin so far usually refers to the frontend, we probably have some other environment variables in use for the api origin?

apps/e2e-tests/jest-e2e.config.js
```diff
@@ -0,0 +1,17 @@
...
+  basename(__dirname),
+);
+
+module.exports = {
+  ...base,
+
+  rootDir: __dirname,
+  displayName: 'e2e-tests',
+
+  testMatch: ["<rootDir>/tests/**/*.test.ts"],
```
jeysal, line 16: I think the Jest default pattern should be sufficient?

jeysal (commented):
Yeah this is about what I expected, I guess the interesting bit you'll tackle next will be CI integration i.e. running this on feature envs / on dev to prevent prod deploy on error / on prod to highlight any error?

apps/e2e-tests/config.ts
```diff
@@ -0,0 +1,4 @@
+export const config = {
+  appOrigin: process.env.ASAP_API_URL || 'https://api-dev.hub.asap.science',
```
fampinheiro, line 2: We use app origin to mention the main page.
To avoid expectation mismatch I would change this to `apiOrigin`.

fampinheiro (approved):

fampinheiro (approved):

PR #624, opened 2021-02-10
feat: enable groups and events page on production

fampinheiro:


fampinheiro:
This is conflicting with an existent branch `flags`

PR #625, opened 2021-02-10
Feat: User profile groups integration

kmaid:


packages/react-components/src/organisms/ProfileCardList.tsx
```diff
@@ -61,31 +61,33 @@ interface ProfileCardListProps {
...
-    {children.flatMap(({ card, editLink }, index) => [
-      <div
-        key={`card-${index}`}
-        css={[cardStyles, { gridArea: `${index + 1} card / card` }]}
-      >
-        {card}
-      </div>,
-      editLink?.href !== undefined && (
+    {children
+      .filter(({ card }) => Boolean(card))
```
jeysal, line 65: `true` is also not to be rendered, so maybe you want to validate typeof to be string or number or object?

jeysal (changes requested):
kool. this can stay pretty until the recoil caching refactors start coming in :joy:

packages/react-components/src/organisms/ProfileCardList.tsx
```diff
@@ -49,43 +49,46 @@ const editButtonStyles = css({
...
 interface ProfileCardListProps {
-  children: ReadonlyArray<{
-    card: ReactNode;
-    editLink?: {
-      href: string;
-      label: string;
-      enabled?: boolean;
-    };
-  }>;
+  children: ReadonlyArray<CardData | ReactNode>;
```
jeysal, line 61:
```suggestion
  children: ReadonlyArray<CardData | boolean | null | undefined>;
```

packages/react-components/src/organisms/ProfileCardList.tsx
```diff
@@ -49,43 +49,46 @@ const editButtonStyles = css({
...
-    {children.flatMap(({ card, editLink }, index) => [
-      <div
-        key={`card-${index}`}
-        css={[cardStyles, { gridArea: `${index + 1} card / card` }]}
-      >
-        {card}
-      </div>,
-      editLink?.href !== undefined && (
+    {children
+      .filter((cardData): cardData is CardData => Boolean(cardData))
```
jeysal, line 66:
```suggestion
      .filter((child): child is CardData => child && typeof child === 'object')
```

jeysal (commented):
the thing about `true` vs `false` children asymmetry was not addressed, and the type is v permissive - commented proposed solution

jeysal (approved):

PR #626, opened 2021-02-10
feat: use calendar link on group tools

fampinheiro:


packages/react-components/src/organisms/GroupTools.tsx
```diff
@@ -26,10 +27,13 @@ const button = css({
   },
 });
 
-type GroupToolsProps = Pick<GroupResponse, 'tools'>;
+type GroupToolsProps = Pick<GroupResponse, 'tools'> & {
+  readonly calendar?: GroupResponse['calendars'][0]['id'];
```
jeysal, line 31: `calendarId`, otherwise confusing given that other places call the whole object calendar

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -99,10 +98,14 @@ const resetButtonStyles = css({
...
 
 type CalendarLinkProps = {
+  readonly children?: string;
   readonly id: string;
 };
 
-const CalendarLink: React.FC<CalendarLinkProps> = ({ id }) => {
+const CalendarLink: React.FC<CalendarLinkProps> = ({
+  id,
+  children = 'Subscribe',
```
jeysal, line 107: test

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -99,10 +98,14 @@ const resetButtonStyles = css({
 });
 
 type CalendarLinkProps = {
+  readonly children?: string;
```
jeysal, line 101: s/string/TextChildren/

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -34,7 +35,7 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
 }) => (
   <div css={styles}>
     <GroupInformation tags={tags} description={description} />
-    <GroupTools tools={tools} />
+    <GroupTools calendar={calendars[0] && calendars[0].id} tools={tools} />
```
jeysal, line 38: oof. can we ask design for a better spec? :/ I don't wanna get into a situation where we need to clean up 5 places in the existing product again. Multiple calendars per group isn't even a hypothetical case is it

fampinheiro: The calendar is a  group property and we need to surface it here. You are right that multiple calendars per group isn't a possible case, however we modelled the relationship in a way that allows for that :disappointed:.

We can create a follow up piece of work to address that.

jeysal: I was actually thinking it *is* a possible case, but okay if it's not can we put in a Squidex relationship maximum restriction, and in the backend extract the single item out of the array (throwing if it has more than one, which should not happen because of the squidex restriction)?

fampinheiro: I'm working on this https://trello.com/c/n0bXwLtY/973-ensure-a-group-as-only-one-calendar-and-change-the-models-to-reflect-it next to address it.

jeysal (approved):

apps/frontend/src/network/groups/__tests__/GroupProfile.test.tsx
```diff
@@ -76,7 +76,9 @@ it('generates a different deep link every time to avoid conflicts', async () =>
 });
 
 it('switches to the calendar tab', async () => {
-  const { findByText } = await renderGroupProfile(createGroupResponse());
-  userEvent.click(await findByText(/^calendar/i));
-  expect(await findByText(/subscribe/i, { selector: 'h3' })).toBeVisible();
+  const { findByText, findAllByText } = await renderGroupProfile(
+    createGroupResponse(),
+  );
+  userEvent.click(await findByText(/calendar/i, { selector: 'nav a *' }));
```
fampinheiro, line 82: So this is how you scope the selectors :+1:

PR #631, opened 2021-02-10
build: update Node 12.18 => 14.15

jeysal:


fampinheiro (approved):
I poked around the created env and everything seems o be working properly. I would give it a go.

PR #646, opened 2021-02-16
feat: Add google calendar push notification integration 

peterstarling:
see: https://trello.com/c/YKUn57NX/992-subscribe-to-changes-on-google-calendar


apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+          };
+        }
+
+        if (payload.dataOld.resourceId) {
+          try {
+            await unsubscribe(payload.dataOld.resourceId?.iv, payload.id);
+          } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error(error);
+          }
```
johnytiago, line 62:
```suggestion
         await unsubscribe(payload.dataOld.resourceId?.iv, payload.id).catch( e => console.error(e))
```

This is, most likely, not properly formatted, but the point is, since it is a one-liner, we can make it smaller and maintain the same readability

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+      if (['CalendarsCreated', 'CalendarsUpdated'].includes(event)) {
+        try {
+          const resourceId = await subscribe(payload.data.id.iv, payload.id);
+
+          await calendarController.update(payload.id, {
+            resourceId,
+          });
+        } catch (error) {
+          return {
+            statusCode: 400,
```
johnytiago, line 75: Why would we only throw 400s?

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+  };
+
+  const response = await client.request<{ resourceId: string }>({
+    url,
+    method: 'POST',
+    data,
+  });
+
+  // eslint-disable-next-line no-console
+  console.log(JSON.stringify(response, null, 2));
```
johnytiago, line 120: Please use the logger

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+  const url = `${googleApiUrl}calendar/v3/channels/stop`;
+  const data = {
+    id: channelId,
+    resourceId,
+  };
+
+  const response = await client.request({ url, method: 'POST', data });
+
+  // eslint-disable-next-line no-console
+  console.log(JSON.stringify(response, null, 2));
```
johnytiago, line 148: same here

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  const secret = await client
+    .getSecretValue({ SecretId: googleApiCredentialsSecretId })
+    .promise();
+
+  if (!('SecretString' in secret) || !secret.SecretString) {
+    throw new Error('Invalid credentials');
+  }
+
+  return JSON.parse(secret.SecretString);
```
johnytiago, line 166: Maybe we should wrap this parse in a try catch just to be safe. What do you think?

johnytiago (changes requested):

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,175 @@
...
+  calendarController: CalendarController,
+): Handler =>
+  http(
+    async (request: lambda.Request): Promise<lambda.Response> => {
+      await validateRequest(request);
+
+      const bodySchema = Joi.object({
+        type: Joi.string().required(),
+        payload: Joi.object({
+          data: Joi.object().required(),
```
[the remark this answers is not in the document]
peterstarling: actually i didnt touch this code

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
peterstarling: oh i wouldnt not want to mix .catch with async/await. I really think it's much better to use one or the other

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
peterstarling: thats a fair point, what would you like to throw?

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
peterstarling: well actually i think this has to be a valid json, i dont think it would work otherwise in AWS. And even if it does work and parsing fails we will still know that it was due to the parsing error, so it kinda works no?

Plus we are not really testing this anyway

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
johnytiago: If we don't want to be very specific 502 is probably the way to go.
image

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
johnytiago: The first part is true, but we don't want to be throwing 500s to the FE. I would prefer if we handle the errors. This would be a 502 most likely with some proper logging for debugging if needed.

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -0,0 +1,176 @@
...
+            statusCode: 200,
+            payload: payload.data.id,
+          };
+        }
+
+        if (payload.dataOld.resourceId) {
+          try {
+            await unsubscribe(payload.dataOld.resourceId?.iv, payload.id);
+          } catch (error) {
+            logger(error);
```
johnytiago, line 64: Sorry to be picky, but can we add some context to the log messages so that we don't just get random errors or JSONs on the log streams?

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
peterstarling: oh we do handle them, this test:
https://github.com/yldio/asap-hub/blob/3c2354b72dba45b071c4bc72e0a72e2e7ff921fa/apps/asap-server/test/handlers/webhooks/webhook-calendar-created.test.ts#L81
shows that when the subscription function rejects with any sort of error it is caught and 502 is returned, plus the error message is attached to the response. 

Unless maybe you would like me to catch and rethrow with some more meaningful message than `Unexpected token  in JSON at position...`?

johnytiago: That's right, I was referring to malformed JSON parsing. That would be the only way (I see) to make the parser blow up. Your point to the aws secret manager source is valid. Otherwise it wouldn't be there in the first place.

But still, if we get a fkd up JSON, it will return a 500 because you are calling it here https://github.com/yldio/asap-hub/pull/646/files#diff-ddce2dd2a843e8f1975ec6fe560934ba7a4727d0ac379c0371a81ca171db49abR175 and that factory has not `try/catch` wrapper.

It's up to you to follow the optimistic or pessimistic approach 😉

peterstarling: actually...
https://github.com/yldio/asap-hub/blob/62188a506db39c4515f5eea1b806e19be9dccb0a/apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts#L176
is where the function is passed. It is invoked here:
https://github.com/yldio/asap-hub/blob/62188a506db39c4515f5eea1b806e19be9dccb0a/apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts#L134
and here
https://github.com/yldio/asap-hub/blob/62188a506db39c4515f5eea1b806e19be9dccb0a/apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts#L98

Both of which are within try and catch, see:
https://github.com/yldio/asap-hub/blob/62188a506db39c4515f5eea1b806e19be9dccb0a/apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts#L61
and
https://github.com/yldio/asap-hub/blob/62188a506db39c4515f5eea1b806e19be9dccb0a/apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts#L70

johnytiago: You are totally right! ✅ It was my bad, I thought we were running the function when passing it as an argument.

johnytiago (approved):

johnytiago:
LGMT :shipit:

apps/asap-server/src/config.ts
```diff
@@ -20,3 +23,8 @@ export const environment = ENVIRONMENT
 export const auth0SharedSecret = AUTH0_SHARED_SECRET || 'auth0_shared_secret';
 export const squidexSharedSecret =
   SQUIDEX_SHARED_SECRET || 'squidex_shared_secret';
+export const googleApiUrl = 'https://www.googleapis.com/';
+export const region = REGION || 'eu-west-1';
```
johnytiago, line 27:
```suggestion
export const region = REGION || 'us-east-1';
```

PR #652, opened 2021-02-18
feat: add update and get methods to calendar controller, extend schema

johnytiago:


apps/asap-server/src/controllers/calendars.ts
```diff
@@ -27,11 +27,33 @@ export default class Calendars implements CalendarController {
       items: calendars.map(parseCalendar),
     };
   }
+
+  async getSyncToken(calendarId: string): Promise<string | undefined> {
```
peterstarling, line 31: if you feel like it you can change it to `fetchRaw` as we are going to need to be able to fetch more than just the sync token in the near future

peterstarling (approved):
i approved by you could change the

packages/squidex/src/entities/calendar.ts
```diff
@@ -5,6 +5,8 @@ export interface Calendar {
   id: string;
   color: GoogleLegacyCalendarColor;
   name: string;
+  syncToken?: string;
+  resourceId?: string;
```
johnytiago, line 9:
```suggestion
  resourceId?: string;
  expirationDate?: string;
```

PR #653, opened 2021-02-18
feat: sync google calendar hook

johnytiago:


apps/asap-server/src/controllers/calendars.ts
```diff
@@ -28,6 +31,30 @@ export default class Calendars implements CalendarController {
     };
   }
 
+  async fetchByResouceId(resourceId: string): Promise<RestCalendar> {
```
peterstarling, line 34: for this function signature to be valid you need to make `resourceId` unique

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -1,21 +1,18 @@
-import debug from 'debug';
-import AWS from 'aws-sdk';
+/* eslint-disable no-shadow */
```
peterstarling, line 1: why do we need it exactly?

apps/asap-server/src/handlers/webhooks/webhook-events-updated.ts
```diff
@@ -1,11 +1,105 @@
...
+      }
+
+      const calendar = await calendars
+        .fetchByResouceId(resourceId)
+        .catch((err) => {
+          logger('Error fetching calendar', err);
+          return undefined;
+        });
+
+      if (!calendar?.id) {
```
peterstarling, line 43: I'd go with
```suggestion
      if (!calendar) {
```

apps/asap-server/src/handlers/webhooks/webhook-events-updated.ts
```diff
@@ -1,11 +1,105 @@
...
+        return { statusCode: 502 };
+      }
+      const squidexCalendarId = calendar.id;
+      const googleCalendarId = calendar.data.id.iv;
+
+      const syncToken = await calendars
```
peterstarling, line 49: why not just get it from 
```typescript
const syncToken = calendar.data.syncToken
```
?

apps/asap-server/src/handlers/webhooks/webhook-events-updated.ts
```diff
@@ -1,11 +1,105 @@
...
+    },
+  );
+
+export const handler: Handler = webhookEventUpdatedHandlerFactory(
+  new Calendars(),
+  getJWTCredentials,
+  syncCalendarFactory,
+);
+
+const syncEventFactory = (calendarId: string) => {
```
peterstarling, line 99: since you moved syncCalendar to a separate file, do you think maybe this should be moved with it? just a suggestion, your call

apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -0,0 +1,67 @@
...
+import logger from './logger';
+
+export type SyncCalendarFactory = (
+  syncToken: string | undefined,
+  syncEvent: (event: calendarV3.Schema$Event) => Promise<void>,
+  auth: Auth.GoogleAuth | Auth.OAuth2Client,
+) => (googleCalendarId: string) => Promise<string | undefined | null>;
+
+export const syncCalendarFactory: SyncCalendarFactory = (
+  syncToken: string | undefined,
```
peterstarling, line 11: You are declaring a type right above function definition where it is used. I would suggest removing these types from the definition because they are unnecessary. 

However - if I remember correctly - there was a rule somewhere that forced that, something about being explicit, except in this scenario it is way to explicit. I tried removing those on my local and I don't see any errors but i think it might come up at build time in some check. Give that a go, that is, if you think this is a valid point. Personally I dont like clutter.

If we do have such rule and we want to leave it then the other thing you could do is declare the type like:
```typescript
export type SyncCalendarFactory = typeof syncCalendarFactory
```
Because you only really need this type so that the factory can be used as a dependency

apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -0,0 +1,67 @@
...
+): Promise<string | undefined | null> => {
+  const params: calendarV3.Params$Resource$Events$List = {
+    pageToken: pageToken || undefined,
+    calendarId: googleCalendarId,
+    singleEvents: true, // recurring events come returned as single events
+  };
+
+  const calendar = google.calendar({ version: 'v3', auth });
+
+  const res = await calendar.events.list(params).catch((err) => {
```
peterstarling, line 36: I would prefer not to mix async/await with promise chains. 

Since we are using async functions all over the place I would simply await the response and try/catch the error

apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const fetchEvents = async (
+  googleCalendarId: string,
+  auth: Auth.GoogleAuth | Auth.OAuth2Client,
+  syncEvent: (event: calendarV3.Schema$Event) => Promise<void>,
+  syncToken: string | undefined,
+  pageToken?: string,
+): Promise<string | undefined | null> => {
+  const params: calendarV3.Params$Resource$Events$List = {
+    pageToken: pageToken || undefined,
```
peterstarling, line 29:
```suggestion
    pageToken,
```

apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -0,0 +1,67 @@
+import { google, calendar_v3 as calendarV3, Auth } from 'googleapis';
+import logger from './logger';
+
+export type SyncCalendarFactory = (
+  syncToken: string | undefined,
+  syncEvent: (event: calendarV3.Schema$Event) => Promise<void>,
+  auth: Auth.GoogleAuth | Auth.OAuth2Client,
+) => (googleCalendarId: string) => Promise<string | undefined | null>;
```
peterstarling, line 8: I know you are not really doing anything with those synced events in this PR apart from logging, but knowing that these events need to be saved later I would personally have `SyncCalendarFactory` return not just the token but also a full list of events that it fetched. 

This way we could separate the logic of saving events (or doing anything with them) from the syncing. I know that the function to sync events itself is separate (`syncEventFactory`), so yeah, you can mock it, test it, etc, but it would be much cleaner and easier to test the entire logic with the suggested approach, i think.

apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
```diff
@@ -1,16 +1,204 @@
...
+    getJWTCredentialsMock.mockRejectedValueOnce(new Error('AWS Error'));
+
+    const res = (await handler(
+      apiGatewayEvent(googlePayload),
+    )) as APIGatewayProxyResult;
+
+    expect(res.statusCode).toStrictEqual(502);
+  });
+
+  test('Should return 200 and do full synk when it fails to fetch the synkToken from squidex', async () => {
```
peterstarling, line 94: sync?

apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
```diff
@@ -1,16 +1,204 @@
...
+      client_email: 'random-data',
+      private_key: 'random-data',
+    });
+
+    const res = (await handler(
+      apiGatewayEvent(googlePayload),
+    )) as APIGatewayProxyResult;
+
+    expect(res.statusCode).toStrictEqual(200);
+    const [syncToken] = syncCalendarFactoryMock.mock.calls[0];
```
peterstarling, line 111: do you not find `toHaveBeenCalledWith` easier to read?

apps/asap-server/test/utils/google-sync-calendar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        ...fixtures.listEventsResponse,
+        items: null,
+      },
+    });
+
+    await syncCalendarHandler(calendarId);
+    expect(syncEvent).toBeCalledTimes(0);
+  });
+
+  test('Should not throw when syncEvent fails to update event', async () => {
```
peterstarling, line 64: well actually.. are we sure we dont want the whole thing to throw when it fails to update a single event? Maybe that way the whole lambda would be retried? Just a question, im not really sure.

While I'm on it - I noticed [you used Promise.all](https://github.com/yldio/asap-hub/pull/653/files#diff-316c33d3a09eac72218fdebd1ac2dc43ec0d0667a3fbd4fb872fabbdc7e919b9R47) for syncing events. You then don't actually test the scenario when you have multiple events of which only one fails but you still want to make sure that all the other ones are saved correctly. In fact it would be difficult to test, i think, unless maybe we some time delay/nextTick magic.

Anyway with promise all its possible for one promise to reject almost immediately and thus the execution of the remaining ones may also be terminated (im not entirely sure but i think that's what the [definition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) says. Regardless of whether you test it or not - I would go for .allSettled

apps/asap-server/test/utils/google-sync-calendar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+    });
+    mockList.mockResolvedValueOnce({ data: fixtures.listEventsResponse });
+
+    await syncCalendarHandler(calendarId);
+
+    expect(mockList).toHaveBeenCalledTimes(2);
+    expect(syncEvent).toBeCalledTimes(4);
+  });
+});
```
peterstarling, line 86: I would add a test to verify that the events are actually passed to the `syncEvent` function, probably the most important part here. I know you tested the number of calls already so you could just extended this assertion.

peterstarling (changes requested):
All of these comments are about coding style/pattern except for this one

image

The rest of them feel free to ignore/address whenever you please

apps/asap-server/src/controllers/calendars.ts
johnytiago: That's a nice catch, will do! 😉

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
johnytiago: This is a rule that makes sense when we talk about generic variables with the same name, but in different scopes. Imagine the case of a variable names `result`.
In this case, the variable we define in the outside scope, the import, and the variable inside the function scope, the parameter, are always the same.
Here, naming the variable with a different name will actually make the code harder to read, which is what the rule intends to achieve.

apps/asap-server/src/handlers/webhooks/webhook-events-updated.ts
johnytiago: Just... Thanks. 🤦🏽‍♂️  
I guess, copy and pasting old versions of your code is not the best strategy 😂

apps/asap-server/src/handlers/webhooks/webhook-events-updated.ts
johnytiago: That will happen in my next PR 😉  I just wanted to make it simple on this one.

apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
johnytiago: It's like sync, just a bit more radikal 🤘🏽

apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
johnytiago: I think with that function i need to assert all arguments and I didn't want to assert them all. 
Btw, I removed this test as it no longer makes sense.

apps/asap-server/test/utils/google-sync-calendar.test.ts
johnytiago: Agreed, while I'm not sure how I'm going to test that without changing the function signature, i'm going to use `allSettled`.

peterstarling (approved):

PR #219, opened 2020-09-07
chore: create build dev by default

fampinheiro:
- Use `ASAP` prefix variables on CI
- Explicit `source ci/env-create.sh` when building and deploying feature branches
- Usage of auxiliary env vars prefixed with `SQUIDEX_TEST` for tests
- Rebuild for production not development or feature branches

PR #654, opened 2021-02-19
feat: remove welcome invite validation

fampinheiro:
The registration flow is broken. Removing the preventive validation to enable users to register.

### DIFF
--- a/apps/frontend/src/welcome/Welcome.tsx
+++ b/apps/frontend/src/welcome/Welcome.tsx
@@ -1,35 +1,32 @@
-import React, { useEffect } from 'react';
-import { useParams, useHistory } from 'react-router-dom';
+import React from 'react';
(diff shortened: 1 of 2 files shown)


jeysal:
Is this temporary until the endpoint is restored? maybe we should put it behind a flag or comment out or something. or we just try to remember and revert in the future?

fampinheiro:
The endpoint was commented out https://github.com/yldio/asap-hub/pull/219/files and later removed. I'm still trying to figure our how it took us so long to uncover this. I would say remove and reevaluate.

The code isn't lost and we need a working version ASAP.

jeysal (approved):
Okay, but can you create a ticket and restore the backend endpoint later this week? I was hoping that's not significantly more difficult than commenting out this..

PR #656, opened 2021-02-22
fix: do not preserve scroll position across pages

kmaid:


packages/react-components/src/atoms/__tests__/Anchor.test.tsx
```diff
@@ -122,4 +122,20 @@ describe('for an internal link with a router', () => {
       expect(spyScrollIntoView).toHaveBeenCalledWith({ behavior: 'smooth' }),
     );
   });
+
+  it('triggers onclick event for internal links', () => {
```
jeysal, line 126: shouldn't the test be the opposite of the test above? so does not scroll it into view if it was like e.g. a `?query` link

jeysal: wait, we had to write custom code for this? o.O I thought with normal react router links it would just work

packages/react-components/src/utils.ts
```diff
@@ -31,6 +31,9 @@ export const isInternalLink = (href: string): boolean =>
   globalThis.location &&
   new URL(href, globalThis.location.href).origin === globalThis.location.origin;
 
+export const isHashLink = (href: string): boolean =>
+  globalThis.location && !!new URL(href, globalThis.location.href).hash;
```
jeysal, line 35: I could see this being a problem with `/#`, where `.hash` is `''`

jeysal: as evidenced by one of the tests

kmaid: They both scroll to the top of the page

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -73,3 +78,38 @@ describe('getIconFromUrl', () => {
     expect(getIconFromUrl('not a url')).toBeUndefined();
   });
 });
+
+describe('isHashLink', () => {
+  describe.each`
```
jeysal, line 83: I think one will suffice :joy: and the description is the same every time, so does not need to be a var :sweat_smile:

packages/react-components/src/atoms/__tests__/Anchor.test.tsx
kmaid: No :(

kmaid: https://reactrouter.com/web/guides/scroll-restoration

packages/react-components/src/utils.ts
jeysal: The non-hashlink because of the custom code? :/

packages/react-components/src/templates/__tests__/Layout.test.tsx
```diff
@@ -75,6 +75,8 @@ it('closes the drawer when clicking the overlay', async () => {
 });
 
 it('closes the drawer on navigation', async () => {
+  Element.prototype.scrollTo = jest.fn();
```
kmaid, line 78: Maybe this should be moved to the jest setup file?

jeysal: Don't think so, we usually only mock what's necessary for a specific test

packages/react-components/src/templates/Layout.tsx
```diff
@@ -175,6 +180,12 @@ const Layout: React.FC<LayoutProps> = ({
   useEffect(() => {
     setMenuShown(false);
   }, [location]);
+  useEffect(() => {
+    if (location?.pathname !== prevLocation?.pathname) {
+      const main = document.querySelector('main');
```
jeysal, line 185: nope nope nope

use a ref :D

packages/react-components/src/templates/Layout.tsx
```diff
@@ -163,10 +164,14 @@ const Layout: React.FC<LayoutProps> = ({
   const [menuShown, setMenuShown] = useState(false);
 
   let location: Location | undefined;
+  let prevLocation: Location | undefined;
   // This hook *is* called unconditionally despite what rules-of-hooks says
   /* eslint-disable react-hooks/rules-of-hooks */
   try {
     location = useLocation();
+    prevLocation = usePrevious<Location<History.PoorMansUnknown> | undefined>(
```
jeysal, line 172: unnecessary type arg, can be inferred

jeysal: I hope my mental TSC is not wrong here and it can't, that'd be embarassing ^^

packages/react-components/src/templates/__tests__/Layout.test.tsx
```diff
@@ -86,3 +88,16 @@ it('closes the drawer on navigation', async () => {
...
+  const scrollToMock = jest.fn();
+  Element.prototype.scrollTo = scrollToMock;
+
+  const { getAllByText } = render(
+    <MemoryRouter initialEntries={['/']}>
+      <Layout {...props} />
+    </MemoryRouter>,
+  );
+  userEvent.click(getAllByText(/network/i, { selector: 'nav *' })[0]);
+  expect(scrollToMock).toHaveBeenCalled();
```
jeysal, line 102: needs to check which element was scrolled (put some text children, intercept the `this`, and check that it has the right text content)

jeysal (changes requested):

packages/react-components/src/templates/Layout.tsx
kmaid: Whoops, I did mean to convert it to a ref.

jeysal (approved):

PR #659, opened 2021-02-23
feat: Add time zone fields to events

peterstarling:
Adds support for time-zones for start and end dates in the Event entity


packages/squidex/schema/schemas/events.json
```diff
@@ -60,6 +60,23 @@
           "isRequired": false
         }
       },
+      {
+        "name": "startDateTimeZone",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
```
johnytiago, line 67:
```suggestion
        "isDisabled": true,
```

packages/squidex/schema/schemas/events.json
```diff
@@ -75,6 +92,23 @@
           "isRequired": false
         }
       },
+      {
+        "name": "endDateTimeZone",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
```
johnytiago, line 99:
```suggestion
        "isDisabled": true,
```

johnytiago:
The date timezones shouldn't be editable through the UI

packages/model/src/event.ts
```diff
@@ -8,7 +8,9 @@ export interface EventResponse {
   id: string;
   title: string;
   startDate: string;
+  startDateTimeZone: string;
```
johnytiago, line 11: I thought the TZs wouldn't propagate directly to the FE

jeysal: with the new approach suggested by @peterstarling they do, this would be something like `"Europe/Lisbon"` afaik

johnytiago (approved):

PR #664, opened 2021-02-24
event fixes, multi commit, do not squash

jeysal:
Also threw in a component extraction I need for my ticket already

- refactor(react-components): split EventInfo out of EventCard
- fix(event-info): multi group behavior
- fix(event-info): show end day if different from start day


jeysal:
@kmaid overflow behavior fixed. I made the flex row alignItems:center for the case where there are multiple groups so that it looks less weird. but I think that case is hypothetical and product doesn't expect it to happen

kmaid (approved):

PR #665, opened 2021-02-24
fix: Calendar-created webhook notFound error

peterstarling:
Fixes the issue whereby the hook, upon receiving a calendar entry with an empty calendar ID, attempts to subscribe and fails.


apps/asap-server/test/handlers/webhooks/webhook-calendar-created.test.ts
```diff
@@ -116,7 +137,7 @@ describe('Calendar Webhook', () => {
       )) as APIGatewayProxyResult;
 
       expect(res.statusCode).toStrictEqual(200);
-      expect(subscribe).not.toHaveBeenCalled();
+      expect(unsubscribe).not.toHaveBeenCalled();
```
peterstarling, line 140: thats just something i noticed that should be fixed too

johnytiago (approved):

PR #669, opened 2021-02-25
feat: Add API token to google event subscription

peterstarling:
see: https://trello.com/c/emZkLgNb/1016-add-authentication-to-event-webhook

this will enable us to authenticate requests coming from google api

Other changes:
* add the resourceId removal from the db once unsubscribed from the channel - this will prevent us from seeing errors about a failed unsubscribing attempt after the calendar ID was previously set to null


fampinheiro (approved):

apps/asap-server/src/handlers/webhooks/webhook-calendar-created.ts
```diff
@@ -60,6 +61,10 @@ export const webhookCalendarCreatedHandlerFactory = (
         if (payload.dataOld.resourceId) {
           try {
             await unsubscribe(payload.dataOld.resourceId?.iv, payload.id);
+
+            await calendarController.update(payload.id, {
+              resourceId: null,
```
johnytiago, line 66: Does this work?

PR #670, opened 2021-02-26
fix: Make Calendar ID required

peterstarling:
Makes the calendar ID required and applies a min length validation of 8 characters

### DIFF
--- a/packages/squidex/schema/schemas/calendars.json
+++ b/packages/squidex/schema/schemas/calendars.json
@@ -35,11 +35,12 @@
         "properties": {
           "fieldType": "String",
           "pattern": "^[a-zA-Z0-9.!#$%&’*+\\/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:.[a-zA-Z0-9-]+)*$",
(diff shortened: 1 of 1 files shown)


fampinheiro (approved):

packages/squidex/schema/schemas/calendars.json
```diff
@@ -35,11 +35,12 @@
         "properties": {
           "fieldType": "String",
           "pattern": "^[a-zA-Z0-9.!#$%&’*+\\/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:.[a-zA-Z0-9-]+)*$",
+          "minLength": 8,
```
fampinheiro, line 38: Do we need min length for calendars?

johnytiago: we can set this to 1, TBH, we just dont want empty strings

peterstarling: yeah so they wouldnt put an empty string or something. You would think that wouldn't be possible with the regex but well, squidex

PR #673, opened 2021-02-26
feat: insert events into squidex

johnytiago:


apps/asap-server/src/utils/sync-google-event.ts
```diff
@@ -0,0 +1,74 @@
...
+        dateTime: Joi.string(),
+        timeZone: Joi.string(),
+      }).or('date', 'dateTime'),
+    }).unknown();
+
+    const { error, value } = schema.validate(event);
+
+    if (error) {
+      logger('Ignored event update, validation error:', error);
+      return Promise.resolve();
```
peterstarling, line 39: i though you were catching errors around syncEvent here: https://github.com/yldio/asap-hub/blob/cc96526794046211fc91fee8b59d92326b535974/apps/asap-server/src/utils/sync-google-calendar.ts#L47

so why not just throw when a validation error occurs?

apps/asap-server/src/utils/sync-google-event.ts
```diff
@@ -0,0 +1,74 @@
...
+        googleEvent.end.dateTime ||
+        googleEvent.end.date ||
+        new Date(0).toISOString(),
+      endDateTimeZone: googleEvent.end.timeZone || defaultTimezone,
+      status: googleEvent.status,
+      calendar: [calendarId],
+      tags: [],
+    };
+
+    return eventsController.upsert(googleEvent.id, squidexEvent);
```
peterstarling, line 62: i don't expect you to rewrite it now but this is what I mentioned to you later that it would be good to separate actually inserting stuff into DB from syncing. Like to me the process is:
* get the token (unless its a full sync)
* get the event data from google api
* insert the event data into API
Currently however the last two intertwine making it a little harder to reason about (and potentially test). 

There is an upside to this approach though - for multiple google sync iterations the events are upserted after each iteration rather than after all of them complete - although I dont think this actually makes any difference to us

apps/asap-server/test/utils/sync-google-event.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        endDateTimeZone: 'Europe/London',
+        status: 'confirmed',
+        calendar: ['squidex-calendar-id'],
+        tags: [],
+      },
+    );
+  });
+
+  describe('Validation test', () => {
+    test('Should not upsert when validation fails - empty object', async () => {
```
peterstarling, line 61: why not throw

peterstarling (approved):
its an approval but i strongly encourage refactoring the sync bit later

PR #675, opened 2021-02-26
Feat: Group upcoming events functionality

kmaid:


packages/flags/src/index.ts
```diff
@@ -2,7 +2,8 @@ export type Flag =
   | 'PERSISTENT_EXAMPLE'
   | 'USER_PROFILE_EDIT_SKILLS'
   | 'USER_PROFILE_EDIT_WORKS'
-  | 'UPCOMING_EVENTS';
+  | 'UPCOMING_EVENTS'
+  | 'GROUP_UPCOMING_EVENTS';
```
jeysal, line 6: not sure it needs another flag, but okay. in that case I think we want to enable it straight away though? the other one is also already enabled

apps/frontend/src/network/groups/__tests__/Upcoming.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    ...createListEventResponse(2),
+    items: createListEventResponse(2).items.map((item, index) => ({
+      ...item,
+      title: `Event title ${index}`,
+    })),
+  });
+  expect(
+    getAllByRole('heading', { level: 3 }).map((heading) => heading.textContent),
+  ).toEqual(['Event title 0', 'Event title 1']);
+});
```
jeysal, line 69: This needs tests for
* The hrefs generated (I know I know, with the routing package restructuring this would be declarative and not require tests, but we're not there :/ )
* The time being passed as `after`, and IMO doing that you might as well add the `subtractHours` already. What we should def ensure is that this and the `events/Upcoming` are consistent in that regard

jeysal (changes requested):

jeysal (approved):

PR #677, opened 2021-03-01
feat: Add a scheduled job to resubscribe the calendars

peterstarling:
see: https://trello.com/c/YCqoAePq/1024-add-a-cron-job-to-resubscribe-to-calendar-event-changes

### DIFF
--- a/apps/asap-server/src/handlers/jobs/resubscribe-calendars.ts
+++ b/apps/asap-server/src/handlers/jobs/resubscribe-calendars.ts
@@ -24,7 +24,7 @@ export const resubscribeCalendarsHandlerFactory = (
   });
 
   const calendarIds = calendars.map((calendar) => calendar.id);
-  logger(`Received the following calendars to resubscribe`, calendarIds);
+  logger(`Received the following calendars to resubscribe: %o`, calendarIds);
 
   await Promise.allSettled(
     calendars.map(async (calendar) => {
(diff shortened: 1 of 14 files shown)


apps/asap-server/src/handlers/jobs/resubscribe-calendars.ts
```diff
@@ -0,0 +1,64 @@
...
+export const resubscribeCalendarsHandlerFactory = (
+  calendarController: CalendarController,
+  unsubscribe: UnsubscribeFromEventChanges,
+  subscribe: SubscribeToEventChanges,
+): ScheduledHandlerAsync => async () => {
+  const now = DateTime.local();
+  const duration24h = Duration.fromObject({ day: 1 });
+  const calendars = await calendarController.fetchRaw({
+    maxExpiration: now.plus(duration24h).toMillis(),
+    take: 100,
```
peterstarling, line 22: FYI i dont think we expect a 100 calendars to all expire at the same time and if they do - we will resubscribe only the first 100 then on the next day - the remaining ones

packages/squidex/src/squidex.ts
```diff
@@ -94,6 +94,12 @@ export class Squidex<T extends { id: string; data: Record<string, unknown> }> {
         };
       }
 
+      // eslint-disable-next-line no-console
```
johnytiago, line 97: Isn't the logger on the http error handler not working? 
https://github.com/yldio/asap-hub/blob/master/packages/services-common/src/framework/lambda.ts#L70
Old lambdas, still run through this code. If squidex errors are not showing up even with `DEBUG=http` the issue should get fixed there.

serverless.js
```diff
@@ -202,6 +202,25 @@ module.exports = {
         }}`,
       },
     },
+    resubscribeCalendars: {
+      handler:
```
johnytiago, line 206: Is 16 seconds of default `timeout` enough for this lambda? We have quite a few orchestrations to do. Maybe we can bump this a bit, just to be safe. What do you think?

johnytiago:
Removing the WIP label as this no longer in progress

johnytiago (approved):

PR #678, opened 2021-03-01
feat: adds token verification to webhook

johnytiago:


apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
```diff
@@ -52,11 +52,41 @@ describe('Event Webhook', () => {
...
+          ...googlePayload.headers,
+          'x-goog-channel-token': undefined,
+        },
+      }),
+    )) as APIGatewayProxyResult;
+
+    expect(res.statusCode).toStrictEqual(401);
+  });
+
+  test('Should return 403 when x-goog-channel-token is not set', async () => {
```
peterstarling, line 69: you mean when its not the same?

peterstarling (approved):

apps/asap-server/test/handlers/webhooks/webhook-events-updated.test.ts
johnytiago, line 69:
```suggestion
  test('Should return 403 when x-goog-channel-token is not the same', async () => {
```

PR #680, opened 2021-03-01
feat(event): join event section

jeysal:
I haven't figured out what's going on with the fake timers, so we may need to merge this and I sort out the tests later


apps/asap-server/src/entities/event.ts
```diff
@@ -14,7 +17,9 @@ export const parseGraphQLEvent = (item: GraphqlEvent): EventResponse => {
     ) || [];
   const startDate = DateTime.fromISO(item.flatData!.startDate!);
   const now = DateTime.local();
-  const oneDayDuration = Duration.fromObject({ day: 1 });
+  const oneDayDuration = Duration.fromObject({
```
kmaid, line 20: naming

kmaid (approved):

PR #682, opened 2021-03-01
ci: don't use Jest cache

jeysal:


jeysal:
10:52 (even though it's just one data point for now ofc) does indeed seem better than the 12-17 minutes the last master test jobs averaged :+1:

jeysal:
Would be interesting to take another shot at working Jest caching for much better results, but another time

PR #691, opened 2021-03-03
feat: add googleID to events

johnytiago:
https://trello.com/c/aHO7UP9B/1047-add-missing-events-to-the-hub

### DIFF
--- a/apps/asap-server/src/utils/sync-google-calendar.ts
+++ b/apps/asap-server/src/utils/sync-google-calendar.ts
@@ -9,10 +9,9 @@ export type SyncCalendarFactory = (
 ) => (googleCalendarId: string) => Promise<string | undefined | null>;
 
 interface SyncEvent {
-  (
-    event: calendarV3.Schema$Event,
-    defaultCalendarTimezone: string,
-  ): Promise<unknown>;
+  (event: calendarV3.Schema$Event, defaultCalendarTimezone: string): Promise<
+    unknown
+  >;
 }
 
 export const syncCalendarFactory: SyncCalendarFactory = (
--- a/apps/asap-server/src/utils/sync-google-calendar.ts
+++ b/apps/asap-server/src/utils/sync-google-calendar.ts
@@ -8,10 +8,9 @@ export type SyncCalendarFactory = (
 ) => (googleCalendarId: string) => Promise<string | undefined | null>;
 
 interface SyncEvent {
-  (
-    event: calendarV3.Schema$Event,
-    defaultCalendarTimezone: string,
-  ): Promise<unknown>;
+  (event: calendarV3.Schema$Event, defaultCalendarTimezone: string): Promise<
+    unknown
+  >;
 }
 
 export const syncCalendarFactory: SyncCalendarFactory = (
--- a/apps/asap-server/src/utils/sync-google-calendar.ts
+++ b/apps/asap-server/src/utils/sync-google-calendar.ts
@@ -37,12 +37,9 @@ const fetchEvents = async (
     pageToken: pageToken || undefined,
     calendarId: googleCalendarId,
     singleEvents: true, // recurring events come returned as single events
+    timeMin: new Date('2020-10-01').toISOString(),
+    timeMax: DateTime.utc().plus({ months: 6 }).toISO(),
   };
-
-  if (!syncToken) {
-    params.timeMin = new Date('2020-10-01').toISOString();
-    params.timeMax = DateTime.utc().plus({ months: 6 }).toISO();
-  }
 
   const calendar = google.calendar({ version: 'v3', auth });
 
--- a/apps/asap-server/src/utils/sync-google-calendar.ts
+++ b/apps/asap-server/src/utils/sync-google-calendar.ts
@@ -3,19 +3,20 @@ import logger from './logger';
 
 export type SyncCalendarFactory = (
   syncToken: string | undefined,
-  syncEvent: (
-    event: calendarV3.Schema$Event,
-    defaultCalendarTimezone: string,
-  ) => Promise<void>,
+  syncEvent: SyncEvent,
   auth: Auth.GoogleAuth | Auth.OAuth2Client,
 ) => (googleCalendarId: string) => Promise<string | undefined | null>;
 
-export const syncCalendarFactory: SyncCalendarFactory = (
-  syncToken: string | undefined,
-  syncEvent: (
+interface SyncEvent {
+  (
     event: calendarV3.Schema$Event,
     defaultCalendarTimezone: string,
-  ) => Promise<void>,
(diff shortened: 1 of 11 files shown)


apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -40,6 +38,11 @@ const fetchEvents = async (
     singleEvents: true, // recurring events come returned as single events
   };
 
+  if (!syncToken) {
+    params.timeMin = new Date('2020-10-01').toISOString();
```
peterstarling, line 42: why don't you use DateTime for both dates?

apps/asap-server/src/utils/sync-google-calendar.ts
```diff
@@ -1,21 +1,22 @@
...
-    event: calendarV3.Schema$Event,
-    defaultCalendarTimezone: string,
-  ) => Promise<void>,
+  syncEvent: SyncEvent,
   auth: Auth.GoogleAuth | Auth.OAuth2Client,
 ) => (googleCalendarId: string) => Promise<string | undefined | null>;
 
+interface SyncEvent {
+  (event: calendarV3.Schema$Event, defaultCalendarTimezone: string): Promise<
+    unknown
```
peterstarling, line 13: probably best to return `Promise<void>` if you don't really expect it to return anything, otherwise it's a bit misleading

Looking at 
https://github.com/yldio/asap-hub/pull/691/files#diff-316c33d3a09eac72218fdebd1ac2dc43ec0d0667a3fbd4fb872fabbdc7e919b9R64
it appears that the signature should actually be a little different and the return type should contain something?

apps/asap-server/test/controllers/events.test.ts
```diff
@@ -498,57 +499,117 @@ describe('Event controller', () => {
...
+        .query({
+          $top: 1,
+          $filter: filter,
+        })
+        .reply(500);
+
+      await expect(events.fetchByGoogleId(googleId)).rejects.toThrow();
+    });
+
+    test('Should return null when squidex returns an empty array', async () => {
```
peterstarling, line 589: shouldn't these two methods behave the same when the event isn't found?

https://github.com/yldio/asap-hub/pull/691/files#diff-31921fb03563a6ca9a330c07a76c3cb5b9d69de98cd20c01997542447353f47bR379

apps/asap-server/test/utils/sync-google-calendar.test.ts
```diff
@@ -17,6 +17,16 @@ jest.mock('googleapis', () => ({
   },
 }));
 
+jest.mock('luxon', () => ({
+  DateTime: {
+    utc: () => ({
+      plus: () => ({
+        toISO: () => 'six-months-from-now',
```
peterstarling, line 24: I would probably stick to mocking the Date.now constructor, like in some other files. What you do here is you mock the method at the end of the method chain, which is wrong for two reasons:
1) You don't really allow for any refactoring, if somebody want's to use a different way to get that date they have to change the mocks
2) You are not actually testing a very important logic - we don't know if the period was added correctly. I would probably pick a date, make Date.now return it, calculate the +6 months myself and check that the calculated ISO is in line with it

apps/asap-server/test/utils/sync-google-event.test.ts
```diff
@@ -26,12 +50,28 @@ describe('Sync calendar util hook', () => {
...
-  test('Should upsert event - with dateTime', async () => {
+  test('Should throw when a remote operation throws', async () => {
+    eventControllerMock.fetchByGoogleId.mockRejectedValueOnce(
+      new Error('Squidex'),
+    );
+    await expect(syncEvent(event, defaultCalendarTimezone)).rejects.toThrow();
+
+    eventControllerMock.fetchByGoogleId.mockResolvedValueOnce(restEvent);
+    eventControllerMock.update.mockRejectedValueOnce(new Error('Squidex'));
+    await expect(syncEvent(event, defaultCalendarTimezone)).rejects.toThrow();
```
peterstarling, line 65: I know its a lot of copy and paste but I think we should split this into 3 different tests, its not a HUGE deal but generally speaking a single test should only test one piece of logic (some people even say that there should only be one assertion per test but thats crazy)

apps/asap-server/test/utils/sync-google-event.test.ts
```diff
@@ -69,56 +110,53 @@ describe('Sync calendar util hook', () => {
...
-      await syncEvent(
-        { ...event, summary: undefined },
-        defaultCalendarTimezone,
-      );
-      expect(eventControllerMock.upsert).not.toHaveBeenCalled();
+    test('Should reject when validation fails - missing fields', async () => {
+      await expect(
+        syncEvent({ ...event, id: undefined }, defaultCalendarTimezone),
+      ).rejects.toThrow();
+      await expect(
```
peterstarling, line 138: same as previously, we probably shouldn't have multiple invocations of that function and multiple assertions all in one test. 

I do realise how repetitive that might be, you may want to use `test.each` instead as part of the test will always be the same so its really just about testing different input

peterstarling (changes requested):

apps/asap-server/src/utils/sync-google-calendar.ts
johnytiago: The native one is a pattern we are used to see and know.

apps/asap-server/src/utils/sync-google-calendar.ts
johnytiago: I could make this a RestEvent but I had plans to make this generic and extract this piece of code. 🤷🏻‍♂️

apps/asap-server/test/utils/sync-google-calendar.test.ts
johnytiago: Agreed. I will change it

apps/asap-server/src/utils/sync-google-calendar.ts
peterstarling: nah, im not convinced, i stand by what i said, mixing two different date constructors literally in two lines of code does not sound like a good practice, especially since we know at this point that using `Date` constructor with `new` is not testable.

apps/asap-server/src/utils/sync-google-calendar.ts
peterstarling: well yeah you could just give it a return type that is not directly linked to our data store, something generic.

peterstarling (approved):

apps/asap-server/src/utils/sync-google-calendar.ts
johnytiago: IT's a fixed date. That date does not change

PR #692, opened 2021-03-04
feat: Don't truncate titles on event details page

kmaid:


apps/storybook/src/EventCard.stories.tsx
```diff
@@ -12,6 +12,7 @@ export default {
 
 const props = (): ComponentProps<typeof EventCard> => ({
   ...createEventResponse(),
+  title: text('Event Name', 'Example Event'),
```
jeysal, line 15: Don't think this is necessary since this behavior can already be tested in the lower level `EventInfo.stories`. Can you instead add a knob for the new `Title Character Limit` to `EventInfo.stories`? Then lgtm

apps/storybook/src/EventInfo.stories.tsx
```diff
@@ -9,10 +9,11 @@ export default {
 };
 
 const props = (): ComponentProps<typeof EventInfo> => ({
-  thumbnail: text('Thumbnail', 'https://via.placeholder.com/150'),
+  thumbnail: text('Thumbnail', 'https://placekitten.com/150/150'),
```
kmaid, line 12: 🐈

jeysal (approved):
Good enough ^^

PR #698, opened 2021-03-05
feat: event 'hidden' flag

peterstarling:
Add a new 'hidden' flag to the event entity plus enables the filter inside event control so they are removed from the response by default.

see: https://trello.com/c/DSugpat0/1064-as-an-admin-i-want-to-be-able-to-hide-cms-events-from-fe-and-keep-them-hidden

### DIFF
--- a/apps/asap-server/src/controllers/events.ts
+++ b/apps/asap-server/src/controllers/events.ts
@@ -54,6 +54,8 @@ export default class Events implements EventController {
         [],
       );
 
+    filters.push('data/hidden/iv ne true');
+
     if (after) {
       filters.push(`data/endDate/iv gt ${after}`);
     }
--- a/packages/squidex/schema/schemas/events.json
+++ b/packages/squidex/schema/schemas/events.json
@@ -412,7 +412,85 @@
           "label": "Status",
           "hints": "",
           "placeholder": "",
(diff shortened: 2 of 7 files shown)


packages/squidex/schema/schemas/events.json
```diff
@@ -342,7 +342,7 @@
         ]
       },
       {
-        "name": "Hidden",
+        "name": "hidden",
```
peterstarling, line 345: the 'Hidden' property never should have got into master however it did as part of:
https://github.com/yldio/asap-hub/commit/13c01e96218e013a0dbd9939bfca47f386fa8f17

apps/asap-server/src/controllers/events.ts
```diff
@@ -54,6 +54,8 @@ export default class Events implements EventController {
         [],
       );
 
+    filters.push('data/hidden/iv eq false');
```
johnytiago, line 57: If you use `ne true` we don't need to migrate the old events. Does `eq false` match the `null` as well?

peterstarling: damn, i totally didn't think of that, nice one!

johnytiago (approved):
LGTM

PR #702, opened 2021-03-05
docs: Update readme with onboarding details

migcarva:
This PR does the following changes:
- updates the README with a "Setting up your development environment" section, with more details on how to get a new user running
- adds a script to run the frontend app from the project's root

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,27 @@ For a list of individual package and particularly app scripts, look inside the r
 - `yarn test` - This will lint and test all packages and apps in the repository. You may want to run this with `--watch` or other [Jest CLI options](https://jestjs.io/docs/en/cli.html).
 - `yarn test:*` - There are further test configurations being run on CI that are usually slower or more specific and thus not suitable to run during day-to-day development. You can execute those scripts manually. Some of them may require installing native modules via `yarn rebuild` first.
 
+## Setting up your development environment
+
+For you, a newcomer, to be running your development setup, you'll need to complete the following steps:
+#### Create a new user
+- Log in to https://cloud.squidex.io/app/asap-hub-dev
+- In `Content/Users`, create a new user
+- While creating, or editing the user, on `Functional/Connections`, add a new item to store your UUID (you can generate a random one [here](https://www.uuidgenerator.net/version4)) and save that (the save button is the blue one on top!). You'll need that UUID later, so save it.
+
+#### Get everything running
+- create a `.env` file and update it with the necessary details (ask someone for this)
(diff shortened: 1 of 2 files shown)


kmaid (approved):
LGTM

package.json
```diff
@@ -23,6 +23,7 @@
     "watch:typecheck": "yarn build:typecheck -w",
     "start": "yarn workspaces foreach -vpi -j 1000 run start",
     "start:backend": "yarn workspace @asap-hub/asap-server start:server",
+    "start:frontend": "yarn workspace @asap-hub/frontend start",
```
jeysal, line 26: exactly what I have been running as well :sweat_smile: 
could also do storybook?

README.md
```diff
@@ -31,6 +31,27 @@ For a list of individual package and particularly app scripts, look inside the r
 - `yarn test` - This will lint and test all packages and apps in the repository. You may want to run this with `--watch` or other [Jest CLI options](https://jestjs.io/docs/en/cli.html).
 - `yarn test:*` - There are further test configurations being run on CI that are usually slower or more specific and thus not suitable to run during day-to-day development. You can execute those scripts manually. Some of them may require installing native modules via `yarn rebuild` first.
 
+## Setting up your development environment
+
+For you, a newcomer, to be running your development setup, you'll need to complete the following steps:
+#### Create a new user
```
jeysal, line 37: heading levels out of order :D

jeysal (commented):
The bit about user creation is nice, no clue how Piotr managed to do that, maybe with help :P

package.json
migcarva: I'm using TMUX to have all those tabs open. It's a bit too much :D
Should we just use: https://github.com/kimmobrunfeldt/concurrently ??

jeysal: Also using tmux. I would say everyone can decide that bit themselves, I think it's personal preference. I personally wouldn't use a script that spawn things in parallel because I want to have the output streams separately. Also, you always need different scripts active depending on what you're doing

README.md
migcarva: fixed

package.json
migcarva: Yes, I do agree. I only see a use for it to run both babel and typechecking, but you can just use `&&` :D

migcarva: Srotybook added

jeysal: Yeah and mixing Babel output into typecheck would suck, typecheck needs to be readable

johnytiago (approved):

PR #704, opened 2021-03-05
build: do not require `build` before `test`

jeysal:
This is especially nice for CI parallelization.
Piotr in particular will be very pleased :D


.gitlab-ci.yml
```diff
@@ -96,17 +96,17 @@ build:native:
   needs: []
   stage: test
 
-test:packages:
+check:packages:
```
jeysal, line 99: These used to be named `check` when I created them, I named them that again because with more and more test scripts it's otherwise hard to see which ones are actually running tests and which are just checking something else

.gitlab-ci.yml
```diff
@@ -202,6 +205,7 @@ deploy:sls:production:
     - build:native
     - test:unit
     - test:browser
+    - test:build-output
     - sync:squidex:production
```
jeysal, line 209: Ideally we would also require `verify:development` here so that master runs a sequential
`deploy dev - verify dev - deploy prod - verify prod` and prod deployment is prevented if dev verification fails. But requires a bit of stage restructuring

jeysal: Also totally not related to the change I'm making

apps/auth0-rules/package.json
```diff
@@ -20,7 +20,6 @@
     "@babel/preset-env": "^7.12.11",
     "@babel/preset-typescript": "^7.12.7",
     "@babel/runtime-corejs3": "^7.12.5",
-    "@types/asap-hub__auth0-rule": "workspace:*",
```
jeysal, line 23: That typings package was very irregular and caused problems with the mapping, I killed it and moved its typings somewhere better

jest/jest-base.config.js
```diff
@@ -1,18 +1,34 @@
...
+      'babel-jest',
+      { configFile: require.resolve('../babel-cjs.config.js') },
+    ],
+  },
 
+  moduleNameMapper: {
+    '^@asap-hub/([^/]+)$': '@asap-hub/$1/src',
+  },
   modulePathIgnorePatterns: [
-    '<rootDir>/build([^/]*)/',
```
jeysal, line 13: Not sure if these ever worked properly since the root dir is different in every project. This should be more stable

.gitlab-ci.yml
peterstarling: yeah that the plan, see https://trello.com/c/FMU3iWrH/933-basic-e2e-smoke-tests

the tickets is now done but ill separate that bit from it which is not finished and it will be in order to do what you said

peterstarling (approved):
to me this is very much LGTM

PR #705, opened 2021-03-08
chore: add e2e verification step before prod deployment

peterstarling:
Includes a verification step after dev deployment and before the production deployment

see: https://trello.com/c/e6yfHIr1/1076-add-e2e-test-to-deployment-pipeline


jeysal:
LOL this is exactly what I tried to include in my PR though and why it failed and then I decided to just focus on the `no build before test` for now :joy: 
The stage order must be respected for ... reasons. Idk what stages are supposed to be useful for anyway, just a DAG of jobs like e.g. in Gradle would make more sense IMO but yeah

peterstarling:
> LOL this is exactly what I tried to include in my PR though and why it failed and then I decided to just focus on the `no build before test` for now 
> The stage order must be respected for ... reasons. Idk what stages are supposed to be useful for anyway, just a DAG of jobs like e.g. in Gradle would make more sense IMO but yeah

yeah i was hoping for it to work exactly like circleci but oh well

johnytiago (approved):

PR #709, opened 2021-03-09
feat(events): add locked states for past events

johnytiago:
https://trello.com/c/aEpe05ua/1060-marking-meeting-materials-of-an-event-as-permanently-unavailable-tick-box


apps/asap-server/src/controllers/events.ts
```diff
@@ -174,8 +174,20 @@ flatData{
   status
   tags
   title
-  calendar{
-    flatData{
+  notesLocked
+  notes
```
johnytiago, line 178: notes were not being fetched for events 🤦🏽

packages/squidex/schema/schemas/events.json
```diff
@@ -80,7 +80,7 @@
         }
       },
       {
-        "name": "MeetingMaterials",
```
johnytiago, line 83: Renamed this prop since it was messing up the GraphQL schema on squidex 🙄

apps/asap-server/test/controllers/events.test.ts
```diff
@@ -390,6 +390,95 @@ describe('Event controller', () => {
...
 
+  describe('Past event materials states', () => {
+    const eventId = 'group-id-1';
+    const eventBase = JSON.parse(
+      JSON.stringify(
+        fetchEventsResponse.data.queryEventsContentsWithTotal.items[0],
+      ),
+    );
+
+    test('Should return locked details if details are locked on the CMS', async () => {
```
peterstarling, line 401: to me `locked` means that whatever the value was set it is locked in place and no longer susceptible to change, whereas the actual state this represents is `permanently unavailable`. I'd definitely prefer to change it `permanentlyUnavailable` or just `unavailable` (still better than locked imo)

peterstarling: also I think the test should say that once the features is marked as `locked` (hopefully soon changed to `permanently-unavailable`) the value returned is `null`. I still don't see how `lock` and `null` are related. Making changes to this logic in the feature will be much easier if you follow these two comments, I think.

apps/asap-server/src/entities/event.ts
```diff
@@ -33,26 +30,58 @@ export const parseGraphQLEvent = (item: GraphqlEvent): EventResponse => {
...
+    presentationLocked,
+    meetingMaterialsLocked,
+    notes,
+    videoRecording,
+    presentation,
+    meetingMaterials,
+  } = item.flatData!;
+
+  const notesRes =
+    notesLocked || (!notes && isStale)
```
peterstarling, line 48: why don't use extract this logic into some other function and reuse it across all 4? I feel like it could be refactored as well. Something like...
```typescript
const getVal = (isPermUnavailable: boolean, val?: T): T | null | undefined => {
   if (isPermUnavailable === true) {
      return null;
   }

   if (!va; && isStale) {
  (....)
}

```

apps/asap-server/test/controllers/events.test.ts
```diff
@@ -390,6 +390,95 @@ describe('Event controller', () => {
...
+      expect(result).toEqual({
+        ...eventResponse,
+        notes: null,
+        videoRecording: null,
+        presentation: null,
+        meetingMaterials: null,
+      });
+    });
+
+    test('Should return locked details if it enough time as passed and details are empty', async () => {
```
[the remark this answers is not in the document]
peterstarling: Also how do you mean 'enough time has passed'? Does the fixture have an end date which is in the past? I would still probably replace the endDate with some fixed time inside the test just to make it obvious

apps/asap-server/test/controllers/events.test.ts
```diff
@@ -390,6 +390,95 @@ describe('Event controller', () => {
...
+        ...eventResponse,
+        endDate: now,
+        notes: undefined,
+        videoRecording: undefined,
+        presentation: undefined,
+        meetingMaterials: [],
+      });
+    });
+  });
+
```
peterstarling, line 481: You seem to be missing a test for when a value is actually returned (not `locked`, not stale, not undefined, just present) - unless it's already somewhere else?

peterstarling: You could also test what happens when the `locked` flags or the materials themselves are not present in the flatData, as I believe will be the case for older events before we added these fields?

peterstarling (changes requested):

apps/asap-server/test/controllers/events.test.ts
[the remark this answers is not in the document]
johnytiago: Yeah, I can make that obvious. IT's 10 days now.

apps/asap-server/test/controllers/events.test.ts
johnytiago: The first one should be covered, since the coverage is 100% for the entity. I will see if I need to make the assertion explicit

peterstarling (approved):

PR #715, opened 2021-03-10
chore: Apply correct order to squidex sync in ci

peterstarling:
This puts the `squidex sync:dev` step before the dev deployment and adds a dependency on the sync itself thus making it required to run the entire pipeline.

### DIFF
--- a/ci/.gitlab-ci.squidex.yml
+++ b/ci/.gitlab-ci.squidex.yml
@@ -26,12 +26,12 @@ setup:squidex-image:
     - sq config add asap-hub $SQUIDEX_CLIENT_ID $SQUIDEX_CLIENT_SECRET -u $SQUIDEX_BASE_URL
     - sq config use asap-hub
     - sq sync in packages/squidex/schema -t schemas -t rules
-  stage: prepare
   rules:
     - if: $CI_COMMIT_BRANCH == 'master'
 
 sync:squidex:dev:
   <<: *tmpl_sync
+  stage: prepare-dev
   rules:
     - if: $CI_COMMIT_BRANCH == 'master'
       when: manual
@@ -39,5 +39,6 @@ sync:squidex:dev:
 
 sync:squidex:production:
   <<: *tmpl_sync
+  stage: prepare-prod
   environment:
     name: production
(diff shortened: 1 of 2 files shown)


ci/.gitlab-ci.squidex.yml
```diff
@@ -26,18 +26,15 @@ setup:squidex-image:
     - sq config add asap-hub $SQUIDEX_CLIENT_ID $SQUIDEX_CLIENT_SECRET -u $SQUIDEX_BASE_URL
     - sq config use asap-hub
     - sq sync in packages/squidex/schema -t schemas -t rules
-  stage: prepare
   rules:
     - if: $CI_COMMIT_BRANCH == 'master'
 
 sync:squidex:dev:
   <<: *tmpl_sync
-  rules:
```
johnytiago, line 35: Can we hold on to removing this until we have PR Squidex Env?
For now just running this every time you merge to master, is a win

peterstarling: do you mean you would like me to leave it as a manual job? or remove the dependency? because in this PR i created the dependency [here](https://github.com/yldio/asap-hub/pull/715/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R190) which means that since the squidex-dev job is manual the whole pipeline will be held until it's approved.

johnytiago: Sorry I wasn't clear. What I mean is that this job should only run when in master. You can remove the manual trigger and don't allow it to fail. Just don't make it run as part of every PR, otherwise we'll smash each others dev schema work in progress.

Once each PR runs on an isolated environment then we can just run this every time since you'll not be messing with someone's work.

peterstarling: oh you probably missed it but it already has a master only filter that it inherits from `tmpl_sync` template, i removed the rule that was overwriting it and adding a manual approval

johnytiago: You are damn right. I missed that part. Given this is the case I'm resolving this.
Yet, it is strange that your `tmpl:sync` differs from the one in master but the change don't appear in the github diff.

johnytiago: Maybe that's just because you have a merge base that is older than the commit I meanwhile merged to master fixing the sync template and git can merge the changes no prob.
But for the record I'm referring to using the right app name on the sync script https://github.com/yldio/asap-hub/blob/master/ci/.gitlab-ci.squidex.yml#L26

johnytiago (approved):

PR #717, opened 2021-03-10
feat: Improve error reporting and logging

peterstarling:


johnytiago:
I think we might be over engineering the problem here.
Have you thought of just using and express logger? https://github.com/pinojs/express-pino-logger
I don't really like to see logger dependencies being explicitly injected all the way to the controller.

peterstarling:
> I think we might be over engineering the problem here.
> Have you thought of just using and express logger? https://github.com/pinojs/express-pino-logger
> I don't really like to see logger dependencies being explicitly injected all the way to the controller.

hmmm... why is it over-engineered? Haven't we been using debug logger all over the place? I'd argue that using logging library without DI lowers the quality of the code. I'd often like to assert that an error was logged for instance. But yeah we don't obviously need to inject `loggerMock` in our tests if we don't want to, we can just use winston logger and turn off logging in tests entirely, but I chose to inject mock for the reasons stated earlier. 

I've looked into this pino thing as well as pino-express but I chose more lightweight and more popular library. I'd probably go for https://www.npmjs.com/package/express-winston but it doesn't seem to provide anything more than we already have with the lightweight implementation that I used. Plus we are still having to adjust the transports for lambda anyway.

apps/asap-server/.eslintrc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   rules: {
     'no-use-before-define': 'off',
     'no-unused-vars': 'off',
+    'lines-between-class-members': 'off',
```
peterstarling, line 5: the reason i want to remove it is because it moans about class properties not being separated with spaces which is crazy imho

johnytiago:
I've given it another thought. The bit I was considering over-engineer is just what it is. If we want to have a request-contextualised logger in the controllers and utils, then we need to inject it as a dependency. 🤷🏽‍♂️  I think having DI everywhere, for everything is not pretty, considering that some of this dependencies are very easy to mock/test, but that is a whole other debate.

Yet, I feel like we are still not addressing the problem the right way. We are using express, yet we are not taking advantage of any of its features. If you use the express version of Winston you don't need to directly inject the logger as a dependency to the error handler middleware, or other middleware we might use.

The other thing I'm not particularly keen on is Winston's defaults. It's a big lib, with a lot of features, most of which we don't use and therefore require us [too make considerable boiler-plating just to get JSON logged to the console](https://github.com/yldio/asap-hub/pull/717/files#diff-c929d43a53daad0a15eab7daad6d7f76a9677f3b4abe65250510c6a73427b19aR12). Whereas Pino's default and interface is much much friendlier.

apps/asap-server/.eslintrc.js
johnytiago: Yeah, I don't understand this rule either

peterstarling:
> I've given it another thought. The bit I was considering over-engineer is just what it is. If we want to have a request-contextualised logger in the controllers and utils, then we need to inject it as a dependency. 🤷🏽‍♂️ I think having DI everywhere, for everything is not pretty, considering that some of this dependencies are very easy to mock/test, but that is a whole other debate.
> 
> Yet, I feel like we are still not addressing the problem the right way. We are using express, yet we are not taking advantage of any of its features. If you use the express version of Winston you don't need to directly inject the logger as a dependency to the error handler middleware, or other middleware we might use.
> 
> The other thing I'm not particularly keen on is Winston's defaults. It's a big lib, with a lot of features, most of which we don't use and therefore require us [too make considerable boiler-plating just to get JSON logged to the console](https://github.com/yldio/asap-hub/pull/717/files#diff-c929d43a53daad0a15eab7daad6d7f76a9677f3b4abe65250510c6a73427b19aR12). Whereas Pino's default and interface is much much friendlier.

ok, ill change to pino, you might be right it may not require this stupid console log formatter to work with lambda out of the box so its worth a try. With winston the problem was the missing request-id. 

But we still need to pass the dependency everywhere the logger is used, no? I think that's what you said, that you agree with that bit.

Request contextualised - do you mean a logger instance per request? yeah im not actually doing it know, im only relying on lambdas to print the request ID next to the log line. Are you thinking we should inject the request ID somewhere?

johnytiago:
Cool 🙌🏽 The initialisation of the logger is probably get simpler that way. 

Regarding requestId, feel free to use whichever you prefer. Pino also generates reqIds and rotates them by default. Eventually, when we grab tracing we might want to revisit this and make traces use the same request ID as the logs, but for now either one will do.

I thought you were injecting the logger dependency because you need it to contain the requestID, so that the logs are request-contextualised. If that is not the case, then I don't see the reason to have it as dependency. We can just import it straight from the utils. And I'm not sure about this, but we might even get context into the logger if we use async-hooks https://itnext.io/request-id-tracing-in-node-js-applications-c517c7dab62d.

peterstarling:
> Cool 🙌🏽 The initialisation of the logger is probably get simpler that way.
> 
> Regarding requestId, feel free to use whichever you prefer. Pino also generates reqIds and rotates them by default. Eventually, when we grab tracing we might want to revisit this and make traces use the same request ID as the logs, but for now either one will do.
> 
> I thought you were injecting the logger dependency because you need it to contain the requestID, so that the logs are request-contextualised. If that is not the case, then I don't see the reason to have it as dependency. We can just import it straight from the utils. And I'm not sure about this, but we might even get context into the logger if we use async-hooks https://itnext.io/request-id-tracing-in-node-js-applications-c517c7dab62d.

We can't inject request ID like that because the instantiation of the app happens outside of the handler so it's the same logger instance per every request anyway. Like I said earlier we could create some sort of global variable which holds request Id (and, potentially, some metadata).

I would like to inject logger as dependency because it then makes it testable (i would like to assert that an error is logged in some of your handlers) but I see your point, we could just export an instance as a module and simply import that without having to use the DI at all.

peterstarling:
> Cool 🙌🏽 The initialisation of the logger is probably get simpler that way.
> 
> Regarding requestId, feel free to use whichever you prefer. Pino also generates reqIds and rotates them by default. Eventually, when we grab tracing we might want to revisit this and make traces use the same request ID as the logs, but for now either one will do.
> 
> I thought you were injecting the logger dependency because you need it to contain the requestID, so that the logs are request-contextualised. If that is not the case, then I don't see the reason to have it as dependency. We can just import it straight from the utils. And I'm not sure about this, but we might even get context into the logger if we use async-hooks https://itnext.io/request-id-tracing-in-node-js-applications-c517c7dab62d.

unfortunately the exact same thing happens with pino:
image
and the solution is just as ugly as with winston (having to swap stdout for regular console logs). But we can still stick to pino if you want, ill add the console bit as suggested [here](https://github.com/pinojs/pino/issues/648#issuecomment-506786115) and see if it works.

Don't worry about async hooks for now - we don't need it in lambdas and as far as I know we are not getting off lambdas any time soon, are we?

Anyway would you like me to import logger as a module and use it without the DI? I think then I could just change the log level to silent so it does not log anything at all in tests. The only thing is - like I mentioned before I would like to assert that the error or warning is logged in some cases so I still might want to inject the logger at least into the event handlers. How do you feel about that?

johnytiago:
Last catch up on this on Monday. We got a few loose ends here that we should address 😉

apps/asap-server/src/middleware/error-handler.ts
```diff
@@ -1,15 +1,18 @@
 import { ErrorRequestHandler } from 'express';
-import debug from 'debug';
 import { isBoom } from '@hapi/boom';
+import { Logger } from 'pino';
 
-const logger = debug('asap-server');
-
-export const errorHandler: ErrorRequestHandler = (err, req, res, next) => {
+export const errorHandlerFactory = (logger: Logger): ErrorRequestHandler => (
```
johnytiago, line 5: Instead of changing this to a factory. ErrorHandler could easily just use the logger on injected on the request e.q: `req.log`

apps/asap-server/src/utils/sync-google-event.ts
```diff
@@ -36,12 +36,12 @@ export const syncEventFactory = (
...
 
     if (error) {
-      logger('Ignored event update, validation error:', error);
+      logger.error('Ignored event update, validation error:', error);
       return Promise.reject(error);
     }
 
     const googleEvent = value as GoogleEvent;
-    logger('google event', googleEvent);
+    logger.info('google event', googleEvent);
```
johnytiago, line 44: Not sure if this `info`s shouldn't be `debug`. They make more sense for debugging purposes. What do you think?

johnytiago:
This is looking great, mate! Good job! 🙌🏽 

> *From the ticket*
> - Logs should be filterable by log level, or other parameter (i.e. `userId`)

Could you investigate how hard it would be for us to add the userId to the log message?

peterstarling:
> This is looking great, mate! Good job! 🙌🏽
> 
> > _From the ticket_
> > 
> > * Logs should be filterable by log level, or other parameter (i.e. `userId`)
> 
> Could you investigate how hard it would be for us to add the userId to the log message?

The code might look great but the output is far worse:
image

It doesn't prefix the log lines with request ID, instead the pino-lambda lib only adds the text message there (not sure what the benefit of that would be anyway).

We could try to add requestId and userId inside the log object with async hooks or something but i feel like it would be easier to just replace the console.stout thing with regular console.log just to get it out

johnytiago:
It seems like we're not using `pino-lambda` properly. We're supposed to initialised the lib with the event & context in order to build proper logs.

image

Yet, doing just this won't solve all our issues. We still need to get request context to the `controllers` and so on. To prevent injecting this logger as a dependency we must use `async_hooks`. I was reading this [blogpost on using async hooks for logging](https://blog.besson.co/nodejs_async_hooks_to_get_per_request_context/) and I was mind blown.

We can catch up on this tomorrow again! 😉

peterstarling:
> It seems like we're not using `pino-lambda` properly. We're supposed to initialised the lib with the event & context in order to build proper logs.
> 
> image
> 
> Yet, doing just this won't solve all our issues. We still need to get request context to the `controllers` and so on. To prevent injecting this logger as a dependency we must use `async_hooks`. I was reading this [blogpost on using async hooks for logging](https://blog.besson.co/nodejs_async_hooks_to_get_per_request_context/) and I was mind blown.
> 
> We can catch up on this tomorrow again!

Yeah I missed that bit about actually tying it to the event and context. 

I think setting up async hooks is not actually such a big deal in express cause there is many available tutorials, it's just the testing part that puts me off a little bit but I have tested it before so it shouldn't be such a big deal either. I think I am leaning more towards async hooks than storing per-request info in some global cache than I did before.

I think there are some performance implications to using async hooks but surely not in our scale.

BTW i dont think it has a lot to do with the DI, you know. You may have realised that we are only creating a single app instance once, outside of the handler, so either way we would have to use some sort of global state to hold the current requestId which btw wouldn't work if we moved away from lambdas. Async-hooks are definitely a way to go, DI or not.

peterstarling:
actually mate I dont even think we need async hooks for now:
image

The request Id and all the request info are all inside the log object, we could try to pick the user info later if we want to but i think this is done

apps/asap-server/src/middleware/error-handler.ts
peterstarling: yeah I mean it could.. people often use this DI pattern and attach libs to the request object. But to be fair i prefer this approach as it is much easier to test so I'd prefer to leave it as is

.gitlab-ci.yml
```diff
@@ -8,12 +8,14 @@ include:
     - if: $CI_COMMIT_BRANCH == 'master'
   variables: &tmpl_master_variables
     SLS_STAGE: production
+    LOG_LEVEL: error
```
johnytiago, line 11: Can we move this variable assignments to `serverless.js` like we did to the `DEBUG` variable? https://github.com/yldio/asap-hub/blob/master/serverless.js#L51

peterstarling: there:
https://github.com/yldio/asap-hub/compare/72608c553d667092afcf3caccefa3f3e2fb6143d..09d56336b2b5bb59a987ece7a32dce5d9aa2440d

johnytiago: Perf! 👍🏽

johnytiago:
As of our conversation, please make sure you:
- Hide sensitive information from the logs, such as **authorization headers**;
- Test the error handler to make sure the logger is called, to make use of DI. Or, if you prefer not to, import the logger instance directly from `utils` and the Factory pattern;

peterstarling:
Authorization header is now hidden:

apps/asap-server/src/utils/logger.ts
```diff
@@ -1,4 +1,14 @@
...
+import noir from 'pino-noir';
+import { logLevel } from '../config';
+
+export const redaction = noir(['req.headers.authorization'], '*');
+
+const logger = pino({
+  level: logLevel,
+  formatters: {
+    level: (label: string) => ({ level: label }),
+  },
```
johnytiago, line 11:
```suggestion
  },
  redaction: noir(['req.headers.authorization'], '*'),
```

Is there a reason we can just set redaction by default?

peterstarling: it doesnt work if you add it to the logger directly, i have tried and it looks like they actually recommend adding it to express-pino which is pino-http equivalent

johnytiago (approved):
LGTM 🚀

PR #721, opened 2021-03-10
fix: fetch event timezones

johnytiago:


peterstarling (approved):
The fact that such fix requires no test changes indicates that the test coverage isn't great. But yeah we know that, and so far we've not been able to find a good way to test squidex integration. Just something to think about.

PR #725, opened 2021-03-11
Fix: Team and User rapid search crashes

kmaid:
Stops race condition occurring while searching on Users and Teams. 
Removed a bunch of useHTTP usages.
Removed legacy fixtures
Improved Testing
etc


packages/react-components/src/templates/EventPage.tsx
```diff
@@ -58,7 +58,7 @@ const EventPage: React.FC<EventPageProps> = ({
         <JoinEvent {...props} />
         <EventAbout {...props} />
       </Card>
-      <RichTextCard text={notes} title="Meeting materials" />
+      <RichTextCard text={notes} title="Notes" />
```
kmaid, line 61: Missed requirement in another ticket.

apps/frontend/src/network/__tests__/Network.test.tsx
```diff
@@ -130,10 +118,14 @@ describe('the network page', () => {
...
       fireEvent.click(toggle);
       expect(searchBox.value).toEqual('test123');
-      await waitFor(() => {
-        const { searchParams } = new URL(teamsReq!.path, 'http://api');
-        expect(searchParams.get('search')).toBe('test123');
-      });
+      await waitFor(() =>
+        expect(mockGetTeams).toHaveBeenLastCalledWith(
+          expect.not.objectContaining({
+            filters: expect.anything(),
```
kmaid, line 124: So I am checking filters isn't set but I would also like to check search query is set. Any idea if i can do both together?

apps/frontend/src/network/teams/__tests__/About.test.tsx
```diff
@@ -70,3 +75,20 @@ describe('the proposal', () => {
     expect(getByText(/proposal/i).closest('a')!.href).toMatch(/someproposal$/);
   });
 });
+
+it('renders the team groups card when a group is present', async () => {
```
kmaid, line 79: Missed test and mock causing error in logs

apps/frontend/src/network/__tests__/Network.test.tsx
jeysal: tbh, just get the object out `const [[options]] = mockGetTeams.mock.calls.slice(-1)` and make separate assertions on it. That'll be the most readable I suspect

apps/frontend/src/network/teams/__tests__/About.test.tsx
```diff
@@ -70,3 +75,20 @@ describe('the proposal', () => {
...
+        ...createGroupResponse(),
+        name: 'Example Group 123',
+      },
+    ],
+    total: 1,
+  });
+  const { getByText } = await renderTeamAbout({
+    team: { ...createTeamResponse(), proposalURL: undefined },
+  });
+  await waitFor(() => expect(mockedGetTeamGroups).toHaveBeenCalled());
```
jeysal, line 92: using `await findByText` instead of `getByText` avoids an explicit wait

jeysal (commented):
I'm kind of torn on the missing cov. On the one hand it's not a super critical code path since an error on the list endpoints is not expected and will break the app one way or another, on the other hand we're now at 91% branch cov project-wide, way below the other cov numbers, and it seems very easy to test :/ (put in error boundary rendering something, `mockRejectedValue`, assert that error boundary thing is rendered). wdyt?

apps/frontend/src/network/teams/__tests__/About.test.tsx
kmaid: I couldn't get this to work. I don't think its worth fighting with

jeysal:
fixed that mock matching for you :see_no_evil: slightly annoying that the GetListOptions allow both `undefined` and `[]` meaning the same thing and both are used

jeysal (approved):

PR #726, opened 2021-03-12
feat: event content is stale after 14 days

johnytiago:
This PR is an iteration on the requirements of this ticket: https://trello.com/c/aEpe05ua/1060-marking-meeting-materials-of-an-event-as-permanently-unavailable-tick-box.
- As stated in the comments. The tickbox takes precedence over contente (was already implemented)
- Content is considered permanently unavailable if empty and event is more than 14 days old

### DIFF
--- a/apps/asap-server/src/entities/event.ts
+++ b/apps/asap-server/src/entities/event.ts
@@ -31,7 +31,7 @@ export const parseGraphQLEvent = (item: GraphqlEvent): EventResponse => {
     : undefined;
 
   const endDate = DateTime.fromISO(item.flatData!.endDate!);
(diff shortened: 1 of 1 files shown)


jeysal:
@johnytiago there seems to be inconsistencies in the requirements - in multiple tickets I've seen that this number of days was specified as 10 days, only in the one you linked it seems to be 14 days. Can you clarify with B which number is supposed to be correct and consistently applied everywhere?

johnytiago:
Hey @jeysal, 👋🏽 Yep. This was a requirement change from B after she spoke with the client. More context can be found in the comments.
Do you have any reference to the 10 days in the FE I should be concerned with?

jeysal:
The number is nowhere in code yet, so don't worry. I'll adjust a few more tickets

PR #727, opened 2021-03-12
build(deps): update TS 3=>4

jeysal:
and CRA to a higher patch as a requirement for TS 4 support


apps/asap-server/src/entities/group.ts
```diff
@@ -15,11 +14,9 @@ import { parseDate, createURL } from '../utils/squidex';
 
 export const parseGraphQLGroup = (item: GraphqlGroup): GroupResponse => {
   const createdDate = parseDate(item.created).toISOString();
-  const teams: GroupTeam[] = (item.flatData?.teams || []).map((t) => {
-    const team: TeamResponse = parseGraphQLTeam(t, []);
-    delete team.members;
```
jeysal, line 20: @johnytiago @peterstarling older TS did not catch this - deleting `team.members` is not allowed because `members` is a mandatory property on `GroupTeam`. Is this safe to remove or why was this in there?

jeysal: @johnytiago @peterstarling ping. I would assume it's not necessary since we're literally just deleting data in conflict with type declarations? And deployment seems to work

jeysal: Oh, the tests fail. I guess I'll throw in destructuring instead of delete, don't need code knowledge for that :sweat_smile:

PR #728, opened 2021-03-12
fix: Avatar iv null bug

peterstarling:
https://trello.com/c/FBMD06GE/1104-team-lee-is-breaking-network


kmaid:
Maybe this could use a test?

johnytiago:
It does @kmaid. I'm on it! 😉

kmaid:
Are responses from Squidex typed? That probably needs fixed too

peterstarling:
i thought we were monkey patching to get that fix out of the way. Proper fix requires not just a test but also changes to our typings which will cascade a series of type errors that need to be addressed

peterstarling:
et voila, here is the test

apps/asap-server/src/controllers/teams.ts
```diff
@@ -94,7 +94,11 @@ const transformRestTeamMember = (
     role: get(user, 'data.teams.iv', []).find(
       (t: { id: string[] }) => t.id[0] === teamId,
     ).role,
-    avatarUrl: user.data.avatar && createURL(user.data.avatar.iv)[0],
+    avatarUrl:
+      (user.data.avatar &&
+        user.data.avatar.iv &&
+        createURL(user.data.avatar.iv)[0]) ||
+      undefined,
```
johnytiago, line 101:
```suggestion
      user.data.avatar?.iv ?  createURL(user.data.avatar.iv)[0]) : undefined,
```

johnytiago (approved):

PR #729, opened 2021-03-12
Adding CTA (contact PM) to group about template

migcarva:

- adds a CtaCard to GroupProfileAbout template
- checks whether there's a Group as one or more leaders with the role "Project Manager", and if there's any show the Call to Action, providing the PM's email to the mailTo link
- changes the signature of the createMailTo function, allowing for a single email (string) or a list of emails to be passed
- extracts two functions from the above function, to create both a URL to a single email or a list of emails


kmaid:
Please raise this as a pull request on this repo so CI runs!

PR #730, opened 2021-03-12
refactor: extract routing to package

jeysal:
Even the unfinished draft is pretty much un-reviewable — we can go through how the new system works once it's ready.

I noticed while starting to build a routing system that pretty much exactly what I was building was already built by someone, so I used that package. It's a bit obscure (little use), but any code I would write would be, too, and I didn't want to NIH, esp. considering that it would have been quite some effort. We can always fork if it becomes necessary.

This will need a lot of manual testing.

Some hrefs are still left, e.g. when the href prop also indicates whether an edit link should be shown at all, or in weird cases like the terms/privacy policy where the routing would need a rework anyway. This does not claim completeness, but the important bits in the complex sections, in particular with parameters in routes, are done.

In the future, it might also make sense to use the query param support of `typesafe-routes` to cover cases like `NetworkPage` with its search query param.


jeysal:
Tom and I had a call about it, and tested everything on the live environment

PR #731, opened 2021-03-15
feat: add CTA (Contact PM) to group about template

migcarva:

- adds a CtaCard to GroupProfileAbout template
- checks whether there's a Group as one or more leaders with the role "Project Manager", and if there's any show the Call to Action, providing the PM's email to the mailTo link
- changes the signature of the createMailTo function, allowing for a single email (string) or a list of emails to be passed
- extracts two functions from the above function, to create both a URL to a single email or a list of emails

### DIFF
--- a/packages/react-components/src/__tests__/mail.test.ts
+++ b/packages/react-components/src/__tests__/mail.test.ts
@@ -6,8 +6,18 @@ import {
 } from '../mail';
 
 describe('createMailTo', () => {
-  it('generates a mailto link', () => {
-    expect(new URL(createMailTo('test@example.com')).protocol).toBe('mailto:');
+  it('generates a mailto link, when an single email is provided', () => {
+    const mailTo = new URL(createMailTo('test@example.com'));
+    expect(mailTo.protocol).toBe('mailto:');
+    expect(mailTo.href).toBe('mailto:test@example.com');
+  });
+
+  it('generates a mailto link, when multiple emails are provided', () => {
+    const mailTo = new URL(
+      createMailTo(['test@example.com', 'test@example.com']),
+    );
+    expect(mailTo.protocol).toBe('mailto:');
+    expect(mailTo.href).toBe('mailto:test@example.com,test@example.com,');
   });
 
   it('escapes the email address', () => {
--- a/packages/react-components/src/__tests__/mail.test.ts
+++ b/packages/react-components/src/__tests__/mail.test.ts
@@ -9,13 +9,13 @@ describe('createMailTo', () => {
   it('generates a mailto link, when an single email is provided', () => {
     const mailTo = new URL(createMailTo('test@example.com'));
     expect(mailTo.protocol).toBe('mailto:');
-    expect(mailTo.href).toBe('mailto:test@example.com');
+    expect(mailTo.pathname).toBe('test@example.com');
   });
 
   it('generates a mailto link, when a list with a single email is provided', () => {
     const mailTo = new URL(createMailTo(['test@example.com']));
     expect(mailTo.protocol).toBe('mailto:');
-    expect(mailTo.href).toBe('mailto:test@example.com');
+    expect(mailTo.pathname).toBe('test@example.com');
   });
 
   it('generates a mailto link, when multiple emails are provided', () => {
@@ -23,7 +23,7 @@ describe('createMailTo', () => {
       createMailTo(['test@example.com', 'test@example.com']),
     );
     expect(mailTo.protocol).toBe('mailto:');
-    expect(mailTo.href).toBe('mailto:test@example.com,test@example.com');
+    expect(mailTo.pathname).toBe('test@example.com,test@example.com');
   });
 
   it('escapes the email address', () => {
--- a/packages/react-components/src/__tests__/mail.test.ts
+++ b/packages/react-components/src/__tests__/mail.test.ts
@@ -12,6 +12,12 @@ describe('createMailTo', () => {
     expect(mailTo.href).toBe('mailto:test@example.com');
   });
 
+  it('generates a mailto link, when a list with a single email is provided', () => {
+    const mailTo = new URL(createMailTo(['test@example.com']));
+    expect(mailTo.protocol).toBe('mailto:');
+    expect(mailTo.href).toBe('mailto:test@example.com');
+  });
+
   it('generates a mailto link, when multiple emails are provided', () => {
     const mailTo = new URL(
       createMailTo(['test@example.com', 'test@example.com']),
--- a/packages/react-components/src/__tests__/mail.test.ts
+++ b/packages/react-components/src/__tests__/mail.test.ts
@@ -23,7 +23,7 @@ describe('createMailTo', () => {
       createMailTo(['test@example.com', 'test@example.com']),
     );
     expect(mailTo.protocol).toBe('mailto:');
-    expect(mailTo.href).toBe('mailto:test@example.com,test@example.com,');
+    expect(mailTo.href).toBe('mailto:test@example.com,test@example.com');
   });
 
   it('escapes the email address', () => {
--- a/packages/react-components/src/mail.ts
+++ b/packages/react-components/src/mail.ts
@@ -4,12 +4,13 @@ interface MailOptions {
 }
 
 export const createMailTo = (
-  email: string,
+  emails: string | string[],
   { subject, body }: MailOptions = {},
 ): string => {
-  const mailTo = new URL(
-    `mailto:${email.split('@').map(encodeURIComponent).join('@')}`,
-  );
+  const mailTo =
+    typeof emails === 'string'
+      ? convertEmailToMailTo(emails)
+      : convertEmailListToMailTo(emails);
 
   if (subject) mailTo.searchParams.set('subject', subject);
   if (body) mailTo.searchParams.set('body', body);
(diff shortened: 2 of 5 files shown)


packages/react-components/src/__tests__/mail.test.ts
```diff
@@ -6,8 +6,18 @@ import {
 } from '../mail';
 
 describe('createMailTo', () => {
-  it('generates a mailto link', () => {
-    expect(new URL(createMailTo('test@example.com')).protocol).toBe('mailto:');
+  it('generates a mailto link, when an single email is provided', () => {
```
kmaid, line 9: We should be testing an array with one item

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -32,14 +35,33 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
...
   membersSectionId,
-}) => (
-  <div css={styles}>
-    <GroupInformation tags={tags} description={description} />
-    <GroupTools calendarId={calendars[0] && calendars[0].id} tools={tools} />
-    <div id={membersSectionId}>
-      <GroupMembersSection teams={teams} leaders={leaders} />
+}) => {
+  const pmsEmails = leaders
+    .filter((pm) => pm.role.match(/project manager/i))
```
kmaid, line 40: Should not me fuzzy matching on enum/union

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -32,14 +35,33 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
...
+    <div css={styles}>
+      <GroupInformation tags={tags} description={description} />
+      <GroupTools calendarId={calendars[0] && calendars[0].id} tools={tools} />
+      <div id={membersSectionId}>
+        <GroupMembersSection teams={teams} leaders={leaders} />
+      </div>
+      {pmsEmails.length !== 0 && (
+        <CtaCard
+          href={
+            pmsEmails.length === 1
```
kmaid, line 53: This should be dealt with by the helper function

kmaid (changes requested):

packages/react-components/src/mail.ts
```diff
@@ -4,12 +4,13 @@ interface MailOptions {
...
 export const createMailTo = (
-  email: string,
+  emails: string | string[],
   { subject, body }: MailOptions = {},
 ): string => {
-  const mailTo = new URL(
-    `mailto:${email.split('@').map(encodeURIComponent).join('@')}`,
-  );
+  const mailTo =
+    typeof emails === 'string'
```
kmaid, line 11: I would just convert the string into an array with one entry. I would then use join so you don't have the trailing comma problem or require the other two functions

packages/react-components/src/__tests__/mail.test.ts
```diff
@@ -6,8 +6,18 @@ import {
...
+    expect(mailTo.protocol).toBe('mailto:');
+    expect(mailTo.href).toBe('mailto:test@example.com');
+  });
+
+  it('generates a mailto link, when multiple emails are provided', () => {
+    const mailTo = new URL(
+      createMailTo(['test@example.com', 'test@example.com']),
+    );
+    expect(mailTo.protocol).toBe('mailto:');
+    expect(mailTo.href).toBe('mailto:test@example.com,test@example.com,');
```
kmaid, line 20: Does the format require a trailing comma for multiple addresses?

packages/react-components/src/templates/__tests__/GroupProfileAbout.test.tsx
```diff
@@ -58,3 +58,28 @@ it('assigns given id to the members section for deep linking', () => {
     /members/i,
   );
 });
+
+it('renders a call to action button, when a PM is defined on the group', () => {
```
kmaid, line 62: We should also test that the CTA card is not shown when there are no project managers

packages/react-components/src/__tests__/mail.test.ts
migcarva: it does

packages/react-components/src/mail.ts
migcarva: That's a good idea

packages/react-components/src/templates/GroupProfileAbout.tsx
migcarva: I had just `createMailTo(pmsEmails)`, and that's the best alternative

packages/react-components/src/templates/GroupProfileAbout.tsx
migcarva: Can you give me more details?

packages/react-components/src/__tests__/mail.test.ts
kmaid: mailto:test@example.com,test@example.com,

migcarva: ohh, the final one... :D it's not, ofc

kmaid: Yeah both seem to work. I think having a trailing comma is a bug

packages/react-components/src/templates/__tests__/GroupProfileAbout.test.tsx
migcarva: I actually wrote the test, and deleted it... I saw the case not being tested on the Team Page...

packages/react-components/src/templates/GroupProfileAbout.tsx
kmaid: Something like

```suggestion
const pmsEmails = leaders
    .filter(({role}) => role === "Project Manager")
    .map(({user: {email}}) => email);
```

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -32,14 +35,33 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
...
-}) => (
-  <div css={styles}>
-    <GroupInformation tags={tags} description={description} />
-    <GroupTools calendarId={calendars[0] && calendars[0].id} tools={tools} />
-    <div id={membersSectionId}>
-      <GroupMembersSection teams={teams} leaders={leaders} />
+}) => {
+  const pmsEmails = leaders
+    .filter((pm) => pm.role.match(/project manager/i))
+    .map((pm) => pm.user.email);
```
kmaid, line 41:
```suggestion
const pmsEmails = leaders
	    .filter(({role}) => role === "Project Manager")
	    .map(({user: {email}}) => email);
```

kmaid: I think that should work.

packages/react-components/src/__tests__/mail.test.ts
migcarva: Now we are testing both 3 cases

packages/react-components/src/__tests__/mail.test.ts
migcarva: I thought about it, but was on a thight schedule :)

kmaid: This test should fail

packages/react-components/src/mail.ts
```diff
@@ -4,12 +4,12 @@ interface MailOptions {
...
-  email: string,
+  emails: string | string[],
   { subject, body }: MailOptions = {},
 ): string => {
-  const mailTo = new URL(
-    `mailto:${email.split('@').map(encodeURIComponent).join('@')}`,
-  );
+
+  const list = typeof emails === 'string' ? new Array(emails) : [...emails];
+  const mailTo = convertEmailListToMailToUrl(list);
```
kmaid, line 12:
```suggestion
  const mailTo = `mailto:${list.map((email)=>email.split('@').map(encodeURIComponent).join('@')}).join(',')}`,
```

packages/react-components/src/mail.ts
```diff
@@ -34,3 +34,14 @@ export const mailToSupport = (overrides?: MailOptions): string =>
...
+export const convertEmailListToMailToUrl = (list: string[]): URL => {
+  const href = list.reduce(
+    (a, e, i) => {
+      const suffix = i === list.length - 1 ? '' : ',';
+      return `${a}${e.split('@').map(encodeURIComponent).join('@')}${suffix}`;
+    },
+    '',
+  );
+  return new URL(`mailto:${href}`);
+};
```
kmaid, line 47:
```suggestion

```

kmaid: This isn't required anymore but if it was it shouldn't be exported if nothing else uses it

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -32,14 +35,26 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
...
   membersSectionId,
-}) => (
-  <div css={styles}>
-    <GroupInformation tags={tags} description={description} />
-    <GroupTools calendarId={calendars[0] && calendars[0].id} tools={tools} />
-    <div id={membersSectionId}>
-      <GroupMembersSection teams={teams} leaders={leaders} />
+}) => {
+  const pmsEmails = leaders
+    .filter(({ role }) => role.match(/project manager/i))
```
kmaid, line 40:
```suggestion
    .filter(({ role }) => role ==== "Project Manager" as const)
```

kmaid (changes requested):

packages/react-components/src/__tests__/mail.test.ts
migcarva: I'm pushing new code :)

packages/react-components/src/mail.ts
migcarva: I found it more readable and maintainable to have a function to this. I'm exporting it so it can be used elsewhere and/or for testing (if necessary).

jeysal:
Seeing as this has been in review for quite a while and there were still open points, I addressed the remaining points. I agree the implementation was excessively complicated for what could have been a plain `join`. The tests were mostly good though :+1: I also added a knob in Storybook to be able to preview the two different states

kmaid (approved):

jeysal (dismissed):
Umm, did you mess up a rebase/merge here?

PR #732, opened 2021-03-15
Feat: Collapsable Notes

kmaid:
Conditional Collapse

### DIFF
--- a/packages/react-components/src/organisms/RichTextCard.tsx
+++ b/packages/react-components/src/organisms/RichTextCard.tsx
@@ -1,4 +1,5 @@
 import React, { Fragment } from 'react';
+
 import { Card, Headline2 } from '../atoms';
 import RichText from './RichText';
 import { Collapsible } from '../molecules';
--- a/packages/react-components/src/organisms/RichTextCard.tsx
+++ b/packages/react-components/src/organisms/RichTextCard.tsx
@@ -1,16 +1,30 @@
-import React from 'react';
+import React, { ComponentProps } from 'react';
 import { Card, Headline2 } from '../atoms';
 import RichText from './RichText';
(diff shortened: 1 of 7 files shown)


packages/react-components/src/organisms/RichTextCard.tsx
```diff
@@ -1,17 +1,25 @@
...
+  title,
+  text,
+  collapsible = false,
+}) => {
+  const wrapper = collapsible ? Collapsible : Fragment;
+  return text ? (
     <Card>
       <Headline2 styleAsHeading={3}>{title}</Headline2>
-      <RichText text={text} />
+      {wrapper({ children: <RichText text={text} /> })}
```
jeysal, line 20: You wanna do something like `<Wrapper><RichText text={text} /></Wrapper>` (wrapper variable needs to be uppercase for it to work. This literally calls the component function which is dangerous

packages/react-components/src/templates/__tests__/EventPage.test.tsx
```diff
@@ -45,8 +45,8 @@ it('renders the join event section', () => {
 });
 
 it('renders the event notes', () => {
-  const { getByText } = render(<EventPage {...props} notes="My Notes" />);
-  expect(getByText('My Notes')).toBeVisible();
+  const { getByRole } = render(<EventPage {...props} notes="My Notes" />);
+  expect(getByRole('heading', { name: 'Notes', level: 2 })).toBeVisible();
```
jeysal, line 49: What is `name`? o.O

jeysal (approved):

packages/react-components/src/templates/__tests__/EventPage.test.tsx
kmaid: sneaky way to differentiate between headings

PR #734, opened 2021-03-15
feat(devops): free branch resources when PR is merged

johnytiago:
Here is an example of a successful execution of this action: 
https://github.com/yldio/asap-hub/actions/runs/655679124


johnytiago:
This was just an action test. Closing

peterstarling (approved):
if this is tested and it works then go ahead and merge. As far as I can tell there is no way for it to tear down our prod env so all good

PR #735, opened 2021-03-15
chore: disable express header

johnytiago:
The code is open source, but still ¯\_(ツ)_/¯


peterstarling (approved):
I actually found it quite helpful as we also route some of your paths to non express handlers so I was able to tell which one went to the new one but I guess at this stage it isn't as helpful anymore

PR #736, opened 2021-03-16
fix(events): remove label from meeting details

johnytiago:
https://trello.com/c/LvoyDaov/1081-remove-label-fields-from-additional-meeting-materials-be-and-cms


peterstarling (approved):
im going to assume that this is a breaking change that the front end are fine with?

PR #739, opened 2021-03-16
Fix: Remove extra paddings

kmaid:


jeysal (approved):
okay from a technical side, whether it looks to your/James's liking on every page that uses these is of course on you to ensure :sweat_smile:

kmaid:
James says its only a problem on the team member component but @migcarva will be picking it up shortly.

PR #740, opened 2021-03-17
chore: Improve logging 

peterstarling:
Rearranges most of the log parameter order to include all of the data in the logs, as per documentation: https://github.com/pinojs/pino/blob/master/docs/api.md#error

### DIFF
--- a/apps/asap-server/src/utils/instrumented-framework.ts
+++ b/apps/asap-server/src/utils/instrumented-framework.ts
@@ -1,6 +1,7 @@
 /* istanbul ignore file */
 import * as opentracing from 'opentracing';
 import lightstep from 'lightstep-tracer';
+import { PinoLambdaLogger } from 'pino-lambda';
 import { promisify } from 'util';
 import {
   APIGatewayProxyStructuredResultV2,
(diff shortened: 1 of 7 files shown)


apps/asap-server/src/utils/instrumented-framework.ts
```diff
@@ -30,10 +31,15 @@ if (lightstepToken && NODE_ENV !== 'test') {
...
 export const http = (
   fn: (request: lambda.Request) => Promise<lambda.Response>,
+  logger?: PinoLambdaLogger,
 ) => async (
   event: APIGatewayProxyEventV2,
   context?: Context,
 ): Promise<APIGatewayProxyResultV2> => {
+  if (logger) {
+    // casting as non-null as the lambda always gives context
+    logger.withRequest(event, context!);
```
johnytiago, line 41: The parameter is optional. Could you change that? Because it doesn't look great to have an explicit optional param and then saying it's non-null 😄 Either way, you'd need an eslint rule to be able to get this through.

peterstarling: yeah I changed it just now. To be fair I don't think lambda ever omits these and I think the types are simply invalid. I could change it but then every test using the handler would have to attach context (which is fine imo)

johnytiago (approved):

PR #742, opened 2021-03-17
feat: minor design changes

migcarva:
This PR applies the following changes:
- https://trello.com/c/0wj3BE42/1026-header-profile-image-component
- https://trello.com/c/cyPb0idA/1027-dashboard-discover-asap-card-component
- https://trello.com/c/nN27oHQe/1028-dashboard-discover-page
- https://trello.com/c/JtR8pGxZ/1029-square-image-component

### DIFF
--- a/packages/react-components/src/atoms/Avatar.tsx
+++ b/packages/react-components/src/atoms/Avatar.tsx
@@ -82,6 +82,7 @@ const initialsStyle = css({
 const textStyle = css(fontStyles, headlineStyles[3], {
   dominantBaseline: 'central',
   textAnchor: 'middle',
+  fontSize: '1.42em',
 });
 
 const placeholderColorStyle = css({
--- a/packages/react-components/src/atoms/Avatar.tsx
+++ b/packages/react-components/src/atoms/Avatar.tsx
@@ -82,7 +82,7 @@ const initialsStyle = css({
 const textStyle = css(fontStyles, headlineStyles[3], {
   dominantBaseline: 'central',
   textAnchor: 'middle',
-  fontSize: '1.42em',
+  fontSize: '1.5em',
 });
 
 const placeholderColorStyle = css({
--- a/packages/react-components/src/card.ts
+++ b/packages/react-components/src/card.ts
@@ -3,7 +3,6 @@ import {
   vminLinearCalcClamped,
   mobileScreen,
   tabletScreen,
-  vminLinearCalc,
   largeDesktopScreen,
 } from './pixels';
 
@@ -17,7 +16,7 @@ export const paddingStyles = css({
     24,
     'px',
   )}`,
-  paddingRight: `${vminLinearCalc(
+  paddingRight: `${vminLinearCalcClamped(
     mobileScreen,
     24,
     largeDesktopScreen,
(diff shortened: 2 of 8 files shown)


packages/react-components/src/atoms/Avatar.tsx
```diff
@@ -82,6 +82,7 @@ const initialsStyle = css({
 const textStyle = css(fontStyles, headlineStyles[3], {
   dominantBaseline: 'central',
   textAnchor: 'middle',
+  fontSize: '1.42em',
```
jeysal, line 85: strange number, does this come directly from James? Seems roughly similar to our h3 font size that's the only similar thing I can think of

jeysal (approved):
odd number but other than that lgtm

packages/react-components/src/atoms/Avatar.tsx
migcarva: it's odd... James asked it to be changed to `17pt` :/

kmaid:
```suggestion
  fontSize: `${17/perRem}em`,
```

jeysal: That's `1em` though, so does nothing

migcarva: That's not what we want. This is not a conversion from PX to EM, but rather PT to EM.

jeysal: If you try changing e.g. ticking and unticking border in the Avatar story, you see that the text also scales because the circle becomes a bit smaller and larger

jeysal: Perhaps hop onto a quick call with James explaining this and trying to find a base size that works well in any scaling?

jeysal: The `Avatar` layouting is very intricate (I was the last to significantly touch this a couple of months ago). This is because the `Avatar` can scale to various sizes between its min and max depending on how much space the container gives it, plus the `border` can modify it, ...
The placeholder text really being an SVG that is handled the same way as a real avatar image is the only thing that makes it somewhat manageable, so changing this to behave like normal text would be a major effort in re-doing the layout and I'm not even sure it's possible with pure CSS to make a placeholder avatar always behave like the real image avatar.
So yeah if we explain to James that it scales relative to the Avatar size and settle on a size taking that into account that would be great.

migcarva: I guess James only complaint is the text size on the dashboard page, where the avatar is shown with the smallest size (I guess). The proportion looks better with a smaller text, just the oddity of its size. I've already asked him to check if its possible to get us a more "normal" value for the font size.

Can I merge this and proceed and come back to it later?

jeysal: If it's just about making it bigger for now what about 1.5? Something as specific as this will leave someone at some point in the future wondering why it needs to be this exact value and being afraid to touch it 😅

jeysal (approved):
@migcarva feel free to merge :upside_down_face:

PR #745, opened 2021-03-18
Event Calendar List

kmaid:


packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -58,19 +58,27 @@ const orderList = css({
 interface CalendarListProps {
   calendars: ReadonlyArray<CalendarResponse>;
   singleGroup?: boolean;
```
jeysal, line 60: Looks like these are exclusive, specifying both doesn't make sense, so the interface would need to change. But instead maybe just pass in the headline and a bool for whether to show the explanation text?

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -58,19 +58,27 @@ const orderList = css({
...
   singleGroup = false,
+  singleEvent = false,
 }) => (
   <div css={containerStyles}>
     <Card>
       <div css={headerStyles}>
         <Headline3>
-          Subscribe to {singleGroup ? "this group's" : 'Groups on'} calendar
+          Subscribe to{' '}
+          {singleGroup
```
migcarva, line 73: Can we just change the order of this? check `singleEvent`, then `singleGoup`

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -57,20 +57,24 @@ const orderList = css({
 });
 interface CalendarListProps {
   calendars: ReadonlyArray<CalendarResponse>;
-  singleGroup?: boolean;
+  page: 'calendar' | 'event' | 'group';
+  singleEvent?: boolean;
```
jeysal, line 61:
```suggestion
```

jeysal (approved):
one cleanup then lgtm

PR #746, opened 2021-03-18
feat: add CTA (Contact PM) to group about template -2 

kmaid:

adds a CtaCard to GroupProfileAbout template
checks whether there's a Group as one or more leaders with the role "Project Manager", and if there's any show the Call to Action, providing the PM's email to the mailTo link
changes the signature of the createMailTo function, allowing for a single email (string) or a list of emails to be passed
extracts two functions from the above function, to create both a URL to a single email or a list of emails

### DIFF
--- a/apps/storybook/src/GroupProfileAbout.stories.tsx
+++ b/apps/storybook/src/GroupProfileAbout.stories.tsx
@@ -16,7 +16,15 @@ export const Normal = () => (
         ? createGroupResponse({
             leadPiCount: 2,
             projectManagerCount: 2,
-          }).leaders
+          }).leaders.map(({ user }) => ({
+            user: {
+              ...user,
+              teams: user.teams.map((team) => ({ ...team, href: '#' })),
+            },
+            role: boolean('Has PMs', true) ? 'Project Manager' : 'Chair',
+            href: '#',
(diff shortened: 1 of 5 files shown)


apps/storybook/src/GroupProfileAbout.stories.tsx
```diff
@@ -16,7 +16,14 @@ export const Normal = () => (
...
             leadPiCount: 2,
             projectManagerCount: 2,
-          }).leaders
+          }).leaders.map(({ user }) => ({
+            user: {
+              ...user,
+              teams: user.teams,
+            },
+            role: boolean('Has PMs', true) ? 'Project Manager' : 'Chair',
+            teams: user.teams,
```
jeysal, line 25: don't think this property is a thing?

apps/storybook/src/GroupProfileAbout.stories.tsx
```diff
@@ -16,7 +16,14 @@ export const Normal = () => (
         ? createGroupResponse({
             leadPiCount: 2,
             projectManagerCount: 2,
-          }).leaders
+          }).leaders.map(({ user }) => ({
+            user: {
+              ...user,
+              teams: user.teams,
+            },
```
jeysal, line 23:
```suggestion
            user,
```

PR #749, opened 2021-03-18
test(devops): pipeline duration

johnytiago:
This is just a test to compare the duration of the pipeline if we run bundlewatch as separate job. We don't expect much improvements here, since the script is so fast. 🤷🏽‍♂️


jeysal:
It's not *super* terrible, but not great either. 6 seconds of 40 seconds job time spent doing the actual work.

PR #751, opened 2021-03-19
feat: change welcome email's copy

migcarva:

* new copy
* renames the test file to represent the test naming convention of the project
* tests check for CTA to Activate Account and mail to Support

### DIFF
--- a/packages/react-components/src/messages/__tests__/Welcome.test.tsx
+++ b/packages/react-components/src/messages/__tests__/Welcome.test.tsx
@@ -1,28 +1,40 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, RenderResult } from '@testing-library/react';
 
 import Welcome from '../Welcome';
 
-it('renders the welcome template with name and link', () => {
-  const { getByRole } = render(
-    <Welcome
-      firstName="John Doe"
-      link="https://example.com"
-      privacyPolicyHref={'https://hub.asap.science/privacy-policy'}
-      termsHref={'https://hub.asap.science/terms-and-conditions'}
-    />,
-  );
+describe('welcome email template', () => {
+  let result!: RenderResult;
 
-  const heading = getByRole('heading');
-  const cta = getByRole('link', {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/messages/__tests__/Welcome.test.tsx
```diff
@@ -14,5 +14,15 @@ it('renders the welcome template with name and link', () => {
...
   const heading = getByRole('heading');
+  const cta = getByRole('link', {
+    name: /activate account/i,
+  }) as HTMLAnchorElement;
+  const mailToSupport = getByRole('link', {
+    name: /get in touch/i,
+  }) as HTMLAnchorElement;
+
   expect(heading.textContent).toContain('John Doe');
+  expect(cta.href).toBe('https://example.com/');
```
jeysal, line 25: split into multiple test cases with descriptive titles please — one concern per test case

packages/react-components/src/messages/__tests__/Welcome.test.tsx
```diff
@@ -14,5 +14,15 @@ it('renders the welcome template with name and link', () => {
   );
 
   const heading = getByRole('heading');
+  const cta = getByRole('link', {
+    name: /activate account/i,
+  }) as HTMLAnchorElement;
```
jeysal, line 19: To make this case safe, the `selector: 'a'` option to getByRole should also be used

jeysal (changes requested):
couple of test comments other than that lgtm

packages/react-components/src/messages/__tests__/Welcome.test.tsx
migcarva: Multiple test cases aded

packages/react-components/src/messages/__tests__/Welcome.test.tsx
migcarva: Can you check the updated version?

jeysal (approved):

PR #753, opened 2021-03-19
refactor(frontend): handle possible 404s on sub resources

jeysal:
We don't really use this new API client capability yet (all the consumers are components used in contexts where we expect the parent resource has to be present so that you are even able to get to that place in the frontend) and continue to throw (show error boundaries) if these 404s happen.
But still, we should model API clients to represent every expected error with a return value, and only unexpected errors by throwing — at some point somewhere we'll want custom handling for the 404 because we're fetching a sub resource without having tried to fetch the parent yet, and if only then we start modeling the API clients to cover all expected cases, we'll have inconsistencies.


migcarva (approved):
lgtm

kmaid (approved):

PR #754, opened 2021-03-19
feat(events): surface unavailable material status

jeysal:
and surface any material status only on past events

### DIFF
--- a/packages/react-components/src/organisms/EventMaterials.tsx
+++ b/packages/react-components/src/organisms/EventMaterials.tsx
@@ -1,10 +1,18 @@
 import React from 'react';
 import css from '@emotion/css';
-import { EventResponse } from '@asap-hub/model';
+import {
+  EventResponse,
+  EVENT_CONSIDERED_PAST_HOURS_AFTER_EVENT,
+} from '@asap-hub/model';
+import { addHours, parseISO } from 'date-fns';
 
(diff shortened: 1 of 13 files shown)


packages/react-components/src/organisms/EventMaterials.tsx
```diff
@@ -22,22 +25,30 @@ const EventMaterials: React.FC<EventMaterialsProps> = ({
   meetingMaterials,
 }) => (
   <div css={cardsStyles}>
-    {notes === '' ? (
+    {notes === null ? (
```
migcarva, line 28: Was wondering if we should create a new `materials` array, with the actual material type data and title, and then map over it on the render.

migcarva (commented):
👍  just commented on the EventMaterials render method

packages/react-components/src/organisms/EventMaterials.tsx
jeysal: The additional meeting materials work differently since they have a different type, so I thought mapping to generate three of four then adding a fourth would only make it more confusing

jeysal:
Requirement was added to `undefined` materials (I said okay to this): This status should only surface once the event is a past event. On unavailable materials, this requirement was already there.
I added the implementation to this PR because it needs to go in together, otherwise we're showing cards where they are not wanted (upcoming events).

kmaid (approved):

PR #757, opened 2021-03-19
perf(frontend): do not re-fetch group when already fetched via userGroups or teamGroups

jeysal:


jeysal:
Yeah our state might need an abstraction soon. Or we just switch to Rx and do it with less code and fewer branches :thinking:

apps/frontend/src/network/teams/groups/GroupsCard.tsx
```diff
@@ -6,7 +6,7 @@ import { useTeamGroupsById } from './state';
 const GroupsCard: React.FC<{ id: string }> = ({ id }) => {
   const groups = useTeamGroupsById(id);
 
-  if (!groups) {
+  if (groups === 'noSuchTeam') {
```
kmaid, line 9: We get that back from the API?

kmaid (approved):

PR #760, opened 2021-03-22
Fix: spacing issues

migcarva:
- fix: spacing on the team card members value
- fix: padding on news and events card's figure
- fix: remove padding from result header
- fix: filter dropdown padding and margin
- fix: spacing of details in the profile page

### DIFF
--- a/packages/react-components/src/organisms/UserProfileBackground.tsx
+++ b/packages/react-components/src/organisms/UserProfileBackground.tsx
@@ -29,21 +29,16 @@ const linksContainer = css({
   },
 });
 
-const noMarginTextStyle = css({
-  h3: {
-    marginBottom: 0,
-  },
-  p: {
-    marginTop: 0,
-  },
-});
-
 const teamContentStyle = css({
   flex: 1,
 });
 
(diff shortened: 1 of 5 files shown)


packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -29,10 +29,23 @@ const linksContainer = css({
   },
 });
 
+const noMarginTextStyle = css({
```
jeysal, line 32: This intrudes into styling of other components and should not be necessary given the headline and paragraph spacings are defined by the design system, can you please push back on this? I think the most we should offer if it really needs to be less space between is using only the headline margin (by replacing the `Paragraph` with a plain `p` that does not have margin)

jeysal (changes requested):
most of the changes seem good

packages/react-components/src/organisms/UserProfileBackground.tsx
migcarva: Agreed. Discussed this with James and he agreed on the final result

jeysal (approved):

PR #762, opened 2021-03-23
feat: Fix main menu highlighting problems

migcarva:
* adds an isActive callback to NavHaskLink
* sets base route to network

### DIFF
--- a/packages/react-components/src/atoms/NavigationLink.tsx
+++ b/packages/react-components/src/atoms/NavigationLink.tsx
@@ -2,6 +2,14 @@ import React from 'react';
 import { NavHashLink } from 'react-router-hash-link';
 import css from '@emotion/css';
 
+import {
+  discover,
+  network,
+  sharedResearch,
+  news,
+  events,
+} from '@asap-hub/routing';
+
 import { TextChildren } from '../text';
 import { useHasRouter } from '../routing';
 import { lead } from '../colors';
@@ -79,6 +87,7 @@ const NavigationLink: React.FC<NavigationLinkProps> = ({
         activeClassName={activeClassName}
         css={[styles, { [`&.${activeClassName}`]: activePrimaryStyles }]}
         smooth
+        isActive={(match, _) => !!match}
       >
         <p css={textStyles}>
           <span css={iconStyles}>{icon}</span>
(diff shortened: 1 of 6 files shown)


packages/react-components/src/organisms/MainNavigation.tsx
```diff
@@ -59,7 +59,7 @@ const MainNavigation: React.FC = () => (
   <nav css={containerStyles}>
     <ul css={listStyles}>
       <li>
-        <NavigationLink href={network({}).teams({}).$} icon={networkIcon}>
```
jeysal, line 62: I think there are a couple more of these, e.g. in stories or so. Can you search the codebase for `network({}).teams({}).$` and replace it? (Except of course where it's *actually* meant to link directly to teams)

packages/react-components/src/atoms/NavigationLink.tsx
```diff
@@ -79,6 +79,7 @@ const NavigationLink: React.FC<NavigationLinkProps> = ({
         activeClassName={activeClassName}
         css={[styles, { [`&.${activeClassName}`]: activePrimaryStyles }]}
         smooth
+        isActive={(match, _) => !!match}
```
jeysal, line 82: this needs a regression test

jeysal (changes requested):

packages/react-components/src/organisms/MainNavigation.tsx
migcarva: done

packages/react-components/src/atoms/NavigationLink.tsx
migcarva: can you please check the newly added tests?

packages/react-components/src/organisms/MainNavigation.tsx
jeysal: You overshot the goal a bit, some of them were in places that are relying on deeply linking into the teams tab :D let me see I'll add/remove some

jeysal: Yeah looks like literally just the one in a test that was too much though :+1:

packages/react-components/src/atoms/NavigationLink.tsx
jeysal: I had to re-do the tests unfortunately since they were relying on implementation details (a class name). I borrowed from `NavigationLink.test` the way that is used to assert on active links (checking bg color).
Note that may be of interest: The rerendering in active and non-active state to observe the difference is there to ensure the test doesn't just assert on a background color that is *always* there no matter if active or not and give a false sense of security. You'll see this pattern in many places across the component lib

jeysal (approved):

PR #763, opened 2021-03-23
fix: Changing event status to cancelled

peterstarling:
Fixes a bug whereby a deleted event remains confirm upon sync after being deleted from the calendar. 

See comments of: https://trello.com/c/1NRKvLxZ/1102-syncing-with-gcal-is-failing for more details


migcarva (approved):
lgtm

PR #764, opened 2021-03-23
refactor(react-templates): convert to TypeScript and clean up somewhat

jeysal:


apps/asap-cli/src/invite/send-mail.ts
```diff
@@ -1,11 +1,7 @@
 import aws from 'aws-sdk';
+import { welcome } from '@asap-hub/message-templates';
```
jeysal, line 2: new package introduced to make template variables type safe

apps/messages/package.json
```diff
@@ -1,46 +1,37 @@
...
   "engines": {
     "node": "14.15.x"
   },
-  "peerDependencies": {
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+  "scripts": {
+    "build:babel": "../../scripts/build-babel.sh build",
+    "watch:babel": "../../scripts/build-babel.sh watch",
+    "deploy": "node build-cjs/deploy"
```
jeysal, line 11: Note that this now builds templates from compiled JS

jeysal: Converted a lot of files to TS

apps/storybook/src/MessageLayout.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
jeysal, line 1: made react-components export structure and story structure more consistent with rest of app

apps/messages/src/template-plugin.ts
```diff
@@ -32,31 +36,54 @@ const htmlTemplate = (
...
       );
 
       const { html, css } = extractCritical(renderToStaticMarkup(element));
       const s3Html = Object.keys(stats.compilation.assets).reduce(
         (res, asset) => {
           const { base } = path.parse(asset);
-          return res.replace(asset, `${origin}/static/media/${base}`);
+          return res.replace(
+            asset,
+            new URL(`/.messages-static/${base}`, APP_ORIGIN).toString(),
```
jeysal, line 85: This was perhaps the biggest architecture problem - images in the email were linking to assets deployed as part of the frontend, hoping that create-react-app constructed the image path with the `[name][hash:8][ext]` etc in the same way. I made a separately deployed bucket for the static assets in the emails now

jeysal:
Tested and seems it will work once deployed on master :tada:

jeysal:
Of course there would be much more to clean up (and start testing), this is a very arcane build process, but it definitely got better

apps/messages/src/template-plugin.ts
kmaid: Maybe we could inline the images instead?

apps/messages/src/templates/welcome.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { MessageLayout, WelcomeMessage } from '@asap-hub/react-components';
+import { welcome } from '@asap-hub/message-templates';
+
+export default (
+  <MessageLayout
+    privacyPolicyHref={'https://hub.asap.science/privacy-policy'}
+    termsHref={'https://hub.asap.science/terms-and-conditions'}
```
kmaid, line 8: Feels like these should be constants somewhere?

kmaid: It be cool if the routing package could generate fully qualified urls given an environment

kmaid (approved):

apps/messages/src/template-plugin.ts
jeysal:
```
Some bad news about base64 encoded images:

totally blocked by Outlook
not displayed by most webmail services (especially if you use more than one inside the message).
```

jeysal: also would be worse for perf since they aren't lazy loaded.

jeysal: So I get why Filipe did it this way, it was just missing a tiny bit to be reliable and not rely on what the FE deploys :D

apps/messages/src/templates/welcome.tsx
jeysal: Yeah idk if the routing package needs to do that since it's just `new URL(route({}).$, ORIGIN).href`. What I could indeed try to do here is re-use the `APP_ORIGIN` or whatever, I'll look into that. The `/privacy-policy` etc I mentioned briefly when doing the routing refactor, I didn't do that yet because the way that `Content.tsx` works in the FE to render them is really obscure, it's a refactor on my list :D

jeysal: done and rebased ^

apps/messages/src/webpack.config.ts
```diff
@@ -0,0 +1,41 @@
+import path from 'path';
+import { Configuration } from 'webpack';
+import TemplatePlugin from './template-plugin';
+
+const outputDir = path.resolve(
+  require.resolve('@asap-hub/messages/package.json'),
```
johnytiago, line 6: `__dirname` maybe?

jeysal: Reasoning was that it's absolute within the package, not relative to where this file is if it is moved into a subdir of src or anything

PR #765, opened 2021-03-24
feat: Enable tracing (without REST API)

peterstarling:
This PR enables tracing using X-Ray. The request is traced from API Gateway through to the lambda and inside the lambda.

See: https://trello.com/c/hXMKudeg/901-migrate-tracing-logic-to-express-lambda

**notes**
This PR is a lightweight version of https://github.com/yldio/asap-hub/pull/752

PR #769, opened 2021-03-25
feat: Enable tracing

peterstarling:
This PR adds instrumentation with AWS X-ray:

image

It covers request tracing all the way from the gateway to the app plus any http/https calls from the app.

**Important**

The deployment requires deleting base path mapping in API Gateway settings. Removing it manually from the console right before the deployment results in about 5 minute downtime. 

see: https://trello.com/c/hXMKudeg/901-migrate-tracing-logic-to-express-lambda for details


serverless.js
```diff
@@ -39,12 +39,9 @@ module.exports = {
     memorySize: 512,
     region: AWS_REGION,
     stage: SLS_STAGE,
-    httpApi: {
-      payload: '2.0',
-      cors: {
```
johnytiago, line 44: Don't we need CORS for these new endpoints?

.gitlab-ci.yml
```diff
@@ -17,7 +17,7 @@ include:
 
 .tmpl:deploy: &tmpl_deploy
   script:
-    - yarn sls deploy --verbose
+    - yarn sls deploy --stage=$SLS_STAGE --verbose
```
johnytiago, line 20: Wasn't this being passed as environment variable, or you're just making it explicit?

johnytiago:
How did you manage to attach the HTTP requests to the handler segment?

serverless.js
peterstarling: yeah actually im not sure why and how its still working! ill look into it

.gitlab-ci.yml
peterstarling: so this is really strange, i knew you would comment on it (and good thing you did) - serverless started throwing errors about missing opt.stage which i found really strange and this actually fixed it. I'll try to remove it and see if it makes a difference as I made many other changes since then so it might no longer be a problem

peterstarling:
> How did you manage to attach the HTTP requests to the handler segment?

erm.. i dont think i understand?

johnytiago:
image
If I recall correctly, these last 4 requests used to be part of a different trace. Now they *span* out of the apiHandler segment. Where they always like this?

peterstarling:
> image
> If I recall correctly, these last 4 requests used to be part of a different trace. Now they _span_ out of the apiHandler segment. Where they always like this?

no, thats the whole point, By disabling express middleware I was able to get a single trace, which is what we wanted right?

.gitlab-ci.yml
peterstarling: guess it works now, go figure

serverless.js
peterstarling: so actually, i think its because of:
https://github.com/yldio/asap-hub/blob/19514f4c94a620c2ca64755ef7d572cd71628a69/apps/asap-server/src/app.ts#L94
we seem to be getting the following response:
image

Previously with cors enabled in serverless we had:
image

I don't think it makes any difference, at least looking at the the asap hub (no sign of errors). The only thing I could think of is that the webhooks probably do not return the cors headers, but then I don't really think they need to anyway.

peterstarling:
@johnytiago FYI the reason we are unable to UPDATE the base path mapping is because API Gateway V1 and V2 have a different type of mapping meaning its a completely different resource in CloudFormation and such modification is not supported, see:
image
However although they seem to be two completely different resources they still clash because of the same domain.

peterstarling:
later attempts to deploy without downtime showed that there is a lot of risk involved in deployment and little value overall, so I'm closing this PR in favour of https://github.com/yldio/asap-hub/pull/765

PR #772, opened 2021-03-26
chore: change protocols csv order

johnytiago:
These were the changes I had to do in order to run the import on the last version of the csv. 
I expect the date format to change in the future. We should revert the parser if so.


apps/asap-cli/src/import/protocols/insert.ts
```diff
@@ -85,21 +85,17 @@ export default async (data: Protocol): Promise<void> => {
...
     },
     publishDate: {
       iv: new Date(data.created).toISOString(),
     },
     text: {
-      iv: `From Team ${data.team} and authors: ${data.authors.join(
-        ', ',
-      )}. Keywords: ${data.keywords.join(', ')}.`,
+      // eslint-disable-next-line prefer-template
+      iv: `${data.abstract ? data.abstract + '\n || ' : ''}From Team ${
```
peterstarling, line 96: why bother and disable the rule when you can just go with it?

apps/asap-cli/src/import/protocols/parse.ts
```diff
@@ -16,13 +16,31 @@ export default (data: string[]): Protocol => {
...
     abstract,
   ] = data.map((s) => s.trim());
 
+  const [day, month, yearAndTime] = created.split('.');
+
+  let abstractText;
+
+  // try to parse string as JSON or fail gracefully
+  try {
+    abstractText = ['(private protocol)', 'Abstract crypted'].includes(abstract)
```
peterstarling, line 32: i think you're try/catching a little bit more than just JSON.parse here. Are you sure you want to do that?

peterstarling (approved):
to be fair i dont really know whats going on but im sure you tested it and you know it works.

Just a couple of comments, feel free to do what you want with them

apps/asap-cli/src/import/protocols/insert.ts
johnytiago: Yeah, I disabled it to prevent from having nested template strings

apps/asap-cli/src/import/protocols/parse.ts
johnytiago: What else would make it throw here?

peterstarling: reduce, concat, includes

apps/asap-cli/src/import/protocols/insert.ts
peterstarling: yeah i think thats more effort than simply using templates but sure your call

apps/asap-cli/src/import/protocols/parse.ts
johnytiago: Yeah, no worries, I'm fine with this failing silently.

PR #774, opened 2021-03-26
feat(event-materials): special case card if all types unavailable

jeysal:


packages/react-components/src/organisms/EventMaterials.tsx
```diff
@@ -37,6 +38,14 @@ const EventMaterials: React.FC<EventMaterialsProps> = ({
     return null;
   }
 
+  if (
+    [notes, videoRecording, presentation, meetingMaterials].every(
```
kmaid, line 42: Maybe this should be moved somewhere else. I have to do the same check for toasts although i used filter instead of every. Whoops.

kmaid (changes requested):

packages/react-components/src/organisms/EventMaterials.tsx
jeysal: somewhere else as in, `containsOnlyNulls`?

kmaid: I was thinking the array of "materials"

PR #775, opened 2021-03-26
feat: only respond with only group for the calendar

johnytiago:


peterstarling (changes requested):
you need to test a scenario when more than a single group is returned:

image

johnytiago:
@peterstarling I noticed that this is already being tested since the fixtures already import referencingGroups as an array https://github.com/yldio/asap-hub/blob/master/apps/asap-server/test/fixtures/events.fixtures.ts#L60

We can make it more explicit if you prefer, but coverage is not going to change 😉

peterstarling:
> @peterstarling I noticed that this is already being tested since the fixtures already import referencingGroups as an array https://github.com/yldio/asap-hub/blob/master/apps/asap-server/test/fixtures/events.fixtures.ts#L60
> 
> We can make it more explicit if you prefer, but coverage is not going to change

yeah im not too bothered about the coverage, personally i prefer to have a test case that highlights which feature is being tested, but it's obviously your call. 

Our current test fixture uses multiple groups attached so I would probably convert it to just one by default (the desired data state) and then have another test with multiple groups and finally one without any groups - i think with the current squidex setup all 3 scenarios are valid, no?

packages/model/src/event.ts
```diff
@@ -37,7 +37,7 @@ export interface EventResponse {
   meetingLink?: string;
 
   calendar: CalendarResponse;
-  groups: GroupResponse[];
+  group?: GroupResponse;
```
peterstarling, line 40: since this is a contract change i would hope that there is a test that requires modification to account for the change

peterstarling (changes requested):

peterstarling:
Finally - since there is now a requirement that a calendar may have only ONE group attached (or none) - should the relationship be reverted to reflect it? as we know we are unable to enforce uniqueness on the `calendar` field of the `group` entity. With how things are set up at the moment it seems very likely that at some point an admin assigns the same calendar to multiple groups and only one of them is actually returned in an event.

Unless we think that is not such a big deal?

johnytiago:
> > @peterstarling I noticed that this is already being tested since the fixtures already import referencingGroups as an array https://github.com/yldio/asap-hub/blob/master/apps/asap-server/test/fixtures/events.fixtures.ts#L60
> > We can make it more explicit if you prefer, but coverage is not going to change
> 
> yeah im not too bothered about the coverage, personally i prefer to have a test case that highlights which feature is being tested, but it's obviously your call.
> 
> Our current test fixture uses multiple groups attached so I would probably convert it to just one by default (the desired data state) and then have another test with multiple groups and finally one without any groups - i think with the current squidex setup all 3 scenarios are valid, no?

Gotcha. I'm pushing an explicit version of the tests. 👍🏽

johnytiago:
> Finally - since there is now a requirement that a calendar may have only ONE group attached (or none) - should the relationship be reverted to reflect it? as we know we are unable to enforce uniqueness on the `calendar` field of the `group` entity. With how things are set up at the moment it seems very likely that at some point an admin assigns the same calendar to multiple groups and only one of them is actually returned in an event.
> 
> Unless we think that is not such a big deal?

Regarding this, I'm not too concerned. The code is safe, and it's up to them to make sure they only reference from a single group. Squidex already has mechanisms that allow users to see where a document is being reference, which should easy debugging for them if this is ever an issue. 
Changing the relationship now would involve a lot of work and probably cause more harm good. 🤔

apps/asap-server/test/fixtures/events.fixtures.ts
```diff
@@ -131,7 +133,7 @@ export const listEventResponse: ListEventResponse = {
         color: '#125A12',
         name: 'Tech 1 - Sequencing/omics',
       },
-      groups: queryGroupsExpectation.items as GroupResponse[],
+      group: queryGroupsExpectation.items[0] as GroupResponse,
```
peterstarling, line 136: i dont think you need casting here

packages/model/src/event.ts
peterstarling: yeah all good

apps/asap-server/src/entities/event.ts
```diff
@@ -11,10 +11,10 @@ import { parseGraphQLGroup } from './group';
 
 export const parseGraphQLEvent = (item: GraphqlEvent): EventResponse => {
   const calendar = parseGraphQLCalendar(item.flatData!.calendar![0]);
-  const groups =
-    item.flatData!.calendar![0].referencingGroupsContents?.map((group) =>
-      parseGraphQLGroup(group),
-    ) || [];
+  const group =
+    item.flatData!.calendar![0].referencingGroupsContents?.map((calGroup) =>
```
peterstarling, line 15: since we are only taking the first element now rather then map the items then pop the first one you could simply take the first one then call `parseGraphQLGroup` with that single element

apps/asap-server/test/controllers/events.test.ts
```diff
@@ -648,6 +649,70 @@ describe('Event controller', () => {
     });
   });
 
+  describe('Event groups', () => {
+    const eventId = 'group-id-1';
+
+    test('Should return one group when event calendar is referenced by multiple groups', async () => {
```
peterstarling, line 655:
```suggestion
    test('Should return the first group when event calendar is referenced by multiple groups', async () => {
```
It may seem like a non significant detail but actually this test outlines the exact requirement that we got. It's then more obvious that when there is more than one group attached we take whichever one came back first from squidex (which means its completely random)

peterstarling (changes requested):

apps/asap-server/src/entities/event.ts
johnytiago: I started by doing that, but code was a bit more confusing. Considering that multiple references to the same calendar is the edge case, these should not have a performance impact. This way is easier to read and to test 😉

peterstarling: mapping a single element in an array that is then converted to a single element is easier to read than simply invoking the function with a single argument? If you find it difficult to read - take the first element to a variable first then call the function. Your call, but i dont personally agree

PR #777, opened 2021-03-26
feat(devops): setup squidex application for each PR

johnytiago:
https://trello.com/c/0rFHWx1i/1121-run-integration-test-on-every-pr


jest/jest-base.config.js
```diff
@@ -6,7 +6,8 @@ const root = dirname(require.resolve('asap-hub/package.json'), '..');
 const testPathIgnorePatterns = [
   '\\.build-output-test\\.(js|jsx|ts|tsx)$',
   '\\.browser-test\\.(js|jsx|ts|tsx)$',
-  '\\.e2e\\.(js|jsx|ts|tsx)$',
+  '\\.e2e-tests\\.(js|jsx|ts|tsx)$',
+  'integration',
```
jeysal, line 10: Would be good to keep marking different test types with consistent file name schemas:
```suggestion
  '\\.e2e-test\\.(js|jsx|ts|tsx)$',
  '\\.integration-test\\.(js|jsx|ts|tsx)$',
```

The test you added could be `apps/asap-server/test/fetch-users.integration-test.ts`

jeysal: Actually seems there's a little more to be changed to achieve that, I'll push something, let me know if you have concerns with it

jeysal: Also remains to be seen if it's green first try or I have to make corrections :P

jeysal (commented):
It's a good start, but a little worrying to see how much python scripting was required to do this already :D now imagine running these for every test file to get every test an isolated environment so that scaling up to hundreds of integration/e2e tests becomes possible :see_no_evil: 

I'll let Piotr review the op-y bits, so most of this. Something that I find a little odd is that a lot of the logic is just run on CI and not as part of the npm scripts. It seems this will make it hard to run these locally?

johnytiago:
>It's a good start, but a little worrying to see how much python scripting was required to do this already :D now imagine running these for every test file to get every test an isolated environment so that scaling up to hundreds of integration/e2e tests becomes possible 🙈

For now I don't expect to run *every single test* in isolation. With the amount of setup needed, I'm not sure that is even feasible.

> I'll let Piotr review the op-y bits, so most of this. Something that I find a little odd is that a lot of the logic is just run on CI and not as part of the npm scripts. It seems this will make it hard to run these locally?

Me an Piotr were discussing this the other day. We couldn't think of an easy way to setup and run the tests with the python setup. We though the best approach would be to revisit the docker local setup and bring that up to date.

jest/jest-base.config.js
johnytiago: Thanks for fixing the Jest setup. 😉

peterstarling (changes requested):

peterstarling (approved):

apps/asap-server/test/integration/fetch-users.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      email: { iv: 'john.doe@asap.science' },
+      role: { iv: 'Grantee' },
+      connections: { iv: [] },
+      questions: { iv: [] },
+      teams: { iv: [] },
+      avatar: { iv: [] },
+      skills: { iv: [] },
+    };
+
+    await squidex.create(user);
```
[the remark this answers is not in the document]
johnytiago: Maybe I should. Will fix that! 😉

PR #778, opened 2021-03-29
feat: team page styling update

migcarva:
- updates the heading styles
- changes the way team members are displayed
- updates buttons disabled state

### DIFF
--- a/packages/react-components/src/button.tsx
+++ b/packages/react-components/src/button.tsx
@@ -151,6 +151,7 @@ const disabledStyles = css({
   cursor: 'unset',
 
   svg: {
+    filter: 'grayscale(1)',
     stroke: tin.rgb,
   },
 });
--- a/packages/react-components/src/button.tsx
+++ b/packages/react-components/src/button.tsx
@@ -124,6 +124,7 @@ const primaryStyles = css({
     backgroundColor: pine.rgb,
     borderColor: pine.rgb,
     boxShadow: 'none',
+    color: paper.rgb,
   },
 });
 const secondaryStyles = css({
@@ -142,7 +143,7 @@ const secondaryStyles = css({
   },
 });
 const disabledStyles = css({
-  color: tin.rgb,
+  color: lead.rgb,
   backgroundColor: silver.rgb,
   borderColor: steel.rgb,
   boxShadow: 'none',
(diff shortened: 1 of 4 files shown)


packages/react-components/src/button.tsx
```diff
@@ -124,6 +124,7 @@ const primaryStyles = css({
     backgroundColor: pine.rgb,
     borderColor: pine.rgb,
     boxShadow: 'none',
+    color: paper.rgb,
```
jeysal, line 127: why is it necessary?

packages/react-components/src/templates/TeamProfileHeader.tsx
```diff
@@ -88,13 +93,17 @@ const TeamProfileHeader: React.FC<TeamProfileHeaderProps> = ({
...
+                <li
+                  key={memberId}
+                  css={[listItemStyles, { left: `-${i * 3}px` }]}
+                >
                   <Anchor href={network({}).users({ userId: memberId }).$}>
                     <Avatar
                       firstName={firstName}
                       lastName={lastName}
                       imageUrl={avatarUrl}
+                      // border
```
jeysal, line 106: ?

jeysal (changes requested):
mostly lgtm just two odd things caught my eye

packages/react-components/src/button.tsx
migcarva: Without that, the text color on active buttons will be wrong (a darker one). Please, check https://trello.com/c/DR4BaNwq/1113-text-on-primary-green-buttons-goes-black-on-click-should-stay-white

jeysal: Yep, I can see what it tries to achieve, but could you reproduce that? For me it always stays white. Is it browser-specific because of certain user agent stylesheets or something?

jeysal: On master already always stays white, that is

migcarva: I can replicate it, yes.

migcarva: macOS - Brave browser: confirmed
Ubuntu - Chrome browser: confirmed

Which OS and browser are you using?

jeysal: I've just checked and it seems it comes from `Anchor` `resetStyles`, so it is only there for `buttonStyle primary` `Link`s

jeysal (approved):

PR #783, opened 2021-03-30
Fix growing hideshow

kmaid:


packages/react-components/src/atoms/Button.tsx
```diff
@@ -10,7 +10,6 @@ const linkStyles = css({
   display: 'inline',
   '> svg': {
     minHeight: '1em',
-    height: '100%',
```
migcarva, line 13: This does not explain everything, right? Why is the parent growing??

kmaid: The parent without this. Its like padding is added each time its resized and then the SVG takes 100%

jeysal (approved):

PR #784, opened 2021-03-31
feat: Create schema migration runner

peterstarling:
Creates a tool for running schema migrations. Can be used with squidex or any other types of migrations.

Work required for: https://trello.com/c/K7U0AkBH/1157-rename-shared-output-schemas-text-field-to-description

### DIFF
--- a/apps/asap-server/scripts/create-migration.sh
+++ b/apps/asap-server/scripts/create-migration.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+input=$1
+
+if [ -z "$input" ]; then
+    echo "Missing argument - migration name"
+    exit
+fi
+
+timestamp=$(date +%s%N | cut -b1-13)
+migration="$timestamp-$input"
+
+echo "Generating migration $migration..."
+
+cp ./scripts/migration.template ./src/migrations/$migration.ts
+
+echo "Done"
--- a/apps/asap-server/scripts/migration.template
+++ b/apps/asap-server/scripts/migration.template
@@ -1,0 +1,8 @@
+/* istanbul ignore file */
+import { RestResearchOutput, Results, Squidex } from '@asap-hub/squidex';
+import { Migration } from '../handlers/webhooks/webhook-run-migrations';
+
+export default class MoveResearchOutputTextToDescription extends Migration {
+  up = async (): Promise<void> => {};
+  down = async (): Promise<void> => {};
+}
--- a/apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
+++ b/apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
@@ -16,7 +16,7 @@ export const runFactory = (
 ): Handler => async () => {
   const getMigrationPaths = getMigrationPathsFromDirectoryFactory(readDir);
   const filterUnexecutedMigrations = filterUnexecutedMigrationsFactory(client);
-  const saveExecutedMigrations = saveExecutedMigrationsFactory(client);
+  const saveExecutedMigration = saveExecutedMigrationFactory(client);
   const getMigrationsFromPaths = getMigrationsFromPathsFactory(
     importModule,
     logger,
@@ -41,8 +41,10 @@ export const runFactory = (
     try {
       await migration.up();
       executedMigrations.push(migration.getPath());
-
       logger.info(`Executed migration '${migration.getPath()}`);
+
+      await saveExecutedMigration(migration.getPath());
+      logger.debug(`Saved migration progress`);
     } catch (error) {
       logger.error(
         error,
@@ -51,10 +53,6 @@ export const runFactory = (
       break;
     }
   }
-
-  logger.debug('Finished executing migrations, saving progress...');
-
-  await saveExecutedMigrations(executedMigrations);
 
   logger.info(`Executed and saved ${executedMigrations.length} migrations`);
 };
@@ -199,16 +197,14 @@ const filterMigrationsFactory = (client: Squidex<RestMigration>) => (
   });
 };
 
-const saveExecutedMigrationsFactory = (
-  client: Squidex<RestMigration>,
-) => async (migrations: string[]): Promise<void> => {
-  for (const migration of migrations) {
-    await client.create({
-      name: {
-        iv: migration,
-      },
-    });
-  }
+const saveExecutedMigrationFactory = (client: Squidex<RestMigration>) => async (
+  migration: string,
+): Promise<void> => {
+  await client.create({
+    name: {
+      iv: migration,
+    },
+  });
 };
 
 const removeExecutedMigrationFactory = (
--- a/apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
+++ b/apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
@@ -1,13 +1,27 @@
-import { RestMigration, Squidex } from '@asap-hub/squidex';
+import { Query, RestMigration, Squidex } from '@asap-hub/squidex';
 import { isBoom } from '@hapi/boom';
 import { Handler } from 'aws-lambda';
-import { promises } from 'fs';
+import { promises as fsPromise } from 'fs';
+import { Logger } from 'pino';
 import { migrationDir } from '../../config';
-import logger from '../../utils/logger';
+import pinoLogger from '../../utils/logger';
 
 const squidexClient = new Squidex<RestMigration>('migration');
 
-export const run: Handler = async () => {
+export const runFactory = (
+  logger: Logger,
+  client: Squidex<RestMigration>,
+  readDir: typeof fsPromise.readdir,
+  importModule: ImportModuleFromPath,
+): Handler => async () => {
+  const getMigrationPaths = getMigrationPathsFromDirectoryFactory(readDir);
+  const filterUnexecutedMigrations = filterUnexecutedMigrationsFactory(client);
+  const saveExecutedMigrations = saveExecutedMigrationsFactory(client);
+  const getMigrationsFromPaths = getMigrationsFromPathsFactory(
+    importModule,
+    logger,
+  );
+
   const migrationPaths = await getMigrationPaths();
   const unexecutedMigrationPaths = await filterUnexecutedMigrations(
     migrationPaths,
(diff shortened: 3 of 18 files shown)


apps/asap-server/scripts/create-migration.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+input=$1
+
+if [ -z "$input" ]; then
+    echo "Missing argument - migration name"
+    exit
+fi
+
+timestamp=$(date +%s%N | cut -b1-13)
```
johnytiago, line 10:
```suggestion
timestamp=$(date +%s)
```
Since this is epoch time, we don't need to worry about cutting the date shorter.
The `%N` wasn't doing anything both on my mac and on linux busybox docker img. Am I doing smth wrong?

apps/asap-server/scripts/migration.template
```diff
@@ -0,0 +1,8 @@
+/* istanbul ignore file */
```
johnytiago, line 1: `migration.template` => `migration.template.ts` ? Or will the build start complaining?

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
+import { Query, RestMigration, Squidex } from '@asap-hub/squidex';
+import { isBoom } from '@hapi/boom';
+import { Handler } from 'aws-lambda';
+import { promises as fsPromise } from 'fs';
+import { Logger } from 'pino';
+import { migrationDir } from '../../config';
+import pinoLogger from '../../utils/logger';
```
johnytiago, line 7:
```suggestion
import logger from '../../utils/logger';
```

johnytiago: If you agree, don't forget to change the references bellow, too

johnytiago: I just saw now, there is another `logger` inside the function scope.

johnytiago:
I was thinking about this. Maybe running this inside a lambda probably isn't the best option (even though it will have to do for now). My concern is that, since the lambda maximum timeout is 15min, this potentially could result in migrations being stopped half way 😬 We, therefore need to ensure that this lambdas timout is extended to the maximum. Additionally we could prevent the same lambda from running multiple migrations, or that the migrations are idempotent. Both this last suggestions are not guaranteed to solve the problem.

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    await removeExecutedMigration(migrationPath);
+
+    logger.info(`Rolled back and removed migration '${migration.getPath()}'`);
+  }
+};
+
+const getMigrationPathsFromDirectoryFactory = (
+  readDir: typeof fsPromise.readdir,
+) => async () => (await readDir(migrationDir)).sort();
```
johnytiago, line 117: How are the migrations getting into the filesystem of the lambda? We sure these are not getting excluded by tree-shaking?

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  return migrations;
+};
+
+const filterUnexecutedMigrationsFactory = (client: Squidex<RestMigration>) =>
+  filterMigrationsFactory(client);
+
+const filterMigrationsFactory = (
+  client: Squidex<RestMigration>,
+  unexecuted = true,
```
johnytiago, line 177:
```suggestion
  executed = false,
```
To prevent double negation, I think it's easier to read :)

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
...
+  ): Promise<T[]> => {
+    const results = await Promise.all(arr.map(predicate));
+
+    return arr.filter((_v, index) => results[index]);
+  };
+
+  return asyncFilter(migrationPaths, async (migration) => {
+    try {
+      await client.fetchOne({
+        filter: { path: 'data/name/iv', op: 'eq', value: migration },
```
johnytiago, line 191: Not sure about using the file name as document key. But I don't see another option also. 🤷🏻‍♂️

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
...
+  logger.info(`Executed and saved ${executedMigrations.length} migrations`);
+};
+
+export const rollbackFactory = (
+  logger: Logger,
+  client: Squidex<RestMigration>,
+  readDir: typeof fsPromise.readdir,
+  importModule: ImportModuleFromPath,
+): Handler => async () => {
+  // const getMigrationPaths = getMigrationPathsFromDirectoryFactory(readDir);
```
johnytiago, line 68: We can remove this

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,257 @@
...
+      throw error;
+    }
+  });
+};
+
+const saveExecutedMigrationsFactory = (
+  client: Squidex<RestMigration>,
+) => async (migrations: string[]): Promise<void> => {
+  for (const migration of migrations) {
+    await client.create({
```
johnytiago, line 209: This probably needs error handling & logging. If one create fails we don't want to lose the state of the migrations executed.

johnytiago: This got me thinking. We probably don't want to run any further migrations if we can't store the state in the end. Probably we should instead migrate & save state for each iteration, instead of migrating all & save all. What to you think

johnytiago:
The webhooks are lacking authorisation. Basic auth like we have for the auth0 rules would work

apps/asap-server/scripts/create-migration.sh
peterstarling: converts nano to mili:
image

apps/asap-server/scripts/migration.template
peterstarling: starts moaning, yeah

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
peterstarling: yeah

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
peterstarling: nah i tested it, somehow it resolves dynamic imports

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
peterstarling: i could leave out the extension? like if we were ever to run js files or something, i dont know

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
peterstarling: agreed

peterstarling:
> The webhooks are lacking authorisation. Basic auth like we have for the auth0 rules would work

actually.. do we want to have it as a webhook? my plan was to be able to run it from github actions as they support manual triggers but since we have access to AWS console we can just run the lambda manually. This PR does not expose it as an action anyway so why bother with auth and stuff for now, we can just create a tech debt ticket to enable it and for now just ship as a standalone lambda, what do you reckon?

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
peterstarling: as requested https://github.com/yldio/asap-hub/pull/784/commits/e3aedfdd8c716989ce66e9000602d18016ef7d30

sorry for force pushing but I rebased over master :D

johnytiago:
> actually.. do we want to have it as a webhook? my plan was to be able to run it from github actions as they support manual triggers but since we have access to AWS console we can just run the lambda manually. This PR does not expose it as an action anyway so why bother with auth and stuff for now, we can just create a tech debt ticket to enable it and for now just ship as a standalone lambda, what do you reckon?

Oh, yeah! I remember we talking about it. Lets try that

peterstarling:
image
working

johnytiago:
I've noticed. Maybe we should disable editing the migrations on the UI.

peterstarling:
> I've noticed. Maybe we should disable editing the migrations on the UI.

the Name field of the Migrations schema has been disabled all along: 
https://github.com/yldio/asap-hub/pull/784/files#diff-58ea096db6c12088237597da49125f6c0624555311181cd9893f1992003ddc64R21

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -0,0 +1,249 @@
...
+      logger.error(
+        error,
+        `Error executing the rollback of migration ${migration.getPath()}`,
+      );
+      return;
+    }
+
+    logger.debug('Finished executing rollback, saving progress...');
+
+    await removeExecutedMigration(migrationPath);
```
johnytiago, line 105: We should move the delete statement inside the try/catch. There's not error handling for this function

johnytiago: On the same concerns me in the initial setup for each function:
- Rollback
```
  const getLatestMigrationPathFromDb = getLatestMigrationPathFromDbFactory(
    client,
  );
  const getMigrationsFromPaths = getMigrationsFromPathsFactory(
    importModule,
    logger,
  );
  const removeExecutedMigration = removeExecutedMigrationFactory(client);

  const migrationPath = await getLatestMigrationPathFromDb();
```

- Run:
```
  const getMigrationPaths = getMigrationPathsFromDirectoryFactory(readDir);
  const filterUnexecutedMigrations = filterUnexecutedMigrationsFactory(client);
  const saveExecutedMigration = saveExecutedMigrationFactory(client);
  const getMigrationsFromPaths = getMigrationsFromPathsFactory(
    importModule,
    logger,
  );

  const migrationPaths = await getMigrationPaths();
  const unexecutedMigrationPaths = await filterUnexecutedMigrations(
    migrationPaths,
  );
```
All this IO is not error handled. What do you think?

peterstarling: the factory building is obviously not doing any work so i dont think it needs to be try catched, with regards to the reading of paths - again i think if this goes wrong and fails i want the lambda to fail. Ideally we would have some sensible alerting in place but since we run it manually ourselves it will just go red which is desired. Or do you mean i should just try catch it and throw something a little more meaningful?

serverless.js
```diff
@@ -231,6 +231,16 @@ module.exports = {
         }}`,
       },
     },
+    runMigrations: {
+      handler:
+        'apps/asap-server/build-cjs/handlers/webhooks/webhook-run-migrations.run',
+      timeout: 120,
```
johnytiago, line 237:
```suggestion
      timeout: 900,
```

serverless.js
```diff
@@ -231,6 +231,16 @@ module.exports = {
...
+    },
+    rollbackMigrations: {
+      handler:
+        'apps/asap-server/build-cjs/handlers/webhooks/webhook-run-migrations.rollback',
+      timeout: 120,
```
johnytiago, line 242:
```suggestion
      timeout: 900,
```

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
johnytiago: Maybe we can do with native logging for the setup bit, we want it to crash nonetheless. Let's just handle the `delete` call. 😉

johnytiago (approved):

PR #786, opened 2021-04-01
perf(frontend): prefetch various lists or entities that are linked to

jeysal:
This is just how far I've gotten until now, of course there are many
more opportunities. Prefetches this adds:

* From Dashboard, prefetch network (teams) and calendars
* From Calendars, prefetch upcoming and past events
* From Upcoming and Past Events, prefetch each other, and calendars
* From Group Upcoming and Past Events, prefetch each other

Remove various duplicate states from `GetListOptions` that can cause
recoil cache misses.

This restructure also happens to fix a bug where group event search
was doing nothing because the search query wasn't passed through.


apps/frontend/src/dashboard/Dashboard.tsx
```diff
@@ -24,6 +27,14 @@ const Dashboard: React.FC<Record<string, never>> = () => {
   const { firstName, id, teams } = currentUser;
   const { loading, data: dashboard } = useDashboard();
 
+  usePrefetchTeams({
```
jeysal, line 30: Search for `usePrefetch` to see the interesting bits ^^

apps/frontend/src/api-util.ts
```diff
@@ -1,25 +1,15 @@
...
 export type GetListOptions = {
-  searchQuery?: string;
-  filters?: string[];
-  currentPage?: number | null;
-  pageSize?: number | null;
-  sort?: {
-    sortBy: string; // column name ie startDate or endDate
-    sortOrder: 'asc' | 'desc';
-  };
+  searchQuery: string;
```
jeysal, line 4: This is the part about reducing duplicate 'nullish' states. most of the lines in the PR are a consequence of this essentially

kmaid (approved):

PR #788, opened 2021-04-07
feat: research output list/detail view toggle

kmaid:


apps/frontend/src/hooks/pagination.ts
```diff
@@ -1,19 +1,36 @@
...
   const searchParams = new URLSearchParams(useLocation().search);
   const currentPage = Number(searchParams.get('currentPage')) ?? 0;
 
+  const resetPaginationSearchParams = new URLSearchParams(searchParams);
+  resetPaginationSearchParams.delete('currentPage');
+
+  const listView = searchParams.get(viewParam) === listViewValue;
+  const listViewParams = new URLSearchParams(resetPaginationSearchParams);
+  listViewParams.set(viewParam, listViewValue);
+  const cardViewParams = new URLSearchParams(resetPaginationSearchParams);
```
jeysal, line 19: detail not card for wording consistency

apps/frontend/src/hooks/pagination.ts
```diff
@@ -1,19 +1,36 @@
...
-export const usePaginationParams = (pageSize = DEFAULT_PAGE_SIZE) => {
+export const usePaginationParams = () => {
   const history = useHistory();
   const searchParams = new URLSearchParams(useLocation().search);
   const currentPage = Number(searchParams.get('currentPage')) ?? 0;
 
+  const resetPaginationSearchParams = new URLSearchParams(searchParams);
+  resetPaginationSearchParams.delete('currentPage');
+
+  const listView = searchParams.get(viewParam) === listViewValue;
```
jeysal, line 16: `isListView`

packages/react-components/src/icons/card-view.tsx
```diff
@@ -0,0 +1,30 @@
...
+/* istanbul ignore file */
+
+const cardViewIcon = (
+  <svg
+    width={24}
+    height={24}
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    stroke="#4D646B"
```
jeysal, line 12: shiet just realized we probably have a lot of color duplication in these icons. meh, can probably do one pass replacing them with the defined colors at some point

packages/react-components/src/molecules/ListControls.tsx
```diff
@@ -0,0 +1,85 @@
...
+  listViewHref,
+  cardViewHref,
+}) => (
+  <div css={containerStyles}>
+    <NavHashLink
+      to={cardViewHref}
+      activeClassName={activeClassName}
+      css={[styles, { [`&.${activeClassName}`]: activeStyles }]}
+      smooth
+      isActive={() => !listView}
```
jeysal, line 63: why is this necessary? The nav link should do that by default based on `to`

apps/frontend/src/hooks/pagination.ts
kmaid: I think i snuck in another commit. They changed the designs on me

packages/react-components/src/molecules/ListControls.tsx
jeysal: that would also reduce props to two

packages/react-components/src/molecules/__tests__/ListControls.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { StaticRouter } from 'react-router-dom';
+
+import ListControls from '../ListControls';
```
jeysal, line 5: test active style (have many examples for that in the codebase)

packages/react-components/src/molecules/ListControls.tsx
kmaid: The URL will change with pagination and searchQuery params. I don't know how it would match without this

jeysal: right, and search query is preserved, but current page isn't I assume?

apps/frontend/src/hooks/pagination.ts
jeysal: Well, or if it's card view now, use that term everywhere :D

packages/react-components/src/molecules/ListControls.tsx
jeysal: The thing is, then we don't actually want a `NavLink`, but just a normal `Link`

jeysal: What I would suggest instead though: Still use `NavLink`, and still omit the `listView` prop, but in `isActive` look at that one specific `viewParam` and compare it to determine the active state

jeysal: That should mean little change from now because we're still using `NavLink`, but in a way that makes sense (you can tell because it actually uses the param to the isActive callback :D), and we still manage to omit one prop

apps/frontend/src/hooks/pagination.ts
kmaid: So the default view is Card View. The new view is List View. I think the naming is consistent

jeysal: But there are still things named `detail` left

apps/frontend/src/hooks/__tests__/pagination.test.ts
```diff
@@ -51,6 +61,33 @@ describe('usePaginationParams', () => {
       urlSearchParamsToObject(result.current.useLocationResult.search),
     ).toEqual({});
   });
+
+  it('generates details and list view params', () => {
```
jeysal, line 65:
```suggestion
  it('generates card and list view params', () => {
```

packages/react-components/src/molecules/ListControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const containerStyles = css({
+  display: 'grid',
+  gridAutoFlow: 'column',
+  columnGap: '24px',
+  width: 'fit-content',
+  margin: `${12 / perRem}em 0`,
+});
+
+const styles = css({
```
jeysal, line 17:
```suggestion
const linkStyles = css({
```

packages/react-components/src/molecules/ListControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+  readonly listViewHref: string;
+  readonly cardViewHref: string;
+}
+const ListControls: React.FC<ListControlsProps> = ({
+  isListView,
+  listViewHref,
+  cardViewHref,
+}) => (
+  <div css={containerStyles}>
+    <Link to={cardViewHref} css={styles}>
```
jeysal, line 56:
```suggestion
    <Link to={cardViewHref} css={linkStyles}>
```

packages/react-components/src/molecules/ListControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <p css={[textStyles, !isListView && activeStyles]}>
+        <span css={iconStyles}>{cardViewIcon}</span>
+        Card View
+      </p>
+    </Link>
+    <Link to={listViewHref} css={styles}>
```
jeysal, line 62:
```suggestion
    <Link to={listViewHref} css={linkStyles}>
```

packages/react-components/src/molecules/__tests__/ListControls.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { getByText, rerender } = render(
+    <ListControls
+      cardViewHref="/card?123"
+      listViewHref="/list?321"
+      isListView={false}
+    />,
+    { wrapper: MemoryRouter },
+  );
+  expect(
+    getComputedStyle(getByText(/card/i, { selector: 'p' })).fontWeight,
```
jeysal, line 30: Avoid `selector` since it's overly specific (we don't need to test that it's in a `p`), and use `findParentWithStyle`

jeysal (commented):
cool, a couple minor local suggestions then lgtm :upside_down_face:

packages/react-components/src/molecules/__tests__/ListControls.test.tsx
kmaid: The problem was get by text was picking up the SVG title. Thats why i put the selector

jeysal: Ah ok

jeysal (approved):

PR #793, opened 2021-04-09
Shared Research List

kmaid:


packages/react-components/src/templates/__tests__/SharedResearchPageBody.test.tsx
```diff
@@ -29,10 +29,25 @@ const props: Omit<ComponentProps<typeof SharedResearchPageBody>, 'children'> = {
   renderPageHref: (index) => `#${index}`,
 };
 
-it('renders multiple library cards', () => {
-  const { queryAllByRole } = render(<SharedResearchPageBody {...props} />, {
-    wrapper: StaticRouter,
-  });
+it('renders multiple shared outputs cards in card view', () => {
```
jeysal, line 32: It looks like these do not actually check any difference so they would pass with `isListView` toggled the other way?

jeysal (changes requested):
I haz quite a few remarks but all local stuff

packages/react-components/src/molecules/ExternalLink.tsx
```diff
@@ -33,6 +35,7 @@ const styles = css({
     },
   },
 });
+const paddingStyles = css({ padding: `0 ${11 / perRem}em` });
```
jeysal, line 38: `12 - borderRadius` I assume?

packages/react-components/src/organisms/SharedResearchList.tsx
```diff
@@ -0,0 +1,82 @@
...
+const containerStyles = css({
+  margin: 0,
+  padding: 0,
+  listStyle: 'none',
+});
+const itemStyles = css({
+  borderBottom: `1px solid ${steel.rgb}`,
+  display: 'grid',
+  gridTemplateRows: 'auto 12px auto',
+  '* ~ div:last-of-type': {
```
jeysal, line 20: I assume you mean this? 
```suggestion
  '&:last-of-type': {
```

jeysal (approved):

PR #797, opened 2021-04-12
feat(ROMS): enable all ROMS types and filters on FE

johnytiago:
Ticket: https://trello.com/c/H9FxenFD/1161-add-shared-output-filter-options


jeysal (approved):
Was any of these changes not forced by the TypeScript compiler? Or did it point you to all these locations? Just checking because if the former, we'll improve the types.

johnytiago:
> Was any of these changes not forced by the TypeScript compiler? Or did it point you to all these locations? Just checking because if the former, we'll improve the types.

@jeysal I think all of this was TS forced.

PR #799, opened 2021-04-13
feat: Convert serverless config to TypeScript

peterstarling:
As per: https://trello.com/c/S5SCkWK5/1134-convert-serverlessjs-file-to-typescript

Also cleans up the config file a bit by removing unnecessary template literals etc.


tsconfig.json
```diff
@@ -7,10 +7,10 @@
     "emitDeclarationOnly": true,
     "target": "es2015",
 
-    "lib": ["esnext"],
+    "lib": ["esnext", "dom"],
```
jeysal, line 10: This is not correct for the majority of packages, and almost all packages inherit from this config.

tsconfig.json
```diff
@@ -26,6 +26,11 @@
 
     "noErrorTruncation": true
   },
-  "files": [],
-  "include": []
+  "include": [
```
jeysal, line 29: `apps` should not be in the root build, there are `tsconfig`s specifically for them

jeysal (changes requested):
The `tsconfig` changes made to support using `ts-node` here are not correct. Please just use `babel-node` instead, it makes things a lot easier :see_no_evil:

tsconfig.json
peterstarling: yah that was dumb, sorry for being a n00b :cry:

tsconfig.json
peterstarling: yeah thats fine, i dont think i actually wanted to check that in, its not even required for the serverless build

PR #801, opened 2021-04-13
chore(devops): move env vars to SSM

johnytiago:
Ticket: https://trello.com/c/fbEazk54/1046-move-secrets-from-ci-to-secrets-manager

### DIFF
--- a/serverless.js
+++ b/serverless.js
@@ -82,28 +82,6 @@ module.exports = {
               }*`,
             ],
           ],
-          'Fn::Join': [
-            ':',
-            [
-              'arn:aws:secretsmanager',
-              { Ref: 'AWS::Region' },
-              { Ref: 'AWS::AccountId' },
-              'secret',
-              `auth0-shared-secret-${
-                SLS_STAGE === 'production' ? 'prod' : 'dev'
-              }`,
-            ],
-          ],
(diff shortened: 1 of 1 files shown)


serverless.js
```diff
@@ -84,6 +82,28 @@ module.exports = {
...
             ],
           ],
+          'Fn::Join': [
+            ':',
+            [
+              'arn:aws:secretsmanager',
+              { Ref: 'AWS::Region' },
+              { Ref: 'AWS::AccountId' },
+              'secret',
+              `auth0-shared-secret-${
```
peterstarling, line 92: i dont think you actually created these secrets?

serverless.js
```diff
@@ -149,6 +169,14 @@ module.exports = {
           },
         },
       ],
+      environment: {
+        AUTH0_CLIENT_ID: `\${ssm:auth0-client-id-${
```
peterstarling, line 173: was your intention to hide these? basically right now these two are exposed as follows:
image
if you dont think these values are secret then they can stay in SSM and be injected into the environment through the environment variable, just like you did it.

peterstarling (changes requested):

peterstarling (approved):

PR #802, opened 2021-04-14
chore(users): improve cms user schema UI

johnytiago:
This improvement is not tracked on a ticket. This was a request from B from the morning before the sprint started.


peterstarling (approved):
no requirements means no acceptance criteria to check so im just approving hoping that you know what you're doing.

And if it deploys to dev that means it works so..

johnytiago:
Yeah, this is just a UI change in the CMS. Without Integration tests there's no way we assert this things don't break the app 🤷🏽‍♂️

PR #803, opened 2021-04-14
refactor(sls): convert config to TS

jeysal:
Closes #799


peterstarling:
Yeah this doesn't seem to work - I only noticed now that we still use webpack to transpile serverless handlers (and serverless.js file), albeit only the js files, and create a bundle. 

In that case we still need to update the tsconfig file, maybe just for the serverless deployment run (as in - attach a tsconfig file that overrides the default one), which is what i am going to do. 

Since we need to transpile the `link-plugins` file i see little point of doing it with babel-node only then to run serverless with webpack config that still uses the tsconfig file. In fact i am now leaning towards pointing the serverless handlers to their ts versions, using ts-node to run the `link-plugins` transpilation and then ditching babel-node for asap-server entirely.

Opinions?

jeysal:
Piotr feel free to close one of the PRs, not sure which one you were gonna use

PR #806, opened 2021-04-15
fix(roms): add types to CMS schema

johnytiago:
https://trello.com/c/H9FxenFD/1161-add-shared-output-filter-options


peterstarling (approved):
i guess since our types and tests already cover these RO types anyway this change does not require any extra code (it looks weird when you look at the PR not knowing that)

PR #807, opened 2021-04-15
Fix: group card's title

migcarva:
Fix a typo on the Title of Profile's Group Card.

Note:
- Are we testing this twice??

### DIFF
--- a/apps/frontend/src/network/users/groups/__tests__/GroupsCard.test.tsx
+++ b/apps/frontend/src/network/users/groups/__tests__/GroupsCard.test.tsx
@@ -63,7 +63,7 @@ it('is rendered when there are groups', async () => {
   );
   await waitFor(() => {
     expect(queryByText(/loading/i)).not.toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


apps/frontend/src/network/users/groups/__tests__/GroupsCard.test.tsx
```diff
@@ -63,7 +63,7 @@ it('is rendered when there are groups', async () => {
   );
   await waitFor(() => {
     expect(queryByText(/loading/i)).not.toBeInTheDocument();
-    expect(queryByText(/test’ groups/i)).toBeInTheDocument();
+    expect(queryByText(/test’s groups/i)).toBeInTheDocument();
```
jeysal, line 66: Yeah I guess this is overly specific.
Something like this may work, you'll also have to change in at least one other place in the file up there :point_up: 
```suggestion
    expect(queryAllByText(/groups/i)).toBeInTheDocument();
```

jeysal (approved):

PR #808, opened 2021-04-15
refactor: Convert serverless file to typescript

peterstarling:
see: https://trello.com/c/S5SCkWK5/1134-convert-serverlessjs-file-to-typescript

closes: https://github.com/yldio/asap-hub/pull/803

## Notable changes:
- running serverless still requires tsconfig as it compiles the serverless.ts file with ts-node
- the PR makes serverless bundle with webpack when running serverless (no build step needed)


serverless.ts
```diff
@@ -16,28 +16,28 @@ const {
...
 const service = paramCase(pkg.name);
-const plugins = [
+export const plugins = [
   'serverless-s3-sync',
   'serverless-iam-roles-per-function',
-  ...(NODE_ENV === 'production'
-    ? ['serverless-plugin-tree-shake']
-    : ['serverless-offline']),
+  'serverless-webpack',
+  // 'serverless-plugin-tree-shake',
```
johnytiago, line 29: No more tree-shaking?

johnytiago:
Looks, good. Was there a difference in handler size?

serverless.ts
peterstarling: i actually have to remove the comment

peterstarling:
> Looks, good. Was there a difference in handler size?

So...
image
These are serverless bundle folders (originally named `.serverless` after packaging), with sizes. Basically the `no-webpack` one is the way we build now, then the other two are from this PR, one with tree-shaking the other one without. 

First of all it looks like tree shaking, at least with webpack, makes no difference in terms of the bundle size. Weird.

Second - webpack approach reduces the size of a bundle by more than 50%, I think thats a lot, although im not sure if it performs any better though, so it may not even be relevant.

johnytiago:
That's great and weird at the same time. 
I think ~17Mb was the size of the lambdas we got before merging them with express 🤔 
I'm going to give it a thumbs up, but I'd be more comfortable if @jeysal took a look at it.

johnytiago (approved):

serverless.ts
jeysal: :point_up:

serverless/webpack.config.js
```diff
@@ -1,7 +1,36 @@
...
   target: 'node',
   mode: 'production',
+  module: {
+    rules: [
+      {
+        test: /\.(js|jsx|ts|tsx)$/,
+        exclude: /node_modules/,
+        use: {
+          loader: require.resolve('babel-loader'),
+          options: {
```
jeysal, line 16: check if necessary, I think it shouldn't be since default babel config resolution should work

serverless/webpack.config.js
```diff
@@ -1,7 +1,36 @@
...
+        },
+      },
+    ],
+  },
+  optimization: {
+    // We do not want to minimize our code.
+    minimize: false,
+  },
+  resolve: {
+    extensions: ['*', '.js', '.jsx', '.ts', '.tsx'],
```
jeysal, line 28: What does `'*'` do?

serverless/webpack.config.js
```diff
@@ -1,7 +1,36 @@
...
+          loader: require.resolve('babel-loader'),
+          options: {
+            ...babelOptions,
+          },
+        },
+      },
+    ],
+  },
+  optimization: {
+    // We do not want to minimize our code.
```
jeysal, line 24: Thinking about it, why, actually? Did the minimizer cause bugs for us? Or did it break source mapping?

jeysal (commented):
Cool yeah I agree with this approach, we used to use `serverless-webpack` and it was quite a bit more stable than the tree shake thing (of course, it's webpack, which is not exactly little used) :sweat_smile: 
`serverless` running ts-node internally isn't ideal but can live with it, not worth the effort now

serverless/webpack.config.js
peterstarling: yeah, got the javascript heap our of memory error, at least on my local.

I'll try to give it a go, maybe in the CI it will pass although I would prefer to have a setup that doesn't crash on my local :D

Also I think by default Uglify is called to do the minification, ill try to give a different plugin a go too. It's been a known fact for a while that a smaller bundle has a positive performance impact on lambdas (albeit increases memory usage slightly, but we are not charged by that). See: https://medium.com/capital-one-tech/applying-minification-and-uglification-to-aws-lambda-functions-dbc7ad75241

serverless/webpack.config.js
peterstarling: yeah fair point, ill remove it, pretty sure it does.. nothing.

serverless/webpack.config.js
peterstarling: this is weird, this:
image
actually creates a bundle of 24MB, with uglification off it's only 17MB. Not sure if there is any point to explore if further as the default minification/uglification hangs anyway and this one is clearly worthless.

package.json
```diff
@@ -132,7 +139,11 @@
     "serverless-offline": "patch:serverless-offline@^6.8.0#./patches/serverless-offline.patch",
     "serverless-plugin-tree-shake": "^1.1.7",
     "serverless-s3-sync": "^1.15.0",
+    "serverless-webpack": "^5.4.1",
     "typescript": "~4.1.0",
     "webpack": "^5.1.0"
+  },
+  "dependencies": {
+    "@hapi/joi": "^17.1.1"
```
jeysal, line 147: This and its `@types` should probably be in `asap-server`?

package.json
```diff
@@ -68,6 +68,9 @@
     "serverless-s3-sync": {
       "unplugged": true
     },
+    "serverless-webpack@5.4.1": {
```
jeysal, line 71:
```suggestion
    "serverless-webpack": {
```

jeysal (commented):
Just commented on some small newly introduced other than that LGTM

package.json
```diff
@@ -102,11 +106,14 @@
     "@emotion/jest": "^11.0.0",
     "@jeysal/babel-plugin-ramda": "^2.0.1",
     "@peculiar/webcrypto": "^1.1.1",
+    "@serverless/typescript": "^2.34.0",
     "@testing-library/jest-dom": "^5.11.9",
+    "@types/hapi__joi": "^17.1.0",
```
jeysal, line 111:
```suggestion
```

jeysal (approved):

jeysal (approved):
:tada:

PR #811, opened 2021-04-15
 feat(migrations): migrate publishDate to addedDate

johnytiago:
Ticket: https://trello.com/c/ipMAHg9b/1160-migrate-research-output-original-publishing-date-to-added-date?menu=filter&filter=label:BE
> **this PR only addressed the first 2 requirements** 

- fix(migrations): fix date function & old migration
  - makes the date command work on OSX and Linux
  - use script relative paths 
  - simplifies the template
  - fixes pagination on old migraition script (should not run for old envs though)
- feat(migrations): add addedDate migration script
  - successfully ran script agains temp environment
image

### DIFF
--- a/apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
+++ b/apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
@@ -22,6 +22,7 @@ export default class MoveResearchOutputTextToDescription extends Migration {
       for (const researchOutput of result.items) {
         const publishDate = researchOutput.data.publishDate?.iv;
         if (!publishDate) {
+          // eslint-disable-next-line no-continue
           continue;
         }
 
@@ -35,6 +36,7 @@ export default class MoveResearchOutputTextToDescription extends Migration {
           });
         } catch (err) {
           logger.error(err, `Error migrating RO: ${researchOutput.id}`);
+          // eslint-disable-next-line no-continue
           continue;
         }
       }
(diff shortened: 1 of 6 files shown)


apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
```diff
@@ -0,0 +1,48 @@
...
+        try {
+          const addedDate = publishDate.includes(' ')
+            ? new Date(`${publishDate} UTC`).toISOString() // for dates in format 'Month Year'
+            : new Date(publishDate).toISOString(); // For dates in ISO format
+
+          await squidexClient.patch(researchOutput.id, {
+            addedDate: { iv: addedDate },
+          });
+        } catch (err) {
+          logger.error(err, `Error migrating RO: ${researchOutput.id}`);
```
peterstarling, line 37: it's a little risky because this way if errors are encountered rerunning the migration will require rollback (which actually isn't so bad either since the rollback doesn't do anything :D), but im just pointing it out

apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
```diff
@@ -0,0 +1,48 @@
...
+          continue;
+        }
+      }
+
+      pointer += 10;
+    } while (pointer < result.total);
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  down = async (): Promise<void> => {};
```
peterstarling, line 47: you could nullify the `addedDate` field here, just in case the migration does something stupid and we want to rollback quickly

peterstarling: but im sure there is probably an easy way to remove all the data from that field in squidex as well so..

peterstarling (approved):

apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
johnytiago: Sure! Nice suggestion!

apps/asap-server/src/migrations/1618488149-publish-date-to-added-date.ts
johnytiago: This block is here just in case there's an error creating a date. (e.g. `publishDate = 'foo bar'`) In this case I don't want to stop the migration because of a bad record. Just log it and move on.
If the patch throws, then that's also fine. Nothing was changed in that case.

peterstarling (approved):
i'd obviously test it in dev first (both run and rollback) but if it works then merge merge merge

johnytiago:
Safe date conversion test:
- Data
image
- Result
image

Rollback test: 
image

PR #812, opened 2021-04-16
feat: Rename short-text to admin-notes

peterstarling:
https://trello.com/c/YUtW7fLX/1159-rename-short-text-field-to-admin-notes

Creates a new field - `adminNotes`, and moves all the data from the `shortText` field. 

A follow up PR is expected to remove the `shortText` field

### DIFF
--- a/apps/asap-server/src/migrations/1618575365-move-short-text-to-admin-notes.ts
+++ b/apps/asap-server/src/migrations/1618575365-move-short-text-to-admin-notes.ts
@@ -1,0 +1,56 @@
+/* istanbul ignore file */
+import { RestResearchOutput, Results, Squidex } from '@asap-hub/squidex';
+import { Migration } from '../handlers/webhooks/webhook-run-migrations';
+
+export default class MoveResearchOutputTextToDescription extends Migration {
+  up = async (): Promise<void> => {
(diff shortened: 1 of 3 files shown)


apps/asap-server/src/migrations/1618575365-move-short-text-to-admin-notes.ts
```diff
@@ -0,0 +1,56 @@
...
+      result = await squidexClient.fetch({
+        $top: 10,
+        $skip: pointer,
+        $orderby: 'created asc',
+      });
+
+      for (const researchOutput of result.items) {
+        await squidexClient.patch(researchOutput.id, {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          adminNotes: null as any, // types dont cover null values
```
johnytiago, line 49: Because we're not deleting the `shortText` just yet we can use it in the rollback and just keep a failsafe condition of null for once it is gone
`adminNotes: researchOutput.data.shortText || null as any, // types dont cover null values`

johnytiago: I guess I need some coffee. Forget about it. 😅

johnytiago (approved):

PR #813, opened 2021-04-16
Research Output page.

kmaid:
Link to research output page.
Stories


apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -68,31 +69,54 @@ const renderComponent = async () => {
...
+      ...createResearchOutputResponse(),
+      type: 'Proposal',
+      team: {
+        displayName: 'Proposal Team',
+        id: '123',
+      },
+      title: 'Proposal title!',
+    });
+    const { getByRole } = await renderComponent();
+    expect(getByRole('heading').textContent).toEqual('Proposal title!');
```
jeysal, line 84: If the non-proposal one has a specific assertion that wouldn't work with a proposal (tags), can the proposal one also have a specific assertion that wouldn't work with a non-proposal?

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -68,31 +69,54 @@ const renderComponent = async () => {
...
+      tags: ['Example Tag'],
+      title: 'Proposal title!',
+    });
+    const { getByRole, getByText } = await renderComponent();
+    expect(getByText(/Example Tag/i)).toBeVisible();
+    expect(getByRole('heading', { level: 1 }).textContent).toEqual(
+      'Proposal title!',
+    );
+  });
 });
```
jeysal, line 120: The describe and it titles are a bit of a mess with the new split, to suggest a format maybe something like
describe 'a non-proposal research output'
  it 'is rendered with tags'

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -0,0 +1,125 @@
...
+  tags,
+  link,
+  lastModifiedDate,
+}) => (
+  <div css={containerStyles}>
+    <BackLink href={backHref} />
+    <div css={cardsStyles}>
+      <Card>
+        <div css={headerStyles}>
+          <div css={typeStyles}>
```
jeysal, line 82: Is there not quite a bit in common here, do you reckon we could split this out?

jeysal (commented):
cool

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -68,31 +69,53 @@ const renderComponent = async () => {
   return result;
 };
 
-it('renders the research output', async () => {
-  mockGetResearchOutput.mockResolvedValue({
-    ...createResearchOutputResponse(),
-    title: 'Proposal title!',
+describe('proposal research output', () => {
+  it('renders a research proposal with team', async () => {
```
jeysal, line 73:
```suggestion
describe('a proposal research output', () => {
  it('renders with its team', async () => {
```

jeysal (approved):

PR #814, opened 2021-04-16
build(backend): use Babel for local dev as well

jeysal:


apps/asap-server/package.json
```diff
@@ -8,14 +8,15 @@
   "scripts": {
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
-    "start:server": "tsnd --transpile-only ./src/server.ts",
+    "start": "node-dev ./src/server.ts",
```
jeysal, line 11: I renamed this to `start` for consistency with the other apps (and because `server` is already in the package name so it should be clear). In the root it's of course still aliased `start:backend`

jeysal:
This'll make it all consistent :sparkles: 
Note: While the local server starts for me, currently every request fails, whether logged in or not, whether guarded endpoints or just terms-and-conditions. However, this is also the case for me on `master`. Maybe I have something messed up in my setup, either way it's on master and not caused by this  PR :sweat_smile:

peterstarling (approved):
nice nice chicken rice

PR #821, opened 2021-04-21
feat: Make role optional in the invite script

peterstarling:
see: https://trello.com/c/hU9PRAES/1179-invitation-script-should-invite-any-user-irrespective-of-role


johnytiago (approved):
LGTM

PR #822, opened 2021-04-21
feat: Create Dev Squidex Backup

peterstarling:
see: https://trello.com/c/1oHlF3sU/1204-create-and-store-dev-squidex-data-dump


peterstarling:
This is WIP because:
* trigger was changed to push - does not look like manual triggers are supported off a branch on github
* S3 resource cration was enabled for this PR (should be dev only)
* bucket name was fixed to this PR (should be dev)
* syncing only the schema from squidex

Also I am unsure about this approach of installing cli here - i dont think there is a docker image with the cli available but we might need to create one ourselves and push it OR make this part a github private action. TBD

.github/workflows/backups.yml
```diff
@@ -0,0 +1,41 @@
+name: Create database backup DEV
+on: workflow_dispatch
+jobs:
+  make_backup:
+    runs-on: ubuntu-18.04
+    container:
+      image: registry.gitlab.com/yldio/asap-hub/squidex-utils
+    env:
+      SQUIDEX_VERSION: '7.4'
```
johnytiago, line 9:
```suggestion
```

.github/workflows/backups.yml
```diff
@@ -0,0 +1,41 @@
...
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      BUCKET_NAME: 'asap-hub-dev-data-backup'
+    steps:
+      - name: Download the backup artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: squidex-dev-backup-artifact-${{ github.run_id }}
+          path: backup/
+      - name: Get the name of the file
+        run: FILE_NAME=$(ls ./backup | head -n 1) && echo "FILE_NAME=$FILE_NAME" >> $GITHUB_ENV
```
johnytiago, line 39:
```suggestion
        run: export FILE_NAME=$(ls ./backup | head -n 1)
```
Since the file name is not going to be used outside of this job we can use bash native assignments

peterstarling: it is being used in the next step though, so i think putting in $GITHUB_ENV is still required

ci/.gitlab-ci.integration.yml
```diff
@@ -5,17 +5,17 @@ variables:
 
 setup:integration-image:
   image: docker:19.03.12
-  stage: .pre
+  stage: build
```
peterstarling, line 8: why move it to build, if i may ask?

ci/.gitlab-ci.squidex.yml
```diff
@@ -12,11 +12,9 @@ setup:squidex-image:
     - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
     - docker build -t $SQUIDEX_SYNC_DOCKER_IMAGE . -f dev/base.Dockerfile
     - docker push $SQUIDEX_SYNC_DOCKER_IMAGE
-  only:
-    refs:
-      - master
-    changes:
-      - dev/base.Dockerfile
+  rules:
```
peterstarling, line 15: whats going on here?

ci/.gitlab-ci.integration.yml
johnytiago: This new image depends on the base one (i.e. `dev/base.Dockerfile`) It has to wait for it to finish before building this one. 
Since this image will be used in the integration job during testing, this felt like the logical place to put it. But to be honest, coming back to this today, I makes even more sense to build this image in the same step as the first one (although, that has a cost in terms of pipeline duration, but at least we leverage docker caching 🤷🏻‍♂️)

ci/.gitlab-ci.squidex.yml
johnytiago: I was trying to make the build only run when there's actual changes to those files. I don't think I was successful

ci/.gitlab-ci.integration.yml
peterstarling: so... can we not move it to the same stage and than create a dependency between the two jobs?

ci/.gitlab-ci.squidex.yml
peterstarling: i see, and it didn't work the way it was previously done?

johnytiago: Nope. This probably has something to doe with the fact that our CI is triggered via webhook (from github) and not from a push. https://gitlab.com/gitlab-org/gitlab/-/issues/27318#note_248733490

ci/.gitlab-ci.integration.yml
johnytiago: Yeah, let me try that

johnytiago: We can create dependencies between jobs in the same stage. I either keep them in the same job or leave the second on during the build stage. 
Considering that `only: changes` doesn't work with our setup, they will always both run. So keeping them separated would be only for logical purposes.

johnytiago (approved):

PR #824, opened 2021-04-22
fix(local-setup): simplify local squidex setup

johnytiago:
Ticket: https://trello.com/c/WH7OSCwI/1190-integration-tests-local-setup


jeysal:
sounds nice. I suppose the only barrier to local dev that will then remain is data import / Auth0 linking?

peterstarling (approved):

PR #826, opened 2021-04-23
feat(filter): analytics

jeysal:
Recommend reviewing the two commits separately:

- refactor(react-components): extract Filter out of SearchAndFilter
- feat(filter): analytics

### DIFF
--- a/packages/react-components/src/organisms/Filter.tsx
+++ b/packages/react-components/src/organisms/Filter.tsx
@@ -1,0 +1,95 @@
+import css from '@emotion/css';
+import React, { useEffect, useState } from 'react';
+
+import { CheckboxGroup } from '.';
+import { Button, Caption } from '../atoms';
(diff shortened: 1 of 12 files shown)


packages/react-components/src/organisms/Filter.tsx
```diff
@@ -0,0 +1,115 @@
...
+  onChangeFilter = noop,
+  filterOptions,
+  filterTitle,
+}) => {
+  const [menuShown, setMenuShown] = useState(false);
+  useEffect(() => {
+    setMenuShown(false);
+  }, [filterTitle]);
+
+  const [debouncedFilters] = useDebounce(filters, 5000);
```
jeysal, line 71: It may be worth extracting a hook to do the debounce and the equality check stuff for us, only leaving us to construct the different data object in each component. Depends on how many of these we'll have I guess. Either way, don't have the time before my holiday :sweat_smile:

PR #828, opened 2021-04-26
chore(docker): share same squidex ubuntu base image

johnytiago:
Ticket: https://trello.com/c/n3mFLdTU/1240-refactor-docker-images-setup

- Prevents installing the squidex cli in multiple places
- Stop using local setup image as base image. This one doesn't even need to be published, it's only used by DCompose.


dev/Dockerfile
```diff
@@ -1,20 +1,8 @@
 FROM ubuntu:18.04
```
peterstarling, line 1: I would name this file something else if have a similar one next to it with a dedicated name.

Just something to tell what its purpose is, unless you want to add a comment here instead

peterstarling (approved):
just a small comment, feel free to address it or merge

dev/Dockerfile
johnytiago: 👍🏽

PR #831, opened 2021-04-28
fix: Quotation marks search bug

peterstarling:
Sanitise search inputs by escaping to hex.

see: https://trello.com/c/7pUQV9hP/1242-%F0%9F%90%9B-searching-with-apostrophes-breaks-the-search

## Notes
* The sanitisation should, imho, happen inside the client as its client specific. The reason I did not do it there is because the client currently only accepts a formatted query, and query generation happens at the controller level anyway. 
* Following up on the above - I think it would be much better to use `variables` part of the graphql queries rather than embed all the params in the query - a tech debt ticket can be found here: https://trello.com/c/Ahlq8rgV/1259-move-graphql-query-parameters-to-the-variables
* We use the, technically deprecated, built in `escape` function to escape these characters - so far i have not been able to find any better npm package that would achieve the same result, plus it doesn't look like its going to go away any soon

### DIFF
--- a/apps/asap-server/src/utils/squidex.ts
+++ b/apps/asap-server/src/utils/squidex.ts
@@ -12,3 +12,6 @@ export const createURL = (assets: string[]): string[] =>
 
 export const parseDate = (date: string): Date =>
   DateTime.fromISO(date).toJSDate();
+
+export const sanitiseForSquidex = (text: string): string =>
+  text.replace(/'/g, "''").replace(/"/g, '\\"');
--- a/apps/asap-server/src/utils/squidex.ts
+++ b/apps/asap-server/src/utils/squidex.ts
@@ -14,4 +14,4 @@ export const parseDate = (date: string): Date =>
   DateTime.fromISO(date).toJSDate();
 
 export const sanitiseForSquidex = (text: string): string =>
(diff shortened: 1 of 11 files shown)


peterstarling:
image
image
image

migcarva (approved):
lgtm

apps/asap-server/src/utils/squidex.ts
```diff
@@ -12,3 +12,6 @@ export const createURL = (assets: string[]): string[] =>
 
 export const parseDate = (date: string): Date =>
   DateTime.fromISO(date).toJSDate();
+
+export const sanitiseForSquidex = (text: string): string =>
+  escape(text.replace(/'/g, "''"));
```
kmaid, line 17: https://www.w3schools.com/jsref/jsref_escape.asp isn't this deprecated?

peterstarling: image

peterstarling: the reason i used it is because there doesn't seem to be any package that does what we want. Well there is xss-filters but i think its a bit of an overkill.

johnytiago (approved):

PR #836, opened 2021-04-29
feat(actions): backup prod action

johnytiago:
Ticket: https://trello.com/c/cBQQw8Uy/1226-create-and-store-prod-squidex-data-dump

The expire policy on the bucket applies to `dev` which makes sense. 

**Before merging, the branch condition has to be removed from cloudformation**

### DIFF
--- a/.github/workflows/dev.backup.yml
+++ b/.github/workflows/dev.backup.yml
@@ -1,0 +1,29 @@
+name: DEV - Create database backup
+on:
+  schedule:
+    - cron: '0 7 * * FRI' # Every Friday at 7am UTC
+  workflow_dispatch:
+jobs:
+  make_backup:
+    runs-on: ubuntu-18.04
+    container:
+      image: registry.gitlab.com/yldio/asap-hub/squidex-utils
+    env:
+      SQUIDEX_APP: 'asap-hub-dev'
+      BUCKET_NAME: 'asap-hub-dev-data-backup'
+      SQUIDEX_CLIENT_ID: ${{ secrets.SQUIDEX_CLIENT_ID_DEV }}
+      SQUIDEX_CLIENT_SECRET: ${{ secrets.SQUIDEX_CLIENT_SECRET_DEV }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    steps:
+      - name: Add the configuration
+        run: sq config add $SQUIDEX_APP $SQUIDEX_CLIENT_ID $SQUIDEX_CLIENT_SECRET
+      - name: Run sync
+        run: sq sync out ./backup
+      - name: Tar the files
+        run: |
+          FILE_NAME=$(date +%s| cut -b1-13)-squidex-dev-backup.tar.gz && \
+          tar -czf $FILE_NAME backup/ && \
+          echo "FILE_NAME=$FILE_NAME" >> $GITHUB_ENV
+      - name: upload to S3
+        run: aws s3api put-object --private --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
--- a/.github/workflows/dev.backup.yml
+++ b/.github/workflows/dev.backup.yml
@@ -26,4 +26,4 @@ jobs:
           tar -czf $FILE_NAME backup/ && \
           echo "FILE_NAME=$FILE_NAME" >> $GITHUB_ENV
       - name: upload to S3
-        run: aws s3api put-object --private --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
+        run: aws s3api put-object --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
--- a/.github/workflows/prod.backup.yml
+++ b/.github/workflows/prod.backup.yml
@@ -3,7 +3,6 @@ on:
   schedule:
     - cron: '0 7,19 * * *' # Twice a day: 7am and 7pm UTC
   workflow_dispatch:
-  push:
 jobs:
   make_backup:
     runs-on: ubuntu-18.04
@@ -11,7 +10,7 @@ jobs:
       image: registry.gitlab.com/yldio/asap-hub/squidex-utils
     env:
       SQUIDEX_APP: 'asap-hub'
-      BUCKET_NAME: 'asap-hub-836-data-backup'
+      BUCKET_NAME: 'asap-hub-prod-data-backup'
       SQUIDEX_CLIENT_ID: ${{ secrets.SQUIDEX_CLIENT_ID_PROD }}
       SQUIDEX_CLIENT_SECRET: ${{ secrets.SQUIDEX_CLIENT_SECRET_PROD }}
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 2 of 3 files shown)


.github/workflows/dev.backup.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           tar -czf $FILE_NAME backup/ && \
           echo "FILE_NAME=$FILE_NAME" >> $GITHUB_ENV
       - name: upload to S3
-        run: aws s3api put-object --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
+        run: aws s3api put-object --private --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
```
peterstarling, line 29: no need, by default they are private

.github/workflows/dev.backup.yml
```diff
@@ -1,7 +1,7 @@
-name: Create database backup DEV
+name: DEV - Create database backup
```
peterstarling, line 1: I actually prefer the suffix than prefix

.github/workflows/prod.backup.yml
```diff
@@ -0,0 +1,29 @@
...
+    container:
+      image: registry.gitlab.com/yldio/asap-hub/squidex-utils
+    env:
+      SQUIDEX_APP: 'asap-hub'
+      BUCKET_NAME: 'asap-hub-prod-data-backup'
+      SQUIDEX_CLIENT_ID: ${{ secrets.SQUIDEX_CLIENT_ID_PROD }}
+      SQUIDEX_CLIENT_SECRET: ${{ secrets.SQUIDEX_CLIENT_SECRET_PROD }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    steps:
```
peterstarling, line 18: this is basically a copy of the dev workflow just with different params. You have probably looked into it but isn't there a way to make this reusable and not have to duplicate the code somehow? Unless only through an action? in which case we may want to consider make this job a private action

serverless.ts
```diff
@@ -274,6 +274,22 @@ const serverlessConfig: AWS = {
       IsDev: {
         'Fn::Equals': ['${self:provider.stage}', 'dev'],
       },
+      IsProd: {
```
peterstarling, line 277: what do you need `isProd` for?

serverless.ts
```diff
@@ -479,10 +495,19 @@ const serverlessConfig: AWS = {
       },
       DataBackupBucket: {
         Type: 'AWS::S3::Bucket',
-        Condition: 'IsDev',
+        Condition: 'IsDevOrProd',
         DeletionPolicy: 'Retain',
         Properties: {
           BucketName: '${self:service}-${self:provider.stage}-data-backup',
+          AccessControl: 'Private',
```
peterstarling, line 502: its private by default

serverless.ts
```diff
@@ -479,10 +495,19 @@ const serverlessConfig: AWS = {
...
+          LifecycleConfiguration: {
+            Rules: [
+              {
+                Id: 'delete-after-3-months',
+                ExpirationInDays: 90,
```
peterstarling, line 507: do we want the same rules for prod and dev?

peterstarling (changes requested):

.github/workflows/prod.backup.yml
johnytiago: I looked into it. I don't think the interface is clear enough. Plus, we want different cron frequencies for dev and prod.
Also I don't want to overcomplicate simple logic just because we want to keep it DRY

.github/workflows/dev.backup.yml
johnytiago: I find that when you have multiple items in a list the prefix makes is easier to find :)

serverless.ts
johnytiago: We just want to create a bucket for `prod` or `dev`.  This condition is used on the `IsDevOrProd` condition.

.github/workflows/dev.backup.yml
johnytiago: I thought it would be good to make this explicit. 🤔 I can remove it

serverless.ts
johnytiago: I thought it would be good to make this explicit. 🤔 I can remove it

serverless.ts
johnytiago: I think it's better they share these settings than just having expiration for prod

johnytiago: For now I don't mind them sharing the expiration date, but we can split in the future

.github/workflows/prod.backup.yml
peterstarling: i mean.. this is a huge amount of repetition so its not 'over complicating just because we want to keep it DRY'.

You've only just raised a PR for Docker changes and asked me to incorporate them because you didn't want to duplicate Docker config.

I would look into making this an action with params, as far as I am aware the workflow triggers and env vars can still be very much independent of the steps.

serverless.ts
peterstarling: oh i see, yeah, i guess that works too

.github/workflows/prod.backup.yml
johnytiago: Is this your attempt to avenge my complaint from your PR? 😄 

> You've only just raised a PR for Docker changes and asked me to incorporate them because you didn't want to duplicate Docker config.

This is incorrect. I suggested we could make the workflow simpler by leveraging docker. Use just one job instead of two. To which you even agreed.

Like I said, we want the backups to happen at different rates, which is not possible with just one action.

peterstarling: I dont understand - actions are this https://docs.github.com/en/actions/creating-actions/about-actions - how is this related to workflow schedule rates? Unless im missing something here..

johnytiago: We want the dev action to run once a week (https://github.com/yldio/asap-hub/pull/836/files#diff-6c5c55f22f119d90b63a7c93fa6ab557f0e95b1a8eac1b98c0ca8ccac95f9954R4) and the prod one to run twice a day (https://github.com/yldio/asap-hub/pull/836/files#diff-91c4c62ed059201e451aa08d9e9abfd2af8e1ad997cd25ebc940ac44a2a3402cR4). 

Would you have a way to have the same action trigger at two different rates for different environments?

peterstarling: from the documentation
```
Actions are individual tasks that you can combine to create jobs and customize your workflow. You can create your own actions, or use and customize actions shared by the GitHub community.
```

This means that your `dev.backup.yml` and `prod.backup.yml` instead of having all of these steps duplicated in both files:
```yml
      - name: Add the configuration
        run: sq config add $SQUIDEX_APP $SQUIDEX_CLIENT_ID $SQUIDEX_CLIENT_SECRET
      - name: Run sync
        run: sq sync out ./backup
      - name: Tar the files
        run: |
          FILE_NAME=$(date +%s| cut -b1-13)-squidex-dev-backup.tar.gz && \
          tar -czf $FILE_NAME backup/ && \
          echo "FILE_NAME=$FILE_NAME" >> $GITHUB_ENV
      - name: upload to S3
        run: aws s3api put-object --private --bucket $BUCKET_NAME --key $FILE_NAME --body $FILE_NAME --storage-class STANDARD_IA
```
Would only have a step like this:
```yml
    - uses: actions/javascript-action@v1.0.1
 ```
As far as I am aware this has nothing to do with workflow triggers.

peterstarling (approved):
just get rid of the PUSH trigger

PR #837, opened 2021-04-29
Fix date

kmaid:


johnytiago:
Doesn't this need tests?

kmaid:
I am going to do a follow up PR to move the format strings into the date.ts file

PR #838, opened 2021-04-29
feat(devops): setup & teardown PR squidex

johnytiago:
Ticket: https://trello.com/c/zcTu8S1x/1189-create-and-teardown-per-pr-squidex-application

- Moved dependency instalation to Docker to make CI jobs faster & more convinient
- Integration tests' Squidex app is named `PR_NUM-CI_JOB_ID` to prevent colision with 
the now long-lived PR Squidex app that is named after the PR Number
- Teardown happens on GH actions since these are the only ones aware that the PR is closed


.gitlab-ci.yml
```diff
@@ -136,19 +136,28 @@ test:build-output:
 deploy:sls:branch:
   <<: *tmpl_branch
   <<: *tmpl_deploy
```
peterstarling, line 138: do we need tmp_deploy if we are overwriting with `script`? i guess its for the `dependencies` but maybe this should be copied across, otherwise it kinda implies the reuse of more functionality than it really does

peterstarling (changes requested):
is this tested?

johnytiago:
> is this tested?

Heck yeah! https://838.hub.asap.science/

I notice that rules are not syncing as they should. But I will address that soon

.gitlab-ci.yml
johnytiago: The usage of these templates is quite questionable but this deploy one is the worst. No job uses it fully, either replacing the script or dependencies. It probably makes sense to just remove it. 🤷🏻‍♂️

peterstarling (approved):

PR #840, opened 2021-05-03
feat: Add teams to research-output response

peterstarling:
This PR adds teams do research-output response. 

Previously only a single team was present. However - because the RO response itself is also being attached to other various responses (ie Group response or Team response), this PR attaches them only for teams, not for groups.

This means that a team->RO->teams traversal is possible, but group->team->RO->teams is not. In order to adhere to the same contract - for the Group response only the parent team is being copied across into the `teams` property.

https://trello.com/c/FarA5LEp/1130-as-an-admin-i-want-to-associate-an-output-to-more-than-one-team-to-correctly-address-authorship

### DIFF
--- a/apps/asap-server/src/controllers/research-outputs.ts
+++ b/apps/asap-server/src/controllers/research-outputs.ts
@@ -33,7 +33,7 @@ function transform(
   };
 }
 
-export const GraphQLQueryResearchOutput = `
+export const getGraphQLQueryResearchOutput = (withTeams = false): string => `
 id
 created
 lastModified
@@ -50,7 +50,20 @@ flatData{
       name
     }
   }
-}`;
+}
+${
+  (withTeams &&
+    `referencingTeamsContents {
+  id
+  created
+  lastModified
+  flatData {
+    displayName
+  }
+}`) ||
+  ''
+}
+`;
 
 export default class ResearchOutputs implements ResearchOutputController {
   researchOutputs: InstrumentedSquidex<RestResearchOutput>;
--- a/apps/asap-server/src/controllers/research-outputs.ts
+++ b/apps/asap-server/src/controllers/research-outputs.ts
@@ -33,7 +33,11 @@ function transform(
   };
 }
 
-export const getGraphQLQueryResearchOutput = (withTeams = false): string => `
+export const getGraphQLQueryResearchOutput = ({
+  withTeams = false,
+}: {
+  withTeams: boolean;
+}): string => `
 id
 created
 lastModified
--- a/apps/asap-server/src/controllers/research-outputs.ts
+++ b/apps/asap-server/src/controllers/research-outputs.ts
@@ -56,16 +56,16 @@ flatData{
   }
 }
 ${
-  (withTeams &&
-    `referencingTeamsContents {
+  withTeams
+    ? `referencingTeamsContents {
   id
   created
   lastModified
   flatData {
     displayName
   }
-}`) ||
-  ''
+}`
+    : ''
 }
 `;
 
--- a/apps/asap-server/src/controllers/research-outputs.ts
+++ b/apps/asap-server/src/controllers/research-outputs.ts
@@ -34,7 +34,7 @@ function transform(
 }
 
 export const getGraphQLQueryResearchOutput = ({
-  withTeams = false,
+  withTeams,
 }: {
   withTeams: boolean;
 }): string => `
--- a/apps/asap-server/src/controllers/research-outputs.ts
+++ b/apps/asap-server/src/controllers/research-outputs.ts
@@ -11,20 +11,25 @@ import { sanitiseForSquidex } from '../utils/squidex';
 
 function transform(
   output: RestResearchOutput,
-  team?: RestTeam,
+  teams: RestTeam[],
 ): ResearchOutputResponse {
-  const teamProps = team
-    ? {
-        team: {
-          id: team.id,
-          displayName: team.data.displayName.iv,
-        },
-      }
-    : {};
+  const teamProps =
+    teams.length > 0
+      ? {
+          team: {
+            id: teams[0].id,
+            displayName: teams[0].data.displayName.iv,
+          },
+        }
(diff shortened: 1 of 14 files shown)


apps/asap-server/src/entities/team.ts
```diff
@@ -43,7 +43,15 @@ export const parseGraphQLTeam = (
   const outputs: ResearchOutputResponse[] = flatOutputs
     .map((o) => {
       const output = parseGraphQLResearchOutput(o as GraphqlResearchOutput);
-      return { ...output, team: { id: team.id, displayName } };
+
+      return {
+        ...output,
+        team: { id: team.id, displayName },
+        teams: (o as GraphqlResearchOutput).flatData?.teams?.map((t) => ({
```
peterstarling, line 50: this is where we either grab the `teams` response when they are present or otherwise use the parent `team`. 

For Group response for instance no `teams` are being fetched so only a single team will always be returned as part of the `teams` property (even though more might be present in the DB)

peterstarling:
Research Output response:
image

Team response, showing RO with parent teams:
image

Group response showing RO with only a single parent team:
image

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -11,24 +11,29 @@ import { sanitiseForSquidex } from '../utils/squidex';
...
-    : {};
+  const teamProps =
+    teams.length > 0
+      ? {
+          team: {
+            id: teams[0].id,
+            displayName: teams[0].data.displayName.iv,
+          },
+        }
+      : {};
```
peterstarling, line 24: needs to throw when no team was found, as it did previously

apps/asap-server/test/controllers/teams.test.ts
```diff
@@ -303,51 +303,90 @@ describe('Team controller', () => {
...
-          {
-            ...fetchTeamByIdExpectation.members[0],
-            avatarUrl: undefined,
-          },
-        ],
-      };
+      test('Should parse the team user correctly when the avatar is undefined', async () => {
+        const teamId = 'team-id-1';
+
+        delete (user.data as any).avatar;
```
johnytiago, line 354: Why do we need casting here?

apps/asap-server/test/controllers/teams.test.ts
```diff
@@ -303,51 +303,90 @@ describe('Team controller', () => {
...
-        .reply(200, graphQlTeamResponse)
-        .get(`/api/content/${config.appName}/users`)
-        .query({
-          $filter: `data/teams/iv/id eq '${teamId}'`,
-        })
-        .reply(200, userResponse);
+      test('Should parse the team user correctly when the avatar is null', async () => {
+        const teamId = 'team-id-1';
+
+        user.data.avatar.iv = null as any;
```
johnytiago, line 318: Since we need to mutate the base object, we maybe need a deep copy, or these tests could use different objects.
Won't the tests fail if you run the `delete` one before the assign no `null`?

apps/asap-server/test/controllers/teams.test.ts
```diff
@@ -303,51 +303,90 @@ describe('Team controller', () => {
...
+          ],
+        };
 
-      expect(result).toEqual(expectedResponse);
+        expect(result).toEqual(expectedResponse);
+      });
     });
 
-    test('Should parse the team user correctly when the avatar is undefined', async () => {
+    test('Should return more than a single team on the nested research-output', async () => {
```
johnytiago, line 388: I know you wanted to make it clear that we have a test case for the new functionality you added, but maybe we should make the assertion more explicit, otherwise it is just a duplicate with: https://github.com/yldio/asap-hub/blob/238050457b37380277be8eb4524699bd66498447/apps/asap-server/test/controllers/teams.test.ts#L251

johnytiago (changes requested):

apps/asap-server/test/controllers/teams.test.ts
peterstarling: well in this case the type of `user` is actually being inferred from an object here:

https://github.com/yldio/asap-hub/pull/840/files/238050457b37380277be8eb4524699bd66498447#diff-79015ebee0b55ad2336fccbd627b61384c730904ec40c7cf86a3ad57c904d90eR307

however - even with extra typings to make the `user` become `RestUser` the `avatar` is still marked as required and so we need to cast it as `any` before deleting the property. 

We do know the types are invalid for squidex lists and we have a ticket to address it here:

https://trello.com/c/oJqbA3z9/1106-update-typings-to-accommodate-for-iv-null-squidex-response

johnytiago: Alright! 👍🏽

apps/asap-server/test/controllers/teams.test.ts
peterstarling: yes, hence the copy:
https://github.com/yldio/asap-hub/pull/840/files/238050457b37380277be8eb4524699bd66498447#diff-79015ebee0b55ad2336fccbd627b61384c730904ec40c7cf86a3ad57c904d90eR307

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -45,7 +50,20 @@ flatData{
       name
     }
   }
-}`;
+}
+${
+  (withTeams &&
```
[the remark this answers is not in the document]
peterstarling: you think? why?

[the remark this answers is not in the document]
johnytiago: This is an `if/else` 🤷🏻‍♂️

apps/asap-server/test/controllers/teams.test.ts
johnytiago: I see, but my point was not file-wide, was only for the `avatar` tests. 
I think this is fine for now

johnytiago (approved):

PR #842, opened 2021-05-05
User Profile Empty States

kmaid:


packages/react-components/src/organisms/ProfileSkills.tsx
```diff
@@ -11,12 +13,21 @@ type ProfileSkillsProps = {
...
-}) => (
-  <Card>
-    <Headline2 styleAsHeading={3}>Expertise and Resources</Headline2>
-    {skillsDescription ? <Paragraph>{skillsDescription}</Paragraph> : null}
-    <TagList tags={skills} />
-  </Card>
-);
+}) => {
+  const { isOwnProfile } = useContext(UserProfileContext);
+  return skills.length || isOwnProfile ? (
```
jeysal, line 18: What if `!skills.length && !isOwnProfile && skillsDescription`, wouldn't this incorrectly hide the whole thing?

packages/react-components/src/organisms/UserProfileBiography.tsx
```diff
@@ -13,32 +15,42 @@ const stretchOnMobile = css({
...
-            View Biosketch
-          </Link>
-        </div>
-      )}
-    </div>
-  </Card>
-);
+}) => {
+  const { isOwnProfile } = useContext(UserProfileContext);
+  return biography || isOwnProfile ? (
```
jeysal, line 26: What if no bio but biosketch is there?

packages/react-components/src/organisms/__tests__/UserProfileBackground.test.tsx
```diff
@@ -110,3 +140,30 @@ it('renders approach if present', () => {
 
   expect(queryAllByText(/interests/i).length).toBeGreaterThan(0);
 });
+
+it('renders approach placeholder ownProfile is true', () => {
```
jeysal, line 144:
```suggestion
it('renders approach placeholder on your own profile'', () => {
```

packages/react-context/src/__tests__/user-profile.test.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import { useUserProfileContext, UserProfileContext } from '../user-profile';
+
+const TestComponent: React.FC<{ index?: number }> = () => {
```
jeysal, line 6: TBH I think we don't need the wrapping hook? And then this test is literally just testing React context, so it can go away? Or I guess to preserve code coverage the one thing we could test (and do test on other contexts) is the expected default value

jeysal (commented):
Looks clean, only thing I'd ask is to rephrase the various test case names to be natural (so easier to read and more user-focused) — "does x for isOwnProfile true" does not make for a good sentence :sweat_smile:

packages/react-components/src/organisms/UserProfileBiography.tsx
kmaid: I just maintained what was there before?

packages/react-components/src/organisms/ProfileSkills.tsx
kmaid: Like the other place i have preserved the behaviour. Without a requirement in the ticket i am reluctant to change

jeysal:
Oh, the weird `skills.length` / `biography` checks were previously just in the upper component. Okay

jeysal (approved):

PR #848, opened 2021-05-07
Profile header empty state and tests

kmaid:


packages/react-components/src/molecules/UserProfilePersonalText.tsx
```diff
@@ -31,19 +31,31 @@ const paragraphStyles = css({
 type UserProfilePersonalTextProps = Pick<
   UserResponse,
   'institution' | 'jobTitle' | 'location' | 'teams' | 'role'
->;
+> & {
+  isOwnProfile?: boolean;
```
jeysal, line 35: Didn't we just introduce a context for this?

packages/react-components/src/molecules/__tests__/SocialIcons.test.tsx
```diff
@@ -53,3 +54,19 @@ it('renders icon and link', () => {
...
+
+it('renders social icons without links for own profile', () => {
+  const { getByTitle } = render(
+    <UserProfileContext.Provider value={{ isOwnProfile: true }}>
+      <SocialIcons />
+    </UserProfileContext.Provider>,
+  );
+  const element = getByTitle('Research Gate');
+  expect(element).toBeInTheDocument();
+  expect(element).not.toHaveAttribute('href');
```
jeysal, line 71:
```suggestion
  expect(element.closest('a')).not.toHaveAttribute('href');
```
otherwise it doesn't say much

packages/react-components/src/molecules/UserProfilePersonalText.tsx
jeysal: Realized the question may sound snarky, so: I think we should not introduce a prop that is passed through, but directly consume the context, right? Its default value is also `false`

kmaid: It's used on the people listing card and we don't want to display the help text on the listing page assumably

packages/react-components/src/molecules/__tests__/SocialIcons.test.tsx
kmaid: Nice spot.

packages/react-components/src/molecules/UserProfilePersonalText.tsx
jeysal: But there the default value of the context (`false`) will just come into play so it's fine innit?

jeysal (commented):
Code lgtm but I have one concern why I think we can't actually merge this now: We are showing greyed out social icons on your profile, but you can't actually edit them yet. Was this discussed on product side? Maybe something to bring up with product?

kmaid:
image

jeysal (approved):

PR #850, opened 2021-05-07
feat(roms): add new fields and adds defauls migration

johnytiago:
Ticket: https://trello.com/c/gK3a6puv/1132-update-roms-schema

Not adding a rollback script since it doesn't make sense here.
Rolling back in the future would just be dangerous as it would delete data.

### DIFF
--- a/apps/asap-server/src/migrations/1620402284-default-fields-roms.ts
+++ b/apps/asap-server/src/migrations/1620402284-default-fields-roms.ts
@@ -14,7 +14,7 @@ type OldRestResearchOutput = Omit<
   usedInAPublication?: RestResearchOutput['data']['usedInAPublication'];
 };
 
-export default class MoveResearchOutputTextToDescription extends Migration {
+export default class SetResearchOutputDefaultFields extends Migration {
   up = async (): Promise<void> => {
     const squidexClient = new Squidex<RestResearchOutput>('research-outputs', {
       unpublished: true,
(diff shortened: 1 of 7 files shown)


packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -213,6 +213,186 @@
...
+        },
+        "isLocked": false,
+        "isHidden": false,
+        "isDisabled": false,
+        "partitioning": "invariant"
+      },
+      {
+        "name": "externalAuthors",
+        "properties": {
+          "isRequired": false,
```
peterstarling, line 350: should we not make this required so its always an empty array and not an optional property? it would make it easier to type and use but it may not be in line with our previous approach to nested fields

packages/squidex/src/entities/research-output.ts
```diff
@@ -12,6 +17,15 @@ export interface ResearchOutput {
   publishDate?: string;
   tags?: string[];
   adminNotes?: string;
+  subtype?: ResearchOutputSubtype;
```
peterstarling, line 20: looking at the requirements in the ticket it's not exactly super clear  to me if the field should contain only a single selected subtype or many. I'd probably clarify that with our PO before we merge this. I can see how the client may want to select several subtypes (like tags) - but i might be wrong.

packages/squidex/src/entities/research-output.ts
```diff
@@ -12,6 +17,15 @@ export interface ResearchOutput {
...
+  accessInstructions?: string;
+  sharingStatus: ResearchOutputSharingStatus;
+  asapFunded: DecisionOption;
+  usedInAPublication: DecisionOption;
+  externalAuthors?: {
```
peterstarling, line 25: i see how `tags` are declared in the same way - ie an optional property containing an array - but wouldn't it be easier to make it compulsory and keep it empty when there are no authors? Again, just asking. It's generally annoying when arrays are optional, nullable and can also be empty (although here null is not possible)

apps/asap-server/src/migrations/1620402284-default-fields-roms.ts
```diff
@@ -0,0 +1,68 @@
...
+type OldRestResearchOutput = Omit<
+  RestResearchOutput,
+  'sharingStatus' | 'asapFunded' | 'usedInAPublication'
+> & {
+  sharingStatus?: RestResearchOutput['data']['sharingStatus'];
+  asapFunded?: RestResearchOutput['data']['asapFunded'];
+  usedInAPublication?: RestResearchOutput['data']['usedInAPublication'];
+};
+
+export default class MoveResearchOutputTextToDescription extends Migration {
```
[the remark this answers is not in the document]
johnytiago: I missed that, thanks

packages/squidex/schema/schemas/research-outputs.json
johnytiago: Since this is a field that could get entries manually deleted, this one could easily end up in the `null` state we witnessed on the `avatar`. This way we will write safer code, what do you think? Either way, we can always send it as an array for the FE

packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -213,6 +213,186 @@
...
+          "isRequiredOnPublish": false,
+          "isHalfWidth": false,
+          "fieldType": "String",
+          "editor": "Dropdown",
+          "inlineEditable": false,
+          "isUnique": false,
+          "allowedValues": [
+            "ASAP annual meeting",
```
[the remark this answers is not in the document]
johnytiago: Even though I can see where you're coming from, I don't think it is a good idea to attach our typings to this _computer-generated, ever-changing, Squidex schemas_ 😄 We've seen the schema change its structure many times.

I think it would be better if we just try to generate the typings from the GraphQL schema instead.

Additionally, I wouldn't want to do such changes as part of this PR. :)

packages/squidex/src/entities/research-output.ts
johnytiago: Confirmed: single subtype.

peterstarling (approved):

PR #852, opened 2021-05-10
feat: Create last-updated-partial field for research-output

peterstarling:
see: https://trello.com/c/QNJQsmTQ/1169-return-research-outputs-last-updated-timestamp-to-fe

### DIFF
--- a/packages/squidex/schema/schemas/research-outputs.json
+++ b/packages/squidex/schema/schemas/research-outputs.json
@@ -220,7 +220,7 @@
         "name": "lastUpdatedPartial",
         "isHidden": false,
         "isLocked": false,
-        "isDisabled": false,
+        "isDisabled": true,
         "partitioning": "invariant",
         "properties": {
           "fieldType": "DateTime",


packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -10,7 +10,9 @@
       "label": "Research Outputs",
       "validateOnPublish": false
     },
-    "scripts": {},
+    "scripts": {
+      "update": "const modified = Object.keys(ctx.data)\n        .filter(key => JSON.stringify(ctx.data[key]) !== JSON.stringify(ctx.oldData[key]))\n        .filter(key => key !== 'publishDate' && key !== 'adminNotes');\n        \nif (modified.length > 0) {\n    ctx.data.test2.iv = JSON.stringify(modified);\n    ctx.data.lastUpdatedPartial.iv = new Date().toISOString().split('.')[0]+\"Z\";\n    replace();\n}\n\n"
```
peterstarling, line 14:
```typescript
const modified = Object.keys(ctx.data)
  .filter(
    (key) => JSON.stringify(ctx.data[key]) !== JSON.stringify(ctx.oldData[key])
  )
  .filter((key) => key !== "publishDate" && key !== "adminNotes");

if (modified.length > 0) {
  ctx.data.test2.iv = JSON.stringify(modified);
  ctx.data.lastUpdatedPartial.iv = new Date().toISOString().split(".")[0] + "Z";
  replace();
}
```

johnytiago: Can we remove the test bit?
Also, I don't think you need to remove the miliseconds from the date. https://support.squidex.io/t/implemented-support-iso-8601-missing-miliseconds/2191

apps/asap-server/src/entities/research-output.ts
```diff
@@ -15,6 +15,8 @@ export const parseResearchOutput = (
   publishDate: output.data.publishDate?.iv,
   addedDate: output.data.addedDate?.iv,
   tags: output.data.tags?.iv || [],
+  lastUpdatedPartial:
```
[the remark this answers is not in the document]
peterstarling: I'm not sure actually, these are two different fields, what if at some point in our admin panel we want to show when the resource was actually modified - including changes to `adminNotes` and `publishDate`? I know right now we don't have an admin panel (well its squidex) but I would prefer to plan ahead for things like this.

To me this is a whole new field. If for now we want to change it in the UI - then let's make that replacement in the UI but BE should imho show these two separately.

packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -213,6 +215,23 @@
           "isRequiredOnPublish": false,
           "isHalfWidth": false
         }
+      },
+      {
+        "name": "lastUpdatedPartial",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
```
johnytiago, line 223:
```suggestion
        "isDisabled": true,
```

Just missed this bit. The field should not be editable, right?

PR #856, opened 2021-05-12
feat: Add teams to user invite script

peterstarling:
completes the last requirement of: https://trello.com/c/hU9PRAES/1179-invitation-script-should-invite-any-user-irrespective-of-role

### DIFF
--- a/apps/asap-cli/src/invite/index.ts
+++ b/apps/asap-cli/src/invite/index.ts
@@ -61,7 +61,7 @@ export const inviteUsers = async (
     .flatMap((team) => team?.id)
     .filter((team) => typeof team !== 'undefined');
 
-  let teamMap: { [key: string]: string } | undefined = undefined;
+  let teamMap: { [key: string]: string } | undefined;
   if (userTeamIds.length > 0) {
     const teamQuery: Query = {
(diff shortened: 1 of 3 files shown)


apps/asap-cli/src/invite/index.ts
```diff
@@ -102,12 +129,22 @@ export const inviteUsers = async (
...
+        let userTeams;
+        if (teamMap) {
+          userTeams = user.data.teams.iv
+            ?.flatMap((team) => team.id)
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            .map((id) => teamMap![id])
+            .join(', ');
+        }
+        console.log(
+          `Invited user ${user.data.email.iv} ${userTeams && `(${userTeams})`}`,
```
johnytiago, line 141:
```suggestion
          `Invited user ${user.data.email.iv} ${userTeams || ""}`,
```
Would that be the same?

johnytiago (approved):

apps/asap-cli/src/invite/index.ts
peterstarling: No

PR #859, opened 2021-05-13
No results page

kmaid:


packages/react-components/src/organisms/ResultList.tsx
```diff
@@ -66,10 +69,29 @@ const ResultList: React.FC<ResultListProps> = ({
...
-    </section>
+    {numberOfItems > 0 ? (
+      <>
+        <main css={mainStyles}>{children}</main>
+        <section css={pageControlsStyles}>
+          <PageControls {...pageControlsProps} />
+        </section>
+      </>
+    ) : (
+      <section
```
jeysal, line 80: Seems this is the `<main>` in that branch

packages/react-components/src/organisms/ResultList.tsx
```diff
@@ -66,10 +69,29 @@ const ResultList: React.FC<ResultListProps> = ({
...
+        css={{
+          textAlign: 'center',
+        }}
+      >
+        <Paragraph primary>
+          <strong>No matches found!</strong>
+          <br />
+          <span css={{ color: lead.rgb }}>
+            We're sorry, we couldn't find results to match your search. Please
```
jeysal, line 89: urgh, this could appear with `searchQuery === ''`. but if it's specced this way in the ticket :shrug:

jeysal (commented):
lgtm, would recommend changing one `section` to a `main`. also for the love of god collapse that css object that makes 1 line be 5 instead :joy:

packages/react-components/src/organisms/ResultList.tsx
kmaid: I clarified this with Beatriz and its been noted in the ticket to behave like this

packages/react-components/src/organisms/ResultList.tsx
```diff
@@ -53,23 +53,42 @@ const ResultList: React.FC<ResultListProps> = ({
...
+          </strong>
+        </Paragraph>
+      )}
       {cardViewHref && listViewHref && (
-        <ListControls
-          isListView={isListView}
-          cardViewHref={cardViewHref}
-          listViewHref={listViewHref}
-        />
+        <div css={{ justifySelf: 'flex-end' }}>
```
jeysal, line 64: There's no flex anymore, so seems this div is dead?

jeysal (approved):

packages/react-components/src/organisms/ResultList.tsx
kmaid: it right aligns the controls

PR #861, opened 2021-05-13
fix: include migration scripts on webpack bundle

johnytiago:


peterstarling:
fixed in https://github.com/yldio/asap-hub/pull/888

PR #863, opened 2021-05-14
feat: Degrees on people listing page

kmaid:


jeysal (approved):
oof, this alone would cause me to remove my degree from my profile ^^

PR #865, opened 2021-05-17
Fix bug with close buttons not going to the correct place

kmaid:


jeysal (changes requested):
Yeah this is probably better that replicating the weird tab matching logic everywhere. Could try to abstract it but no concrete great idea atm.
However, regression test please :upside_down_face:

apps/frontend/src/network/users/__tests__/UserProfile.test.tsx
```diff
@@ -260,6 +260,41 @@ describe('a header edit button', () => {
...
+
+    const {
+      getByText,
+      getByTitle,
+      queryByText,
+      findByLabelText,
+      findByText,
+    } = await renderUserProfile(userProfile);
+
+    // Open and close on research tab
```
jeysal, line 277:
```suggestion
    // Open and close on research tab
    userEvent.click(await findByText(/research/i, { selector: 'nav *' }));
```
just because otherwise this also implicitly tests what the default tab is

apps/frontend/src/network/users/__tests__/UserProfile.test.tsx
```diff
@@ -260,6 +260,41 @@ describe('a header edit button', () => {
     );
   });
 
+  it('can open and close   personal info remaining on the tab', async () => {
```
jeysal, line 263:
```suggestion
  it('remains on the same tab after closing a modal', async () => {
```
just emphasizing what the crucial thing we're testing is

jeysal (approved):

PR #867, opened 2021-05-17
fix(team): member links

kmaid:


packages/react-components/src/templates/TeamProfileHeader.tsx
```diff
@@ -98,7 +98,9 @@ const TeamProfileHeader: React.FC<TeamProfileHeaderProps> = ({
                   key={memberId}
                   css={[listItemStyles, { left: `-${i * 3}px` }]}
                 >
-                  <Anchor href={network({}).users({ userId: memberId }).$}>
```
jeysal, line 101: This typechecked????

packages/react-components/src/templates/TeamProfileHeader.tsx
```diff
@@ -109,9 +111,11 @@ const TeamProfileHeader: React.FC<TeamProfileHeaderProps> = ({
               ))}
             <li css={extraUsersStyles}>
               {members.length > MAX_MEMBER_AVATARS && (
-                <Avatar
-                  placeholder={`+${members.length - MAX_MEMBER_AVATARS}`}
-                />
+                <Anchor href={`${route.about({}).$}#TEAM_MEMBERS_SECTION`}>
```
jeysal, line 114: This hardcoding has two problems:
1. Easy to break by changing strings to be inconsistent
2. No guaranteed uniqueness e.g. when component is mounted multiple times

To solve these, look at how `GroupProfile` in the FE generates a unique id and passes it to the link and down to the element. The lowest common ancestor of the two usages can do the same here (it's probably even in `react-components`, TeamProfilePage or something).

Also, see `GroupProfile.test` for a clean way to test at the common ancestor that the impl avoids the two problems

packages/react-components/src/templates/TeamProfileHeader.tsx
jeysal: I suppose it's not limited to the params that actually exist. Unfortunate, but meh, the real problem I guess is that the `.user` part was missing, which can't really be avoided by typechecks. Once the `.user` part with the param is added, one will notice that there shouldn't be two routes that get the `userId` param

jeysal (changes requested):
Requesting change to a less brittle fragment handling as we have elsewhere. The other comment is mostly talking to myself :sweat_smile:

jeysal (approved):

PR #868, opened 2021-05-17
fix(asap-cli): use array for ids, add error handling

johnytiago:
Ticket: https://trello.com/c/hU9PRAES/1179-invitation-script-should-invite-any-user-irrespective-of-role

### DIFF
--- a/apps/asap-cli/src/invite/index.ts
+++ b/apps/asap-cli/src/invite/index.ts
@@ -17,12 +17,9 @@ interface HTTPError extends Error {
 
 interface inviteUsersFn {
   /* eslint-disable no-unused-vars */
-  (
-    role?: string,
-    reinvite?: boolean,
-    take?: number,
-    skip?: number,
-  ): Promise<void>;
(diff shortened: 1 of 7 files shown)


apps/asap-cli/src/invite/index.ts
```diff
@@ -59,7 +59,7 @@ export const inviteUsers = async (
   const userTeamIds = usersToInvite
     .flatMap((user) => user.data.teams.iv)
     .flatMap((team) => team?.id)
-    .filter((team) => typeof team !== 'undefined');
```
johnytiago, line 62: This doesn't change the logic.

apps/asap-cli/src/invite/index.ts
```diff
@@ -59,7 +59,7 @@ export const inviteUsers = async (
...
+    .filter(Boolean) as string[];
```
johnytiago, line 62: This doesn't change the logic.

peterstarling (approved):
have you tested it? does this fix the error?

PR #870, opened 2021-05-20
feat: Show additional research-output information

peterstarling:
as per: https://trello.com/c/W40XJ3QT/1233-show-aditional-information-on-an-output-detail-page

Adds the following to research-output endpoint responses:
* usedInPublication
* asapFunded
* sharingStatus

### Notes

* `usedInPublication` was used in the response, omitting the article `a` which is present in the DB

### DIFF
--- a/apps/asap-server/src/entities/research-output.ts
+++ b/apps/asap-server/src/entities/research-output.ts
@@ -1,4 +1,8 @@
-import { ResearchOutputResponse } from '@asap-hub/model';
+import {
+  DecisionOption,
+  DecisionOptionSimple,
+  ResearchOutputResponse,
+} from '@asap-hub/model';
 import { GraphqlResearchOutput, GraphqlTeam } from '@asap-hub/squidex';
 import { parseDate } from '../utils/squidex';
 import { parseGraphQLUser } from './user';
@@ -50,6 +54,7 @@ export const parseGraphQLResearchOutput = (
     ...optionalAuthors,
     ...optionalTeams,
     sharingStatus: output.flatData?.sharingStatus || 'Public',
+    asapFunded: convertNotSureToUndefined(output.flatData?.asapFunded),
   };
 };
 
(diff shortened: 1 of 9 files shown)


apps/asap-server/src/entities/research-output.ts
```diff
@@ -49,6 +53,11 @@ export const parseGraphQLResearchOutput = (
     accessInstructions: output.flatData?.accessInstructions || undefined,
     ...optionalAuthors,
     ...optionalTeams,
+    sharingStatus: output.flatData?.sharingStatus || 'Public',
```
johnytiago, line 56:
```suggestion
    sharingStatus: output.flatData?.sharingStatus || 'Network Only',
```

The default is "Network Only"

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -213,6 +213,86 @@ describe('ResearchOutputs controller', () => {
       expect(result.type).toEqual('Proposal');
     });
 
+    test('Should default sharingStatus to Public when missing', async () => {
```
johnytiago, line 216: The default is Network Only

johnytiago:
`Yes`, `No` and `Not Sure` are a UX requirement on the CMS, but it doesn't make sense to pollute our responseTypes with them. They represent a boolean response, which is what I think they should be.

IMO: 
```
usedInPublication?: boolean;
asapFunded?: boolean;
sharingStatus: SharingStatus;
```

apps/asap-server/src/entities/research-output.ts
```diff
@@ -58,3 +63,8 @@ const parseGraphqlTeamLite = (
   id: graphqlTeam.id,
   displayName: graphqlTeam.flatData?.displayName || '',
 });
+
+const convertNotSureToUndefined = (
```
johnytiago, line 67:
```suggestion
const convertDecisionToBoolean = (
```
Or a name that better represents what the function does?

johnytiago (approved):

PR #871, opened 2021-05-20
refactor: use automatic JSX runtime

jeysal:
and update Emotion


jeysal:
@kmaid really you don't need to review but only to be happy about no more React imports

kmaid (approved):
Ive approved but the build needs to pass

jeysal:
urgh didn't check properly after a rebase

PR #872, opened 2021-05-21
chore: Updated vulnerable dependency

peterstarling:
Updates node-fetch to the latest beta due to security warnings from dependabot


johnytiago (approved):
Ughh! Is it ok for us to be using beta versions 👀 ?

peterstarling:
> Ughh! Is it ok for us to be using beta versions  ?

dependabot seems to like it :D

jeysal:
Just sticking a `^` to it is a bit ambitious, betas aren't that known to not introduce further breaking changes ^^

PR #876, opened 2021-05-21
chore(RO): change RO fields order

johnytiago:
Ticket: https://trello.com/c/ef7Mw780/1237-reorder-cms-fields-for-research-outputs

- feat(CI): apply schema to PR squidex app
- chore(RO): change RO fields order

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -156,6 +156,8 @@ deploy:sls:branch:
     - ./ci/config-squidex-rules.sh ./backup/rules
     # create-app.py configures sq to new app
     - sq sync in backup
+    # apply schema changes on top of dev backup
+    - sq sync in packages/squidex/schema -t schemas --delete
     - source ci/env-setup.sh
     - yarn sls deploy --verbose
 
--- a/packages/squidex/schema/schemas/research-outputs.json
+++ b/packages/squidex/schema/schemas/research-outputs.json
@@ -17,6 +17,20 @@
     "fieldsInLists": ["title", "type"],
     "fields": [
       {
+        "name": "Overview",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
+        "partitioning": "invariant",
+        "properties": {
+          "fieldType": "UI",
+          "editor": "Separator",
+          "isRequired": false,
+          "isRequiredOnPublish": false,
+          "isHalfWidth": false
+        }
+      },
+      {
         "name": "title",
         "isHidden": false,
         "isLocked": false,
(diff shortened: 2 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -156,6 +156,8 @@ deploy:sls:branch:
     - ./ci/config-squidex-rules.sh ./backup/rules
     # create-app.py configures sq to new app
     - sq sync in backup
+    # apply schema changes on top of dev backup
+    - sq sync in packages/squidex/schema -t schemas --delete
```
peterstarling, line 160: hm.... i dont mind adding it to the PR but... do you think maybe this should run on every commit on a non master branch? Imagine a scenario:

1. An engineer raises a branch with some code changes (not schema)
2. That branch is deployed, dev backup restored and no other changes applied (as per the change you are proposing)
3. That engineer than decides to make changes to the schema and pushes a commit
4. No changes are applied as this jobs trigger is a new PR only (right?)

packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -60,27 +74,94 @@
           "placeholder": "",
           "isRequired": true,
           "isRequiredOnPublish": false,
-          "isHalfWidth": false
+          "isHalfWidth": true
```
peterstarling, line 77: this intended? (nothing in the ticket description which is why i ask)

peterstarling, line 77: this intended?

peterstarling (changes requested):
The diff is a little difficult to follow (i dont remember how to turn on a more intelligent diff and whether it would help here) so it's worth looking over the changes you proposed (since they may have been generated by syncing the schema with sq cli).

.gitlab-ci.yml
johnytiago: This should do just that.
We have this verification step, but every commit inside a PR will pass it.
https://github.com/yldio/asap-hub/pull/876/files/acb2c10995ae13dc5b197168a309436e79997011#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7L14

packages/squidex/schema/schemas/research-outputs.json
johnytiago: Yeah. This makes a field only take half of the page on the CMS

johnytiago: Yeah. This makes a field only take half of the page on the CMS

peterstarling (approved):

PR #884, opened 2021-05-25
feat(react-components): various small design fixes

kmaid:
https://trello.com/c/RwcwXNgV/1328-add-ellipsis-to-type-and-subtype-and-introduce-layout-limitations

### DIFF
--- a/packages/react-components/src/organisms/ResultList.tsx
+++ b/packages/react-components/src/organisms/ResultList.tsx
@@ -15,7 +15,18 @@ const headerStyles = css({
   display: 'grid',
   gridAutoFlow: 'column',
   alignItems: 'center',
+  [`@media (max-width: ${mobileScreen.max}px)`]: {
+    gridAutoFlow: 'row',
+  },
 });
+
+const controlStyles = css({
+  justifySelf: 'end',
+  [`@media (max-width: ${mobileScreen.max}px)`]: {
+    justifySelf: 'unset',
+  },
+});
+
 const mainStyles = css({
   justifySelf: 'stretch',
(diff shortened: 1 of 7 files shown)


packages/react-components/src/organisms/ResultList.tsx
```diff
@@ -15,7 +15,18 @@ const headerStyles = css({
   display: 'grid',
```
jeysal, line 15: make this a flex-wrap for auto wrapping, will be much simpler and will know when to wrap more exactly than a breakpoint 🙃

packages/react-components/src/organisms/SharedResearchOutputHeaderCard.tsx
```diff
@@ -83,7 +83,19 @@ const SharedResearchOutputHeaderCard: React.FC<SharedResearchOutputHeaderCardPro
             </li>
             {subTypes.map((subType, i) => (
               <li css={typeStyles} key={`subtype-${i}`}>
-                <TagLabel>{subType}</TagLabel>
+                <TagLabel>
+                  <span
+                    css={{
+                      display: '-webkit-box',
```
jeysal, line 89: Yeah I don't think this is right. It should be something with `white-space: nowrap`, like UsersList and TeamsList which also abbreviate with an ellipsis once too long for the line. It should also be part of `TagLabel` itself right, because the current behavior of `TagLabel` when the content is just broken and not desirable in any case afaik

jeysal (changes requested):

jeysal (approved):

PR #887, opened 2021-05-28
feat: Add user onboarded state

peterstarling:
First stage of https://trello.com/c/CyzhVfx5/1299-8-add-user-state-to-cms-and-block-api-calls

Adds the `onboarded` flag to the `User`. Once the migration is run and all the existing as well as new users have got the flag set to `true` we can then implement the auth rule to reject the ones with a `false` value

## Notable changes
* adds a modification to auth0 rule which needs to be manually deployed

### DIFF
--- a/apps/asap-server/test/utils/migrations.test.ts
+++ b/apps/asap-server/test/utils/migrations.test.ts
@@ -1,0 +1,84 @@
+import { RestUser, Results, Squidex } from '@asap-hub/squidex';
+import { applyToAllItemsInCollection } from '../../src/utils/migrations';
+
+const mockFetch: jest.MockedFunction<Squidex<RestUser>['fetch']> = jest.fn();
+const mockConstructor = jest.fn();
+
+jest.mock('@asap-hub/squidex', () => ({
+  Squidex: class Squidex {
+    constructor(collection: string) {
+      mockConstructor(collection);
+    }
+    fetch = mockFetch;
+  },
+}));
+
+const restUserMock = {
+  id: 'userId',
+  onboarded: true,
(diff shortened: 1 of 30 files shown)


apps/asap-server/test/utils/migrations.test.ts
```diff
@@ -0,0 +1,84 @@
...
+}));
+
+const restUserMock = {
+  id: 'userId',
+  onboarded: true,
+  displayName: 'Cristiano Ronaldo',
+  created: '',
+  lastModified: '',
+  data: {},
+} as any as RestUser;
```
kmaid, line 23: How come this had to be casted?

kmaid (changes requested):
The fixtures are really repetitive. Ideally there would be a base one you overwrite specific props in a test. Could some of these fixtures be refactored to use your generator?

apps/asap-server/test/utils/migrations.test.ts
peterstarling: because of missing `data`, it just felt unnecessary to add it all in line

yeah fixtures are a pain, i would like to have fixture generators for most of the return types and then used across other generators when nesting is required but oh well, i dont think this is the right time to refactor

kmaid: At least we removed an additional fixture. When should this tech debt be addressed? I think doing it in parts during other tickets is the best way forward. Tech debt tickets don't provide any user value (usually)

harlan-kohll:
These changes look good to me.

kmaid (approved):

PR #888, opened 2021-05-28
fix: Migration file import

peterstarling:
Fixes the issue with failed import and directory listing.

see: https://trello.com/c/dO0omSAG/1312-fix-migration-execution-flow


peterstarling:
image
well it works...

peterstarling:
image
so does the rollback

johnytiago (approved):

PR #892, opened 2021-05-31
feat: Block non-onboarded users

peterstarling:
Second stage of: https://trello.com/c/CyzhVfx5/1299-8-add-user-state-to-cms-and-block-api-calls

This will effectively block all private endpoints from accessing by non-onboarded users.

Should not affect our current or new users as they should be automatically set to `onboarded: true` as per the migration in https://github.com/yldio/asap-hub/pull/887


harlan-kohll (approved):
LGTM

peterstarling:
closing as we removed this requirement from the story

PR #893, opened 2021-06-01
fix: Team pagination bug

peterstarling:
Fixes the bug with team pagination - the number of teams for each user should be fixed - this PR sets it to maximum 20.


harlan-kohll (approved):
LGTM!

PR #894, opened 2021-06-01
Feat: Website 1 & 2 for social icons

kmaid:
https://trello.com/c/ZGYvylID/1300-surface-website-1-and-website-2-on-user-profile

### DIFF
--- a/packages/react-components/src/molecules/SocialIcons.tsx
+++ b/packages/react-components/src/molecules/SocialIcons.tsx
@@ -12,29 +12,43 @@ import {
   researcherIdIcon,
   orcidSocialIcon,
   googleScholarIcon,
+  globeIcon,
 } from '../icons';
 import { perRem } from '../pixels';
 import { tin } from '../colors';
 
+const ROW_GAP = 12;
 const socialContainerStyles = css({
-  display: 'grid',
-  columnGap: `${12 / perRem}em`,
-  gridAutoFlow: 'column',
+  display: 'flex',
+  flexWrap: 'wrap',
+  paddingRight: `-${ROW_GAP / perRem}em`,
 });
 
 const iconStyles = css({
   paddingTop: `${15 / perRem}em`,
   paddingBottom: `${17 / perRem}em`,
+  paddingRight: `${ROW_GAP / perRem}em`,
 });
+
+const fillInactiveStyles = css({ svg: { fill: tin.rgb } });
+const strokeInactiveStyles = css({ svg: { stroke: tin.rgb } });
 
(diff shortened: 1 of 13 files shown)


packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -28,6 +27,7 @@ const containerStyles = css({
     ".             edit-personal-info" ${24 / perRem}em
     "personal-info personal-info     " auto
     "contact       edit-contact-info " auto
```
jeysal, line 29: Yeah I'm afraid I can't approve this with the duplicate `<SocialIcons>` :sweat_smile: it would be the very first time we duplicate DOM elements to do a `display: none` switch at a breakpoint.
It would seem to me that in order to have the one `SocialIcons` that is immediately inside the grid always placed correctly, we could extend the desktop grid's `1fr` column to two columns: `auto 1fr` or `max-content 1fr` or so, `contact social`. Then we have only one `SocialIcons` (the "non-inline" one) that we can place on both mobile and desktop.
We can pair on this tmrw

packages/react-components/src/molecules/SocialIcons.tsx
```diff
@@ -12,29 +12,43 @@ import {
...
+const fillInactiveStyles = css({ svg: { fill: tin.rgb } });
+const strokeInactiveStyles = css({ svg: { stroke: tin.rgb } });
+
 const SocialIconLink: React.FC<{
   link?: string;
   icon: JSX.Element;
+  strokeInactive?: boolean;
   isOwnProfile: boolean;
-}> = ({ link, icon, isOwnProfile }) =>
+}> = ({ link, icon, isOwnProfile, strokeInactive = false }) =>
```
jeysal, line 41: Urgh yeah we have another problem here, I remember seeing this briefly. The user of this should not have to decide how to transform the icon, each icon should know that on its own. This will also require some though probably

packages/react-components/src/molecules/SocialIcons.tsx
```diff
@@ -12,29 +12,35 @@ import {
...
   paddingBottom: `${17 / perRem}em`,
+  paddingRight: `${ROW_GAP / perRem}em`,
 });
 
+const inactiveStyles = css({ svg: { fill: tin.rgb, stroke: tin.rgb } });
+
 const SocialIconLink: React.FC<{
   link?: string;
   icon: JSX.Element;
+  strokeInactive?: boolean;
```
jeysal, line 38: dead prop?

jeysal (approved):

PR #895, opened 2021-06-02
feat(shared-research-output): "Used in a Publication" copy change

jeysal:


kmaid (approved):

jeysal:
urgh I even had a feeling this would break a test but then thought naaaahhh I'm sure it will be a generic enough regex

PR #896, opened 2021-06-02
feat: Block non-onboarded users from private endpoints

peterstarling:
Blocks non-onboarded users from accessing all private endpoints except for user related ones.
Also renames some poorly named fixtures

see: https://trello.com/c/Xws38Tfi/1297-5-redirect-user-to-profile-page


apps/asap-server/src/middleware/permission-handler.ts
```diff
@@ -0,0 +1,10 @@
+import Boom from '@hapi/boom';
+import { RequestHandler } from 'express';
+
+export const permissionHandler: RequestHandler = async (req, _res, next) => {
+  if (!req.loggedUser || req.loggedUser.onboarded !== true) {
```
harlan-kohll, line 5: Would `loggedInUser` perhaps be clearer than `loggedUser`? As my first thought when reading this was to do with logging

peterstarling: et voila https://github.com/yldio/asap-hub/pull/896/commits/80293c3f189ecaf6d35466e42380eebbab453396

harlan-kohll (approved):
LGTM

PR #897, opened 2021-06-02
Feat: Extends edit personal information modal

kmaid:
https://trello.com/c/KX24uBln/639-5-edit-social-media-links

### DIFF
--- a/packages/react-components/src/form.ts
+++ b/packages/react-components/src/form.ts
@@ -61,6 +61,7 @@ type ValidationTarget =
   | HTMLSelectElement;
 export function useValidation<T extends ValidationTarget>(
   customValidationMessage: string,
+  getValidationMessage?: (validityState: ValidityState) => string | void,
 ): {
   validationMessage: string;
   validationTargetProps: Pick<HTMLProps<T>, 'ref' | 'onBlur' | 'onInvalid'>;
@@ -83,9 +84,17 @@ export function useValidation<T extends ValidationTarget>(
     validationTargetProps: {
       ref: inputRef,
       onBlur: (event: FocusEvent<ValidationTarget>) =>
-        setValidationMessage(event.currentTarget.validationMessage),
+        setValidationMessage(
+          (getValidationMessage &&
+            getValidationMessage(event.currentTarget.validity)) ||
+            event.currentTarget.validationMessage,
+        ),
       onInvalid: (event: FormEvent<ValidationTarget>) => {
-        setValidationMessage(event.currentTarget.validationMessage);
+        setValidationMessage(
+          (getValidationMessage &&
+            getValidationMessage(event.currentTarget.validity)) ||
+            event.currentTarget.validationMessage,
+        );
         event.preventDefault();
       },
     },
--- a/packages/react-components/src/form.ts
+++ b/packages/react-components/src/form.ts
@@ -86,7 +86,7 @@ export function useValidation<T extends ValidationTarget>(
       onBlur: (event: FocusEvent<ValidationTarget>) => {
         setValidationMessage(
           (getValidationMessage &&
-            event.currentTarget.validity.valid &&
+            !event.currentTarget.validity.valid &&
             getValidationMessage(event.currentTarget.validity)) ||
             event.currentTarget.validationMessage,
         );
@@ -95,7 +95,7 @@ export function useValidation<T extends ValidationTarget>(
       onInvalid: (event: FormEvent<ValidationTarget>) => {
         setValidationMessage(
           (getValidationMessage &&
-            event.currentTarget.validationMessage &&
+            !event.currentTarget.validity.valid &&
             getValidationMessage(event.currentTarget.validity)) ||
             event.currentTarget.validationMessage,
         );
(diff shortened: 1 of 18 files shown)


packages/validation/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
jeysal, line 1: ???

packages/validation/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "module": "build/index.js",
+  "main": "build-cjs/index.js",
+  "types": "build/index.d.ts",
+  "scripts": {
+    "build:babel": "../../scripts/build-babel.sh build",
+    "watch:babel": "../../scripts/build-babel.sh watch"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs3": "^7.14.0",
+    "typescript": "~4.2.4"
```
jeysal, line 17: ???

packages/react-components/src/templates/ContactInfoModal.tsx
```diff
@@ -1,49 +1,232 @@
...
 
   return (
     <EditModal
       backHref={backHref}
       title="Your contact details"
       dirty={newEmail !== email}
-      onSave={() => onSave(newEmail)}
+      onSave={() =>
+        onSave({
+          contactEmail: newEmail,
```
jeysal, line 75: no `|| undefined`?

packages/react-components/src/templates/__tests__/ContactInfoModal.test.tsx
```diff
@@ -95,3 +97,29 @@ it('disables the form elements while submitting', async () => {
...
+  ${'Website 1'}      | ${'not url'} | ${'valid website'}
+  ${'Website 2'}      | ${'not url'} | ${'valid website'}
+  ${'ResearcherID'}   | ${'http://'} | ${'valid ResearcherID'}
+  ${'Twitter'}        | ${'http://'} | ${'valid Twitter handle'}
+  ${'Github'}         | ${'http://'} | ${'valid Github username'}
+  ${'LinkedIn'}       | ${'http://'} | ${'valid LinkedIn username'}
+  ${'Researchgate'}   | ${'http://'} | ${'valid Research Gate Profile ID'}
+  ${'Google Scholar'} | ${'http://'} | ${'valid Google Scholar Profile ID'}
+`(
+  'shows invalid message $message for $label input',
```
jeysal, line 112:
```suggestion
  'shows validation message "$message" for $label input',
```

packages/validation/src/user.ts
```diff
@@ -0,0 +1,3 @@
+export const USER_SOCIAL_WEBSITE = /^http(s?):\/\/.+/i;
+export const USER_SOCIAL_RESEARCHER_ID = /^[a-z]{1,3}-\d{4}-20[0-2]\d$/i;
```
jeysal, line 2:
```suggestion
export const USER_SOCIAL_RESEARCHER_ID = /^[a-z]{1,3}-\d{4}-\d{4,}$/i;
```
yeah idk if this was specified in the ticket but if it was we should push back :D I don't want a regex that requires maintenance in a maximum of 8.5yrs :D

packages/validation/src/user.ts
```diff
@@ -0,0 +1,3 @@
...
+export const USER_SOCIAL_NOT_URL = /^(?!\s*http(s?):\/\/).+/i;
```
jeysal, line 3: `USER_SOCIAL_NOT_WEBSITE` would be consistent with the other thing, and more correct since other URLs are matched.
It's good that they are prefixed with USER_SOCIAL I think, the regexes as per requirements are hardly well-thought out so this method of 'website checking' should stay local to user socials :D

jeysal (changes requested):

packages/react-components/src/molecules/TeamsList.tsx
```diff
@@ -101,7 +101,7 @@ const TeamsList: React.FC<TeamsListProps> = ({
             key={`sep-${i}`}
           >
             {inline || teamIcon}
-            <div css={inline && teamInlineStyles}>
+            <div css={inline ? teamInlineStyles : undefined}>
```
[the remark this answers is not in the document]
kmaid: None, emotion doesn't like false being passed to it. It complains in the test output

packages/react-components/src/form.ts
```diff
@@ -82,10 +83,22 @@ export function useValidation<T extends ValidationTarget>(
     validationMessage,
     validationTargetProps: {
       ref: inputRef,
-      onBlur: (event: FocusEvent<ValidationTarget>) =>
-        setValidationMessage(event.currentTarget.validationMessage),
+      onBlur: (event: FocusEvent<ValidationTarget>) => {
+        setValidationMessage(
+          (getValidationMessage &&
+            event.currentTarget.validationMessage &&
```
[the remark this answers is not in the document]
kmaid: This was intentional so the validation message gets reset to ''. Otherwise we can problems where the validation message shows up immediately.

jeysal (approved):

PR #898, opened 2021-06-02
feat(frontend): redirect users to own profile until onboarded

jeysal:


jeysal:
Just realized I need to exempt the logout route too

kmaid (approved):

PR #902, opened 2021-06-04
refactor(react-components): rename TagLabel to Pill and extract PillList

jeysal:


jeysal:
@kmaid would be good if you could get this into master on Monday after review since I'll be away. Not that I expect conflicts but always ideal not to have branches lying around

kmaid (approved):

PR #904, opened 2021-06-08
Feat: Skills edit modal

kmaid:
https://trello.com/c/a2kdryza/417-skills-and-expertise-modal-overview-only


apps/frontend/src/network/users/Research.tsx
```diff
@@ -78,6 +79,11 @@ const Research: React.FC<ResearchProps> = ({ user }) => {
               />
             </Frame>
           </Route>
+          <Route path={path + route.editSkills.template}>
+            <Frame title="Edit Expertise and resources">
```
jeysal, line 83:
```suggestion
            <Frame title="Edit Expertise and Resources">
```

packages/react-components/src/templates/__tests__/SkillsModal.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+it('renders default values into text inputs', () => {
+  const { getByLabelText } = render(
+    <SkillsModal {...props} skillsDescription="example description" />,
+    { wrapper: StaticRouter },
+  );
+  expect(getByLabelText(/overview/i)).toHaveValue('example description');
+});
+
+it('triggers the save function', async () => {
+  const jestFn = jest.fn();
```
jeysal, line 31: naming....
```suggestion
  const handleSave = jest.fn();
```

jeysal (approved):
minor comments then lgtm

jeysal:
Oh and as always the tags placeholder being there and telling you to add tags to your profile, while the modal does not allow you to add tags to your profile, seems hugely problematic to me, but I trust you are coordinating this with product

apps/frontend/src/network/users/Research.tsx
kmaid: I copy paste all copy but I think your correct 🤷

jeysal: This is not copy though is it, we make up the `<title>`s ourselves

PR #906, opened 2021-06-08
Edit modal your role in asap network

migcarva:


packages/react-components/src/atoms/RequiredAsterisk.tsx
```diff
@@ -0,0 +1,13 @@
...
+  | {
+      readonly required?: false;
+      readonly omitRequiredAsterisk?: never;
+    }
+  | { readonly required?: true; readonly omitRequiredAsterisk?: boolean };
+
+const RequiredAsterisk: React.FC<RequiredAsteriskProps> = ({
+  required = false,
+  omitRequiredAsterisk = false,
+}) => (required && !omitRequiredAsterisk && <>*</>) || null;
```
jeysal, line 11: These two booleans are really just one, `false, false`, `false, true`, `true, true` all do the same (no asterisk), only `true, false` means an asterisk. So it can just be one boolean `required`

migcarva: That would mean moving the logic to display the asterisk in the first place to the parent Labeled element, which would mean more repeated code. Up to you and @kmaid

jeysal: Not necessarily, it would just mean reducing props from two bools to one. Although I'm also not opposed to taking the component out because yeah it does seem a bit excessive to have a 'render a * or nothing' component

migcarva: Originally, the components had the title (label) as well. So it would render the label, and the required icon if necessary. This final version is Tom's suggestion. We can ditch the component altogether, we are actually only using it in the LabeledTextField and LabeledTextArea. I thought we would use it in more scenarios.

kmaid: Yeah, now its reduced to two usages this additional component is too verbose. Do we even need omit Asterisk now we have removed required from email? We could probably remove it from password too and just have the invalid username and password error message

packages/react-components/src/templates/TeamMembershipModal.tsx
```diff
@@ -87,16 +87,23 @@ const TeamMembershipModal: React.FC<TeamMembershipModalProps> = ({
...
               maxLength={200}
               enabled={!isSaving}
               value={newApproach}
               onChange={setNewApproach}
             />
             <LabeledTextArea
+              required
               title="Your responsibilities"
               placeholder="Example: Randy will be responsible for applying basic principles he developed from studies of a simple eukaryote, yeast, to investigate the mechanisms of intracellular vesicular transport and biogenesis of extracellular vesicles (exosomes) in cultured human cells. His team's current work is devoted to understanding how proteins and RNA are sorted into extracellular vesicles and how these molecules may be delivered to target cells in relation to normal and pathological functions."
+              // customValidationMessage={
```
kmaid, line 103: These comments should be removed

kmaid (changes requested):

kmaid (approved):

PR #908, opened 2021-06-11
Feat: Change copy

kmaid:
https://trello.com/c/afgH4HvJ/1379-change-the-error-message-for-website-1-and-2-on-edit-my-profile-modal


jeysal (commented):
I personally think we should ask if we can change to `starting with http:// or https://`. Or @migcarva do you think there would be a reason not to :sweat_smile:

PR #910, opened 2021-06-11
feat(edit-personal-info): position TextField => Typeahead

jeysal:


jeysal:
Okay so this version is much much nicer, just one small hack left. Reason why I couldn't get something like this to work originally was I had `Input` component identity change on every render so the `input` field would lose focus / cursor position all the time

packages/react-components/src/atoms/Dropdown.tsx
```diff
@@ -1,111 +1,16 @@
...
 import { dropdownChevronIcon } from '../icons';
-import { perRem, lineHeight } from '../pixels';
-import { ember, steel, mint, pine, tin, rose, lead, silver } from '../colors';
-import { Option } from '../select';
+import { Option, reactSelectStyles } from '../select';
 
 const containerStyles = css({
   flexBasis: '100%',
 });
-
```
jeysal, line 21: extracted to share between Dropdown and Typeahead

packages/react-components/src/select.ts
```diff
@@ -1,4 +1,103 @@
...
+  borderWidth,
+  paddingLeftRight,
+} from './form';
+import { perRem, lineHeight } from './pixels';
+
 export interface Option<V extends string> {
   value: V;
   label: string;
 }
+
```
jeysal, line 16: ...extracted to here

packages/react-components/src/form.ts
```diff
@@ -83,21 +88,25 @@ export function useValidation<T extends ValidationTarget>(
     validationMessage,
     validationTargetProps: {
       ref: inputRef,
-      onBlur: (event: FocusEvent<ValidationTarget>) => {
+      onBlur: () => {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
kmaid, line 92: Might be worth the comment or putting in the handle gracefully and ignore coverage.

packages/react-components/src/molecules/LabeledTypeahead.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type LabeledTypeaheadProps = {
+  readonly title: ReactNode;
+} & Exclude<ComponentProps<typeof Typeahead>, 'id'>;
+const LabeledTypeahead: React.FC<LabeledTypeaheadProps> = ({
+  title,
+  ...typeaheadProps
+}) => (
+  <div css={{ paddingBottom: `${18 / perRem}em` }}>
+    <Label forContent={(id) => <Typeahead {...typeaheadProps} id={id} />}>
```
kmaid, line 13: Do we need the required thing?

jeysal: What required thing, sorry?

packages/react-components/src/form.ts
jeysal: Tbh I'd like it to throw if for some reason this is null. It's a path that should be impossible to produce. If it doesn't throw, it's harder to figure out why validation isn't working

packages/react-components/src/molecules/LabeledTypeahead.tsx
jeysal: Clarified, was about the asterisk, but it's there already

jeysal: pushed an additional refactor commit to straighten out asterisk handling across input types

kmaid (approved):

PR #912, opened 2021-06-14
feat: Onboard user

peterstarling:
see: https://trello.com/c/XnOmu8Ni/1301-enable-complete-users-to-publish-their-profile

## Notable changes
- added a validation function for profile completeness
- refactored some tests, added extra coverage for the `patch` method in the user controller
- applied the validation rules in the route as they dont seem to belong to the controller method

### DIFF
--- a/apps/asap-server/src/routes/user.route.ts
+++ b/apps/asap-server/src/routes/user.route.ts
@@ -170,7 +170,7 @@ export const userRouteFactory = (
 
     const result = await userController.update(userId, userProfileUpdate);
 
-    if (typeof onboarded === 'boolean') {
+    if (onboarded === true) {
       if (!isUserOnboardable(result).isOnboardable) {
         throw Boom.badData('User profile is not complete');
       }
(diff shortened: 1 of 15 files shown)


apps/asap-server/src/routes/user.route.ts
```diff
@@ -166,7 +166,22 @@ export const userRouteFactory = (
       throw Boom.forbidden();
     }
 
-    const result = await userController.update(userId, update);
+    const { onboarded, ...userProfileUpdate } = update;
+
+    const result = await userController.update(userId, userProfileUpdate);
+
+    if (typeof onboarded === 'boolean') {
```
kmaid, line 173: This is kind of weird. You could send false and it would reject.

kmaid (approved):

apps/asap-server/src/routes/user.route.ts
peterstarling: i dont think i understand?

harlan-kohll (approved):
LGTM!

PR #913, opened 2021-06-14
feat: Add Algolia config 

peterstarling:
Adds Application ID to frontend and asap-server

see: https://trello.com/c/ZgizdZzU/1329-set-up-algolia-search-engine

### DIFF
--- a/apps/frontend/src/config.ts
+++ b/apps/frontend/src/config.ts
@@ -1,3 +1,4 @@
 export const GTM_CONTAINER_ID = process.env.REACT_APP_GTM_CONTAINER_ID;
 export const API_BASE_URL =
   process.env.REACT_APP_API_BASE_URL || 'http://localhost:3333';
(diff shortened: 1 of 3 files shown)


peterstarling:
image
@kmaid i dont really know how to test it - i added an `echo` to print out the value of the env var and it looks good (i have since removed the echo)

apps/frontend/src/config.ts
```diff
@@ -1,3 +1,4 @@
 export const GTM_CONTAINER_ID = process.env.REACT_APP_GTM_CONTAINER_ID;
 export const API_BASE_URL =
   process.env.REACT_APP_API_BASE_URL || 'http://localhost:3333';
+export const ALGOLIA_APP_ID = process.env.ALGOLIA_APP_ID || '';
```
kmaid, line 4: This should default to dev id (for local development)

kmaid (approved):

PR #915, opened 2021-06-15
feat: Add monitoring

peterstarling:
adds monitoring with sentry

see: https://trello.com/c/4SDrrakq/1123-set-up-monitoring-and-alerting-for-asap-server


apps/asap-server/src/handlers/api-handler.ts
```diff
@@ -40,3 +41,13 @@ export const apiHandler = serverlessHttp(app, {
     logger.withRequest(event, context);
   },
 });
+
+Sentry.AWSLambda.init({
+  dsn: "https://c61a36e015ef481bb29e0bd2a9f84aa3@o850903.ingest.sentry.io/5817859",
```
kmaid, line 46: Does this mean all environments including localhost will report errors to the same place?

peterstarling: perhaps i should have made ti clear that this was a POC

PR #917, opened 2021-06-17
Updating TextField to support both right and left indicators

migcarva:
The following changes were introduced:
- As discussed with @kmaid, we used the custom indicator of the text field input, and made it work for both the right and left sides concurrently, which wasn't the case before.
- The loading and validity check icons were removed, as requested by product.
- The Password field had to be updated to reflect the changes and use the new interface for the Text Field input.

(Requirements: https://trello.com/c/P6AT7U22/1120-add-a-custom-icon-for-x-on-all-search-bars)

### DIFF
--- a/packages/react-components/src/atoms/TextField.tsx
+++ b/packages/react-components/src/atoms/TextField.tsx
@@ -97,17 +97,11 @@ const containerStyles = css({
   position: 'relative',
 });
 
-const getIndicatorPadding = (aspectRatio: number, position: Position) => {
-  const padding = `${
+const getIndicatorPadding = (icon: React.ReactElement) => {
+  const aspectRatio = getSvgAspectRatio(icon);
+  return `${
     (paddingLeftRight + indicatorSize * aspectRatio + indicatorPadding) / perRem
   }em`;
-  return position === 'right'
-    ? css({
-        paddingRight: padding,
-      })
-    : css({
-        paddingLeft: padding,
-      });
 };
 
 const getIndicatorStyles = (aspectRatio: number, position: Position) =>
@@ -139,7 +133,7 @@ type TextFieldProps = {
   readonly customValidationMessage?: string;
 
   readonly leftIndicator?: React.ReactElement;
-  readonly rightIndicator?: React.ReactElement | null;
+  readonly rightIndicator?: React.ReactElement;
 
   readonly labelIndicator?: React.ReactElement | string;
   readonly getValidationMessage?: Parameters<typeof useValidation>[1];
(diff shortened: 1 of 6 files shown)


packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -35,12 +45,34 @@ const SearchField: React.FC<SearchProps> = (props) => {
   }, [debouncedValue, props.placeholder, props.value]);
 
   return (
-    <div css={{ padding: `${18 / perRem}em 0` }}>
+    <div
+      css={{
+        padding: `${18 / perRem}em 0`,
```
kmaid, line 50: Please move this out of the component into a css variable

packages/react-components/src/atoms/TextField.tsx
```diff
@@ -214,41 +189,44 @@ const TextField: React.FC<TextFieldProps> = ({
...
           !labelIndicator && { gridColumn: '1 / span 2' },
-          customIndicator &&
-            customIndicatorPadding(
-              getSvgAspectRatio(customIndicator),
-              customIndicatorPosition,
-            ),
-          loading && loadingStyles,
+
+          leftIndicator &&
+            getIndicatorPadding(getSvgAspectRatio(leftIndicator), 'left'),
```
kmaid, line 204: I think this could be simplified further. Return the padding string and set it to padding left or right where appropriate

kmaid (changes requested):

packages/react-components/src/atoms/TextField.tsx
```diff
@@ -156,9 +138,9 @@ type TextFieldProps = {
   readonly indicateValid?: boolean;
   readonly customValidationMessage?: string;
 
-  readonly loading?: boolean;
-  readonly customIndicator?: React.ReactElement;
-  readonly customIndicatorPosition?: Position;
+  readonly leftIndicator?: React.ReactElement;
+  readonly rightIndicator?: React.ReactElement | null;
```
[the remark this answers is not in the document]
migcarva: Yep, it was from a previous test, thanks :)

packages/react-components/src/molecules/__tests__/SearchField.test.tsx
```diff
@@ -35,6 +35,25 @@ describe('GTM data', () => {
       expect.anything(),
     );
   });
+  it('shows an custom X to clear the query', () => {
+    const { getByRole } = render(<SearchField placeholder="p" value="q" />);
```
kmaid, line 39: This test should be improved. You should use rerender. First without a value and ensuring the button isn't present. The second rerender with the value and the button is present

packages/react-components/src/molecules/__tests__/SearchField.test.tsx
```diff
@@ -35,6 +35,25 @@ describe('GTM data', () => {
...
+  it('clears the query after the custom X is pushed', () => {
+    const onChange = jest.fn();
+    const { getByRole } = render(
+      <SearchField placeholder="p" value="q" onChange={onChange} />,
+    );
+    fireEvent.click(getByRole('button'));
+    expect(onChange).toHaveBeenCalledWith('');
+
+    act(() => {
+      jest.advanceTimersByTime(30 * 1000);
```
kmaid, line 51: This is brittle. Look for waitFor to solve this

packages/react-components/src/molecules/__tests__/SearchField.test.tsx
```diff
@@ -35,6 +35,25 @@ describe('GTM data', () => {
...
+    });
+    expect(window.dataLayer?.reduce(Object.assign)).not.toHaveProperty(
+      SEARCH_EVENT,
```
kmaid, line 54: We should not be testing analytics in this test.

kmaid (approved):

PR #918, opened 2021-06-21
Contact CTA card BE implementation

harlan-kohll:
Implement the BE component of the Contact CTA card, as detailed here: https://trello.com/c/UJXl8wzr/1234-add-contact-cta-to-output-details-page

The BE should return a deduplicated array of PMs emails for the relevant shared RO.

### DIFF
--- a/apps/asap-cli/src/import/protocols/insert.ts
+++ b/apps/asap-cli/src/import/protocols/insert.ts
@@ -99,7 +99,7 @@ export default async (data: Protocol): Promise<void> => {
     sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
     asapFunded: { iv: 'Not Sure' as DecisionOption },
     usedInAPublication: { iv: 'Not Sure' as DecisionOption },
-    pmsEmails: { iv: null }
+    pmsEmails: { iv: null },
   };
 
   const [e1, teamId] = await Intercept(fetchTeamByName(data.team, promises));
--- a/apps/asap-cli/src/import/protocols/insert.ts
+++ b/apps/asap-cli/src/import/protocols/insert.ts
@@ -99,6 +99,7 @@ export default async (data: Protocol): Promise<void> => {
     sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
     asapFunded: { iv: 'Not Sure' as DecisionOption },
     usedInAPublication: { iv: 'Not Sure' as DecisionOption },
+    pmsEmails: { iv: null }
   };
 
   const [e1, teamId] = await Intercept(fetchTeamByName(data.team, promises));
--- a/apps/asap-cli/src/import/protocols/insert.ts
+++ b/apps/asap-cli/src/import/protocols/insert.ts
@@ -83,7 +83,7 @@ const addProtocolToTeam = async (
 
 export default async (data: Protocol): Promise<void> => {
   const promises: Cache = {};
-  const protocol: RestResearchOutput['data'] = {
+  const protocol: Omit<RestResearchOutput['data'], 'pmsEmails'> = {
     type: { iv: 'Protocol' },
     title: { iv: data.name },
     link: { iv: data.link },
@@ -99,7 +99,6 @@ export default async (data: Protocol): Promise<void> => {
     sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
     asapFunded: { iv: 'Not Sure' as DecisionOption },
     usedInAPublication: { iv: 'Not Sure' as DecisionOption },
-    pmsEmails: { iv: null },
   };
 
   const [e1, teamId] = await Intercept(fetchTeamByName(data.team, promises));
--- a/apps/asap-cli/src/import/protocols/insert.ts
+++ b/apps/asap-cli/src/import/protocols/insert.ts
@@ -6,6 +6,7 @@ import {
   RestUser,
   RestResearchOutput,
   RestTeam,
+  CreateResearchOutput,
 } from '@asap-hub/squidex';
 import { DecisionOption, ResearchOutputSharingStatus } from '@asap-hub/model';
 import { DateTime } from 'luxon';
@@ -15,9 +16,12 @@ import log from '../../logger';
 
 const users = new Squidex<RestUser>('users', { unpublished: true });
 const teams = new Squidex<RestTeam>('teams', { unpublished: true });
-const protocols = new Squidex<RestResearchOutput>('research-outputs', {
-  unpublished: true,
-});
+const protocols = new Squidex<RestResearchOutput, CreateResearchOutput>(
+  'research-outputs',
+  {
+    unpublished: true,
+  },
+);
 
 interface Cache {
   [key: string]: Promise<RestUser | null>;
--- a/apps/asap-cli/test/import/protocols.fixture.ts
+++ b/apps/asap-cli/test/import/protocols.fixture.ts
@@ -1,4 +1,9 @@
-import { RestResearchOutput, RestUser, RestTeam } from '@asap-hub/squidex';
+import {
+  CreateResearchOutput,
+  RestResearchOutput,
+  RestUser,
+  RestTeam,
+} from '@asap-hub/squidex';
 import { DecisionOption, ResearchOutputSharingStatus } from '@asap-hub/model';
 
 export const createProtocolsRequest: Omit<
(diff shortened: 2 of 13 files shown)


apps/asap-cli/src/import/protocols/insert.ts
```diff
@@ -99,6 +99,7 @@ export default async (data: Protocol): Promise<void> => {
     sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
     asapFunded: { iv: 'Not Sure' as DecisionOption },
     usedInAPublication: { iv: 'Not Sure' as DecisionOption },
+    pmsEmails: { iv: null }
```
harlan-kohll, line 102: Make tsc happy

apps/asap-cli/test/import/protocols.fixture.ts
```diff
@@ -18,6 +18,7 @@ export const createProtocolsRequest: RestResearchOutput['data'] = {
   sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
   asapFunded: { iv: 'Not Sure' as DecisionOption },
   usedInAPublication: { iv: 'Not Sure' as DecisionOption },
+  pmsEmails: { iv: null }
```
harlan-kohll, line 21: Make tsc happy

apps/asap-server/src/entities/research-output.ts
```diff
@@ -40,6 +40,24 @@ export const parseGraphQLResearchOutput = (
       }
     : {};
 
+
+  function userIsPMInTeam(user: GraphqlUser, team: GraphqlTeam): boolean {
+    const teamsMatchAndUserIsPM = (innerTeam: UserTeamConnection<GraphqlTeam>) =>
+      innerTeam.id[0].id === team.id && innerTeam.role === 'Project Manager';
+
+    const filteredTeams = user.flatData?.teams?.filter(teamsMatchAndUserIsPM);
```
peterstarling, line 53: so... there is a small problem with your types - they are not aligned with the graphql query. Well not entirely at least, let me explain.

In here 
https://github.com/yldio/asap-hub/blob/6328a1b8cdd7753da8d2fdf652f12919a847aeb2/apps/asap-server/src/controllers/research-outputs.ts#L60-L64

you specify that you only want an `email` from a `User` but later on you have it declared as 

https://github.com/yldio/asap-hub/blob/f18503f576e1abeacb0a0a702b083309a5e96a0f/packages/squidex/src/entities/team.ts#L24

which suggests that we should get a full `GraphqlUser` from the relationship where in fact we only ever get the `email`. You probably did it cause you followed this:

https://github.com/yldio/asap-hub/blob/68ba60814eb35ea05cfc9b15e9d928a87f411919/packages/squidex/src/entities/research-output.ts#L49

Which makes sense (I'm pretty sure I told you to follow similar implementations) but unfortunately in your situation it will not work. The type logically is correct because every property of a `GraphqlUser` is marked as optional - except we know for a fact only `email` will ever be present so your filters will not work (`.teams` will always come as `undefined`).

Also when you say that your `referencingUsersContents` is of type `GraphqlUser` you are actually making it more difficult to filter users by a role as you are then having to traverse back through to the team the user belongs to. Instead you could create a new type for the user connection where alongside user email a `role` will be added. Check out this query:
```
{
  queryResearchOutputsContents {
    id
    referencingTeamsContents {
      referencingUsersContents {
        flatData {
          email
          role
        }
      }
    }
  }
}
```
in here our `referencingUsersContents` should point to a type which only has two properties `email` and `role` and you can use the role to limit it to just the PMs.

peterstarling (changes requested):

apps/asap-server/src/entities/research-output.ts
```diff
@@ -40,6 +40,24 @@ export const parseGraphQLResearchOutput = (
       }
     : {};
 
+
+  function userIsPMInTeam(user: GraphqlUser, team: GraphqlTeam): boolean {
```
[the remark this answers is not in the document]
harlan-kohll: Fixed

apps/asap-server/src/entities/research-output.ts
harlan-kohll: I initially tried
```
flatData {
  email
  role
}
```
but it seems to come back with the wrong kind (?) of role: either `Guest` or `Grantee` and looking in `apps/asap-cli/src/import/users/parse.ts`, I guess possibly `Staff` also. The kind of role needed is in `packages/model/src/team.ts`. where `"Project Manager"` is defined on `TeamRole`. That being said in the PR env `pmsEmails` is always coming back empty `[]` so there may be an issue with the query.

peterstarling: i explained why:
image

try adding 
```
teams {
    role
    id {
      id
    }
}
```
to the flatData (where you said you tried adding just `role`)

harlan-kohll: This works

harlan-kohll: Manual testing:

From `https://918.hub.asap.science/shared-research`:

The two PM emails for the first shared RO `pmsEmails: ["test-email1@asap.science", "harlan.kohll@yld.io"]` are returned and the non-PM member is not.

apps/asap-cli/src/import/protocols/insert.ts
```diff
@@ -99,6 +99,7 @@ export default async (data: Protocol): Promise<void> => {
     sharingStatus: { iv: 'Network Only' as ResearchOutputSharingStatus },
     asapFunded: { iv: 'Not Sure' as DecisionOption },
     usedInAPublication: { iv: 'Not Sure' as DecisionOption },
+    pmsEmails: { iv: null },
```
peterstarling, line 102: this may make tsc happy but it doesn't look valid. We do not actually insert pm emails, do we, so i think that the type of the object is incorrect. Perhaps we need to omit this property in order to not have to provide it?

peterstarling (changes requested):

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -382,6 +383,18 @@ describe('ResearchOutputs controller', () => {
       expect(result.authors).toEqual(expectedAuthorsResponse);
     });
 
+    test('Should return a list of PM emails', async () => {
```
peterstarling, line 386: testing coverage may be ok but i think there is a few cases we might want to add, for instance:

1) A research output belongs to multiple teams, some of which have a PM
2) A research output belongs to a team with multiple PMs
3) A research output belongs to teams which have the same PMs - the PM email should be deduplicated

I don't know which of these your test fixture covers but personally i like the tests to be explicit, so i would still add all three with a specific test setup and then have the fixture and expectations updated only for some other generic tests

peterstarling (changes requested):

apps/asap-cli/src/import/protocols/insert.ts
harlan-kohll: Fixed with the new types

apps/asap-server/test/controllers/research-outputs.test.ts
harlan-kohll: 1. This is kind of difficult to test explicitly because it's implicit in the fixtures
2. Again, a bit tricky to write a test for this because we just get back an array of emails, difficult to break it down further. But I have renamed one of the emails `multiple-pms-on-same-team@example.com` to make it clearer
3. Have added a test for deduplication

peterstarling: yeah but you are not limited to the fixtures right, you can use the fixture but then update it specifically for the test case. Im sure its fairly easy to test, you just update the squidex response types (the ones in fixtures) so that they reflect the specific secnario no?

apps/asap-cli/src/import/protocols/insert.ts
```diff
@@ -83,7 +87,7 @@ const addProtocolToTeam = async (
 
 export default async (data: Protocol): Promise<void> => {
   const promises: Cache = {};
-  const protocol: RestResearchOutput['data'] = {
+  const protocol: Omit<RestResearchOutput['data'], 'pmsEmails'> = {
```
peterstarling, line 90: actually you could swap it for the new CreateResearchOutput type

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -382,6 +383,41 @@ describe('ResearchOutputs controller', () => {
...
+      const researchOutputResponse = getSquidexResearchOutputGraphqlResponse();
+      nock(config.baseUrl)
+        .post(`/api/content/${config.appName}/graphql`, {
+          query: buildGraphQLQueryResearchOutput(researchOutputId),
+        })
+        .reply(200, { data: researchOutputResponse });
+
+      const result = await researchOutputs.fetchById(researchOutputId);
+
+      // Both these PMs are duplicated in the fixture
```
peterstarling, line 412: instead of stating that in the comment i would leave the previous fixture as is and have a different setup specifically in this test

peterstarling (approved):
Approving albeit with comments, i dont want this PR to drag for too long but i still think you could update the tests (plus the type thing). Your call though

apps/asap-cli/src/import/protocols/insert.ts
harlan-kohll: there's no `type` field on `CreateResearchOutput`, it complains

PR #919, opened 2021-06-21
Feat:Multi Select - Skills modal

kmaid:
https://trello.com/c/oz83gWY6/1309-skills-and-expertise-modal-tags-group


apps/frontend/src/network/users/Research.tsx
```diff
@@ -12,6 +12,7 @@ import { network } from '@asap-hub/routing';
 import { usePatchUserById } from './state';
 import Frame from '../../structure/Frame';
 import GroupsCard from './groups/GroupsCard';
+import skillSuggestions from './skill-suggestions';
```
migcarva, line 15: since we are not using these suggestions for anything else, and they are pretty much an integral part of the skills modal, shouldn't we just move this import there? Or do you see this being fetched from the API someday?

kmaid: I gave some thought to this. React Components is primarily for presentation logic so it didn't really fit there. I know that the client wants to be able to add tags themselves which will require an API or this list could get too big so I put it here. Its quite big and i considered making another loadable just for the modal but it seemed overkill

migcarva (approved):
👍

PR #921, opened 2021-06-22
Fix: Contact info modal isn't showing data

kmaid:
https://trello.com/c/v0Ars5ZT/1403-user-cannot-edit-previously-saved-websites


apps/frontend/src/network/users/Editing.tsx
```diff
@@ -29,6 +29,7 @@ const Editing: React.FC<EditingProps> = ({ user, backHref }) => {
       <Route exact path={path + route.editContactInfo.template}>
         <Frame title="Edit Contact Information">
           <ContactInfoModal
+            {...user}
```
kmaid, line 32: 😔

migcarva (approved):
lgtm

PR #924, opened 2021-06-24
Feat: Add authors to list

kmaid:
https://trello.com/c/7S1d8oEy/1245-show-authors-on-outputs-list

### DIFF
--- a/packages/react-components/src/molecules/UsersList.tsx
+++ b/packages/react-components/src/molecules/UsersList.tsx
@@ -59,10 +59,14 @@ interface UsersListProps {
       >
     | ExternalAuthor
   >;
+  max?: number;
 }
-const UsersList: FC<UsersListProps> = ({ users }) => (
+const UsersList: FC<UsersListProps> = ({
+  users,
+  max = Number.POSITIVE_INFINITY,
+}) => (
   <ul css={listStyles}>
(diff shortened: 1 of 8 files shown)


packages/react-components/src/molecules/UsersList.tsx
```diff
@@ -81,6 +85,14 @@ const UsersList: FC<UsersListProps> = ({ users }) => (
         )}
       </li>
     ))}
+    {users.length - max > 0 && (
```
migcarva, line 88: users.length > max?

packages/react-components/src/molecules/__tests__/UsersList.test.tsx
```diff
@@ -42,3 +42,35 @@ it('does not link external users', () => {
...
+        users={Array.from({ length: 4 }).map(
+          () =>
+            ({
+              displayName: 'John Doe',
+            } as ExternalAuthor),
+        )}
+      />,
+    );
+    expect(getAllByRole('listitem').length).toEqual(4);
+    expect(getByText('+1')).toBeVisible();
```
migcarva, line 60: Should we first getByRole the 'img', and then check the text? I don't see it breaking that easily, but would be better.

migcarva (approved):
lgtm

PR #933, opened 2021-07-05
fix: Profile picture upload bug

peterstarling:
Extends a global API payload limit to 10MB as per API Gateway limitation: https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html

see: https://trello.com/c/U5vpkDJP/1264-i-am-not-able-to-upload-a-new-profile-picture

### DIFF
--- a/apps/asap-server/src/app.ts
+++ b/apps/asap-server/src/app.ts
@@ -111,7 +111,7 @@ export const appFactory = (libs: Libs = {}): Express => {
   app.use(httpLogger);
   app.use(tracingHandler);
   app.use(cors());
-  app.use(express.json());
+  app.use(express.json({ limit: '10MB' }));
 
   /**
    * Public routes --->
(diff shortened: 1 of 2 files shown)


kmaid (changes requested):
Shouldn't we write a test for this as it is a regression we missed?
Do the terraform scripts need to change?
Was this a consequence of moving to express?

peterstarling:
> Shouldn't we write a test for this as it is a regression we missed?
Yeah we could actually, but im just testing for now, sorry - I should marked it as WIP again.
> Do the terraform scripts need to change?
We only use terraform for the gitlab runners - our infrastructure is all CloudFormation
> Was this a consequence of moving to express?
Yes

kmaid:
I have never worked with cloud formation. Only terraform. My understanding is terraform generates cloud formation files. Does this mean we don't have our infrastructure as code?

kmaid (approved):

peterstarling:
> Does this mean we don't have our infrastructure as code?

We do, it's CloudFormation -> which is in yaml/json. We just dont user terraform

PR #936, opened 2021-07-08
feat: Create Algolia PR environments

peterstarling:
* Creates a CI job to store Algolia index dump from dev index
* Sets up a new Algolia index in Algolia dev account on every new PR raised, then tears it down
* Updates the env var setup for PR/Dev/Prod for Algolia App ID, sets up a new Algolia index var

see: https://trello.com/c/BvZW2XBK/1342-create-pr-environments


peterstarling:
@kmaid 
image
i think this setup works, at least for the PR env it does

apps/asap-server/scripts/restore-algolia-index.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const stream = fs
+    .createReadStream(process.env.ALGOLIA_BACKUP)
+    .pipe(StreamArray.withParser());
+  let chunks = [];
+
+  await stream
+    .on('data', ({ value }) => {
+      chunks.push(value);
+      if (chunks.length === 10000) {
```
harlan-kohll, line 19: Where does this number come from?

apps/asap-server/scripts/create-algolia-backup.js
```diff
@@ -0,0 +1,28 @@
+const algoliasearch = require('algoliasearch');
+const fs = require('fs/promises');
+
+const main = async () => {
+  console.log({
+    appId: process.env.ALGOLIA_APP_ID,
+    apiKey: process.env.ALGOLIA_API_KEY,
```
kmaid, line 7: Should we be revealing secrets in logs?

kmaid: probably should also include ALGOLIA_INDEX

apps/asap-server/scripts/create-algolia-backup.js
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  const index = algoliasearch(
+    process.env.ALGOLIA_APP_ID,
+    process.env.ALGOLIA_API_KEY,
+  ).initIndex(process.env.ALGOLIA_INDEX);
+
+  let hits = [];
+
+  await index.browseObjects({
```
kmaid, line 17: Won't this consume a lot of memory? What happened to the copy index function?

kmaid (changes requested):
Could these files also be typescript and run through TS-Node

apps/asap-server/scripts/restore-algolia-index.js
peterstarling: oh its an arbitrary number as per algolia guidebook: https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/how-to/sending-records-in-batches/#using-the-api

apps/asap-server/scripts/create-algolia-backup.js
peterstarling: we should not reveal them, no.

As for the TS - seems like an unnecessary effort. There is algolia CLI which technically enables us to do all that (as in exports and imports) but i kinda wanted to deliver a more custom solution, not sure if that was the right choice. Whats your opinion on that @harlan-kohll @kmaid

apps/asap-server/scripts/create-algolia-backup.js
peterstarling: well copying index is copying index and this script serves a different purpose which is to create a db dump into a file

peterstarling: but to answer your other question - it probably will, i could batch it and use streams but for that we'd need a writeable object/array type of stream and it just seemed a little unnecessary given that we dont really have that many research outputs. Do you think i should look into it anyway?

apps/asap-server/scripts/create-algolia-backup.js
kmaid: If it covers our usecase then yes I think we should. The less code we write the less we have to maintain

apps/asap-server/scripts/create-algolia-backup.js
kmaid: I think it would be better for it not to fail at somepoint in the future

kmaid (approved):

PR #938, opened 2021-07-09
feat: Add error logs to Squidex client

peterstarling:


peterstarling:
not needed now but referenced the issue in https://trello.com/c/DgUU3RPu/1436-replace-instrumentation-code-from-asap-server-with-a-logger

PR #939, opened 2021-07-09
FIx: Fix orcid

kmaid:


packages/model/src/user.ts
```diff
@@ -142,7 +142,7 @@ export interface UserPatchRequest {
   skillsDescription?: string;
   questions?: string[];
   teams?: Pick<UserTeam, 'id' | 'approach' | 'responsibilities'>[];
-  social?: UserSocialLinks;
+  social?: Omit<UserSocialLinks, 'orcid'>;
```
peterstarling, line 145: why dont you remove orcid from the `UserSocialLinks` instead? I don't think it's actually being used anywhere

peterstarling (changes requested):

packages/model/src/user.ts
kmaid: Because it is being used

peterstarling: where?

kmaid: Social Icons & Social modal & fixtures tests etc

peterstarling: yeah i see but how does that make any sense - logically - if orcid isn't actually available there? like.. unless this is meant to be a hotfix i would go through those fixtures and remove all the references

kmaid: It was supposed to move into socialLinks. It is available there to read but not to write it would seem

peterstarling (approved):

PR #940, opened 2021-07-09
Feat: Async Typeahead

kmaid:
https://trello.com/c/StkDlFFF/1351-change-input-field-for-institution

### DIFF
--- a/apps/frontend/src/network/users/__mocks__/api.ts
+++ b/apps/frontend/src/network/users/__mocks__/api.ts
@@ -6,6 +6,7 @@ import {
   ListUserResponse,
 } from '@asap-hub/model';
 import { GetListOptions } from '../../../api-util';
+import { InstitutionsResponse } from '../api';
 
 export const getUser = jest.fn(
   async (id: string): Promise<UserResponse> => ({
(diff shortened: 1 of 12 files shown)


peterstarling (changes requested):
have you considered removing autocomplete from the input field? otherwise it does this for me in chrome
image

as for the tests and implementation - there a few patters i probably wouldnt have gone for but i see this is how its done in this repo so it all looks legit to me

apps/frontend/src/network/users/__mocks__/api.ts
```diff
@@ -38,3 +39,22 @@ export const getUsers = jest.fn(
   async ({ pageSize }: GetListOptions): Promise<ListUserResponse> =>
     createListUserResponse(pageSize ?? 10),
 );
+
+export const getInstitutions = jest.fn(
```
peterstarling, line 43: why bother define the default implementation if i may ask? I see you apply a custom resolve response anyway https://github.com/yldio/asap-hub/pull/940/files#diff-3108a0410b24cdad1ffe15c3d2688b42bfbfc211b3a79eca00b5f4e9f67d1078R77

peterstarling (changes requested):

apps/frontend/src/network/users/__mocks__/api.ts
kmaid: for the other tests or using the api somewhere else

PR #944, opened 2021-07-15
Add lab catalog number to ROMS outputs

harlan-kohll:
Add field labCatalogNumber to Research Outputs.


apps/asap-server/src/entities/research-output.ts
```diff
@@ -77,6 +77,7 @@ export const parseGraphQLResearchOutput = (
     description: output.flatData?.description || '',
     tags: output.flatData?.tags || [],
     publishDate: output.flatData?.publishDate || undefined,
+    labCatalogNumber: output.flatData?.labCatalogNumber || '',
```
peterstarling, line 80: why do you default to an empty string? i see the field isnt required either in the schema or in the response so i dont think this is valid. It should just be returned as undefined if missing

peterstarling (changes requested):

apps/asap-server/test/fixtures/teams.fixtures.ts
```diff
@@ -418,6 +422,7 @@ export const graphQlTeamResponse: { data: ResponseFetchTeam } = {
             flatData: {
               link: null,
               publishDate: null,
+              labCatalogNumber: 'http://example.com',
```
peterstarling, line 425: i see you've changed the defaulting to undefined but you didn't change the tests - this is how i realised you never really tested the scenario when the `labCatalogNumber` is missing. I'd remove it from one of the ROs here and then update the expectation too just to cover that scenario

peterstarling (changes requested):

peterstarling (approved):

PR #948, opened 2021-07-19
feat: Remove location field

peterstarling:
Removes the `location` from:
- BE fixtures and types
- FE types
- user import and invite script 
see: https://trello.com/c/jquoG31D/1373-loc4-remove-location


apps/asap-cli/src/import/users/parse.ts
```diff
@@ -59,7 +58,6 @@ export default (data: string[]): Data => {
     responsibilities,
     skills,
     asapRole,
-    location,
     ,
```
kmaid, line 61: ?

packages/squidex/schema/schemas/users.json
```diff
@@ -280,26 +280,6 @@
           "isHalfWidth": true
         }
       },
-      {
```
kmaid, line 283: I assume this means we lose the data in production. Have we confirmed this with bea?

kmaid (approved):

packages/squidex/schema/schemas/users.json
peterstarling: image
I think so - i mean the ticket clearly says to verify that the data was migrated to city/country and its crossed out now, but ill speak with B just in case

peterstarling: although - we do weekly backups so it wont be gone gone

PR #953, opened 2021-07-21
feat: Add accession field to research-outputs

peterstarling:
https://trello.com/c/GvN2W66A/1398-store-accession-for-each-roms-output

### DIFF
--- a/packages/squidex/schema/schemas/research-outputs.json
+++ b/packages/squidex/schema/schemas/research-outputs.json
@@ -431,6 +431,27 @@
         }
       },
       {
+        "name": "accession",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
+        "partitioning": "invariant",
+        "properties": {
+          "fieldType": "String",
+          "pattern": "^(\\w+\\d+(\\.\\d+)?)|(NP_\\d+)$",
+          "isUnique": false,
+          "inlineEditable": false,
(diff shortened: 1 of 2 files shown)


packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -430,6 +430,27 @@
...
       },
+      {
+        "name": "accession",
+        "isHidden": false,
+        "isLocked": false,
+        "isDisabled": false,
+        "partitioning": "invariant",
+        "properties": {
+          "fieldType": "String",
+          "pattern": "^(\\w+\\d+(\\.\\d+)?)|(NP_\\d+)$",
```
harlan-kohll, line 441: In the ticket it says that should match one of two regexes (think this is just the first one), could combine them into one big one with an |

harlan-kohll (approved):

packages/squidex/schema/schemas/research-outputs.json
peterstarling: i am aware, but like i mentioned in the ticket - it looks like every provided example passes the first one. It appears that the second one is a subset of the first pattern and so I think that unless we can come up with an example of an ID that only passes the second regex we shouldnt include it

harlan-kohll: OK cool

PR #954, opened 2021-07-21
feat: Add DOI identifier to research-outputs

peterstarling:
https://trello.com/c/Ct1o4197/1395-2-store-doi-for-each-roms-output


apps/asap-server/test/integration/fetch-users.integration-test.ts
```diff
@@ -1,35 +1,44 @@
...
-      email: 'john.doe@asap.science',
+      email: randomEmail,
       role: 'Grantee' as Role,
       degree: 'MPH',
     };
 
     await createUser(user);
 
-    const result = await users.fetch({});
+    const result = await users.fetch({
```
peterstarling, line 30: or we could just fetch the user by ID? although then we would have to specify the ID on creation so maybe some other time

peterstarling (commented):
i cant really approve my own PR :D but basically i think i like your approach to the integration tests. I have created a ticket to set up basic integration tests for each of our entities but this doesn't mean that yours shouldnt already land on master.

apps/asap-server/test/integration/dummy.integration-test.ts
```diff
@@ -0,0 +1,7 @@
+export {};
```
peterstarling, line 1: you probably want to remove this file?

peterstarling:
@harlan-kohll i think you need to review this :D

harlan-kohll (approved):

PR #965, opened 2021-07-26
Surface identifiers on FE 2 (to get pipeline moving)

harlan-kohll:
Add accession and RRID to research outputs endpoint


peterstarling (approved):
looks about right, have you tested the response in the PR env?

PR #969, opened 2021-07-27
1394 expand additional information section

jhonny-m:
https://trello.com/c/FUqUFy3Q/1394-expand-additional-information-section

### DIFF
--- a/packages/react-components/src/templates/SharedResearchOutput.tsx
+++ b/packages/react-components/src/templates/SharedResearchOutput.tsx
@@ -2,12 +2,14 @@ import { ComponentProps } from 'react';
 import { css } from '@emotion/react';
 import { ResearchOutputResponse } from '@asap-hub/model';
 
-import { Card, Headline2, Divider } from '../atoms';
-import { perRem } from '../pixels';
+import { Card, Headline2, Divider, Anchor } from '../atoms';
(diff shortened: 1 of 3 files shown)


packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -2,12 +2,14 @@ import { ComponentProps } from 'react';
 import { css } from '@emotion/react';
 import { ResearchOutputResponse } from '@asap-hub/model';
 
-import { Card, Headline2, Divider } from '../atoms';
-import { perRem } from '../pixels';
+import { Card, Headline2, Divider, Anchor } from '../atoms';
```
ana-yldio, line 5: Why are you using Anchor instead of ExternalLink?

jhonny-m: The expected component and the `ExternalLink` component look quite different, I wasn't sure if changing the `ExternalLink` component to support the styles needed for this one was the best approach

ana-yldio: fair enough, have a look at https://github.com/yldio/asap-hub/pull/970
On Storybook, the component External Link is different from the specs, it matches the External Link on Card spec

ana-yldio (approved):

PR #970, opened 2021-07-28
fix: update link styles

ana-yldio:
https://trello.com/c/gWaDbP7V/880-as-a-user-i-want-to-better-identify-links-states-hover-visited-clicked

This only updates the link's colours and text-decoration. There's an inconsistency regarding the icon placement between the components External Links and External Link on Card (see below) that will be tackled on a different ticket. 

image (1)
image

### DIFF
--- a/packages/react-components/src/atoms/Link.tsx
+++ b/packages/react-components/src/atoms/Link.tsx
@@ -8,9 +8,15 @@ import { Anchor } from '.';
 
 const styles = css({
   textDecoration: 'underline',
+  ':hover': {
+    textDecoration: 'none',
+  },
 });
 export const themeStyles: Record<ThemeVariant, SerializedStyles> = {
-  light: css({ color: fern.rgb, ':active': { color: pine.rgb } }),
+  light: css({
+    color: fern.rgb,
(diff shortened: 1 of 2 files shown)


kmaid (approved):

packages/react-components/src/atoms/Link.tsx
```diff
@@ -8,9 +8,15 @@ import { Anchor } from '.';
 
 const styles = css({
   textDecoration: 'underline',
+  ':hover': {
+    textDecoration: 'none',
```
jhonny-m, line 12: shouldn't the hover keep the underline?

packages/react-components/src/molecules/ExternalLink.tsx
```diff
@@ -61,7 +66,7 @@ const ExternalLink: React.FC<ExternalLinkProps> = ({
     <Anchor href={href}>
       <div css={styles}>
         {icon}
-        <div css={textStyles}>{label}</div>
+        <div css={[textStyles, linkStyles]}>{label}</div>
```
jhonny-m, line 69: I think we need to select one of the themes of `linkStyles` to apply an style

PR #972, opened 2021-07-28
feat: Add authentication for Algolia

peterstarling:
https://trello.com/c/rd7Z4bX0/1331-enable-authentication-for-the-search


peterstarling:
image
test evidence

apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
```diff
@@ -0,0 +1,51 @@
+import { SearchClient } from 'algoliasearch';
+import Joi from '@hapi/joi';
+import { framework as lambda } from '@asap-hub/services-common';
+import { UserMetadataResponse } from '@asap-hub/model';
+import { http } from '../../../utils/instrumented-framework';
```
kmaid, line 5: Are we moving away from the instrumented framework?

apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const user = await userController.fetchByCode(code);
+    let apiKey: string | null;
+
+    try {
+      apiKey = algoliaClient.generateSecuredApiKey(algoliaSearchApiKey, {
+        validUntil: Date.now() + 36001, // which is one second over the TTL of the ID token
+      });
+    } catch (e) {
+      apiKey = null;
```
kmaid, line 41: I thought we agreed we would fail the auth request if getting the token fails. Having a user unable to browse listing pages is pretty critical. I was in favour of handling this on another endpoint in the frontend over it being possibly null

kmaid (changes requested):

kmaid:
The new env var needs to be added to the .env.example

apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
peterstarling: we did and like i said on slack - because i cant test auth0 flow without it being deployed to dev (and so - prod) i want to use this first, merge, test and then remove the null - i specifically used null not undefined for this

apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
```diff
@@ -0,0 +1,170 @@
...
+          pathParameters: {
+            code: 'welcomeCode',
+          },
+        }),
+      )) as APIGatewayProxyResult;
+
+      expect(result.statusCode).toStrictEqual(401);
+    });
+
+    test('returns 401 when request is not using Basic Auth', async () => {
```
harlan-kohll, line 48: If possible can we use digest or bearer auth here instead?

peterstarling: we could, but probably not in this PR, as this PR does not address auth0 auth method. Feel free to raise as a Tech debt though

harlan-kohll (approved):

PR #973, opened 2021-07-28
1394 expand additional information section 2

jhonny-m:
https://trello.com/c/FUqUFy3Q/1394-expand-additional-information-section
[previous pr](https://github.com/yldio/asap-hub/pull/969)

### DIFF
--- a/packages/react-components/src/atoms/Link.tsx
+++ b/packages/react-components/src/atoms/Link.tsx
@@ -22,6 +22,28 @@ export const themeStyles: Record<ThemeVariant, SerializedStyles> = {
   dark: css({ color: paper.rgb, ':active': { color: paper.rgb } }),
 };
 
+const iconThemeStyle: Record<ThemeVariant, SerializedStyles> = {
+  light: css({
+    svg: {
+      stroke: fern.rgb,
+    },
+    ':hover': {
+      svg: {
+        stroke: pine.rgb,
+      },
+    },
+    ':active': { svg: { stroke: fern.rgb } },
+  }),
+  grey: css({
+    svg: { stroke: fern.rgb },
+    ':active': { svg: { stroke: pine.rgb } },
+  }),
+  dark: css({
+    svg: { stroke: paper.rgb },
+    ':active': { svg: { stroke: paper.rgb } },
+  }),
+};
+
 interface NormalLinkProps {
   readonly theme?: ThemeVariant;
 
@@ -44,6 +66,7 @@ type LinkProps = {
   readonly children: ReactNode;
   readonly href: string | undefined;
   readonly label?: string;
+  readonly applyIconTheme?: boolean;
 } & (NormalLinkProps | ButtonStyleLinkProps);
 const Link: React.FC<LinkProps> = ({
   children,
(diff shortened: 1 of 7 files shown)


packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -52,72 +80,153 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
...
+                </li>
+              </>
+            )}
+            {doi === undefined || (
+              <>
+                <Divider />
+                <li css={additionalInformationEntryStyles}>
+                  <strong>Identifier (DOI)</strong>
+                  <span css={additionalInformationValueStyles}>
+                    <Anchor href={`https://doi.org/${doi}`}>
```
kmaid, line 159: Could you please construct this url using new URL().toString(). We do this everywhere to ensure a link is always valid

packages/react-components/src/templates/__tests__/SharedResearchOutput.test.tsx
```diff
@@ -129,4 +129,55 @@ describe('additional information', () => {
...
+  it('contains labCatalogNumber and builds the correct href', () => {
+    const labCatalogNumberLink = 'https://example.com';
+    const labCatalogNumberText = '0000-0004-9946-3696';
+    const { getByTestId, rerender } = render(
+      <SharedResearchOutput
+        {...props}
+        labCatalogNumber={labCatalogNumberLink}
+      />,
+    );
+    expect(getByTestId('labCatalogNumber')?.firstElementChild?.tagName).toMatch(
```
kmaid, line 166: We don't use test ids anywhere. We do this to improve our accessibility. I would take the test id out and use something like getByText('https://example.com').closetest('a').toHaveAttribtue('href', 'example.com')

kmaid (changes requested):
See comments

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -52,72 +80,153 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
...
-}) => (
-  <div css={containerStyles}>
-    <BackLink href={backHref} />
-    <div css={cardsStyles}>
-      <SharedResearchOutputHeaderCard {...props} />
-      {(description || !!tags.length) && (
-        <Card>
-          {description && (
+}) => {
+  const isLabCatalogNumberALink =
```
kmaid, line 89: I don't like this. I would do something like 
```suggestion
  const isLabCatalogNumberALink = ()=> {
  try {
    url = new URL(string);
  } catch (_) {
    return false;  
  }
  return true; 
}
```

and stick it in packages/react-components/src/utils.ts

kmaid (approved):

packages/react-components/src/atoms/Link.tsx
```diff
@@ -56,10 +79,11 @@ const Link: React.FC<LinkProps> = ({
   primary = false,
   small = false,
   enabled = true,
+  applyIconTheme = false,
```
kmaid, line 82: Could this be automated using isExternalLink?

kmaid (approved):

PR #974, opened 2021-07-28
Add admin notes section to user schema

harlan-kohll:
Add admin notes section to user schema

### DIFF
--- a/packages/squidex/schema/schemas/users.json
+++ b/packages/squidex/schema/schemas/users.json
@@ -679,6 +679,43 @@
         }
       },
       {
+        "name": "separatorAdminNotes",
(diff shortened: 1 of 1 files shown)


harlan-kohll:
Note: this field should not be returned from the BE

packages/squidex/schema/schemas/users.json
```diff
@@ -678,6 +678,43 @@
           "isHalfWidth": false
         }
       },
+      {
+        "name": "separatorAdminNotes",
```
peterstarling, line 682: what is this field?

peterstarling (approved):

PR #980, opened 2021-07-29
feat: Algolia authentication - remove fail-safe feature

peterstarling:
Follow up on https://trello.com/c/rd7Z4bX0/1331-enable-authentication-for-the-search

### DIFF
--- a/apps/asap-server/src/config.ts
+++ b/apps/asap-server/src/config.ts
@@ -44,3 +44,4 @@ export const algoliaAppId = ALGOLIA_APP_ID || 'LVYWOPQ0A9';
 export const algoliaSearchApiKey = ALGOLIA_SEARCH_API_KEY || '';
 export const algoliaResearchOutputIndex =
   ALGOLIA_RESEARCH_OUTPUT_INDEX || 'asap-hub_research_outputs_dev';
+export const algoliaApiKeyTtl = 36060;
--- a/apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
+++ b/apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
@@ -7,7 +7,6 @@ import { UserController } from '../../../controllers/users';
 import validateRequest from '../../../utils/validate-auth0-request';
 import { Handler } from '../../../utils/types';
 import { algoliaSearchApiKey } from '../../../config';
-import logger from '../../../utils/logger';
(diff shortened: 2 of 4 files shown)


apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts
```diff
@@ -31,16 +30,9 @@ export const fetchUserByCodeHandlerFactory = (
...
-    try {
-      apiKey = algoliaClient.generateSecuredApiKey(algoliaSearchApiKey, {
-        validUntil: Date.now() + 36060, // which is one minute over the TTL of the ID token
-      });
-    } catch (e) {
-      apiKey = null;
-      logger.error(e, 'Error during generating the key for Algolia search');
-    }
+    const apiKey = algoliaClient.generateSecuredApiKey(algoliaSearchApiKey, {
+      validUntil: Date.now() + 36060, // which is one minute over the TTL of the ID token
```
harlan-kohll, line 34: Think there was another comment about this - can we put the `36060` into a `const`?

peterstarling: we could but erm.. what difference does it make :D? i guess i can put it in the config although to be fair its not a config attribute you would ever need to change and if you do its still global. But yeah if thats what you guys want

apps/asap-server/src/config.ts
```diff
@@ -44,3 +44,4 @@ export const algoliaAppId = ALGOLIA_APP_ID || 'LVYWOPQ0A9';
 export const algoliaSearchApiKey = ALGOLIA_SEARCH_API_KEY || '';
 export const algoliaResearchOutputIndex =
   ALGOLIA_RESEARCH_OUTPUT_INDEX || 'asap-hub_research_outputs_dev';
+export const algoliaApiKeyTtl = 36060;
```
kmaid, line 47: casing 😅

peterstarling: whats wrong with it in this var name?

kmaid: ALGOLIA_API_KEY_TTL

kmaid: meh I don't know what the convention is in the backend. Thats how I would normally expect to find a constant

kmaid (approved):

PR #982, opened 2021-07-29
fix: validates team tool url protocol

ana-yldio:
https://trello.com/c/Gxvd1zId/1038-keep-users-from-inputting-incomplete-urls-on-team-tools

### DIFF
--- a/packages/react-components/src/templates/__tests__/ToolModal.test.tsx
+++ b/packages/react-components/src/templates/__tests__/ToolModal.test.tsx
@@ -39,6 +39,33 @@ it('renders default values into inputs', () => {
   `);
 });
 
+it('allows url with http protocol', () => {
+  const { getAllByRole } = render(
+    <ToolModal {...props} url="http://example.com/tool" />,
+    { wrapper: StaticRouter },
+  );
+
+  expect(getAllByRole('textbox')[0]).toBeValid();
+});
+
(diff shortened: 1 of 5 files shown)


packages/react-components/src/templates/__tests__/ToolModal.test.tsx
```diff
@@ -39,6 +39,50 @@ it('renders default values into inputs', () => {
   `);
 });
 
+it('allows url with https protocol', () => {
+  const { getAllByRole, queryByText } = render(<ToolModal {...props} />, {
+    wrapper: StaticRouter,
+  });
+  const inputUrl = getAllByRole('textbox')[0];
```
kmaid, line 46: Can we get by label text? [0] is quite brittle if we have to add another textbox

packages/validation/test/teams.test.ts
```diff
@@ -0,0 +1,12 @@
+import { TEAM_TOOL_URL } from '../src/teams';
+
+describe('ss', () => {
```
kmaid, line 3: ss?

kmaid (changes requested):
looks good super minor things

packages/validation/test/teams.test.ts
ana-yldio: sorry!

kmaid (approved):

PR #986, opened 2021-08-03
feat: add sentry to the frontend

migcarva:


.gitlab-ci.yml
```diff
@@ -62,12 +62,15 @@ build:ts:
     - export REACT_APP_API_BASE_URL=$ASAP_API_URL
     - export REACT_APP_ALGOLIA_INDEX=$ALGOLIA_INDEX
     - export REACT_APP_ALGOLIA_APP_ID=$(aws ssm get-parameter --name "algolia-app-id-dev" --query Parameter.Value --output text)
+    - yarn install --immutable --immutable-cache
```
peterstarling, line 65: just like we said i think its best to move all this into a different job that uses this build artifact, in there you can run yarn install to get the binary you need but i dont think yarn install should be run in the CI before bundling

peterstarling (changes requested):

.gitlab-ci.yml
```diff
@@ -63,8 +63,9 @@ build:ts:
     - export REACT_APP_API_BASE_URL=$ASAP_API_URL
     - export REACT_APP_ALGOLIA_INDEX=$ALGOLIA_INDEX
     - export REACT_APP_ALGOLIA_APP_ID=$(aws ssm get-parameter --name "algolia-app-id-dev" --query Parameter.Value --output text)
+    - export REACT_APP_RELEASE=$CI_PIPELINE_ID
```
kmaid, line 66: These are baked into the artifact and have to be done at build

.gitlab-ci.yml
```diff
@@ -198,6 +199,18 @@ verify:branch:
   script:
     - ASAP_API_URL=$API_URL yarn test:e2e
 
+verify:report:frontend:release:
+  <<: *tmpl_master
```
kmaid, line 203: I think this means this will only happen on master

.gitlab-ci.yml
```diff
@@ -198,6 +199,18 @@ verify:branch:
   script:
     - ASAP_API_URL=$API_URL yarn test:e2e
 
+verify:report:frontend:release:
```
kmaid, line 202: I have done this after a successful dev development at which point its already a release.

.gitlab-ci.yml
```diff
@@ -198,6 +199,18 @@ verify:branch:
...
+  <<: *tmpl_master
+  stage: verify-dev
+  needs:
+    - build:ts
```
peterstarling, line 206: it doesn't really affect the pipeline but you only really need `deploy:sls:development`

apps/frontend/scripts/sentry-release.js
```diff
@@ -0,0 +1,31 @@
...
+      urlPrefix: '~/static/js',
+      rewrite: false,
+    });
+    console.log('Finalizing release');
+    await cli.releases.finalize(release);
+  } catch (e) {
+    console.error('Source maps uploading failed:', e);
+  }
+}
+createReleaseAndUpload();
```
peterstarling, line 31: i see you are catching the error in the function so i guess there is almost no way for an unhandled promise to occur (unless it fails before line 15). Is that something we really want? Like.. maybe we want to log the error and rethrow, then make sure to exit with non-0 when this happens?

peterstarling (approved):
two minor comments - feel free to address them or not, but what I would say is that I think you should test it and you can do it by simply running the release jump after the sls branch deployment.

PR #987, opened 2021-08-04
fix: update email template

ghost:
https://trello.com/c/I8H0mjmj/1315-new-template-for-invite-email

### DIFF
--- a/apps/storybook/src/WelcomeMessage.stories.tsx
+++ b/apps/storybook/src/WelcomeMessage.stories.tsx
@@ -9,17 +9,17 @@ export default {
   decorators: [MessageLayoutDecorator],
 };
 
-export const Variant1 = () => (
+export const InviteScript = () => (
   <WelcomeMessage
     firstName={text('First Name', 'John')}
     link={text('Link', 'https://example.com/register')}
   />
 );
 
-export const Variant2 = () => (
+export const Invite = () => (
   <WelcomeMessage
     firstName={text('First Name', 'John')}
     link={text('Link', 'https://example.com/register')}
-    variant={'Variant2'}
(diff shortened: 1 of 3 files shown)


apps/storybook/src/WelcomeMessage.stories.tsx
```diff
@@ -9,11 +9,17 @@ export default {
   decorators: [MessageLayoutDecorator],
 };
 
-export const Normal = () => (
+export const Variant1 = () => (
```
kmaid, line 12: I don't like this naming. Can we be more specific? Something that says InviteScriptWelcomeTemplate or something?

kmaid (changes requested):
Did you figure out how these templates are pushed into AWS? Do we need to do anything special to do this?

ana-yldio:
> Did you figure out how these templates are pushed into AWS? Do we need to do anything special to do this?

The deploy to prod runs the SES job, and then we need to check on AWS the template is there

packages/react-components/src/messages/__tests__/Welcome.test.tsx
```diff
@@ -29,4 +29,30 @@ describe('welcome email template', () => {
...
+    rerender(
+      <Welcome
+        firstName="John Doe"
+        link="https://example.com"
+        variant={'InviteWelcomeTemplate'}
+      />,
+    );
+
+    cta1 = queryByRole('link', { name: /activate account/i });
+    cta2 = queryByRole('link', { name: /create account/i });
```
kmaid, line 53:
```suggestion
```

I dont think you need these. i think the first two can be const. Not 100% though

kmaid (approved):

PR #988, opened 2021-08-04
fix: removes the edit button from recent works

ana-yldio:
https://trello.com/c/rIK8W6v6/1498-remove-the-orcid-recent-works-pencil

### DIFF
--- a/apps/frontend/src/network/users/__tests__/About.test.tsx
+++ b/apps/frontend/src/network/users/__tests__/About.test.tsx
@@ -125,7 +125,4 @@ describe('when editing the biography', () => {
       expect.any(String),
     );
   });
-  it('does not allow to edit the works section', () => {
-    expect(result.queryByLabelText(/edit.+recent.+visib/i)).toBeNull();
-  });
 });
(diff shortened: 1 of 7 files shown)


apps/frontend/src/network/users/__tests__/About.test.tsx
```diff
@@ -125,4 +125,7 @@ describe('when editing the biography', () => {
       expect.any(String),
     );
   });
+  it('does not allow to edit the works section', () => {
```
kmaid, line 128: I don't think we should test for things that aren't there

packages/react-components/src/templates/__tests__/UserProfileAbout.test.tsx
```diff
@@ -41,19 +41,25 @@ it('renders an edit button for the biography', () => {
...
   );
 });
-it('renders an edit button for the recent works visibility', () => {
-  const { getByLabelText } = render(
-    <UserProfileAbout orcidWorks={[]} editOrcidWorksHref="/edit-works" />,
-  );
-  expect(getByLabelText(/edit.+recent.+visib/i)).toHaveAttribute(
-    'href',
-    '/edit-works',
+it('does not render an edit for the recent works', () => {
```
kmaid, line 44: Same as above happy to remove this test

kmaid (changes requested):

kmaid (approved):

PR #991, opened 2021-08-05
Add import-squidex-data.sh script

harlan-kohll:
Put a script wrapper around the squidex import so that it fails the pipeline if `sq sync` fails


ci/import-squidex-data.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+
+# create-app.py configures sq to new app
+sq sync in backup 2>&1 | tee sq-sync-log.txt
```
peterstarling, line 4: do you need to store it in a file? cant you simply store in some env var? surely you could still grep it?

peterstarling (changes requested):

ci/import-squidex-data.sh
harlan-kohll: i started off with `if sq sync in backup | grep` but that hid all of the sync output from the pipeline, this seems to be the easiest way to get gitlab to keep piping the results to the pipeline (in realtime). a bit clunky but the logs are deleted at the end

peterstarling: you could store it in an env variable 
```bash
export output="$(sq sync in backup 2>&1)"
```
also if sq sync in does not actually error during failures do you need to redirect stderr to stdout?

anyway once you have it in the env file i dont think we need to see the output unless it was a failure, in which case you can just echo it (as in - only when it fails).

I'd also add some other echo like 'Importing data...' before the command and 'Import successful' after and i think thats really all that matters. Alternatively on success you could have an env variable which is a flag that you add it it will also display the output on success (for debugging). Otherwise i think this output is just noise, i really think this whole process should be a separate job.

harlan-kohll: have implemented all the above, with the exception of the separate job, was taking me a while to figure out the gitlab yaml stuff. have joined that ticket https://trello.com/c/8trOLvTW/1522-separate-squidex-import-from-the-sls-deploy-to-speed-up-the-pipeline

peterstarling: oh yeah i didnt want you to split it now, of course

ci/import-squidex-data.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env sh
+
+# Set to true if you want to print the sq sync logs on success for debugging
+PRINT_SYNC_LOGS_ON_SUCCESS=false
```
peterstarling, line 4: actually - this should be false by default but we should be able to override it with simple config change (in gitlab) by defining the env variable without having to alter the script

peterstarling (changes requested):

ci/import-squidex-data.sh
harlan-kohll: Added test for the envvar

ci/import-squidex-data.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env sh
+
+# Set this envvar to true in GitLab if you want to print the sq sync logs on success for debugging
+if [ -z "$PRINT_SYNC_LOGS_ON_SUCCESS" ]; then
```
peterstarling, line 4:
```bash
PRINT_SYNC_LOGS_ON_SUCCESS="${PRINT_SYNC_LOGS_ON_SUCCESS:-false}"
```

peterstarling (approved):
one suggestion about the default value - you don't really need to update it though, but I thought I'd show you how to use defaul vals when assigning env vars in bash.

Plus - I don't think you need to unset all these env vars before exiting? i think its just noise, but again, you dont need to modify that if you dont feel like it

PR #993, opened 2021-08-05
feat: Automate user invitation 

peterstarling:


peterstarling:
the plugin doesnt work and creates a whole new REST API so its best to tear the whole stack down, closing

PR #998, opened 2021-08-06

ana-yldio:


apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
```diff
@@ -93,6 +93,7 @@ const apiUser: UserMetadataResponse = {
   questions: [],
   role: 'Grantee',
   algoliaApiKey: 'test-api-key',
+  labs: [],
```
peterstarling, line 96: i dont think we want to add labs to auth0 token - i already think we have too much information on it. You can easily omit this field in `UserMetadataResponse` type

packages/model/src/user.ts
```diff
@@ -81,6 +81,11 @@ export interface OrcidWork {
   lastModifiedDate: string;
 }
 
+export interface UserLab {
```
peterstarling, line 84: to be fair i dont think this needs to be a user-specific type.

Labs unlike other entities are not being returned on their own, and so there is no direct Response type definition inside /entities - I would still probably create a file and have a basic Lab definition there that we might use for teams etc.

apps/asap-server/test/fixtures/users.fixtures.ts
```diff
@@ -361,6 +364,7 @@ export const userResponse: UserResponse = {
   questions: ['Question 1', 'Question 2'],
   avatarUrl: `${config.baseUrl}/api/assets/${config.appName}/squidex-asset-id`,
   role: 'Grantee',
+  labs: [],
```
peterstarling, line 367: thats not really a great fixture if you ask me - you dont need to implement anything for this test to pass, we could just return an empty array. It's ok to test that case too - as in - the empty result, but I would first make sure that we get actual labs here, that goes for both route testing and controller testing (both of which use this fixture)

peterstarling (changes requested):

PR #1000, opened 2021-08-09
chore: Improve gitlab pipeline

peterstarling:
https://trello.com/c/ETFGxtoD/1516-fix-docker-rate-limits-breaking-pipeline
* replace base image with one from gitlab

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,16 +25,17 @@ include:
     - build:ts
     - build:native
 
-image: node:14.17-alpine
+variables:
+  ASAP_HOSTNAME: hub.asap.science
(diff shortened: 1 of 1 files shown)


peterstarling:
I specifically didn't add the image pushing bit to the .pre stage - not only would that increase the amount of calls to dockerhub (just like with other custom images - this one would need to be pulled from dockerhub then tagged and pushed to gitlab registry) but would also feel a little unnecessary. Unlike other custom images this one does not have a whole Docker file with a more complex build - no changes to the repo will alter this image and i think think we need to look for latest node version either.

harlan-kohll (approved):

jhonny-m (approved):

PR #1003, opened 2021-08-10

ana-yldio:
https://github.com/yldio/asap-hub/pull/998


packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -1,14 +1,16 @@
...
 import { UserProfileContext } from '@asap-hub/react-context';
 
 import { Card, Headline2, Headline3, Link } from '../atoms';
 import { mobileScreen, perRem } from '../pixels';
 import UserProfilePlaceholderCard from './UserProfilePlaceholderCard';
+import { getListStrWithSuffix, capitalizeText } from '../utils';
 
-type UserProfileBackgroundProps = UserTeam & Pick<UserResponse, 'firstName'>;
+type UserProfileBackgroundProps = UserTeam &
+  Pick<UserResponse, 'firstName'> & { labs?: Lab[] };
```
jhonny-m, line 13: we can pick the `labs` from the `UserResponse` type

packages/model/src/lab.ts
```diff
@@ -0,0 +1,4 @@
+export interface Lab {
+  id: string;
+  name?: string | null;
```
jhonny-m, line 3: It shoulnd't need the `null`as a type since it is optional

packages/react-components/src/__tests__/utils.test.tsx
```diff
@@ -93,3 +95,30 @@ describe('isLink', () => {
     expect(isLink('https://example.com')).toBeTruthy();
   });
 });
+
+describe('getListStrWithSuffix', () => {
```
jhonny-m, line 99: it would be cool to test the suffix as well

packages/model/src/lab.ts
ana-yldio: I think we need it because of this
https://github.com/yldio/asap-hub/blob/master/packages/squidex/src/entities/common.ts#L19

packages/react-components/src/organisms/UserProfileBackground.tsx
ana-yldio: My thought was to set `labs` as optional because there are components that use `UserProfileBackground` but don't render labs, like `UserProfileStaffBackground`

packages/model/src/lab.ts
kmaid: it shouldn't be null. The backend omits in that case otherwise everything would be either

ana-yldio: right, we coerce `null` to `undefined` when processing the API response 
https://github.com/yldio/asap-hub/blob/e7202ba33eede1e7b391c6c43ce02b298909de27/apps/asap-server/src/entities/user.ts#L97

PR #1006, opened 2021-08-11

ana-yldio:
https://trello.com/c/sW4Z1Iez/1494-show-labs-on-profile-page

https://github.com/yldio/asap-hub/pull/1003
https://github.com/yldio/asap-hub/pull/998


apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
```diff
@@ -77,7 +77,7 @@ const context: RuleContext = {
   authorization: { roles: [] },
 };
 
-const apiUser: UserMetadataResponse = {
+const apiUser: Omit<UserMetadataResponse, 'labs'> = {
```
peterstarling, line 80: i think `UserMetadataResponse` is a type designed specifically for this use case (ie auth0) and its best to remove the `labs` from that type rather than create something here on the fly

peterstarling (changes requested):
one change request - plus i see you modified user fixtures to prove that the user endpoint returns labs - but if you would like the labs to be returned anywhere else (like teams and group endpoints where the user info is returned) i'd probably modify those fixtures too so that the `labs` property returns data. 

I am generally of an opinion that a basic fixture should test that the data is present, and there can be another test to assert all the empty values (like nulls, optional properties and empty arrays)

peterstarling (approved):
check if it actually works in any PR env but if it does i think this is legit

packages/react-components/src/utils.ts
```diff
@@ -63,3 +63,32 @@ export const getIconFromUrl = (url: string): JSX.Element | undefined => {
   });
   return icon?.[1];
 };
+
+export const getListStrWithSuffix = (
```
kmaid, line 67: This should be called getUniqueOxfordCommaString.

packages/react-components/src/utils.ts
```diff
@@ -63,3 +63,32 @@ export const getIconFromUrl = (url: string): JSX.Element | undefined => {
...
+  data: (string | null | undefined)[] = [],
+  suffix: string = '',
+  format?: (str: string) => string,
```
kmaid, line 70: We can do the format on the whole string afterwards.

packages/react-components/src/utils.ts
```diff
@@ -63,3 +63,32 @@ export const getIconFromUrl = (url: string): JSX.Element | undefined => {
...
+    .filter((value): value is string => !!value?.trim())
+    .map((item, idx, arr) => {
+      const itemFormatted = format ? format(item) : item;
+      const itemValue = suffix ? `${itemFormatted} ${suffix}` : itemFormatted;
+      const lastIdx = arr.length - 1;
+      return arr.length === 1 || idx !== lastIdx
+        ? `${itemValue}`
+        : `and ${itemValue}`;
+    })
+    .join(', ');
```
kmaid, line 82: I think this should be written something like 

```
const format = (array, suffix) => {
  const trimmedUnique = Array.from(new Set(array)).map((lab) => `${lab.trim()}${suffix ? ` ${suffix}`: ''}`);
  return trimmedUnique.slice(0, -2).join(', ') +
    (trimmedUnique.slice(0, -2).length ? ', ' : '') +
    trimmedUnique.slice(-2).join(`${trimmedUnique.length > 2 ? ',': ''} and `);
};
// test cases
console.log(format(['a']));
console.log(format(['a', 'b'], 'Lab'));
console.log(format(['a', 'b', 'c', 'c']));
console.log(format(['a', 'b', 'c', 'd']));
```

packages/react-components/src/utils.ts
```diff
@@ -63,3 +63,32 @@ export const getIconFromUrl = (url: string): JSX.Element | undefined => {
...
+
+export const capitalizeText = (str: string = ''): string =>
```
kmaid, line 84: This should be called TitleCase. 

I am not sure we should be transforming an admin controlled list. Might want to push back on this requirement. I am also wondering if this should just be a text transform: Capitalise

kmaid (changes requested):

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -50,11 +52,18 @@ const UserProfileBackground: React.FC<UserProfileBackgroundProps> = ({
   approach = '',
   responsibilities = '',
   proposal,
-
+  labs = [],
```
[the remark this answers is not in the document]
ana-yldio: `UserProfileStaff` renders this component, but it doesn't show labs on the UI, so I made the `labs` field optional.

apps/asap-server/test/controllers/teams.test.ts
```diff
@@ -53,12 +54,9 @@ describe('Team controller', () => {
     identity();
   });
 
-  afterEach(() => {
-    expect(nock.isDone()).toBe(true);
-  });
-
   afterEach(() => {
     nock.cleanAll();
+    expect(nock.isDone()).toBe(true);
```
kmaid, line 59: Clear all the nock fixtures.
Test cleared nock fixtures have been completed?

packages/model/src/team.ts
```diff
@@ -30,6 +31,7 @@ export interface TeamMember {
   email: string;
   role: TeamRole;
   avatarUrl?: string;
+  labs?: Lab[];
```
kmaid, line 34: Should this be optional? Could it be [] if undefined in the transformer?

packages/react-components/src/utils/__tests__/text.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from '../index';
+
+describe('appendSuffix', () => {
+  it('can add a suffix to each value of the list', () => {
+    expect(appendSuffix('one', 'LAB')).toEqual('one LAB');
+  });
+});
+
+describe('getUniqueList', () => {
+  it('is case sensitive', () => {
```
kmaid, line 15: This test isn't required. We only need to test things we right code for. This is testing how things normally work.

packages/react-components/src/utils/__tests__/text.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      expect(getUniqueCommaString([''])).toEqual('');
+      expect(getUniqueCommaString(['one'])).toEqual('one');
+      expect(getUniqueCommaString(['one', 'two'])).toEqual('one and two');
+      expect(getUniqueCommaString(['one', 'two', 'three'])).toEqual(
+        'one, two and three',
+      );
+      expect(getUniqueCommaString([' one ', ' two '])).toEqual('one and two');
+      expect(
+        getUniqueCommaString(['one lab', 'two labs', 'three labs']),
+      ).toEqual('one lab, two labs and three labs');
```
kmaid, line 53:
```suggestion
```
This expectation is a duplicate of line 47.

packages/react-components/src/utils/__tests__/text.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      );
+      expect(
+        getUniqueCommaStringWithSuffix(['one', 'two', 'three'], 'lab'),
+      ).toEqual('one lab, two lab and three lab');
+      expect(
+        getUniqueCommaStringWithSuffix(
+          ['BCN one', 'LON two', 'MAD three'],
+          'lab',
+        ),
+      ).toEqual('BCN one lab, LON two lab and MAD three lab');
```
kmaid, line 73:
```suggestion
```
Duplicate of 65

packages/react-components/src/utils/text.ts
```diff
@@ -0,0 +1,29 @@
+export const appendSuffix = (val: string, suffix: string): string =>
+  suffix ? `${val} ${suffix}` : `${val}`;
+
+export const getUniqueList = (array: string[]): string[] =>
+  Array.from(new Set(array.filter((value) => value.trim())));
+
+export const getUniqueCommaString = (array: string[]): string => {
```
kmaid, line 7:
```suggestion
export const getCommaAndString = (array: string[]): string => {
```

packages/react-components/src/utils/text.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+  }
+};
+
+export const getUniqueCommaStringWithSuffix = (
+  array: string[],
+  suffix: string,
+): string =>
+  getUniqueCommaString(
+    getUniqueList(array).map((s) => `${appendSuffix(s, suffix)}`),
```
kmaid, line 28:
```suggestion
    getUniqueList(array).map((s) => appendSuffix(s, suffix)),
```

packages/react-components/src/utils/text.ts
```diff
@@ -0,0 +1,29 @@
+export const appendSuffix = (val: string, suffix: string): string =>
+  suffix ? `${val} ${suffix}` : `${val}`;
```
kmaid, line 2:
```suggestion
export const appendSuffix = (val: string, suffix: string): string => `${val} ${suffix}`
```

kmaid: suffix is required by this function. Not sure why there is a ternary

kmaid (changes requested):

packages/model/src/team.ts
ana-yldio: `TeamMembers` is created from `RestUser` or `GraphqlUser`. The first one doesn't have a labs field.

https://github.com/yldio/asap-hub/blob/74abb0af41ae1979d653ddc5e24a7f0218753853/apps/asap-server/src/controllers/teams.ts#L138
https://github.com/yldio/asap-hub/blob/74abb0af41ae1979d653ddc5e24a7f0218753853/apps/asap-server/src/controllers/teams.ts#L105

apps/asap-server/test/controllers/teams.test.ts
jhonny-m: shouldn't the order matter?

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -50,11 +52,17 @@ const UserProfileBackground: React.FC<UserProfileBackgroundProps> = ({
...
   firstName,
 }) => {
   const teamHref = network({}).teams({}).team({ teamId: id }).$;
   const { isOwnProfile } = useContext(UserProfileContext);
+
+  const labsList = getUniqueCommaStringWithSuffix(
+    labs.map((lab) => lab.name) || [],
```
jhonny-m, line 62: map should always return an array

jhonny-m (approved):

kmaid (approved):

PR #1008, opened 2021-08-11
Set sentry DSN specifically on development.

kmaid:
Upload source maps for both dev and production builds

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -240,7 +240,8 @@ deploy:sls:production:
   <<: *tmpl_deploy
   artifacts:
     paths:
-      - 'apps/auth-frontend/build'
+      - 'apps/*/build*'
+      - 'packages/*/build*'
   before_script:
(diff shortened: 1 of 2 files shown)


peterstarling (approved):

.gitlab-ci.yml
```diff
@@ -53,6 +53,11 @@ stages:
 
 build:ts:
   image: $INTEGRATION_DOCKER_IMAGE
+  rules:
+    - if: $CI_COMMIT_BRANCH == 'master'
```
kmaid, line 57: @peterstarling it doesn't like this section.

peterstarling: yeah try either
```
rules:
    variables:
        REACT_APP_SENTRY_DSN: ${REACT_APP_SENTRY_DSN_DEV}
    - if: $CI_COMMIT_BRANCH != 'master'
 ```
 
 or just from inside the `script` definition remap the variable there

PR #1009, opened 2021-08-11
fixing scroll issues with onboarding header and small screens

jhonny-m:
https://trello.com/c/I7z0VkYv/815-cannot-scroll-all-the-way-to-the-bottom-on-low-height-screens


kmaid (approved):
Lets do this for now and we can reconsider if it becomes a problem

kmaid (approved):

PR #1010, opened 2021-08-11
Add orcid schema changes and test validation script

harlan-kohll:
Add orcid schema regexes and hints and one-off standalone orcid validation script.
https://trello.com/c/YoLh54Tt/1387-enforce-orcid-on-the-cms


apps/asap-server/validate-orcids-script/validate-orcids.ts
```diff
@@ -0,0 +1,28 @@
+// This is a one-off script. It was run on prod on <date> and the results (invalid ORCIDs) were passed on to product
```
harlan-kohll, line 1: update the date when the script is run

apps/asap-server/validate-orcids-script/validate-orcids.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+  }
+
+  if (noInvalidOrcids) {
+    // eslint-disable-next-line no-console
+    console.log('No users with invalid orcids found.');
+  }
+};
+
+validateOrcids();
```
harlan-kohll, line 28: This script will be run by ssh'ing into prod then `TS_NODE_COMPILER_OPTIONS='{"module": "commonjs"}' yarn ts-node apps/asap-server/validate-orcids-script/validate-orcids.ts`

peterstarling: is there any point of adding this to codebase? i assume this is a one off to make sure all orcids are valid - if not what is the action?

apps/asap-server/test/integration/research-outputs.integration-test.ts
```diff
@@ -51,9 +51,13 @@ describe('Research Outputs', () => {
     try {
       await createResearchOutput(researchOutput);
     } catch (e) {
-      expect(e.name).toBe('HTTPError');
-      expect(e.output.statusCode).toBe(400);
-      expect(e.data).toMatch('doi.iv: Must follow the pattern');
+      const parsedErrorData = JSON.parse(e.data);
```
peterstarling, line 54: whats going on here if i may ask? why the change?

apps/asap-server/test/integration/fetch-users.integration-test.ts
```diff
@@ -3,16 +3,16 @@ import Chance from 'chance';
...
 
 const chance = new Chance();
 const users = new Users();
 
 describe('Users', () => {
   test('Should create and fetch a user', async () => {
-    const randomOrcid = chance.ssn();
     const randomEmail = chance.email();
     const randomName = chance.guid();
+    const randomOrcid = createRandomOrcid();
```
peterstarling, line 15: do we want to add a test with an invalid Orcid rejected by squidex?

peterstarling (changes requested):
is it possible to update orcid through the API or just in squidex? if so - i think it's probably through the patch request of /user route? should you not apply the validation there?

apps/asap-server/test/integration/fetch-users.integration-test.ts
harlan-kohll: Have written the test for invalid orcids but on reflection do we want a test for this? isn't this like testing squidex itself (which we said we would avoid doing in a previous ticket)?

apps/asap-server/test/integration/research-outputs.integration-test.ts
harlan-kohll: the testing data format changed slightly, have updated test accordingly

apps/asap-server/validate-orcids-script/validate-orcids.ts
harlan-kohll: eh, may as well keep the knowledge in there, may help in the future

apps/asap-server/src/entities/user.ts
```diff
@@ -18,7 +18,6 @@ export const userUpdateSchema = Joi.object({
   biography: Joi.string().allow(''),
   country: Joi.string().allow(''),
   city: Joi.string().allow(''),
-  orcid: Joi.string().allow(''),
```
harlan-kohll, line 21: Have deleted these because orcid can't be updated in FE

apps/asap-server/test/integration/fetch-users.integration-test.ts
```diff
@@ -51,4 +51,77 @@ describe('Users', () => {
...
+      jobTitle: 'Project Manager',
+      orcid: 'invalid orcid',
+      institution: 'Instituto Superior Tecnico',
+      email: randomEmail,
+      role: 'Grantee' as Role,
+      degree: 'MPH',
+    };
+
+    try {
+      const result = await createUser(user);
```
peterstarling, line 72: im not a huge fan of try catching inside tests. Could you try and use the jest `expect(...).rejects.toThrow()` syntax?

apps/asap-server/test/integration/fetch-users.integration-test.ts
peterstarling: well since it's the only way we can test that we actually have a validation in place i would say - sure, why not, leave it

apps/asap-server/test/integration/research-outputs.integration-test.ts
peterstarling: i was going to suggest the jest expect...toThrow syntax, same as above, but I am not 100% sure you are able to test all these different error properties nicely with it. Since, i would give it a go, even though we had try/catch in this test previously

apps/asap-server/validate-orcids-script/validate-orcids.ts
peterstarling: cool, then maybe you should consider adding some sort of README next to it or something and provide a command to run it as well some description?

apps/asap-server/src/entities/user.ts
peterstarling: that's fine but how do we know orcid is no longer being passed down to the controller? Like.. I know we could test any other arbitrary property but i wonder if it's actually ignored? I guess you could extend/create a test in `user.route.test.ts` to see if it's really removed from the payload when passed further down to the controller

peterstarling (changes requested):

apps/asap-server/test/routes/user.route.test.ts
```diff
@@ -354,6 +354,15 @@ describe('/users/ route', () => {
       expect(userControllerMock.update).toBeCalledWith(userId, { onboarded });
     });
 
+    test('Should not call the controller method if parameters are not valid', async () => {
```
peterstarling, line 357: i think we should be a little more specific here - i wonder why we even got the 400? is it because orcid is of invalid type or because it should not be in the input at all (i think somewhere above Harlan removed it from the input)

apps/asap-server/validate-orcids-script/validate-orcids.ts
peterstarling: @ana-yldio maybe add the link about how to run this script too? its here in this comment thread

peterstarling (changes requested):

apps/asap-server/test/routes/user.route.test.ts
ana-yldio: We got 400 when passing an orcid field (valid and invalid values). 
I think that's correct because Harlam removed it from the input. 
I'll update it, so it covers both cases 👍

peterstarling (approved):

ana-yldio:
see https://github.com/yldio/asap-hub/pull/1063

PR #1015, opened 2021-08-16
feat: show labs on Team's members and Network

ana-yldio:
https://trello.com/c/fJyOlqEm/1492-show-labs-on-team-members-list-on-team-page
https://trello.com/c/T4c42EIw/1495-show-labs-people-list-network


apps/asap-server/src/controllers/teams.ts
```diff
@@ -217,14 +208,10 @@ export default class Teams implements TeamController {
...
-    const teamUsers = await Promise.all(
-      teams.map((team) => fetchUsers(team.id, this.users.client)),
-    );
-
-    const teamItems = teams.map((team, index) =>
+    const teamItems = teams.map((team) =>
       transformGraphQLTeam(
         team,
-        transformRestTeamMember(teamUsers[index], team.id),
+        transformGraphQLUser(team?.referencingUsersContents || [], team),
```
peterstarling, line 214: actually - since you're changing the logic so much - I wonder if we need to pass down the members separately when they are still part of the `team`? like.. whats the point of extracting the members as a separate argument when the `GraphqlTeam` type has a reference to the team members? As far as I can tell `transformGraphQLTeam` function could get those members from the first argument

apps/asap-server/src/controllers/teams.ts
```diff
@@ -237,26 +224,23 @@ export default class Teams implements TeamController {
...
     const { findTeamsContent: team } = teamResponse;
+
     if (!team) {
       throw Boom.notFound();
     }
 
     return transformGraphQLTeam(
       team,
-      transformRestTeamMember(users, teamId),
+      transformGraphQLUser(
```
peterstarling, line 240: same as in my other comment - whats the point of passing users as a separate argument if the `referencingUsersContents` property is accessible from the `team`?

apps/asap-server/test/controllers/teams.test.ts
```diff
@@ -46,18 +42,12 @@ describe('Team controller', () => {
...
-  const getUserFilterExpectation = (teamId: string): string =>
-    `data/teams/iv/id eq '${teamId}' and data/onboarded/iv eq true`;
-
   beforeAll(() => {
     identity();
   });
 
   afterEach(() => {
     expect(nock.isDone()).toBe(true);
-  });
```
peterstarling, line 58: why are you merging the two `afterEach`? they are here for a reason and the reason is that if it's a single `afterEach` once the first statement throws the other one will not get executed - but we need to clean all mocks each time so it's important that it runs. It can't be run as first however as it will mess up the expectation

peterstarling (changes requested):

apps/asap-server/test/controllers/teams.test.ts
ana-yldio: No reason. I misunderstood this. Thanks for the explanation

packages/react-components/src/molecules/UserProfilePersonalText.tsx
```diff
@@ -63,6 +66,8 @@ const UserProfilePersonalText: FC<UserProfilePersonalTextProps> = ({
             ASAP Staff on <Link href={discover({}).$}>Team ASAP</Link>
           </>
         ) : null}
+        <br />
```
jhonny-m, line 69: we might need to check if labs.length is > 0 to add this line break

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -71,6 +74,12 @@ const UserProfileBackground: React.FC<UserProfileBackgroundProps> = ({
             <p>{role}</p>
           </div>
         </div>
+        {labs && (
```
jhonny-m, line 77: this will not catch the `[]`

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -1,14 +1,16 @@
...
 import { UserProfileContext } from '@asap-hub/react-context';
 
 import { Card, Headline2, Headline3, Link } from '../atoms';
 import { mobileScreen, perRem } from '../pixels';
 import UserProfilePlaceholderCard from './UserProfilePlaceholderCard';
+import { UserProfileLabText } from '../molecules';
 
-type UserProfileBackgroundProps = UserTeam & Pick<UserResponse, 'firstName'>;
+type UserProfileBackgroundProps = UserTeam &
+  Pick<UserResponse, 'firstName'> & { labs?: Lab[] };
```
jhonny-m, line 13: Not sure if it is correct to do this, but when we set up the default value, it makes the labs property optional. So we could use the one from `UserResponse`

harlan-kohll (approved):

packages/react-components/src/molecules/__tests__/UserProfilePersonalText.test.tsx
```diff
@@ -29,7 +29,7 @@ it.each`
 
 it('does not show the location icon if no location is available', () => {
   const { queryByTitle } = render(
-    <UserProfilePersonalText teams={[]} role={'Grantee'} />,
+    <UserProfilePersonalText teams={[]} role={'Grantee'} labs={[]} />,
```
kmaid, line 32: Rather than modify every test each time there is a new prop can you do something like https://github.com/yldio/asap-hub/pull/1015/files#diff-c5e608943b86b21f188f1b72070124978e7cd87f080765ce67cde9ce174f5243R7

apps/storybook/src/NetworkPeople.stories.tsx
```diff
@@ -27,6 +27,10 @@ const peopleProps = (): ComponentProps<typeof NetworkPeople> => {
           role: 'Co-Investigator' as TeamRole,
         },
       ],
+      labs: [
```
kmaid, line 30: Its fine to have these hardcoded like this as you can alter them in a more granular component 👍

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -79,10 +74,10 @@ const UserProfileBackground: React.FC<UserProfileBackgroundProps> = ({
             <p>{role}</p>
           </div>
         </div>
-        {!!labsList.length && (
+        {!!labs.length && (
           <div css={detailsContentStyle}>
             <Headline3 styleAsHeading={5}>Labs</Headline3>
-            {labsList}
+            <UserProfileLabText labs={labs} />
```
kmaid, line 80: Not sure about this. You are now putting in an elipsis when we were not previously.
```suggestion
{getUniqueCommaStringWithSuffix(labs.map(({name}) => name), 'Lab',)}
```

packages/react-components/src/organisms/__tests__/PeopleCard.test.tsx
```diff
@@ -30,6 +30,15 @@ it('renders the date in the correct format', () => {
     `"Joined: 1st January 2021"`,
   );
 });
+it('only show lab information if the user is on a lab', async () => {
+  const { queryByText, rerender } = render(
+    <PeopleCard {...props} displayName="John Doe" />,
```
kmaid, line 35: Don't rely on fixture data. Please set labs in the props. You don't need to set display name as that should be dealt with with the defaults in {...props} and we don't assert on it.

packages/react-components/src/molecules/UserProfileLabText.tsx
```diff
@@ -0,0 +1,17 @@
+import { UserResponse } from '@asap-hub/model';
+import { Ellipsis } from '../atoms';
+import { getUniqueCommaStringWithSuffix } from '../utils';
+
+type UserProfileLabTextProps = Partial<Pick<UserResponse, 'labs'>>;
+
+const UserProfileLabText: React.FC<UserProfileLabTextProps> = ({
```
kmaid, line 7: I don't like this component as it can be replaced with one line of code. I think it's already caused problems like introducing Ellipsis where it shouldn't be (see below).
`<Ellipsis>{getUniqueCommaStringWithSuffix(labs.map(({name}) => name),'Lab')}</Ellipsis>`

packages/react-components/src/molecules/UserProfilePersonalText.tsx
```diff
@@ -40,8 +41,10 @@ const UserProfilePersonalText: FC<UserProfilePersonalTextProps> = ({
   jobTitle,
   teams,
   role,
+  labs = [],
```
kmaid, line 44: Do we still need this default? Lab is required on UserResponse

packages/react-components/src/templates/__tests__/NetworkPeople.test.tsx
```diff
@@ -30,8 +37,12 @@ const props: ComponentProps<typeof NetworkPeople> = {
 };
 
 it('renders multiple people cards', () => {
-  const { queryAllByRole } = render(<NetworkPeople {...props} />);
+  const { queryAllByRole, queryAllByText } = render(
+    <NetworkPeople {...props} />,
+  );
```
kmaid, line 42:
```suggestion
```

packages/react-components/src/templates/__tests__/NetworkPeople.test.tsx
```diff
@@ -30,8 +37,12 @@ const props: ComponentProps<typeof NetworkPeople> = {
...
   expect(
     queryAllByRole('heading').map(({ textContent }) => textContent),
   ).toEqual(['Person A', 'Person B']);
+
+  expect(queryAllByText('Barcelona Lab and Glasgow Lab')).toHaveLength(1);
```
kmaid, line 47:
```suggestion
```

kmaid: The purpose of this test is to check the integration of the people card onto the page. Showing labs has already been tested in people card and therefore already covered

kmaid (changes requested):

packages/react-components/src/organisms/UserProfileBackground.tsx
```diff
@@ -79,10 +78,10 @@ const UserProfileBackground: React.FC<UserProfileBackgroundProps> = ({
             <p>{role}</p>
           </div>
         </div>
-        {!!labsList.length && (
+        {!!labsList?.length && (
```
jhonny-m, line 81: is there a way that labsList is not an string?

packages/react-components/src/molecules/__tests__/UserProfilePersonalText.test.tsx
jhonny-m: is the link correct? @kmaid

kmaid: No, but I have spoken with @ana-yldio so it already changed

kmaid (approved):

jhonny-m (approved):

peterstarling (approved):

packages/react-components/src/molecules/MembersList.tsx
```diff
@@ -104,10 +109,10 @@ const MembersList: React.FC<MembersListProps> = ({
               css={[
                 addToColumnStyles,
                 singleColumn || multiColumnAddToColumnStyles,
-                roleStyles,
+                labStyles,
```
jhonny-m, line 112: we can keep the `roleStyles`and add the `labStyles´ with the padding so we don't set the color in two different places

ana-yldio: In that case, we should change the name to something more generic. I'm using `textStyles`

jhonny-m (approved):

PR #1016, opened 2021-08-16
feat: Automate user invitation

peterstarling:
https://trello.com/c/KoRVJXXk/1322-send-invites-to-each-new-user-to-a-sandbox


harlan-kohll:
Can all the .zips in .yarn be .gitignored? Otherwise looks good.

peterstarling:
> Can all the .zips in .yarn be .gitignored? Otherwise looks good.

the whole point of zero-install is that they are not ignored and we don't run `yarn install` in our build. See: https://yarnpkg.com/features/zero-installs

jhonny-m (commented):
It looks good to me

jhonny-m (approved):

PR #1017, opened 2021-08-17
feat: improve sentry alerts

migcarva:
- allowed and blocked URLs
- sample Rate
- ignore errors list


apps/frontend/src/App.tsx
```diff
@@ -33,6 +33,24 @@ init({
...
+    'hub.asap.science/.auth/static/js', // code served from Auth0
+  ],
+  denyUrls: [
+    /fonts\.googleapis\.com/i, // code served from Google Fonts
+    /extensions\//i, // Chrome extensions
+    /^chrome:\/\//i, // Chrome extensions
+    /gtag\/js\//i, // Google Tag Manager
+  ],
+  ignoreErrors: [
+    'Failed to fetch event list. Expected status 2xx. Received status 401',
```
kmaid, line 50: This should be a regex otherwise it won't catch all of these errors.

apps/frontend/src/App.tsx
```diff
@@ -33,6 +33,24 @@ init({
...
+    'Failed to fetch dynamically imported module: (gtm)',
+    'Failed to fetch dynamically imported module: https://www.googletagmanager.com/gtm.js?id=GTM-M5HZBRZ',
```
kmaid, line 52: Are we sure this isn't an actual issue?

apps/frontend/src/App.tsx
```diff
@@ -33,6 +33,24 @@ init({
...
+    'asap-hub.us.auth0.com', // code served from Auth0
+    'hub.asap.science/.auth/static/js', // code served from Auth0
+  ],
+  denyUrls: [
+    /fonts\.googleapis\.com/i, // code served from Google Fonts
+    /extensions\//i, // Chrome extensions
+    /^chrome:\/\//i, // Chrome extensions
+    /gtag\/js\//i, // Google Tag Manager
+  ],
+  ignoreErrors: [
```
kmaid, line 49: We should be filtering out chunk load errors.

kmaid (changes requested):

apps/frontend/src/App.tsx
migcarva: I couldn’t find any indication of it. I'm removing it from the list anyway.

apps/frontend/src/App.tsx
```diff
@@ -33,6 +33,23 @@ init({
...
+    /Failed to fetch \w+.*? list. Expected status 2xx. Received status 401/g,
+    /Loading chunk [0-9]*? failed/g,
```
kmaid, line 51: Some of the chunks are named eg main. They won't be picked up by this regex. Check the built files for all the examples

kmaid (changes requested):

apps/frontend/src/App.tsx
migcarva: Let's go with this, and see if we have any issues with "named" chunks. I guess not, because the error I saw on Sentry was for a named chunk, and it appeared as a number (welcome.chunk gets us: Loading chunk 55 failed.)

kmaid (approved):

PR #1020, opened 2021-08-19
wip

ana-yldio:


apps/frontend/src/dashboard/api.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const getDashboard = async (
+  authorization: string,
+): Promise<DashboardResponse> => {
+  const resp = await fetch(`${API_BASE_URL}/dashboard`, {
+    headers: { authorization },
+  });
+  if (!resp.ok) {
+    throw new Error(
+      `Failed to fetch research output list. Expected status 2xx. Received status ${`${resp.status} ${resp.statusText}`.trim()}.`,
```
kmaid, line 12: This error message needs to be changed to reflect its for dashboard.

apps/frontend/src/dashboard/__tests__/Dashboard.test.tsx
```diff
@@ -29,7 +30,12 @@ const mockGetResearchOutputsLegacy =
 const renderDashboard = async (user: Partial<User>) => {
   const result = render(
     <Suspense fallback="loading">
-      <RecoilRoot>
+      <RecoilRoot
+        initializeState={({ set, reset }) => {
+          set(refreshDashboardState, Math.random());
+          reset(dashboardState);
```
kmaid, line 36:
```suggestion
```

apps/frontend/src/dashboard/__tests__/Dashboard.test.tsx
```diff
@@ -45,53 +51,39 @@ const renderDashboard = async (user: Partial<User>) => {
...
 
 it('prefetches research outputs (REGRESSION)', async () => {
-  nock(API_BASE_URL)
-    .get('/dashboard')
-    .reply(200, {
-      newsAndEvents: [],
-      pages: [],
-    } as DashboardResponse)
-    .persist();
+  mockGetDashboard.mockResolvedValue({ newsAndEvents: [], pages: [] });
```
kmaid, line 86:
```suggestion
```
You should be able to omit this line since there is a default mock in that __mocks__ directory and its not the focus of this test.

apps/frontend/src/dashboard/__tests__/api.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+jest.mock('../../config');
+
+describe('getDashboard', () => {
+  afterEach(() => {
+    expect(nock.isDone()).toBe(true);
+  });
+
+  afterEach(() => {
+    nock.cleanAll();
+  });
```
kmaid, line 14:
```suggestion
  afterEach(() => {
    expect(nock.isDone()).toBe(true);
    nock.cleanAll();
  });
```

I know this is counter to what you were told in the backend but this will work here and much more readable.

Even if expect(nock.isDone()).toBe(true); fails at worst it will fail the next test in addition but its probably overwritten anyway.

apps/frontend/src/dashboard/state.ts
```diff
@@ -0,0 +1,25 @@
...
+import { DashboardResponse } from '@asap-hub/model';
+import { authorizationState } from '../auth/state';
+import { getDashboard } from './api';
+
+export const fetchDashboardState = selector<DashboardResponse>({
+  key: 'fetchDashboardState',
+  get: async ({ get }) => {
+    get(refreshDashboardState);
+    const dashboard = await getDashboard(get(authorizationState));
+    return dashboard;
```
kmaid, line 11:
```suggestion
    return getDashboard(get(authorizationState));
```

So not sure exactly whats happening here. I think there might be a problem as get(authorizationState) will return a promise? 

If it works you can just return the promise and make the function not async

packages/fixtures/src/news-and-events.ts
```diff
@@ -1,10 +1,11 @@
-import { NewsOrEventResponse } from '@asap-hub/model';
+import { NewsAndEventsType, NewsOrEventResponse } from '@asap-hub/model';
 
 export const createNewsAndEventsResponse = (
   key: string,
+  responseType: NewsAndEventsType = 'News',
 ): NewsOrEventResponse => ({
   id: `uuid-${key}`,
```
kmaid, line 7: Might be worth just adding the `responseType` into the key so we don't get duplicate Ids which i can see is handled above but just for someone in the future.

packages/fixtures/src/news-and-events.ts
```diff
@@ -1,10 +1,11 @@
...
-  type: 'News',
+  type: responseType,
   title: `News ${key} title`,
```
kmaid, line 9: I think we should put `responseType` in here too as events will show up as news in storybook or tests

apps/frontend/src/dashboard/__tests__/api.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it('returns a successfully fetched dashboard', async () => {
+    const dashboardResponse = {
```
kmaid, line 17: Please type this response

kmaid (changes requested):

kmaid (approved):

PR #1026, opened 2021-08-24
fix: Hide tools and non-onboarded users

peterstarling:
see: https://trello.com/c/ehQe5myf/1547-private-team-tab-team-workspace-is-being-exposed-to-all-users

This PR moves the auth logic to the route layer keeping the controller libraries unaware of the logged-user state.

### DIFF
--- a/apps/asap-server/src/controllers/teams.ts
+++ b/apps/asap-server/src/controllers/teams.ts
@@ -102,6 +102,7 @@ export type FetchTeamsOptions = {
   search?: string;
   filter?: string | string[];
   // select team IDs of which tools should be returned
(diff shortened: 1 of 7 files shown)


apps/asap-server/src/controllers/teams.ts
```diff
@@ -44,7 +44,7 @@ flatData {
     url
   }
 }
-referencingUsersContents {
+referencingUsersContents(filter: "data/onboarded/iv eq true") {
```
peterstarling, line 47: with how we do tests in asap-server there is currently no way to write a sensible test case for this type of change (ie query change)

ana-yldio (approved):

PR #1034, opened 2021-08-27
feat: Support events with multiple parents

peterstarling:
ticket: https://trello.com/c/uTXT3Crm/1412-manage-events-that-belong-to-multiple-calendars

Notable changes:
* converts calendar hooks to use the event-bus
* adds a new controller method
* modifies linting rules to allow overloading
* renames the `id` fields in squidex to `google-calendar-id`


apps/asap-server/.eslintrc.js
```diff
@@ -12,5 +12,7 @@ module.exports = {
       'WithStatement',
     ],
     'no-underscore-dangle': ['error', { allow: ['__typename'] }],
+    'no-dupe-class-members': 'off',
+    '@typescript-eslint/no-dupe-class-members': ['error'],
```
peterstarling, line 16: typescript eslint rule is more clever about it and recognises overloads

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -91,6 +102,51 @@ export default class Calendars implements CalendarController {
     return res.items[0];
   }
 
+  fetchById(id: string, options?: { raw: false }): Promise<CalendarResponse>;
```
peterstarling, line 105: you might wonder why i added this method - well at first i thought i'd use it in the calendar-handler to fetch the newest calendar data (atm we get it from the event payload). Turns out we need to stick to what we have because of how google do subscriptions - we have to be able to compare the old and new payload each time to cancel subscriptions etc.

I figured since I wrote the method I might as well leave it here

apps/asap-server/src/utils/sync-google-event.ts
```diff
@@ -61,7 +61,7 @@ export const syncEventFactory = (
       endDateTimeZone: googleEvent.end.timeZone || defaultTimezone,
       status: (googleEvent.status.charAt(0).toUpperCase() +
         googleEvent.status.slice(1)) as EventStatus, // TODO: use lowercase
-      calendar: [calendarId],
+
```
peterstarling, line 64: im making sure that the calendar ID does not get overwritten - although it's not necessary as there is a conditional added (Further down) to skip the update if it comes from a different calendar so only the original owner of the event would be attached

ana-yldio (commented):
LGTM

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
```diff
@@ -301,3 +263,12 @@ describe('Unsubscribing', () => {
     expect(nock.isDone()).toBe(true);
   });
 });
+
+const getEvent = (
```
jhonny-m, line 267: I'd rather see this before the tests

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -91,6 +102,51 @@ export default class Calendars implements CalendarController {
...
+  fetchById(id: string, options?: { raw: true }): Promise<CalendarRaw>;
+  async fetchById(
+    calendarId: string,
```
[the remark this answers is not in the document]
peterstarling: well we could for consistency but i think i would like us to stop using generic ID name since it's confusing. Sometimes its the calendar ID, sometimes some other entity, and then there is an ID attached by squidex or google so

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
peterstarling: you would? I kinda prefer to have all the fixtures and generators either in a different file or at the bottom. I can move to another file actually

jhonny-m (approved):

PR #1035, opened 2021-08-27
Feat: Algolia Schema sync

kmaid:
Schema upload and download


peterstarling (changes requested):
not sure if this is still in progress or not but bear in mind that:

- we are moving to github actions - current algolia index setup for the PR envs lives there and i think so should these two jobs
- we use algolia cli for operations like this and i really dont think a custom script is necessary, see: https://github.com/algolia/algolia-cli#8-set-settings--setsettings

.github/workflows/master-sync.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Development Environment Variables
+        run: |
+          echo "ALGOLIA_APP_ID=$(aws ssm get-parameter --name 'algolia-app-id-dev' --query Parameter.Value --output text)" >> $GITHUB_ENV
+          echo "ALGOLIA_API_KEY=$(aws ssm get-parameter --name 'algolia-api-key-ci-dev' --query Parameter.Value --output text)" >> $GITHUB_ENV
+          echo "::add-mask::$(aws ssm get-parameter --name 'algolia-api-key-ci-dev' --query Parameter.Value --output text)"
+          echo "ALGOLIA_RESEARCH_OUTPUT_INDEX=asap-hub_research_outputs_dev >> $GITHUB_ENV
+      - name: Synchronize Development Schema
+        run: yarn run algolia:upload
+      - name: Setup Production Environment Variables
```
peterstarling, line 24: I'd slit this into two jobs, the prod one would have the `needs: sync_algolia_settings_dev` config

peterstarling (approved):
one minor comment, its up to you if you want to do it

PR #1036, opened 2021-08-31
fix: Separate the Squidex app creation and data import into a new job

ana-yldio:
https://trello.com/c/8trOLvTW/1522-separate-squidex-import-from-the-sls-deploy-to-speed-up-the-pipeline

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -144,10 +144,8 @@ test:build-output:
 ####################
 # stage deploy-dev #
 ####################
-
-deploy:sls:branch:
+deploy:squidex:branch:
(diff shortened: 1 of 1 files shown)


peterstarling (changes requested):
you may want to add another condition here:
https://github.com/yldio/asap-hub/pull/1036/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R208
so that we know that the  `deploy:squidex:branch` job finished. Gitlab is a little funny in that it tends to run another stage before the previous one finishes if there is no `needs` param

peterstarling (approved):

PR #1037, opened 2021-08-31
1324 search for outputs on team page

jhonny-m:
https://trello.com/c/PVXOkq93/1324-search-for-outputs-on-team-page


apps/frontend/src/network/teams/__tests__/Outputs.test.tsx
```diff
@@ -1,39 +1,118 @@
...
+    ...createAlgoliaResearchOutputResponse(2),
+    hits: createAlgoliaResearchOutputResponse(2).hits.map((hit, index) => ({
+      ...hit,
+      title: `Test Output ${index}`,
+    })),
+  });
+  const { container } = await renderOutputs([], '');
+  expect(container.textContent).toContain('Test Output 0');
+  expect(container.textContent).toContain('Test Output 1');
+});
```
ana-yldio, line 118: Shouldn't we add tests to cover if it allows typing search queries and applies filters?

ana-yldio (approved):

apps/frontend/src/shared-research/__tests__/api.test.ts
```diff
@@ -105,6 +105,31 @@ describe('getResearchOutputs', () => {
...
+      teamId: '12345',
+    });
+
+    expect(mockIndex.search).toHaveBeenLastCalledWith(
+      '',
+      expect.objectContaining({ filters: 'team.id:"12345"' }),
+    );
+  });
+
+  it('adds teamId to other filters', async () => {
```
peterstarling, line 120: you may wanna add a test that uses a couple of type filters and a team filter. I would expect it to produce a query with brackets around the types which I dont think it will

peterstarling (changes requested):

apps/frontend/src/shared-research/__tests__/api.test.ts
jhonny-m: Good catch 👍

peterstarling (approved):

apps/frontend/src/network/teams/Outputs.tsx
```diff
@@ -1,5 +1,95 @@
...
+  });
+  const { numberOfPages, renderPageHref } = usePagination(
+    result?.total || 0,
+    pageSize,
+  );
+  return (
+    <TeamProfileOutputs
+      outputs={
+        isEnabled('ALGOLIA_RESEARCH_OUTPUTS')
+          ? result?.items || []
```
kmaid, line 47: is result always potentially undefined?

packages/react-components/src/templates/__tests__/TeamProfileOutputs.test.tsx
```diff
@@ -27,6 +81,13 @@ it('renders output cards', () => {
           ],
         },
       ]}
+      numberOfItems={1}
```
kmaid, line 84: This should really be abstracted into const props = ComponentType<typeof....

apps/frontend/src/api-util.ts
```diff
@@ -5,6 +5,7 @@ export type GetListOptions = {
   filters: Set<string>;
   currentPage: number | null;
   pageSize: number | null;
+  teamId?: string;
```
kmaid, line 8: Not really thrilled by this. This should be moved into a separate type that extends this one specifically for research outputs as teamId won't apply to other endpoints.

kmaid (changes requested):

apps/frontend/src/network/teams/Outputs.tsx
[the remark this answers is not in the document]
jhonny-m: if the feature flag is off the results will be undefined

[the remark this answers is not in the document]
kmaid: Hrmm, rule of hooks means it has to run. Maybe return an empty result rather than undefined

kmaid (approved):

PR #1064, opened 2021-09-21
adding research outputs to user profiles

jhonny-m:
https://trello.com/c/0Pqjcq4c/1325-search-for-outputs-on-profile-page-must


apps/frontend/src/hooks/search.ts
```diff
@@ -40,7 +41,7 @@ export const useSearch = () => {
     searchQuery,
     debouncedSearchQuery,
     setSearchQuery,
-    filters,
+    filters: new Set<string>(currentFilters),
```
kmaid, line 44: We should probably add a unit test for this. While I know what this does now there is no guarantee someone would in the future without a test

apps/frontend/src/network/users/Outputs.tsx
```diff
@@ -1,5 +1,119 @@
...
+    filters,
+    searchQuery,
+    toggleFilter,
+    setSearchQuery,
+    debouncedSearchQuery,
+  } = useSearch();
+
+  return (
+    <article>
+      <ProfileCardList>
```
kmaid, line 72: Do we need to use this? Profile card list is mainly about offering the floating pencils

kmaid (approved):

apps/frontend/src/network/users/Outputs.tsx
jhonny-m: it was to maintain the same spacing as the rest of the page

apps/frontend/src/network/users/__tests__/Outputs.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+it('shows the coming soon text  (REGRESSION)', async () => {
+  disable('RESEARCH_OUTPUTS_ON_AUTHOR_PROFILE');
+  const { getByText } = await renderOutputs();
+  expect(getByText(/more\sto\scome/i)).toBeVisible();
+});
+
+it('renders search and filter', async () => {
+  const { getByRole } = await renderOutputs();
+  expect(
+    (getByRole('searchbox') as HTMLInputElement).placeholder,
```
kmaid, line 78: You can use toHaveAttribute("placeholder") to avoid the cast

apps/frontend/src/network/users/__tests__/UserProfile.test.tsx
```diff
@@ -133,13 +145,33 @@ it('navigates to the background tab', async () => {
...
+      title: `Test Output ${index}`,
+    })),
+  });
+  const { findByText, getByRole } = await renderUserProfile(
+    createUserResponse(),
+  );
+
+  userEvent.click(await findByText(/output/i, { selector: 'nav *' }));
+  expect(
+    (getByRole('searchbox') as HTMLInputElement).placeholder,
```
kmaid, line 170: Same as above

kmaid (approved):

PR #1066, opened 2021-09-22
feat: Add Labs to Research Output response and make them searchable

peterstarling:
Adds labs to RO response and index settings so they become searchable in Algolia.
Also tweaks some types and fixes the bug in PMs email list

see: https://trello.com/c/poF2Q9ay/1572-make-labs-searchable-on-algolia


apps/asap-server/src/entities/research-output.ts
```diff
@@ -109,6 +109,11 @@ export const parseGraphQLResearchOutput = (
     asapFunded: convertDecisionToBoolean(data.asapFunded),
     usedInPublication: convertDecisionToBoolean(data.usedInAPublication),
     pmsEmails: uniquePmsEmails,
+    labs:
+      data.labs?.map((lab) => ({
+        id: lab.id,
+        name: lab.flatData.name || '',
+      })) || [],
   };
```
ana-yldio, line 117: Teams endpoint doesn't default the lab's name to an empty string. It doesn't seem to be a problem now or when we do https://trello.com/c/5OfLSTzo/1491-show-labs-on-roms-detail-page

Do you think the two endpoints (team, rooms) should handle labs data the same way?

peterstarling: totally agreed - you should have probably requested the change not approve :D

PR #1069, opened 2021-09-24
Research Outputs Export

kmaid:
https://trello.com/c/v5T2OpQW/1257-5-export-search-results-to-csv


apps/frontend/src/shared-research/export.ts
```diff
@@ -0,0 +1,102 @@
...
+    Promise<SearchResponse<V>>
+  >,
+  transform: (result: V) => Row,
+) => {
+  let morePages = true;
+  let currentPage = 0;
+  while (morePages) {
+    // We are doing this in chunks and streams to avoid blob/ram limits.
+    // eslint-disable-next-line no-await-in-loop
+    const data = await getResults({
```
ana-yldio, line 93: how we handle when `getResults` fails?

peterstarling (approved):
it does seem to work although B stills needs to AC test it. Testing looks good too

apps/frontend/src/shared-research/export.ts
kmaid: Good catch! we don't and it silently fails. I am asking beatriz how we should handle this

kmaid: image

PR #1079, opened 2021-10-05
Change header location

kmaid:
Update styles


packages/react-components/src/molecules/AssociationList.tsx
```diff
@@ -0,0 +1,135 @@
...
+  inline = false,
+  max = Number.POSITIVE_INFINITY,
+}) => {
+  const icon = type === 'Team' ? teamIcon : labIcon;
+  return (
+    <div
+      css={[
+        containerStyles,
+        inline && containerInlineStyles,
+        associations.length > max && containerSummarizedStyles,
```
jhonny-m, line 96: we use the `associations.length > max` a few times, creating a variable to hold this would clean it up a bit, especially on longer conditions

packages/react-components/src/molecules/AssociationList.tsx
```diff
@@ -0,0 +1,135 @@
...
+      ]}
+    >
+      {((inline && !!associations.length) || associations.length > max) && icon}
```
jhonny-m, line 99: if `associations.length` if falsy will there be a scenario where `associations.length > max` would be truthy?

packages/react-components/src/molecules/AssociationList.tsx
```diff
@@ -0,0 +1,135 @@
...
+      {associations.length > max ? (
+        <>
+          {associations.length} {type}
+          {associations.length === 1 ? '' : 's'}
+        </>
+      ) : (
+        <ul css={[listStyles, inline ? listInlineStyles : listBlockStyles]}>
+          {associations.flatMap(({ displayName, id }, i) => (
```
jhonny-m, line 107: why do we need the flatMap?

jhonny-m (approved):

PR #1080, opened 2021-10-05
fix: User search with special characters

peterstarling:
Replace the native escape function with a custom sanitisation function for odata support

see: https://trello.com/c/IHU7bWU2/1525-search-does-not-work-on-terms-with-odd-characters-such-as-%C3%A8

### DIFF
--- a/apps/asap-server/src/utils/squidex.ts
+++ b/apps/asap-server/src/utils/squidex.ts
@@ -14,7 +14,16 @@ export const parseDate = (date: string): Date =>
   DateTime.fromISO(date).toJSDate();
 
 export const sanitiseForSquidex = (text: string): string =>
-  escape(text.replace(/'/g, "''"));
+  text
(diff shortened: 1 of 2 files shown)


apps/asap-server/src/utils/squidex.ts
```diff
@@ -14,7 +14,16 @@ export const parseDate = (date: string): Date =>
   DateTime.fromISO(date).toJSDate();
 
 export const sanitiseForSquidex = (text: string): string =>
-  escape(text.replace(/'/g, "''"));
+  text
+    .replace(/'/g, "''")
+    .replace(/%/g, '%25')
+    .replace(/"/g, '%22')
+    .replace(/'/g, '%27')
```
jhonny-m, line 21: we could do this in line 18

jhonny-m (approved):

PR #1071, opened 2021-09-27
feat: extends dropdown component

ana-yldio:
https://trello.com/c/27dUYTjL/1566-disable-custom-values-for-country-only-accept-allowed-values

- Updates Personal Info Modal, so the country selector only allow the selection of valid values
- Updates Dropdown component to support validation

PR #1086, opened 2021-10-08
feat: team update sync

ana-yldio:
https://trello.com/c/wQYMzkB8/1568-3-algolia-is-not-receiving-updates-regarding-the-teams-of-a-shared-output-must

### DIFF
--- a/apps/asap-server/src/handlers/teams/index-handler.ts
+++ b/apps/asap-server/src/handlers/teams/index-handler.ts
@@ -1,7 +1,9 @@
 import { EventBridgeEvent } from 'aws-lambda';
 import algoliasearch, { SearchClient } from 'algoliasearch';
 import { TeamsEventType } from '../webhooks/webhook-teams';
-import Team, { TeamController } from '../../controllers/teams';
+import ResearchOutputs, {
+  ResearchOutputController,
+} from '../../controllers/research-outputs';
 import {
   algoliaAppId,
   algoliaIndexApiKey,
@@ -10,7 +12,7 @@ import {
 import logger from '../../utils/logger';
 
 export const indexResearchOutputtByTeamHandler = (
-  teamController: TeamController,
+  researchOutputController: ResearchOutputController,
   algoliaClient: SearchClient,
 ): ((
   event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
@@ -20,18 +22,26 @@ export const indexResearchOutputtByTeamHandler = (
   return async (
     event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
   ): Promise<void> => {
-    const team = await teamController.fetchById(event.detail.payload.id);
+    const outputsIds = event.detail.payload.data.outputs.iv;
 
-    logger.info(`Fetched team with id ${team.id}`);
+    if (outputsIds.length > 0) {
+      await Promise.all(
+        outputsIds.map(async (id) => {
+          logger.info(`Fetch research-output with id ${id}`);
+          const researchOutput = await researchOutputController.fetchById(id);
 
-    if (team.outputs.length > 0) {
-      await algoliaIndex.saveObjects(
-        team.outputs.map((researchOutput) => ({
-          ...researchOutput,
-          objectID: researchOutput.id,
-        })),
+          logger.info(
+            `Fetched research-output ${JSON.stringify(researchOutput)}`,
+          );
+
+          await algoliaIndex.saveObject({
+            ...researchOutput,
+            objectID: researchOutput.id,
+          });
+
+          logger.info(`Saved research-output with id ${id}`);
+        }),
       );
-      logger.info(`Saved outputs of team with id ${team.id}`);
     }
   };
 };
@@ -42,10 +52,13 @@ export type SquidexWebhookTeamPayload = {
     $type: 'EnrichedContentEvent';
     type: 'Created';
     id: string;
+    data: {
+      outputs: { iv: string[] };
+    };
   };
 };
 
 export const handler = indexResearchOutputtByTeamHandler(
-  new Team(),
+  new ResearchOutputs(),
   algoliasearch(algoliaAppId, algoliaIndexApiKey),
 );
--- a/apps/asap-server/src/handlers/teams/index-handler.ts
+++ b/apps/asap-server/src/handlers/teams/index-handler.ts
@@ -1,0 +1,60 @@
+import { EventBridgeEvent } from 'aws-lambda';
+import algoliasearch, { SearchClient } from 'algoliasearch';
+import { TeamResponse } from '@asap-hub/model';
+import { TeamsEventType } from '../webhooks/webhook-teams';
+import Team, { TeamController } from '../../controllers/teams';
+import {
+  algoliaAppId,
+  algoliaIndexApiKey,
+  algoliaResearchOutputIndex,
+} from '../../config';
+import logger from '../../utils/logger';
+
+export const indexHandlerFactory =
+  (
+    teamController: TeamController,
+    algoliaClient: SearchClient,
+  ): ((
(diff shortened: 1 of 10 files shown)


apps/asap-server/src/handlers/teams/index-handler.ts
```diff
@@ -0,0 +1,60 @@
...
+    event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
+  ) => Promise<void>) =>
+  async (
+    event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
+  ): Promise<void> => {
+    const team = await teamController.fetchById(event.detail.payload.id);
+
+    logger.info(`Fetched team with id ${team.id}`);
+
+    await indexResearchOutputs(team, algoliaClient);
```
peterstarling, line 27: I see you tested the handler without mocking this so it's not really super relevant but... 

you could make the `indexResearchOutput` a function that takes a `ResearchOutputResponse` type and use it here and in [this handler](https://github.com/yldio/asap-hub/blob/master/apps/asap-server/src/handlers/research-output/index-handler.ts). Technically that would abstract away the algolia implementation and would enable you to mock it (which may become handy should you need to test some edge cases here).

peterstarling: one more thing - your `indexResearchOutputs` takes a dependency that it inherits from `indexHandlerFactory` - thats not a good pattern, in general, [see here](https://en.wikipedia.org/wiki/Dependency_inversion_principle)

`indexHandlerFactory` is not even interested in `algoliaClient`, so realistically it should instead take a `indexResearchOutput` dependency instead and not care about what that function does.

peterstarling: finally :D you could just leave it as is because your `indexResearchOutputs` is just a wrapper around some logic - it's just strange to see that you're having to pass down the algoliaClient - I guess since your `indexHandlerFactory` is a factory you could add the init phase to it (before returning the function) and define your `indexResearchOutputs` there - that way `algoliaClient` will already be available in the scope

apps/asap-server/test/handlers/teams/index-handler.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  afterEach(() => jest.clearAllMocks());
+
+  test('Should fetch the team and create a record in Algolia for every research output', async () => {
+    teamControllerMock.fetchById.mockResolvedValueOnce(teamResponse);
+
+    await indexHandler(getEvent());
+
+    expect(algoliaIndexMock.saveObjects).toHaveBeenCalledWith(
+      teamResponse.outputs.map((output) => ({
```
peterstarling, line 29: not a huge fan of such logic in tests -especially since you are effectively coping it from the implementing function. 

I would probably reassign `teamResponse.outputs` in the test and then expect these exact objects that you define.

apps/asap-server/test/handlers/teams/index-handler.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('Team Research Outputs Index', () => {
+  const indexHandler = indexHandlerFactory(
+    teamControllerMock,
+    algoliaClientMock,
+  );
+
+  afterEach(() => jest.clearAllMocks());
+
+  test('Should fetch the team and create a record in Algolia for every research output', async () => {
```
peterstarling, line 23: maybe add a test that confirms that algolia client is not called when the team has no associated researchOutputs?

apps/asap-server/test/handlers/webhooks/webhook-teams.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          EventBusName: eventBus,
+          Source: eventSource,
+          DetailType: 'TeamsUpdated',
+          Detail: JSON.stringify(updateTeamEvent()),
+        },
+      ],
+    });
+  });
+
+  test('Should put the teams-updated event into the event bus and return 200', async () => {
```
peterstarling, line 81: its the same test as the above, no?

packages/squidex/schema/rules/rule4.json
```diff
@@ -1,13 +1,13 @@
 {
   "$schema": "./../__json/rule.json",
   "isEnabled": true,
-  "name": "Teams upated",
+  "name": "Teams Updated",
```
peterstarling, line 4: hang on - did we have this rule already? i see it was added as part of [this PR](https://github.com/yldio/asap-hub/pull/1071/files#diff-c7331d70cf82ee31cdc53c03fd9f2e587bc72de2156c271e43c3083588ae16b7) - i take it it was added accidentally? how did it slip through the PR?

packages/squidex/schema/rules/rule4.json
```diff
@@ -1,13 +1,13 @@
...
   "trigger": {
     "triggerType": "ContentChanged",
     "schemas": [
       {
         "schemaId": "teams",
-        "condition": "event.type == 'Created' || event.type == 'Updated'"
+        "condition": "(event.type == 'Updated' || event.type == 'Published') && event.status == 'Published'"
```
peterstarling, line 10: see above comment - basically if we know for sure that the rule was here accidentally you are free to change it but otherwise i would check why it was here in the first place

packages/squidex/schema/rules/rule4.json
ana-yldio: yes, it was a mistake

apps/asap-server/src/handlers/teams/index-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+import { TeamsEventType } from '../webhooks/webhook-teams';
+import Team, { TeamController } from '../../controllers/teams';
+import {
+  algoliaAppId,
+  algoliaIndexApiKey,
+  algoliaResearchOutputIndex,
+} from '../../config';
+import logger from '../../utils/logger';
+
+export const indexHandlerFactory = (
```
peterstarling, line 12: I'd rename this one to `indexResearchOutputtByTeamHandler` and the other one to `indexResearchOutputHandler` or something along these lines cause it will be easier to use the named import this way

apps/asap-server/test/handlers/teams/index-handler.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      objectID: output.id,
+    }));
+
+    await indexHandler(getEvent());
+    expect(algoliaIndexMock.saveObjects).toHaveBeenCalledWith(
+      teamOutputsIndexes,
+    );
+  });
+
+  test('Should not trigger algolia when there are no research outputs', async () => {
```
peterstarling, line 42: `should not trigger algolia save when there are no research outputs associated with the team`

apps/asap-server/test/handlers/teams/index-handler.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      },
+    ];
+
+    outputs.forEach((output) =>
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce(output),
+    );
+
+    const updateEvent = getEvent();
+    updateEvent.detail.payload.data.outputs.iv = outputs.map((ro) => ro.id);
+    await indexHandler(getEvent());
```
peterstarling, line 41: you're not actually using the `updateEvent` from above, im surprised it still works

apps/asap-server/test/handlers/teams/index-handler.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    expect(algoliaIndexMock.saveObject).toHaveBeenCalledTimes(2);
+    expect(algoliaIndexMock.saveObject).toHaveBeenNthCalledWith(1, {
```
peterstarling, line 44: we dont really care about the order of calls do we? i would simply assert that the call was made

ana-yldio:
move to https://github.com/yldio/asap-hub/pull/1086
exceed api calls limit

PR #1087, opened 2021-10-08
Fix: Export for excel 

kmaid:
https://trello.com/c/3djBcMMK/1644-export-is-not-opening-correctly-on-excel-must

### DIFF
--- a/apps/frontend/src/shared-research/__tests__/export.test.ts
+++ b/apps/frontend/src/shared-research/__tests__/export.test.ts
@@ -13,6 +13,7 @@ import {
   researchOutputToCSV,
   createCsvFileStream,
   algoliaResultsToStream,
+  EXCEL_CELL_CHARACTER_LIMIT,
 } from '../export';
 
 const mockWriteStream = { write: jest.fn(), close: jest.fn() };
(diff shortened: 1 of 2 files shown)


apps/frontend/src/shared-research/__tests__/export.test.ts
```diff
@@ -158,20 +176,40 @@ describe('createCsvFileStream', () => {
       a: 'test',
     });
 
-    expect(mockWriteStream.write.mock.calls[0].toString()).toEqual('a,b');
-    expect(mockWriteStream.write.mock.calls[1].toString()).toMatch(
+    expect(mockWriteStream.write.mock.calls[1].toString()).toEqual('a,b');
```
kmaid, line 179: I don't like this. The problem is calls[1] so i cant use toHaveBeenCalledWith. I have had to increase this number because the first write is writeDom so the CSV supports UTF-8 characters which are common in description

apps/frontend/src/shared-research/export.ts
```diff
@@ -26,7 +41,7 @@ export const researchOutputToCSV = (
     .map((team) => team.displayName)
     .sort(caseInsensitive)
     .join(','),
-  labs: (output.labs ?? [])
+  labs: output.labs
```
kmaid, line 44: This was a hack while we didn't have labs on algolia in development.

peterstarling (approved):

PR #1091, opened 2021-10-12
Feat: Change typings for Calendar & Events to GraphQL

kmaid:
https://trello.com/c/TlY40kaa/1606-5-replace-graphql-types-with-the-generated-ones-calendar-and-events


apps/asap-server/src/controllers/events.ts
```diff
@@ -74,35 +87,35 @@ export default class Events implements EventController {
...
+    >(FETCH_EVENTS, {
       filter: filters.join(' and '),
       top: take,
       skip,
       order: orderby,
     });
 
-    const { total, items: events } = queryEventsContentsWithTotal;
+    const total = result.queryEventsContentsWithTotal?.total ?? 0;
+    const events = result.queryEventsContentsWithTotal?.items ?? [];
```
jhonny-m, line 118: I had copied what @peterstarling did on research-outputs, the return will be the same but we are logging a warning when either `queryEventsContentsWithTotal` or the `item` are null

PR #1095, opened 2021-10-13
feat: removes Algolia record when research output deleted/drafted

ana-yldio:
https://trello.com/c/UVHhQ6TX/1656-algolia-is-not-receiving-updates-when-research-outputs-is-delete-or-converted-to-draft

### DIFF
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -30,11 +30,7 @@ export const indexResearchOutputHandler = (
   ): Promise<void> => {
     logger.debug(`Event ${event['detail-type']}`);
 
-    if (
-      ['ResearchOutputCreated', 'ResearchOutputUpdated'].includes(
-        event['detail-type'],
-      )
-    ) {
+    try {
       const researchOutput = await researchOutputController.fetchById(
         event.detail.payload.id,
       );
@@ -47,9 +43,9 @@ export const indexResearchOutputHandler = (
       });
 
       logger.debug(`Saved research-output ${researchOutput.id}`);
-    }
+    } catch (e) {
+      logger.debug(JSON.stringify(e));
 
-    if (event['detail-type'] === 'ResearchOutputDeleted') {
       await algoliaIndex.deleteObject(event.detail.payload.id);
       logger.debug(`Deleted research-output ${event.detail.payload.id}`);
     }
@@ -58,7 +54,7 @@ export const indexResearchOutputHandler = (
 
 export type SquidexWebhookResearchOutputPayload = {
   type:
-    | 'ResearchOutputsCreated'
+    | 'ResearchOutputsPublished'
     | 'ResearchOutputsUpdated'
     | 'ResearchOutputsUnpublished'
     | 'ResearchOutputsDeleted';
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -45,15 +45,9 @@ export const indexResearchOutputHandler = (
       logger.debug(`Saved research-output ${researchOutput.id}`);
     } catch (e) {
       if (e?.output?.statusCode === 404) {
-        try {
-          await algoliaIndex.deleteObject(event.detail.payload.id);
-        } catch (err) {
-          logger.debug(JSON.stringify(err));
-          logger.debug(`Deleted research-output ${event.detail.payload.id}`);
-        }
-      } else {
-        throw e;
+        await algoliaIndex.deleteObject(event.detail.payload.id);
       }
+      throw e;
     }
   };
 };
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -44,11 +44,15 @@ export const indexResearchOutputHandler = (
 
       logger.debug(`Saved research-output ${researchOutput.id}`);
     } catch (e) {
-      logger.debug(JSON.stringify(e));
-
-      if (e.output.statusCode === 404) {
-        await algoliaIndex.deleteObject(event.detail.payload.id);
-        logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+      if (e?.output?.statusCode === 404) {
+        try {
+          await algoliaIndex.deleteObject(event.detail.payload.id);
+        } catch (err) {
+          logger.debug(JSON.stringify(err));
+          logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+        }
+      } else {
+        throw e;
       }
     }
   };
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -49,11 +49,7 @@ export const indexResearchOutputHandler = (
       logger.debug(`Saved research-output ${researchOutput.id}`);
     }
 
-    if (
-      ['ResearchOutputUnpublished', 'ResearchOutputDeleted'].includes(
-        event['detail-type'],
-      )
-    ) {
+    if (event['detail-type'] === 'ResearchOutputDeleted') {
       await algoliaIndex.deleteObject(event.detail.payload.id);
       logger.debug(`Deleted research-output ${event.detail.payload.id}`);
     }
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -46,8 +46,10 @@ export const indexResearchOutputHandler = (
     } catch (e) {
       logger.debug(JSON.stringify(e));
 
-      await algoliaIndex.deleteObject(event.detail.payload.id);
-      logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+      if (e.output.statusCode === 404) {
+        await algoliaIndex.deleteObject(event.detail.payload.id);
+        logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+      }
     }
   };
 };
--- a/apps/asap-server/src/handlers/research-output/index-handler.ts
+++ b/apps/asap-server/src/handlers/research-output/index-handler.ts
@@ -28,24 +28,47 @@ export const indexResearchOutputHandler = (
       SquidexWebhookResearchOutputPayload
     >,
   ): Promise<void> => {
-    const researchOutput = await researchOutputController.fetchById(
-      event.detail.payload.id,
-    );
+    logger.debug(`Event ${event['detail-type']}`);
 
-    await algoliaIndex.saveObject({
-      ...researchOutput,
-      objectID: researchOutput.id,
-    });
+    if (
+      ['ResearchOutputCreated', 'ResearchOutputUpdated'].includes(
+        event['detail-type'],
+      )
+    ) {
+      const researchOutput = await researchOutputController.fetchById(
+        event.detail.payload.id,
+      );
 
-    logger.info(`Saved Research-Output with ID ${researchOutput.id}`);
+      logger.debug(`Fetched research-output ${researchOutput.id}`);
+
+      await algoliaIndex.saveObject({
+        ...researchOutput,
+        objectID: researchOutput.id,
+      });
+
+      logger.debug(`Saved research-output ${researchOutput.id}`);
+    }
+
+    if (
+      ['ResearchOutputUnpublished', 'ResearchOutputDeleted'].includes(
+        event['detail-type'],
+      )
+    ) {
+      await algoliaIndex.deleteObject(event.detail.payload.id);
+      logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+    }
   };
 };
 
 export type SquidexWebhookResearchOutputPayload = {
-  type: 'ResearchOutputsCreated' | 'ResearchOutputsUpdated';
+  type:
+    | 'ResearchOutputsCreated'
+    | 'ResearchOutputsUpdated'
+    | 'ResearchOutputsUnpublished'
+    | 'ResearchOutputsDeleted';
   payload: {
     $type: 'EnrichedContentEvent';
-    type: 'Created';
+    type: 'Published' | 'Updated' | 'Unpublished' | 'Deleted';
     id: string;
   };
 };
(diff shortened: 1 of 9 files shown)


apps/asap-server/src/handlers/research-output/index-handler.ts
```diff
@@ -28,24 +28,43 @@ export const indexResearchOutputHandler = (
...
+
+      await algoliaIndex.saveObject({
+        ...researchOutput,
+        objectID: researchOutput.id,
+      });
+
+      logger.debug(`Saved research-output ${researchOutput.id}`);
+    }
+
+    if (event['detail-type'] === 'ResearchOutputDeleted') {
```
peterstarling, line 52: so there is a problem with deletes in general which i just discovered. Remember I said we should fetch the RO from squidex each time we receive this event? The reason I said it was because of the race conditions. Say somebody unpublishes a RO and then publish it again straight away - we have no guarantee that these notifications will arrive in order - either squidex or AWS may mix it up (and thats normal). If we dont check the RO state at the time of receiving the event then we may accidentally delete the RO because of the unpublish event coming after. 

So I would say that once you receive the Created/Updated event - fetch the RO from source - like you do now, which is great, but when it's deleted - you should still fetch it - if it's unpublished or deleted the call will return a 404 and thats when we know that the object should be removed, otherwise you either leave it or update it with what comes back.

What do you think?

apps/asap-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -323,6 +323,41 @@ export const getResearchOutputEvent = (): WebhookPayload<ResearchOutput> => ({
   },
 });
 
+export const researchOutputEvent = (
```
peterstarling, line 326: this is how fixtures become cluttered - you have an almost identical fixture above - if you want to build another one which is more generic why didn't you remove the one above and replace its references?

peterstarling (changes requested):

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +20,127 @@ describe('Research Output index handler', () => {
...
-      type: 'Created',
-      id: '0ecccf93-bd06-4307-90ea-c153fe495580',
-    },
-  };
+    expect(algoliaIndexMock.deleteObject).toHaveBeenCalledWith(
+      event.detail.payload.id,
+    );
+  });
+
+  test('Should not relay on the order of the events, i.e. unpublished and published', async () => {
```
peterstarling, line 102: i think this test doesn't really assert what you want anyway - do you know when the delete and save methods were called? You would probably want to write the expectations after the first and second invocation.

Anyway I feel like we need to test the following scenarios:

1. User unpublishes and quickly publishes - the events come in the right order
In such scenario we will invoke the handler with the unpublished and published event. Because the user quickly published the entity back - the resource will always be present. We want to make sure that `saveObject` method was called with the last invocation (might be called twice) - we could also make sure that the `deleteObject` method doesn't get called since the entity is always present.

2. User unpublishes and quickly publishes - the events come in the reverse order 
The order of the invocations is reversed however we still want to make sure that the `saveObject` method was called last  (it might even get called twice like previously) and again - we could check that the `deleteObject` method never gets called.

3. User publishes and quickly unpublishes - the events come in the right order
In here we want to make sure that the entity is eventually deleted. When the handlers receive the notifications the entity will not be present - btw can you make sure that the `deleteObject` method doesnt throw when the resource isn't present (i doubt that it does)? Technically in this scenario we could also assert that the `saveObject` method never gets called

4. User publishes and quickly unpublishes - the events come in the reverse order
I think you know where I am going with this so...

Basically we want to make sure that we capture the race conditions - I think your current implementation where the object is saved or deleted irrespective of the received event (it only depends on the current object state) is actually correct but I didn't give it much thought - this will come up in tests anyway.

I will make sure to add info to the ticket that these scenarios need to be tested  - I would recommend that you manually test it too before sending across to AC.

Maybe you can think of other scenarios that we want to test? I generally prefer to have too many tests than too few

apps/asap-server/test/fixtures/research-output.fixtures.ts
peterstarling: no problem - now that you converted it I would recommend that you follow our current practices and add a `get` prefix for the fixture generator because essentially this is a function that returns a fixture, not the fixture itself

apps/asap-server/src/handlers/research-output/index-handler.ts
```diff
@@ -28,24 +28,39 @@ export const indexResearchOutputHandler = (
...
+      await algoliaIndex.saveObject({
+        ...researchOutput,
+        objectID: researchOutput.id,
+      });
+
+      logger.debug(`Saved research-output ${researchOutput.id}`);
+    } catch (e) {
+      logger.debug(JSON.stringify(e));
+
+      await algoliaIndex.deleteObject(event.detail.payload.id);
```
peterstarling, line 49: what if algolia.saveObject throws? or what if it is an intermittent squidex networking error? we have to make sure its a 404 - but i would also double check to see what squidex returns when the resource is not present or is a draft

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+    const event = deleteEvent('ro-1234');
+    await indexHandler(event);
+
+    expect(algoliaIndexMock.deleteObject).toHaveBeenCalledWith(
+      event.detail.payload.id,
+    );
+  });
+
+  describe('Should process all events, and not relay on the order', () => {
+    it('receives on correct order the events created and updated', async () => {
```
peterstarling, line 72: why are you mixing `test` and `it`? I think we should stick to `test`

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+      expect(algoliaIndexMock.deleteObject).not.toHaveBeenCalled();
+      expect(algoliaIndexMock.saveObject).toHaveBeenLastCalledWith({
+        ...researchOutputResponse,
+        title: 'Updated title',
+        objectID: researchOutputResponse.id,
+      });
+
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce({
+        ...researchOutputResponse,
+        title: 'Old title',
```
peterstarling, line 133: this is a bit misleading - how would the second call to the `fetchById` return some 'old title'? I mean it could return anything but you would probably want to say that the `Updated title` is return the second time. In fact I would think that both the calls return the same `Updated title` if we are testing race conditions (so it would be whatever the last state of the resource was).

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+      const roID = 'ro-1234';
+      const createEv = createEvent(roID);
+      const unpublishedEv = unpublishedEvent(roID);
+
+      const researchOutputResponse = {
+        ...getResearchOutputResponse(),
+        id: roID,
+      };
+
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce(
```
peterstarling, line 154: ok so - we know that these functions called individually would do what you have in your test - but the challenge here is this:

when they created and then unpublish very quickly BOTH attempts to fetch the RO by ID return 404 - this is because by the time the 'create' event is received the status already changes in squidex. So your `fetchById` should throw twice with a 404

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+      await indexHandler(createEv);
+      expect(algoliaIndexMock.deleteObject).not.toHaveBeenCalled();
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledWith({
+        ...researchOutputResponse,
+        objectID: researchOutputResponse.id,
+      });
+
+      await indexHandler(unpublishedEv);
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledTimes(1);
+      expect(algoliaIndexMock.deleteObject).toHaveBeenCalledWith(
```
peterstarling, line 167: im actually surprised that this happens - shouldn't the delete only happen when `fetchById` throws? where does it say that it should throw in your test?

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+  test('Should remove the record Algolia when research-output is deleted', async () => {
+    const event = deleteEvent('ro-1234');
+    await indexHandler(event);
+
+    expect(algoliaIndexMock.deleteObject).toHaveBeenCalledWith(
+      event.detail.payload.id,
+    );
+  });
+
+  describe('Should process all events, and not relay on the order', () => {
```
peterstarling, line 71: i think the point of these tests is that these events happen almost at the same time - we are testing race conditions and the fact that they should not rely on the event data (or the order like you said). i think its important that the description outlines that so it should be something along the lines of `Should process the events, handle race conditions and not rely on the order of the events`

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +22,379 @@ describe('Research Output index handler', () => {
...
+
+      await indexHandler(createEv);
+      expect(algoliaIndexMock.deleteObject).not.toHaveBeenCalled();
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledWith({
+        ...researchOutputResponse,
+        title: 'Original title',
+        objectID: researchOutputResponse.id,
+      });
+
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce({
```
peterstarling, line 94: ok so the idea is this - the whole point of this entire block of tests is that we test events that happen close together. What you tested is simply invoking the handler twice when the data changes - you could split your current tests after the first invocation and expectation.

What we want to test is that although we have two different events - the `fetchById` method retyrns the same thing in BOTH invocations. Thats the whole point. We change something in the entity and then we change it again very quickly - so by the time its received by this handler we only see the last version of the entity

apps/asap-server/test/handlers/research-output/index-handler.test.ts
peterstarling: this is still unresolved - and for a reason. Both handler invocations should, once again, encounter 404s

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +23,336 @@ describe('Research Output index handler', () => {
...
+
+      const researchOutputResponse = {
+        ...getResearchOutputResponse(),
+        id: roID,
+      };
+
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce(
+        researchOutputResponse,
+      );
+      researchOutputControllerMock.fetchById.mockRejectedValue(Boom.notFound());
```
peterstarling, line 196: nope - again - the delete happens last in squidex and thats what both the handler invocations are going to see

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +23,336 @@ describe('Research Output index handler', () => {
...
+
+      await indexHandler(createEv);
+      await indexHandler(deleteEv);
+
+      expect(algoliaIndexMock.deleteObject).toHaveBeenCalledTimes(1);
```
peterstarling, line 201: FYI knowing that both times it will see a 404 it probably means that the `deleteObject` method will be called twice - and its fine - what we really care about is that at the end the object is deleted

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +23,336 @@ describe('Research Output index handler', () => {
...
+      expect(algoliaIndexMock.deleteObject).toHaveBeenCalledWith(
+        deleteEv.detail.payload.id,
+      );
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledTimes(1);
```
peterstarling, line 205: no, we dont want this

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +23,336 @@ describe('Research Output index handler', () => {
...
+        unpublishedEv.detail.payload.id,
+      );
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledTimes(1);
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledWith({
+        ...researchOutputResponse,
+        objectID: researchOutputResponse.id,
+      });
+    });
+
+    test('receives the reverse created and unpublished in correct order', async () => {
```
peterstarling, line 166: this test is correct

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -20,36 +23,336 @@ describe('Research Output index handler', () => {
...
+        deleteEv.detail.payload.id,
+      );
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledTimes(1);
+      expect(algoliaIndexMock.saveObject).toHaveBeenCalledWith({
+        ...researchOutputResponse,
+        objectID: researchOutputResponse.id,
+      });
+    });
+
+    test('receives the events created and deleted in reverse order', async () => {
```
peterstarling, line 212: this test is correct

apps/asap-server/src/handlers/research-output/index-handler.ts
```diff
@@ -28,24 +28,45 @@ export const indexResearchOutputHandler = (
...
+        ...researchOutput,
+        objectID: researchOutput.id,
+      });
+
+      logger.debug(`Saved research-output ${researchOutput.id}`);
+    } catch (e) {
+      if (e?.output?.statusCode === 404) {
+        try {
+          await algoliaIndex.deleteObject(event.detail.payload.id);
+        } catch (err) {
```
peterstarling, line 50: you can try catch the `deleteObject` if you like (its not necessary) but then you should still rethrow the error as we want the handler to retry if there is an error

apps/asap-server/src/handlers/research-output/index-handler.ts
```diff
@@ -28,24 +28,45 @@ export const indexResearchOutputHandler = (
...
+          logger.debug(JSON.stringify(err));
+          logger.debug(`Deleted research-output ${event.detail.payload.id}`);
+        }
+      } else {
+        throw e;
```
peterstarling, line 55: is this tested?

PR #1103, opened 2021-10-19
Feat: Sentry-tracing

kmaid:


apps/asap-server/src/middleware/sentry-transaction-id-handler.ts
```diff
@@ -0,0 +1,13 @@
+import { RequestHandler } from 'express';
+import { configureScope } from '@sentry/serverless';
+
+export const sentryTransactionIdFactory =
```
peterstarling, line 4: no need for it to be a factory, no?

apps/asap-server/src/app.ts
```diff
@@ -75,6 +76,8 @@ export const appFactory = (libs: Libs = {}): Express => {
   // Handlers
   const authHandler = libs.authHandler || authHandlerFactory(decodeToken);
   const tracingHandler = tracingHandlerFactory(libs.tracer);
+  const sentryTransactionIdHandler =
```
peterstarling, line 79: you're not testing it anyway so i would just do `app.use(<< handler here>>)` without looking it up in libs

peterstarling (changes requested):
i suppose we dont need to test it right? other than manual test it i mean

PR #1106, opened 2021-10-20
Fix events without associated calendars.

kmaid:
https://trello.com/c/pmZJMWOd/1659-events-without-an-associated-calendar

### DIFF
--- a/ci/integration/scripts/create-app.py
+++ b/ci/integration/scripts/create-app.py
@@ -32,7 +32,7 @@ def syncApp(appName, clientId, clientSecret):
     ))
     os.system('sq config use {appName}'.format( appName = appName ))
     # ATM we have not contents in the repo, but we can easily use this feature to setup fixture data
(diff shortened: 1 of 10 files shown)


ci/integration/scripts/create-app.py
```diff
@@ -15,12 +15,12 @@ def createAPP(appName):
...
     url = SQUIDEX_URL + "/api/apps/" + appName + "/clients"
     r = requests.get(url = url, headers = headers)
-    json_data = r.json() 
+    json_data = r.json()
     return json_data['items'][0]
 
 def promoteClient(appName, client):
     # Make user owner to allow him deleting the app
-    clientUrl = SQUIDEX_URL + "/api/apps/" + appName + "/clients/" + client['id'] 
+    clientUrl = SQUIDEX_URL + "/api/apps/" + appName + "/clients/" + client['id']
```
niezgoda, line 23: You could use the f-strings (if this is python 3.6 and above).
https://realpython.com/python-f-strings/

PR #1116, opened 2021-10-27
feat: show count of labs on Team listing on Network

ana-yldio:
https://trello.com/c/MdzNzcUS/1493-3-show-count-of-labs-on-team-listing-on-network

### DIFF
--- a/packages/react-components/src/organisms/TeamCard.tsx
+++ b/packages/react-components/src/organisms/TeamCard.tsx
@@ -5,8 +5,9 @@ import { network } from '@asap-hub/routing';
 import { Card, Anchor, Paragraph, Headline2 } from '../atoms';
 import { perRem } from '../pixels';
 import { lead } from '../colors';
-import { teamIcon } from '../icons';
+import { teamIcon, labIcon } from '../icons';
 import { TagList } from '../molecules';
+import { getCounterString } from '../utils';
 
 const teamMemberStyles = css({
   color: lead.rgb,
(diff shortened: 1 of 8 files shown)


packages/react-components/src/organisms/__tests__/TeamCard.test.tsx
```diff
@@ -48,3 +49,18 @@ it('pluralises when more than one team member', () => {
...
 });
+
+it('renders a lab count for multiple labs', () => {
+  const { getByText } = render(<TeamCard {...teamCardProps} labCount={13} />);
+  expect(getByText('13 Labs')).toBeVisible();
+});
+
+it('renders a lab count for a single lab using singular form', () => {
+  const { getByText } = render(<TeamCard {...teamCardProps} labCount={1} />);
+  expect(getByText('1 Lab')).toBeVisible();
```
peterstarling, line 60: is this the right way to test singular forms? I wonder if this test wouldn't pass even if plural was used?

packages/react-components/src/utils/text.ts
```diff
@@ -1,4 +1,9 @@
-export const appendSuffix = (val: string, suffix: string): string =>
+export const getCounterString = (count: number, term: string): string => {
+  const int = Number(count.toFixed());
```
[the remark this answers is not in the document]
peterstarling: well technically if we could do 1.5 Lab it would still be singular. I notice that you are not testing decimals yet using `toFixed()` - I guess since there is no way to enforce an `Integer` in TypeScript through types so we have to either throw or round like you did. I would suggest to test the decimals

peterstarling (changes requested):

packages/react-components/src/organisms/__tests__/TeamCard.test.tsx
ana-yldio: I believe that's covered by these tests 
https://github.com/yldio/asap-hub/blob/215fe56c8e8eed1fa5235fe1011fb77b934cb654/packages/react-components/src/utils/__tests__/text.test.tsx#L9-L25

peterstarling (approved):

packages/react-components/src/organisms/TeamCard.tsx
```diff
@@ -5,14 +5,21 @@ import { network } from '@asap-hub/routing';
...
 import { TagList } from '../molecules';
+import { getCounterString } from '../utils';
 
-const teamMemberStyles = css({
+const teamMemberMetaStyles = css({
   color: lead.rgb,
   display: 'flex',
   alignItems: 'center',
   padding: `${12 / perRem}em 0`,
+  ' > *': {
```
kmaid, line 17: Could we just use gap?

kmaid (approved):

PR #1121, opened 2021-11-02
adding noUncheckedIndexedAccess to asap-server

jhonny-m:
https://trello.com/c/e24ooYHm/1549-2-enable-nouncheckedindexedaccess-in-ts-config

### DIFF
--- a/apps/asap-server/src/entities/group.ts
+++ b/apps/asap-server/src/entities/group.ts
@@ -51,18 +51,18 @@ export const parseGraphQLGroup = (
   let tools: GroupTools = {};
   if (item.flatData.tools?.length) {
     const [groupTools] = item.flatData.tools;
-    if (groupTools.slack) {
+    if (groupTools?.slack) {
       tools = { ...tools, slack: groupTools.slack };
     }
 
-    if (groupTools.googleDrive) {
+    if (groupTools?.googleDrive) {
       tools = { ...tools, googleDrive: groupTools.googleDrive };
     }
   }
 
   if (item.flatData.calendars?.length) {
     const url = new URL('https://calendar.google.com/calendar/r');
-    url.searchParams.set('cid', item.flatData.calendars[0].id || '');
+    url.searchParams.set('cid', item.flatData.calendars[0]?.id || '');
     tools = { ...tools, googleCalendar: url.toString() };
   }
 
--- a/apps/asap-server/src/entities/user.ts
+++ b/apps/asap-server/src/entities/user.ts
@@ -207,7 +207,7 @@ export const parseUser = (user: RestUser): UserResponse => {
         throw new Error('Team id cannot be undefined');
       }
       return {
-        id: id[0]!,
+        id: id[0],
         displayName: 'Unknown',
         ...t,
         approach: t.approach ? t.approach : undefined,
--- a/apps/asap-server/src/entities/user.ts
+++ b/apps/asap-server/src/entities/user.ts
@@ -16,6 +16,7 @@ import { GraphqlUser, RestUser } from '@asap-hub/squidex';
 import { parseDate, createURL } from '../utils/squidex';
 import { FetchUserQuery } from '../gql/graphql';
 import { isTeamRole } from './team';
+import logger from '../utils/logger';
 
 export type CMSOrcidWork = OrcidWork;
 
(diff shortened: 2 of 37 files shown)


apps/asap-server/src/entities/group.ts
```diff
@@ -51,18 +51,18 @@ export const parseGraphQLGroup = (
   let tools: GroupTools = {};
   if (item.flatData.tools?.length) {
     const [groupTools] = item.flatData.tools;
```
kmaid, line 53:
```suggestion
    const [groupTools] = item.flatData.tools ?? {};
```

apps/asap-server/src/entities/user.ts
```diff
@@ -109,7 +113,9 @@ export const parseGraphQLUser = (
     ...((item.flatData.social && item.flatData.social[0]) || {}),
     orcid,
   }).reduce((acc, [k, v]) => {
-    if (v == null) return acc;
+    if (v == null) {
```
kmaid, line 116:
```suggestion
    if (v === null) {
```

apps/asap-server/src/entities/group.ts
jhonny-m: why does it needs this?

apps/asap-server/src/entities/user.ts
```diff
@@ -196,13 +202,18 @@ export const parseGraphQLUser = (
...
 export const parseUser = (user: RestUser): UserResponse => {
   const teams: UserTeam[] =
-    user.data.teams?.iv?.map(({ id, ...t }) => ({
-      id: id[0],
-      displayName: 'Unknown',
-      ...t,
-      approach: t.approach ? t.approach : undefined,
-      responsibilities: t.responsibilities ? t.responsibilities : undefined,
-    })) || [];
+    user.data.teams?.iv?.map(({ id, ...t }) => {
```
peterstarling, line 205: why dont you skip the team which does not have the ID instead? maybe log a warning or something too.

apps/asap-server/src/handlers/webhooks/webhook-run-migrations.ts
```diff
@@ -145,7 +146,7 @@ const getLatestMigrationPathFromDbFactory =
 
     const { items: migrations } = await client.fetch(query);
 
-    if (migrations.length === 0) {
+    if (migrations.length === 0 || !migrations[0]) {
```
peterstarling, line 149: i guess you could just use the `!migrations[0]`

apps/asap-server/src/routes/events.route.ts
```diff
@@ -1,7 +1,8 @@
-import { Router, Response } from 'express';
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
peterstarling, line 1: why dont you just use the next-line one instead for just that single assertion?

apps/asap-server/src/routes/groups.route.ts
```diff
@@ -1,9 +1,10 @@
-import { Router } from 'express';
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
peterstarling, line 1: same as above - why dont you disable it for just one line

apps/asap-server/src/routes/events.route.ts
```diff
@@ -25,7 +26,7 @@ export const eventRouteFactory = (eventController: EventController): Router => {
       const { params } = req;
       const { eventId } = framework.validate('parameters', params, paramSchema);
 
-      const result = await eventController.fetchById(eventId);
+      const result = await eventController.fetchById(eventId!);
```
peterstarling, line 29: instead of non-null assertion why dont you use something like this in your route definition

```typescript
eventRoutes.get<{ eventId: string}>('/events/:eventId', async (req, res) => {
```
It's still technically just asserting but at least you have the param definition right next to it so i think it's legit

apps/asap-server/src/routes/groups.route.ts
peterstarling: or dont - just do what i suggested above

apps/asap-server/test/controllers/news.test.ts
```diff
@@ -87,7 +87,7 @@ describe('News controller', () => {
           {
             ...newsSquidexApiResponse.items[0],
             data: {
-              ...newsSquidexApiResponse.items[0].data,
+              ...newsSquidexApiResponse.items[0]!.data,
```
peterstarling, line 90: can you not disable this rule in tests?

peterstarling (changes requested):
can you not disable the rule for the tests?

kmaid (approved):

peterstarling (approved):

PR #1134, opened 2021-11-05
feat: Set up Algolia automatic sync

peterstarling:
https://trello.com/c/ZtoQVDq3/1674-set-up-automatic-syncs-with-algolia

Changes:
* move the custom algolia scripts to asap-cli
* use the algolia api `moveIndex` update the indexes without downtime
* add daily algolia prod sync


kmaid (changes requested):
https://www.algolia.com/doc/api-reference/api-methods/move-index/

I think we should be using the move index approach. This way we can create an index and only when this process is complete we can replace the existing index with no downtime. I also think we should be using the squidex api to iterate through all results if possible

kmaid (approved):

PR #1127, opened 2021-11-03
chore: adds fields to user for migration

markcsims:
Create new fields in User schema:

Skills = expertiseandResourceTags
SkillsDescription = expertiseandResourceDescription
approach = mainResearchInterests

PR #1135, opened 2021-11-08
Chore: rename user fields use new fields

markcsims:
Uses new field after the migration
Removes old field

Continuation of https://github.com/yldio/asap-hub/pull/1127

The last 2 commits.


ana-yldio:
LGTM

PR #1138, opened 2021-11-09
Chore: rename user fields switch over

markcsims:
Uses new field after the migration
Removes old field

Continuation of #1127

The last 2 commits.


apps/asap-server/src/migrations/1636027226-move-repurposed-fields.ts
```diff
@@ -39,8 +50,9 @@ async function migrateTeamFields() {
   await applyToAllItemsInCollection<RestTeam>(
     'teams',
     async (team, squidexClient) => {
+      const { skills } = (team as unknown as Rest<{ skills: string[] }>).data;
```
peterstarling, line 53: i dont think you should change the logic of the old migrations. I see you're having to cast some types (i would try to do as little as possible in that area too) but probably leave the logic as is, if possible

peterstarling (approved):
only a small comment - to be fair this is a lot of file changes. I would say we should QA this properly before merging

apps/asap-server/src/migrations/1636027226-move-repurposed-fields.ts
markcsims: done

PR #1140, opened 2021-11-09
feat: update feature flags by setting a cookie manually

ana-yldio:
https://trello.com/c/obyvqP86/1662-2-allow-feature-flags-to-be-overridden-by-a-cookie-value

### DIFF
--- a/packages/flags/src/index.ts
+++ b/packages/flags/src/index.ts
@@ -26,7 +26,7 @@ export const isEnabled = (flag: Flag): boolean =>
 
 export const getOverrides = (): Flags => overrides;
 
-export const setManualOverrides = (flags?: Record<string, boolean>): void => {
+export const currentOverrides = (flags?: Record<string, boolean>): void => {
   overrides = Object.entries(overrides).reduce<Record<string, boolean>>(
     (acc, [name, val]) => {
       if (typeof val === 'boolean') {
--- a/packages/flags/src/index.ts
+++ b/packages/flags/src/index.ts
@@ -25,6 +25,18 @@ export const isEnabled = (flag: Flag): boolean =>
   false;
 
 export const getOverrides = (): Flags => overrides;
+
+export const setManualOverrides = (flags?: Record<string, boolean>): void => {
+  overrides = Object.entries(overrides).reduce<Record<string, boolean>>(
+    (acc, [name, val]) => {
+      if (typeof val === 'boolean') {
+        acc[name] = flags && flags[name] !== undefined ? flags[name] : val;
+      }
+      return acc;
+    },
+    {},
+  );
+};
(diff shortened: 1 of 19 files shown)


packages/flags/src/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+export const isValidFF = (str: string, flags: Flags) => {
+  try {
+    const parsedValue = JSON.parse(str);
+
+    if (Array.isArray(parsedValue)) return false;
+
+    if (Object.keys(parsedValue).length === 0) return false;
+
+    return Object.entries(parsedValue).some(
+      ([key, val]) =>
```
ana-yldio, line 35: I think that it's ok having additional keys on the cookie value as soon as we don't add them as feature flags. 
What do you think?

packages/flags/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  } catch (e) {
+    return false;
+  }
+};
+export const isFF = (name: string, value: string, flags: Flags): boolean =>
+  name.trim() === FF && isValidFF(value, flags);
+
+export const mergeFlags = (flags: Flags): Partial<Record<Flag, boolean>> => {
+  if (process.browser) {
+    return document.cookie
```
ana-yldio, line 26: It assumes having a single cookie that stores a JSON value `{ [feature_flag]: flag }`
Instead of setting a cookie for each feature flag.

kmaid (changes requested):
We should not be using Dom in a FE/BE agnostic package. This functionality should reside within 	packages/react-context/flags

packages/flags/src/index.ts
```diff
@@ -25,6 +25,18 @@ export const isEnabled = (flag: Flag): boolean =>
   false;
 
 export const getOverrides = (): Flags => overrides;
+
+export const setManualOverrides = (flags?: Record<string, boolean>): void => {
```
kmaid, line 29: Should this just be named setOverrides to keep naming consistent?

kmaid (approved):
Out of curiosity does the flags package still return the correct value when its overridden by a cookie?

packages/flags/src/index.ts
ana-yldio: `setOverrides` is already taken. I renamed it as `currentOverrides`
https://github.com/yldio/asap-hub/blob/5c87fdc52cfbd92bf0449526a704fe9976e2b842/packages/react-context/src/flags.tsx#L50

ana-yldio:
https://github.com/yldio/asap-hub/blob/5c87fdc52cfbd92bf0449526a704fe9976e2b842/packages/flags/src/__tests__/index.test.ts#L35

PR #1141, opened 2021-11-09
Handling errors on Fetch User

jhonny-m:
https://trello.com/c/4JbrKhWt/1702-user-request-throws-when-team-connection-is-undefined


apps/asap-server/test/controllers/users.test.ts
```diff
@@ -293,29 +273,12 @@ describe('Users controller', () => {
...
-        })
-        .reply(200, mockResponse);
-
-      await expect(users.fetchById('user-id')).rejects.toThrow(
-        'Invalid user-team connection',
-      );
-    });
-
-    test('Should throw an error when the team role is invalid', async () => {
+    test('Should filter out team the team role is invalid', async () => {
```
peterstarling, line 276: something is wrong with this description

apps/asap-server/src/entities/user.ts
```diff
@@ -59,37 +59,41 @@ export const userUpdateSchema = Joi.object({
...
-  if (!team) {
-    throw new Error(`User team connection is not defined`);
-  }
-  const displayName = team.flatData?.displayName;
-  const proposal = team.flatData?.proposal;
-
-  if (!item.role || !isTeamRole(item.role)) {
-    throw new Error(`Invalid team role: ${item.role}`);
-  }
+export const parseGraphQLUserTeamConnections = (
```
peterstarling, line 66: im not a huge fan of exposing functions only for tests - this could just be internal to the client and tested in the client test. We are about to drop the nock fixtures anyway for all entities which will make these tests much leaner.
But for now leave it as is

peterstarling (approved):
minor comments

PR #1144, opened 2021-11-11
Fixing tags area not working after right click

jhonny-m:
https://trello.com/c/DCZP4I8H/1559-%F0%9F%90%9B-tags-input-area-stops-working-after-a-right-click

### DIFF
--- a/packages/react-components/src/atoms/__tests__/MultiSelect.test.tsx
+++ b/packages/react-components/src/atoms/__tests__/MultiSelect.test.tsx
@@ -2,8 +2,9 @@ import userEvent from '@testing-library/user-event';
 import { fireEvent, render } from '@testing-library/react';
 import { findParentWithStyle } from '@asap-hub/dom-test-utils';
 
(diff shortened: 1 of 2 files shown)


packages/react-components/src/atoms/__tests__/MultiSelect.test.tsx
```diff
@@ -121,18 +121,15 @@ describe('invalidity', () => {
       />,
     );
     const input = getByRole('textbox');
-
     fireEvent.focusIn(input);
```
ana-yldio, line 124: The test description refers to the `right-click`, but the test uses `focusIn`. 
Does the bug affect other events (i.e. click)?

jhonny-m: I wasn't too sure on how to test it, so asserted on the regular behavior plus the right click. But I am open to suggestions

ana-yldio (approved):

PR #1150, opened 2021-11-16
Feat: New Research output template

kmaid:
https://trello.com/c/CLoRAmDK/1690-3-roms-proposal-template-update-header-separate-body


packages/react-components/src/organisms/SharedResearchAdditionalInformationCard.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+import { Card, Divider, Headline2, Link } from '../atoms';
+import { externalLinkIcon } from '../icons';
+import { mobileScreen, perRem } from '../pixels';
+import { isLink } from '../utils';
+
+const additionalInformationListStyles = css({
+  listStyle: 'none',
+  margin: 0,
+  padding: `${6 / perRem}em 0`,
```
jhonny-m, line 12: Not for this pr, but it would be cool to have a function we call and get the `value + em` like `rem(6)`

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -20,200 +22,71 @@ const cardsStyles = css({
...
+}) => {
+  const isGrantDocument = (
+    ['Grant Document', 'Proposal'] as ResearchOutputType[]
+  ).includes(props.type);
+  return (
+    <div css={containerStyles}>
+      <BackLink href={backHref} />
+      <div css={cardsStyles}>
+        <SharedResearchOutputHeaderCard {...props} />
+        {((description && !isGrantDocument) || !!tags.length) && (
```
jhonny-m, line 49: I would move the `(description && !isGrantDocument)` to a `showDescription` and the `tags.length` to `showTags` and to facilitate understanding the logic happening here

jhonny-m: or `showNormalDescription` or default to differentiate for the Grant document's one

jhonny-m: there is also no need for `!!`

packages/react-components/src/organisms/SharedResearchAdditionalInformationCard.tsx
kmaid: Yeah, I think that might be a good idea tbh

packages/react-components/src/templates/SharedResearchOutput.tsx
kmaid: I have removed the additional !!. The problem with trying to simplify the variable names is that there are three states. Grant Document | Other and then we show hide depending if description is there.

jhonny-m (approved):

PR #1153, opened 2021-11-17
checks if version of calendar is out of date

markcsims:
https://trello.com/c/6MqtYnAv/1411-3-ignore-out-of-sync-updates-to-the-calendar

Checking the version number in the controller.

The version/timestamp could be checked closer to squidex - but we unsubscribe
from google here.


peterstarling (changes requested):

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -187,6 +205,7 @@ export interface CalendarController {
 
   fetchById(id: string, options?: { raw: false }): Promise<CalendarResponse>;
   fetchById(id: string, options?: { raw: true }): Promise<CalendarRaw>;
+  fetchVersion(calendarId: string): Promise<number>;
```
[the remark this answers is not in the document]
markcsims: @peterstarling have a look at the latest commit. fetchById can now return CalendarRawResponse which is an extended CalendarRaw type with version on it.

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -186,9 +187,12 @@ export interface CalendarController {
...
 
   fetchById(id: string, options?: { raw: false }): Promise<CalendarResponse>;
-  fetchById(id: string, options?: { raw: true }): Promise<CalendarRaw>;
+  fetchById(id: string, options?: { raw: true }): Promise<CalendarRawResponse>;
 }
 
 export type CalendarRaw = Calendar & {
   id: string;
 };
+export type CalendarRawResponse = CalendarRaw & {
```
peterstarling, line 196: i think creating the new type is a little confusing - plus using the `Response` suffix indicates this would be, like the rest of the methods, meant for our API 

Simply add the version to the `CalendarRaw` type

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
```diff
@@ -123,44 +127,74 @@ describe('Calendar Webhook', () => {
...
+            resourceId: null,
+          },
+        );
+        expect(subscribe).toHaveBeenCalled();
+      });
 
-      const res = await handler(
-        getEvent('CalendarUpdated', calendarUpdateEvent),
-      );
+      test('Should skip subscription and unsubscribing and return 200 when the version is old', async () => {
```
peterstarling, line 155: are we sure you want to skip unsubscribing if the version is old? technically this means that if we need to unsubscribe but a new version has been saved we will never be able to unsubscribe. Or maybe we dont need to?

peterstarling (changes requested):

apps/asap-server/src/controllers/calendars.ts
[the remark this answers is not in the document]
markcsims: https://github.com/yldio/asap-hub/pull/1153/commits/5745b51c89ede3d01c266a9cbed62399b58e3c47

apps/asap-server/src/controllers/calendars.ts
markcsims: The function is already overloaded and returns 2 different types:

``` 
fetchById(id: string, options?: { raw: false }): Promise<CalendarResponse>;
fetchById(id: string, options?: { raw: true }): Promise<CalendarRaw>;
```

The other type has a `Response` suffix. The 2 are inconsistent.

markcsims: We should change `CalendarResponse`. 

`CalendarRawResponse` removed here: https://github.com/yldio/asap-hub/pull/1153/commits/cb6ad57c40dbdb9675e07c2ba35edb44ec10d571

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
markcsims: The unsubscribe is in 2 parts:

1. `await unsubscribe(payload.dataOld.resourceId?.iv, payload.id);`
2. ```await calendarController.update(payload.id, {
            resourceId: null,
          });```

The first unsubscribes from Google, I think we could do this.
The second sets the resourceid to null in squidex. Could we get in a senario where a later event has been applied, and the resourceid updated to a newer one?

peterstarling: i guess in the current setup we cannot - unless of course we subscribe from one valid calendar to another but that is a whole different scenario of course because then there will be no race condition and no retry attempt to unsubscribe and subscribe to an invalid (fake) calendar (which is what the admins do now)

If they 1) Change the calendar to a non-existent calendarId and then 2) Set it again to the same calendarId as at the beginning  

Then if 2) comes in first the system will detect no difference in the calendarIds (there is a comparison in the code right?) so it will not try to unsubscribe, in fact it will exit immediately

Once 1) comes in, in the old code it would then unsubscribe the calendar and try to subscribe to the new one which would fail resulting in empty subscription (can we test that? maybe there is no point) but with your improvement it should reject the update all together, right?


Question is - can we test it? Without actually running this against squidex it's a silly test really but we might be able to invoke the handler in one test twice with different input to replicate the out-of-order event issue and write some expectations that would at least describe what we are trying to do, what do you reckon?

markcsims: ```2) comes in first the system will detect no difference in the calendarIds (there is a comparison in the code right?)```

There is a comparision in the code, but in step 2 it will compare the valid calendar Id with the invalid one. Leading to it being unsubscribed.

https://github.com/yldio/asap-hub/pull/1153/commits/aab9a64ab168416f4fe7d7b6416b0a221aa65f17

peterstarling (approved):

PR #1158, opened 2021-11-19
Feat/1688 - Create a new type (Grant Document) and two subtypes (Proposal, Report) #1156

niezgoda:
https://trello.com/c/JS4dPCx7/1688-create-a-new-type-grant-document-and-two-subtypes-proposal-report

Requirements
- add the following Type for RO: Grant Document
- add the following Subtypes for RO: Proposal, Report
- make sure we use the Proposal template for Grant Documents
- remove Proposal from Squidex schema but not from our model

Note
- don't remove Proposals search filter until the migration is done

Out of scope
- update the search filter by adding Grant Documents as an option


apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -454,15 +480,15 @@ describe('ResearchOutputs controller', () => {
       test('Should pass the filter parameter as a squidex filter', async () => {
```
peterstarling, line 480: i guess we should probably test searching for Proposals and Grant Documents separately - i would expect the first one to use just the Proposal filter and the other one - both?

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,27 +69,27 @@ const renderComponent = async () => {
   return result;
 };
 
-describe('a proposal research output', () => {
+describe('a Grant Document research output', () => {
   it('renders with its teams', async () => {
```
peterstarling, line 73: i suppose this test should show the difference in rendering proposals or grant documents and other types - i think for the time being you should show that both Grant Documents and Proposal types are rendered differently (i think the test technically show that only the grant document one is different)

apps/frontend/src/shared-research/__tests__/api.test.ts
```diff
@@ -84,12 +84,14 @@ describe('getResearchOutputs', () => {
   it('builds a multiple filter query', async () => {
     await getResearchOutputs(mockIndex, {
       ...options,
-      filters: new Set<ResearchOutputType>(['Article', 'Proposal']),
+      filters: new Set<ResearchOutputType>(['Article', 'Grant Document']),
```
peterstarling, line 87: so if the Proposal is still being left should we not keep this test and add a new one to show how grant documents are searched?

apps/storybook/src/SearchAndFilter.stories.tsx
```diff
@@ -15,7 +15,7 @@ export const Normal = () => (
     searchQuery={text('Search Query', '')}
     filterTitle={text('Filter Title', 'TYPE OF OUTPUTS')}
     filterOptions={[
-      { label: 'Proposal', value: '1' },
+      { label: 'Grant Documentdata/type/iv', value: '1' },
```
peterstarling, line 18: why the `/type/iv`?

apps/frontend/src/shared-research/api.ts
```diff
@@ -35,7 +35,8 @@ export const researchOutputTypeFilters: Record<
   ResearchOutputType,
   { filter: string }
 > = {
-  Proposal: { filter: 'type:Proposal' },
+  'Grant Document': { filter: 'type:Proposal OR type:"Grant Document"' },
+  Proposal: { filter: 'type:Proposal OR type:"Grant Document"' },
```
peterstarling, line 39: so you are leaving Proposal in the api yeah?

peterstarling (changes requested):

peterstarling:
im not sure if i understood this correctly but i think you are leaving proposal in the filter and in the parameters and types so i think all the tests for the proposals should remain untouched

apps/frontend/src/shared-research/__tests__/api.test.ts
niezgoda: You cannot search by the Proposal. It's removed on the FE
image

apps/storybook/src/SearchAndFilter.stories.tsx
niezgoda: whoops.

apps/frontend/src/shared-research/api.ts
niezgoda: That's true

apps/asap-server/src/entities/research-output.ts
```diff
@@ -133,6 +136,7 @@ const isSharingStatus = (
   (sharingStatuses as ReadonlyArray<string>).includes(status);
 
 const isResearchOutputType = (type: string): type is ResearchOutputType =>
+  type !== 'Proposal' &&
```
kmaid, line 139: This seems strange. Why are proposals not a research output type? If proposals are remapped and never exposed to the frontend can we remove this type from the modal?

packages/react-components/src/templates/SharedResearchGrantDocument.tsx
```diff
@@ -0,0 +1,78 @@
+import { css } from '@emotion/react';
```
kmaid, line 1: I don't think this file exists anymore after my PR

kmaid (changes requested):

peterstarling (approved):

kmaid (approved):

PR #1161, opened 2021-11-23
feat: Basic E2E test with playwright

peterstarling:
https://trello.com/c/tUIzHvc8/1673-5-write-e2e-test-as-a-response-to-friday-merge-issue

## Notable changes
* reverts the transaction-id header
* adds the transation-id header to calls for static pages
* adds the e2e test with playwright to test the terms-and-conditions page - simple sanity check
* enables the x-transaction-id header in asap api server


apps/e2e-tests/tests/frontend/terms-and-conditions.test.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from '@playwright/test';
+
+test('basic test', async ({ page }) => {
+  await page.goto('https://1155.hub.asap.science/terms-and-conditions');
+
```
ana-yldio: I think this test will fail once this is merged. 
`https://1155.hub.asap.science/` won't be available

apps/frontend/src/__tests__/api-util.test.tsx
```diff
@@ -41,3 +46,18 @@ describe('createListApiUrl', () => {
...
   });
 });
+
+describe('createSentryHeaders', () => {
+  it('generates a header with a random string', () => {
+    expect(createSentryHeaders()).toEqual({
+      'X-Transaction-Id': expect.anything(),
+    });
+  });
+  it('sets the conte a header with a random string', () => {
```
ana-yldio, line 56: Double-check the description, there's a typo

ana-yldio (changes requested):

apps/e2e-tests/tests/frontend/terms-and-conditions.test.ts
peterstarling: thanks @ana-yldio I guess i should have marked this as WIP :D

apps/e2e-tests/tests/frontend/terms-and-conditions.e2e-test.ts
```diff
@@ -0,0 +1,12 @@
+import { test, expect } from '@playwright/test';
+import { config } from '../../config';
+
+test('Should be able to see the terms-and-conditions page', async ({
+  page,
+}) => {
+  await page.goto(`${config.appUrl}/terms-and-conditions`);
```
jhonny-m, line 7: wouldn't it be better if we clicked the terms-and-conditions link that there is in the login page?

peterstarling: i mean we could - but it does sound like adding bits to the scope. Do you think there is value in that? If there is i am happy to try, its best to make this more meaningful

kmaid: Could we reuse the typed routes from the routing package? The best end2end test follows the exact path a user might take but I already want to bypass authentication so atleast if we used the typed routings we would catch this reference if it was to ever change

kmaid (approved):

ana-yldio (approved):

PR #1172, opened 2021-11-26
chore: remove research outputs from teams endpoint

ana-yldio:
https://trello.com/c/TUdSORUR/1441-remove-research-outputs-from-teams-endpoint

- Deprecates the script for adding protocols
- Removes outputs field from teams' endpoint


apps/asap-server/src/queries/teams.queries.ts
```diff
@@ -119,7 +119,7 @@ export const FETCH_TEAMS = gql`
     $top: Int
     $skip: Int
     $filter: String
-    $withResearchOutputs: Boolean = true
```
peterstarling, line 122: do we ever fetch them WITH reasearch outputs?

packages/squidex/src/entities/team.ts
```diff
@@ -14,13 +13,10 @@ export interface Team<T = string> {
...
   }[];
-  outputs: T[];
 }
 
 export interface RestTeam extends Entity, Rest<Team> {}
 // TODO: REMOVE old proposal type
-export interface GraphqlTeam
-  extends Entity,
-    Graphql<Team<GraphqlResearchOutput | { id: string }>> {
+export interface GraphqlTeam extends Entity, Graphql<Team<{ id: string }>> {
```
peterstarling, line 20: is this still being used anywhere? I think we should have moved to the generated queries by now

apps/asap-server/src/queries/teams.queries.ts
peterstarling: ok so by NOT removing the reasearch outputs and only changing the default value for the Directive you effectively didn't alter the types for the response of the query - this is the downside of these generated queries which, if such directive is ever used, ALWAYS include the conditional resource. 

If there is a valid reason for keeping the directive and the flag here then leave it but otherwise i would remove this part entirely:
https://github.com/yldio/asap-hub/blob/b369fb43363491a9132e1098133ec68f1587a8f2/apps/asap-server/src/queries/teams.queries.ts#L13-L15

peterstarling (changes requested):

peterstarling (approved):

PR #1176, opened 2021-11-29
feat: retry to subscribe to calendars after failure

markcsims:
https://trello.com/c/MncCAJym/1717-retry-to-subscribe-to-calendars-after-faillure

### DIFF
--- a/apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
+++ b/apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
@@ -87,15 +87,13 @@ describe('Calendar handler', () => {
       );
     });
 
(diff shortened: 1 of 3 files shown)


apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
```diff
@@ -87,15 +87,13 @@ describe('Calendar handler', () => {
       );
     });
 
-    test('Should return "ERROR" and alert when the subscription was unsuccessful, but should not throw', async () => {
+    test('Should return 502 and alert when the subscription was unsuccessful', async () => {
```
peterstarling, line 90: to be fair that's not really what it does but I guess you wanted these tests to follow the same pattern - the only I see in the expectation is that it throws an error

peterstarling (approved):

PR #1177, opened 2021-11-30
Add proposal to list of types in Research Outputs

niezgoda:


peterstarling (approved):
add a description. 

Also - are we able to test this? Can you see what happens with an old Proposal document once we add this 'type' back in? is it going to show up as Proposal again?

I think maybe this is unnecessary at this point if you have been able to write a migration and run it, unless you haven't been because of this?

PR #1180, opened 2021-11-30
feat: Migrate all type proposals to type grant document subtype proposal

niezgoda:


apps/asap-server/src/migrations/1637748102-move-type-proposal-to-grant-research.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    pointer += 10;
+  } while (pointer < result.total);
+};
+
+export default class MoveResearchOutputTextToDescription extends Migration {
+  up = async (): Promise<void> => {
+    const squidexClient = new Squidex<RestResearchOutput>('research-outputs');
+
+    await iterate(fetchSquidex(squidexClient, 'Proposal'), (researchOutput) =>
```
peterstarling, line 47: you are building something that has already been built:  https://github.com/yldio/asap-hub/blob/709b1e48126d9665e9970c68f45d1407c2b551e5/apps/asap-server/src/migrations/1622036244-add-default-onboarded-field-to-users.ts#L8

and im pretty sure that it doesn't iterate through all the records because many have been left out.

Although your way does not fetch all the records, just the proposals. Maybe expand the `applyToAllItemsInCollection` to accept a filter? Or iterate over all of them but skip the ones that aren't proposal programatically?

peterstarling (changes requested):

apps/asap-server/src/utils/migrations.ts
```diff
@@ -1,30 +1,37 @@
...
     });
 
     for (const item of result.items) {
       await processEntity(item, squidexClient);
+      processedItems += 1;
     }
 
     pointer += 10;
-  } while (pointer < result.total);
+  } while (pointer <= result.items.length);
```
markcsims, line 34: Will changing this mean old migrations fail to work because they will be relying on the total property?

peterstarling (approved):

markcsims (approved):

PR #1184, opened 2021-12-01
fix: update ro templates

ana-yldio:
https://trello.com/c/5My9r4Bx/1727-2-change-detail-pages-for-presentation

### DIFF
--- a/packages/react-components/src/templates/SharedResearchOutput.tsx
+++ b/packages/react-components/src/templates/SharedResearchOutput.tsx
@@ -63,7 +63,7 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
             )}
           </Card>
         )}
-        {accessInstructions && (
+        {!isGrantDocument && accessInstructions && (
           <Card>
             <div css={{ paddingBottom: `${12 / perRem}em` }}>
               <Headline2 styleAsHeading={4}>Access Instructions</Headline2>
--- a/packages/react-components/src/templates/SharedResearchOutput.tsx
+++ b/packages/react-components/src/templates/SharedResearchOutput.tsx
@@ -38,8 +38,9 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
   contactEmails,
   ...props
 }) => {
-  const isGrantDocument = props.type === 'Grant Document';
-
+  const isGrantDocument = ['Grant Document', 'Presentation'].includes(
+    props.type,
+  );
   return (
     <div css={containerStyles}>
       <BackLink href={backHref} />
--- a/apps/storybook/src/SharedResearchOutput.stories.tsx
+++ b/apps/storybook/src/SharedResearchOutput.stories.tsx
@@ -91,7 +91,6 @@ export const GrantDocument = () => (
   <SharedResearchOutput
     {...props()}
     type="Grant Document"
-    accessInstructions={undefined}
     description={text(
       'Description',
       `<h1>Example</h1>


jhonny-m (commented):
I think @kmaid wanted to keep only one Research Output instead of having to maintain two types, most of the work seems to be separating them

apps/frontend/src/shared-research/ResearchOutput.tsx
```diff
@@ -1,10 +1,25 @@
...
+  SharedResearchOutputGrantDocument,
+} from '@asap-hub/react-components';
 import { sharedResearch, useRouteParams } from '@asap-hub/routing';
+import { ResearchOutputResponse } from '@asap-hub/model';
 
 import { useBackHref } from '../hooks';
 import { useResearchOutputById } from './state';
 import Frame from '../structure/Frame';
 
+const getResearchOutputTemplate = (type: ResearchOutputResponse['type']) => {
```
kmaid, line 13: Can we do something like it was before? https://github.com/yldio/asap-hub/pull/1150/files#diff-d532068a9bbf7564c5df0eae332d711fc99586a2084562d38e21e5e03ba71607L22. Its allot more readable

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,64 +69,206 @@ const renderComponent = async () => {
...
+      description: 'Dataset contents',
+      accessInstructions: 'instructions',
+      asapFunded: true,
+      sharingStatus: 'Public',
+      contactEmails: ['pmmail'],
+    });
+    const { getByText } = await renderComponent();
+    expect(getByText(/dataset title/i)).toBeVisible();
+    expect(getByText(/dataset contents/i)).toBeVisible();
+    expect(getByText(/dataset contents/i)).toBeVisible();
```
kmaid, line 187: I don't understand this assertion. its identical to the one above

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,64 +69,206 @@ const renderComponent = async () => {
...
+    expect(queryByText(/access instructions/i)).toBeNull();
+    expect(queryByText(/additional information/i)).toBeNull();
+    expect(getByText(/contact pm/i).closest('a')).toHaveAttribute(
+      'href',
+      expect.stringMatching(/pmmail/i),
     );
   });
-  it('links to a teams', async () => {
+
+  it('renders the proposal type', async () => {
```
kmaid, line 125: This is probably a regression test.

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,64 +69,206 @@ const renderComponent = async () => {
...
+      contactEmails: ['pmmail'],
     });
-    const { getByRole, getByText } = await renderComponent();
-    expect(getByText(/Example Tag/i)).toBeVisible();
-    expect(getByRole('heading', { level: 1 }).textContent).toEqual(
-      'Not-Grant-Document title!',
+    const { getByText } = await renderComponent();
+    expect(getByText(/protocol title/i)).toBeVisible();
+    expect(getByText(/protocol contents/i)).toBeVisible();
+    expect(getByText(/protocol contents/i)).toBeVisible();
```
kmaid, line 211: Same again

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,64 +69,206 @@ const renderComponent = async () => {
...
+      description: 'Lab Resource contents',
+      accessInstructions: 'instructions',
+      asapFunded: true,
+      sharingStatus: 'Public',
+      contactEmails: ['pmmail'],
+    });
+    const { getByText } = await renderComponent();
+    expect(getByText(/lab resource title/i)).toBeVisible();
+    expect(getByText(/lab resource contents/i)).toBeVisible();
+    expect(getByText(/lab resource contents/i)).toBeVisible();
```
kmaid, line 235: and again

apps/frontend/src/shared-research/__tests__/ResearchOutput.test.tsx
```diff
@@ -69,64 +69,206 @@ const renderComponent = async () => {
...
+      title: 'Bioinformatics title!',
+      description: 'Bioinformatics contents',
+      accessInstructions: 'instructions',
+      asapFunded: true,
+      sharingStatus: 'Public',
+      contactEmails: ['pmmail'],
+    });
+    const { getByText } = await renderComponent();
+    expect(getByText(/bioinformatics title/i)).toBeVisible();
+    expect(getByText(/bioinformatics contents/i)).toBeVisible();
```
kmaid, line 258: Can we not do this as a it.each? This is 5 duplicated tests which essentially test one prop difference.

packages/react-components/src/organisms/__tests__/SharedResearchOutputHeaderCard.test.tsx
```diff
@@ -65,14 +65,40 @@ it('shows authors', () => {
   expect(getByText('John Doe')).toBeVisible();
 });
 
-it('shows labs', () => {
-  const { getByText } = render(
+it('handle labs data', () => {
```
kmaid, line 68:
```suggestion
it('handles labs data', () => {
```

packages/react-components/src/templates/SharedResearchOutputGrantDocument.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+type SharedResearchOutputGrantDocumentProps = Pick<
+  ResearchOutputResponse,
+  'description' | 'tags' | 'contactEmails'
+> &
+  ComponentProps<typeof SharedResearchOutputHeaderCard> & {
+    backHref: string;
+  } & ComponentProps<typeof SharedResearchAdditionalInformationCard>;
+
+const SharedResearchOutputGrantDocument: React.FC<SharedResearchOutputGrantDocumentProps> =
```
kmaid, line 33: I wonder if this name should reflect the difference between the templates which is an expanded description SharedResearchOutputRTFDescription

kmaid (changes requested):

kmaid (commented):
I am not convinced that separating the templates has made this easier to read or maintain what do you think @jhonny-m ?

apps/frontend/src/shared-research/ResearchOutput.tsx
jhonny-m: IMO, we could keep it as only one research output component, since most of the things are going to be the same. with the container, header contact info and just have a component for the different part with is the description tags and access information which are the differences

jhonny-m (commented):
the difference between the components seems so small that it shouldn't justify 2 different components, I would suggest pehaps separating the difference which seems to be the tags and description to a different component if we want so simplify this template

ana-yldio:
@kmaid @jhonny-m I'm afraid I have to disagree, tbh I think it's clearer breaking the logic into two different templates.
But since you both agree on this, I reverted it and just updated the `isGrantDocument` variable,

kmaid (approved):

jhonny-m (approved):

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -39,7 +39,7 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
   ...props
 }) => {
   const isGrantDocument = (
-    ['Grant Document', 'Proposal'] as ResearchOutputType[]
+    ['Grant Document', 'Proposal', 'Presentation'] as ResearchOutputType[]
```
niezgoda, line 42: this needs to be changed after merge https://github.com/yldio/asap-hub/pull/1194

niezgoda (changes requested):

niezgoda (approved):

PR #1185, opened 2021-12-02
Feat: No research outputs for Team and User profile

kmaid:
https://trello.com/c/4qTyEQQ6/1326-8-empty-state-for-users-and-teams-without-a-roms-output


packages/react-components/src/atoms/LinkConditional.tsx
```diff
@@ -0,0 +1,10 @@
+import { ComponentProps } from 'react';
+
+import Link from './Link';
+
+const ConditionalLink: React.FC<ComponentProps<typeof Link>> = ({
+  children,
+  ...props
+}) => (props.href ? <Link {...props}>{children}</Link> : <>{children}</>);
+
+export default ConditionalLink;
```
jhonny-m, line 10: there is a mismatch between the filename and the component name.
but I didn't find where are we using this and why its necessary

packages/react-components/src/templates/UserProfileResearchOutputs.tsx
```diff
@@ -22,11 +23,35 @@ const styles = css({
...
+  hasOutputs,
+  ownUser,
+  firstName,
+  ...props
+}) => (
   <div css={styles}>
-    <SharedResearchList {...props} />
+    {hasOutputs ? (
+      <SharedResearchList {...props} />
+    ) : ownUser ? (
```
jhonny-m, line 44: I prefer to have the parameter set on a different variable, then we could have 
```
<NoOutputsPage {...(ownUser? ownProfileProps: otherProfileProps)}
```

jhonny-m: Same on the team outputs, I am not a fan of a ternary within a ternary

jhonny-m (changes requested):

packages/react-components/src/atoms/LinkConditional.tsx
kmaid: I need it in packages/react-components/src/templates/TeamProfileOutputs.tsx as we may not always have emails to create a mailto with

kmaid: I thought both components would need the link for the pm emails link but we decided against the extra query on users. I could potentially move this into TeamProfileOutputs.tsx

jhonny-m (approved):

PR #1193, opened 2021-12-06
feat: Invalidate cloudfront on deploy

markcsims:
https://trello.com/c/6ix8oOfR/1698-3-clear-fe-cache-on-every-deploy


apps/asap-server/test/handlers/invalidate-cache/index-handler.test.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  CloudFrontClient,
+  CreateInvalidationCommand,
+} from '@aws-sdk/client-cloudfront';
+import { createHandler } from '../../../src/handlers/invalidate-cache/index-handler';
+const mockSend = jest.fn();
+
+jest.mock('@aws-sdk/client-cloudfront', () => ({
```
peterstarling, line 8: i think we generally use the DI in asap-server instead of jest.mock, proxyquire etc but I guess its fine.

apps/asap-server/test/handlers/invalidate-cache/index-handler.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      .useFakeTimers('modern')
+      .setSystemTime(new Date(dateReference).getTime());
+
+    const distributionId = 'fake-distribution-id';
+    const event = {
+      Records: [],
+    };
+    const handler = createHandler(distributionId);
+    await handler(event);
+    expect(CloudFrontClient).toHaveBeenCalledWith({
```
peterstarling, line 33: i know there are different versions you can use but i still have a feeling expecting this is like saying 'the code is what it is'

In fact I would be tempted to not jest this handler at all or just a simple sanity check.

apps/asap-server/src/handlers/invalidate-cache/index-handler.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '@aws-sdk/client-cloudfront';
+import { cloudfrontDistributionId } from '../../config';
+import logger from '../../utils/logger';
+
+export const createHandler =
+  (distributionId: string) =>
+  async (event: S3Event): Promise<void> => {
+    logger.debug(`event: ${JSON.stringify(event)}`);
+    const client = new CloudFrontClient({ apiVersion: '2020-05-31' });
+    const command = new CreateInvalidationCommand({
```
peterstarling, line 14: why do you use CreateInvalidationCommand and not `CreateInvalidation` ?https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudFront.html#createInvalidation-property

serverless.ts
```diff
@@ -399,6 +404,30 @@ const serverlessConfig: AWS = {
         EVENT_SOURCE: 'asap.teams',
       },
     },
+    invalidateCache: {
+      handler:
+        'apps/asap-server/src/handlers/invalidate-cache/index-handler.handler',
```
peterstarling, line 409: index-handler? not sure if the naming is right although i dont actually have a better one

peterstarling (changes requested):
looks good, mostly just questions i need answers too after which i can accept

apps/asap-server/test/handlers/invalidate-cache/index-handler.test.ts
markcsims: It's covering that the command has the right object passed to it. The command is created before being sent. 

The test is relatively simple, it's checking the send method is called with the right command.

serverless.ts
markcsims: I wanted to call it 'apps/asap-server/src/handlers/invalidate-cache/index.handler' but all the files seem to finish with the word handler.

I'll rename it to 'apps/asap-server/src/handlers/invalidate-cache/invalidate-handler.handler'

peterstarling (approved):

PR #1195, opened 2021-12-07
Feat: Set max length on search field

kmaid:
https://trello.com/c/JTHPTPXw/1716-1-prevent-users-from-performing-searches-that-break-algolia-api

### DIFF
--- a/packages/react-components/src/molecules/SearchField.tsx
+++ b/packages/react-components/src/molecules/SearchField.tsx
@@ -29,6 +29,7 @@ const rightIndicatorStyles = css({
     alignItems: 'center',
   },
 });
+const MAX_ALGOLIA_QUERY_LENGTH = 512;
 
 type SearchProps = Pick<
   ComponentProps<typeof TextField>,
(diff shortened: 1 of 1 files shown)


packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -59,6 +59,7 @@ const SearchField: React.FC<SearchProps> = (props) => {
     <div css={clearButtonStyles}>
       <TextField
         {...props}
+        maxLength={512}
```
ana-yldio, line 62: Why are we changing the max length of every text field?

ana-yldio (changes requested):

packages/react-components/src/molecules/SearchField.tsx
kmaid: This does not effect all text fields. This effects all search fields to keep things consistent and has been approved by Beatriz

ana-yldio: Can we abstract the 512 into a var? 
Then, we can remember why are we using 512 instead of any other number

kmaid: I have made this change

ana-yldio (approved):

PR #1198, opened 2021-12-08
chore: add mock server tests to controllers

peterstarling:
https://trello.com/c/CeH5RrUB/1748-use-mock-server-for-testing-all-controllers

This should also test the failure from https://github.com/yldio/asap-hub/pull/1172

### DIFF
--- a/apps/asap-server/src/handlers/webhooks/webhook-connect-by-code.ts
+++ b/apps/asap-server/src/handlers/webhooks/webhook-connect-by-code.ts
@@ -1,9 +1,10 @@
 import Joi from '@hapi/joi';
 import { framework as lambda } from '@asap-hub/services-common';
-import { http } from '../../utils/instrumented-framework';
+import { SquidexGraphql } from '@asap-hub/squidex';
 
 import Users from '../../controllers/users';
 import { Handler } from '../../utils/types';
+import { http } from '../../utils/instrumented-framework';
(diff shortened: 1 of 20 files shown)


niezgoda (approved):

apps/asap-server/src/handlers/webhooks/webhook-sync-orcid.ts
```diff
@@ -30,7 +30,8 @@ export const handler: Handler = http(
       bodySchema,
     ) as WebhookPayload<User>;
 
-    const users = new Users(request.headers);
+    const squidexGraphqlClientMock = new SquidexGraphql();
```
markcsims, line 33: This is a real client, not a mock? I think we need to remove the suffix.

apps/asap-server/src/handlers/webhooks/webhook-connect-by-code.ts
```diff
@@ -24,7 +25,8 @@ export const handler: Handler = http(
       userId: string;
     };
 
-    const users = new Users(request.headers);
+    const squidexGraphqlClientMock = new SquidexGraphql();
```
markcsims, line 28: This is a real client, not a mock? I think we need to remove the suffix.

markcsims (changes requested):

markcsims (approved):

PR #1216, opened 2021-12-10
Seperate SLS Deploy from Package

kmaid:
https://trello.com/c/gDV2zPNE/1701-speed-up-the-deploy-job-in-ci-pipeline

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,6 +40,7 @@ workflow:
 stages:
   - build
   - test
+  - check
   - prepare-dev
   - deploy-dev
   - verify-dev
@@ -71,7 +72,7 @@ prepare:squidex:branch:
     - python ci/integration/scripts/create-app.py
     - export MOST_RECENT_BACKUP=$(aws s3api list-objects-v2 --bucket "asap-hub-dev-data-backup" --query 'sort_by(Contents, &LastModified)[-1].Key' --output=text)
     - aws s3api get-object --bucket "asap-hub-dev-data-backup" --key $MOST_RECENT_BACKUP  $MOST_RECENT_BACKUP
-    - tar -xf $MOST_RECENT_BACKUP
+    - tar -xf $MOST_RECENT_BACKUP #todo
     - source ci/env-setup.sh
     # replaces placeholders in squidex rules.
     - ./ci/config-squidex-rules.sh
@@ -133,13 +134,10 @@ build:sls:package:
     - export SQUIDEX_CLIENT_ID=$SQUIDEX_TEST_CLIENT_ID
     - export SQUIDEX_CLIENT_SECRET=$SQUIDEX_TEST_CLIENT_SECRET
     - source ci/env-setup.sh
+    - yarn rebuild # Duplicated for performance
+    - yarn build:typecheck # Duplicated for performance
+    - yarn build:babel # Duplicated for performance
     - yarn sls package
-  needs:
-    - build:ts
-    - build:native
-  dependencies:
-    - build:ts
-    - build:native
 
 build:native:
   artifacts:
@@ -160,21 +158,6 @@ build:native:
     - echo "no script property found"
   needs: []
   stage: test
-
-check:packages:
-  <<: *tmpl_test
-  script:
-    - yarn install --immutable --immutable-cache
-
-check:constraints:
-  <<: *tmpl_test
-  script:
-    - yarn constraints
-
-check:format:
-  <<: *tmpl_test
-  script:
-    - yarn run lint:format
 
 test:unit:
   <<: *tmpl_test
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -50,6 +51,42 @@ stages:
...
 # stage build #
 ###############
+prepare:squidex:branch:
+  <<: *tmpl_branch
+  image: $INTEGRATION_DOCKER_IMAGE
+  stage: build
+  environment:
+    name: review/$CI_EXTERNAL_PULL_REQUEST_IID
+    url: https://$CI_EXTERNAL_PULL_REQUEST_IID.$ASAP_HOSTNAME
+    auto_stop_in: 3 days
```
peterstarling, line 61: whats this for? i dont think we have a `stop` action that corresponds to this? (the env teardown is in github actions)

unless it has another function?

.gitlab-ci.yml
```diff
@@ -81,6 +118,27 @@ build:ts:
...
+  artifacts:
+    expire_in: 2 weeks
+    paths:
+      - '.serverless'
+  script:
+    - export SQUIDEX_APP_NAME=${CI_EXTERNAL_PULL_REQUEST_IID}
+    - export SQUIDEX_CLIENT_ID=$SQUIDEX_TEST_CLIENT_ID
+    - export SQUIDEX_CLIENT_SECRET=$SQUIDEX_TEST_CLIENT_SECRET
+    - source ci/env-setup.sh
+    - yarn rebuild # Duplicated for performance
```
peterstarling, line 137: this is to rebuild the FE app etc? maybe we should make this dependant on the other build job and have it use its artifact?

.gitlab-ci.yml
```diff
@@ -40,6 +40,7 @@ workflow:
 stages:
   - build
   - test
+  - check
```
peterstarling, line 43: by introducing another stage are we not delaying the work and making the pipeline run for longer?

.gitlab-ci.yml
```diff
@@ -221,7 +255,7 @@ verify:branch:
   needs:
     - build:ts
     - build:native
```
peterstarling, line 257: i guess these could probably can be removed?

ci/.gitlab-ci.integration.yml
```diff
@@ -7,7 +7,7 @@ variables:
 
 build:docker-images:
   image: $BASE_DOCKER_IMAGE
-  stage: .pre
+  stage: .post
```
peterstarling, line 10: this means that we are unable to ever use the image that we build in the same pipeline unless we run at at least once, is this something we want?

ci/.gitlab-ci.integration.yml
```diff
@@ -16,6 +16,7 @@ build:docker-images:
     - docker push $SQUIDEX_BASE_DOCKER_IMAGE
     - docker build -t $INTEGRATION_DOCKER_IMAGE . -f ci/integration/Dockerfile
     - docker push $INTEGRATION_DOCKER_IMAGE
+  needs: []
```
peterstarling, line 19: what do we need this for?

peterstarling (changes requested):

ci/.gitlab-ci.integration.yml
kmaid: We couldn't do that before either. This should be a seperate job for all the reasons i explained in the main channel

.gitlab-ci.yml
kmaid: No, we introduced this additional stage so we could priotize the long running tasks. They will still start as early as possible but give priority to the frontend unit tests which wasn't the case before

ci/.gitlab-ci.integration.yml
kmaid: The way to work around this would be to do any changes to the base docker image in isolation and then open a second PR with whatever you want to do with the improved image

.gitlab-ci.yml
kmaid: No, this is some PNP stuff. We wouldn't duplicate these commands if the machine allocation took less time. In our tests it was upwards of 30 seconds for 10 yarn rebuild

.gitlab-ci.yml
kmaid: Maybe. I don't think we need build:ts but we may need build:native. Build:native does the yarn rebuild. The duplicated stuff isn't persisted as an artifact. They do no harm

ci/.gitlab-ci.integration.yml
jhonny-m: Yeah, this shouldn't be a separate thing

.gitlab-ci.yml
jhonny-m: we didn't change anything in the development pipeline

peterstarling (approved):

.gitlab-ci.yml
kmaid: That too.

PR #1221, opened 2021-12-14
Feat: Speedup production pipeline

kmaid:
In this PR we are speeding up the development and production pipeline. This has been done by having the schema sync occur at the same time as the sls deploy. 

We have also moved all build steps to the start of the build. We have done this by turning our PR deployment jobs into a template and extending these templates with the environment specific configuration.

We have removed the needs condition in favour of using the default gitlab behaviour of having all prior stages jobs successfully completed to try and simplify the configuration

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,10 +41,8 @@ stages:
   - build
   - test
   - check
-  - prepare-dev
   - deploy-dev
   - verify-dev
-  - prepare-prod
   - deploy-prod
   - verify-prod
 
@@ -245,7 +243,6 @@ deploy:sls:development:
   needs:
     - build:ts
     - build:native
-    - sync:squidex:dev
 
 ####################
 # stage verify-dev #
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,9 +15,7 @@ include:
   variables: &tmpl_branch_variables
     SLS_STAGE: $CI_EXTERNAL_PULL_REQUEST_IID
 
-.tmpl:deploy: &tmpl_deploy
-  script:
-    - yarn sls deploy --verbose
+.tmpl:deploy:
   variables: &tmpl_deploy_variables
     NODE_ENV: production
     NODE_OPTIONS: '--max_old_space_size=4096'
@@ -63,7 +61,6 @@ stages:
     - export REACT_APP_SENTRY_DSN=$REACT_APP_SENTRY_DSN
   script:
     - yarn run build
-  needs: []
   stage: build
 
 build:ts:branch:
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -210,42 +240,62 @@ check:format:
...
   <<: *tmpl_branch
   image: $INTEGRATION_DOCKER_IMAGE
   stage: deploy-dev
+  script:
+    - source ci/env-setup.sh
+    - yarn sls deploy --package .serverless --verbose
+
+deploy:sls:branch:
+  extends: .deploy:sls
+  rules:
```
peterstarling, line 270: should we not use the `tmpl_branch` template? some of the jobs for the PR pipeline use it while others (like this one) just have the rule

.gitlab-ci.yml
```diff
@@ -210,42 +240,62 @@ check:format:
...
     <<: *tmpl_branch_variables
     <<: *tmpl_deploy_variables
-  script:
+  before_script:
     - export SQUIDEX_APP_NAME=${CI_EXTERNAL_PULL_REQUEST_IID}
     - export SQUIDEX_CLIENT_ID=$SQUIDEX_TEST_CLIENT_ID
     - export SQUIDEX_CLIENT_SECRET=$SQUIDEX_TEST_CLIENT_SECRET
-    - source ci/env-setup.sh
-    - yarn sls deploy --package .serverless --verbose
   dependencies:
```
peterstarling, line 279: i think this has been here all along (previously the branch template would also specify these are dependencies) - but what if we start deploying to PR before finishing all tests? If this might optimise the pipeline i guess we could try that, unless the tests are not really on critical path anyway in which case it makes sense to keep it simple and wait for the entire stage to finish.

peterstarling (changes requested):
these are mostly comments/questions than hints.

What you should add is a meaningful description of the changes - the previous ones were already difficult to understand and a reference to the ticket may not suffice.

Also it would be great to see the pipeline graph but i guess we wont get it until we merge.

.gitlab-ci.yml
kmaid: I have very mixed feelings about the tmp_branch template. We found it incredibly difficult to read and have it in our list of improvements to tackle. Maybe we could give them more descriptive names?

.gitlab-ci.yml
kmaid: Dependencies are different to needs. Dependecies say that we want to consume an artifact. The ommitted needs is what says all previous stages are required to be successful

.gitlab-ci.yml
peterstarling: yeah sounds good, we could either get rid of the templates entirely all use the same ones for every PR/dev/prod job, just to be able to tell them apart and not mix these up

peterstarling (approved):

.gitlab-ci.yml
peterstarling: i am well aware, which is why i said if we should use needs instead and have it wait for just the selected jobs since we dont need all of them

PR #1233, opened 2021-12-15
Feat: Add pipeline id, job id and when it was triggered into index.html

kmaid:
No Ticket,

Include the build version in the HTML meta


kmaid:
image

peterstarling (approved):

PR #1236, opened 2021-12-17
chore: Remove instrumentation and add logging

peterstarling:
https://trello.com/c/DgUU3RPu/1436-5-replace-instrumentation-code-from-asap-server-with-a-logger

Notable changes:
* removes InstrumentedSquidexGraphql client
* adds Squidex response to error message for when it fails to fetch the auth token from Squidex
* fixes the bug whereby a single instance of an error occurring whilst getting the token resulted in consecutive failures until the app was restarted
* moves token caching to a different file
* makes the get-access-token function testable
*


ana-yldio (approved):

packages/squidex/.eslintrc.js
```diff
@@ -3,5 +3,6 @@ module.exports = {
   rules: {
     'lines-between-class-members': 'off',
     'no-unused-vars': 'off',
+    'max-classes-per-file': 'off',
```
niezgoda, line 6: In the future, we should enable this feature.
And for the number of lines of code :P

packages/squidex/src/auth.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (error instanceof RequestError) {
+      error.message = `Request to Squidex failed (code ${error.code}), response: ${error.response?.body}`;
+    }
+
+    throw error;
+  }
+};
+
+export const getAccessTokenFactory = (): (() => Promise<string>) => {
+  let tokenP: Promise<string> | undefined;
```
niezgoda, line 34: What tokenP means? Token Promise?
Storage is not expensive nowadays 😸

packages/squidex/src/auth.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return async (): Promise<string> => {
+    if (tokenP) {
+      try {
+        const tk1 = await tokenP;
```
niezgoda, line 39: this variables are weird

niezgoda: naming I mean

niezgoda (approved):

PR #1242, opened 2021-12-20
Feat/Algolia works when the user session has timed out

niezgoda:
## Reproduction steps:

- log in
- let your session time out (leave your tab open for 10 hours)
- navigate the hub

(this is only reproducible before https://trello.com/c/BZEGD3jD/1594-users-are-able-to-navigate-the-hub-even-if-they-are-logged-out  is merged)

## current behaviour:

- most pages of the hub are broken
- shared outputs work

## expected behaviour:

- shared outputs don't work (because algolia token should expire when Auth0 expires)

## notes

- a technical should be able to product accept this

### DIFF
--- a/apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
+++ b/apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
@@ -6,7 +6,10 @@ import {
   auth0SharedSecret as secret,
 } from '../../../../src/config';
 import { identity } from '../../../helpers/squidex';
-import { fetchUserByCodeHandlerFactory } from '../../../../src/handlers/webhooks/fetch-by-code/fetch-by-code';
+import {
+  fetchUserByCodeHandlerFactory,
+  getValidUntil,
+} from '../../../../src/handlers/webhooks/fetch-by-code/fetch-by-code';
 import { SearchClient } from 'algoliasearch';
 import { userControllerMock } from '../../../mocks/user-controller.mock';
 import { getUserResponse } from '../../../fixtures/users.fixtures';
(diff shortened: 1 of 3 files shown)


apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
```diff
@@ -20,6 +23,11 @@ describe('Fetch-user-by-code handler', () => {
     algoliaClientMock,
   );
 
+  describe('Check if algolia validUntil is in correct format(seconds)', () => {
```
peterstarling, line 26: I would actually mock/spy the `algoliaClient.generateSecuredApiKey` to see what it is being called with

peterstarling (changes requested):

apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
niezgoda: And we would have to mock everything :)
I have done this to do this in FP manner :)

peterstarling: thats fine, but I don't think the test actually proves that the issue was fixed. I don't mind FP but we need to make sure that `generateSecuredApiKey` was called with the correct expiration Date. With your test I could change this line: https://github.com/yldio/asap-hub/blob/2a0f349e0e2a496750cd5383d3e5f92965fe7e6f/apps/asap-server/src/handlers/webhooks/fetch-by-code/fetch-by-code.ts#L34-L36

to 
```typescript 
validUntil: 123124
```
which is pretty much where we were before when the only test we had was this https://github.com/yldio/asap-hub/blob/5f1253e5f9d4b0f19651a1962e91108a27aca7ba/apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts#L142

peterstarling (changes requested):

apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
peterstarling: and to make it more clear - the way to test such bug is to write a test that fails, then fix it and make sure it goes green

jhonny-m (approved):

ana-yldio (approved):

PR #1246, opened 2021-12-20
fix: Token expiration check

peterstarling:
Fixes the issue whereby a new token was generated each time due to an invalid expiration check

### DIFF
--- a/packages/squidex/src/auth.ts
+++ b/packages/squidex/src/auth.ts
@@ -41,7 +41,7 @@ export const getAccessTokenFactory = (): (() => Promise<string>) => {
           const jwt = decode<JwtToken>(tk1);
           const currentTime = Date.now() / 1000;
 
-          if (currentTime > jwt.exp) {
+          if (currentTime < jwt.exp) {
             return tk1;
           }
         }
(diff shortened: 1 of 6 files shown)


packages/squidex/src/auth.ts
```diff
@@ -41,7 +41,7 @@ export const getAccessTokenFactory = (): (() => Promise<string>) => {
           const jwt = decode<JwtToken>(tk1);
           const currentTime = Date.now() / 1000;
 
-          if (currentTime > jwt.exp) {
+          if (currentTime < jwt.exp) {
```
kmaid, line 44: ooof.

kmaid: <=?

niezgoda: Do we really care about this one second? :D

packages/squidex/test/auth.test.ts
```diff
@@ -40,6 +36,57 @@ describe('Get Access Token', () => {
...
+        )
+        .reply(200, {
+          access_token: mockTokenFresh,
+          expires_in: Math.floor(oneHourFromNow.getTime() / 1000),
+          token_type: 'Bearer',
+          scope: 'squidex-api',
+        });
+
+      await getAccessToken();
+      await getAccessToken();
```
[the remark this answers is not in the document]
peterstarling: yeah thats the whole point. Should only call once when invoked twice

niezgoda (approved):

PR #1255, opened 2021-12-21
feat: update ros schema

ana-yldio:
https://trello.com/c/QPF8nBmH/1786-make-addeddate-mandatory

Migration test: [Before](https://gist.github.com/ana-yldio/91258d2cc47c1384badc8d98aa0994ff)/[After](https://gist.github.com/ana-yldio/892ade19f34453954a12b5e6245fa83a
)


apps/asap-server/src/migrations/1640102001540-set-ros-added-date-default.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class SetResearchOutputAddedDateDefault extends Migration {
+  up = async (): Promise<void> => {
+    const squidexClient = new SquidexRest<RestResearchOutput>(
+      'research-outputs',
+    );
+
+    let pointer = 0;
+    let result: Results<RestResearchOutput>;
+
+    do {
```
peterstarling, line 14: there is a helper function that does this for you, see: https://github.com/yldio/asap-hub/blob/98c32d3efe7b615b8466f4f6f093727b54082741/apps/asap-server/src/utils/migrations.ts#L4

you may wanna consider using that as it has been used before. We've had issues with other attempts at looping through collections which is why I think its quite useful to stick to that :D

peterstarling (changes requested):
looks good in general, but I would change the migration as per my review and test it - can you post test evidence here when you do? Like a screenshot of before and after on a single page, or better, a query to get ROs with empty dateAdded, before and after

peterstarling (approved):

PR #1262, opened 2022-01-05
feat: Remove inactive group calendars

peterstarling:
Removes inactive calendars from the `/calendars` endpoint.

## Notable changes:
- disables pagination for the calendar endpoint
- migrates REST calendar fetch to GraphQL

see: https://trello.com/c/HhKZss3v/1781-admins-can-mark-groups-as-inactive-and-omit-them-from-calendars-page


apps/asap-server/src/entities/group.ts
```diff
@@ -25,7 +25,7 @@ export const parseGraphQLGroup = (
     return team;
   });
   const calendars: CalendarResponse[] = (item.flatData?.calendars || []).map(
-    (c) => parseGraphQLCalendar(c as Calendars), // @todo remove cast
+    (c) => parseGraphqlCalendar(c as Calendars), // @todo remove cast
```
kmaid, line 28: Can we remove this cast now?

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -30,21 +32,50 @@ export default class Calendars implements CalendarController {
...
 
-  async fetch(options: {
-    take: number;
-    skip: number;
-  }): Promise<ListCalendarResponse> {
-    const { take = 50, skip = 0 } = options;
-    const { total, items: calendars } =
-      await this.calendarSquidexRestClient.fetch({
-        take,
+  async fetch(): Promise<ListCalendarResponse> {
```
niezgoda, line 35: When you have basically refactored this code - why have you not moved the business logic (with specific technologies/API behind) into the proper class or file - like move it away from controller?

apps/asap-server/src/controllers/calendars.ts
```diff
@@ -30,21 +32,50 @@ export default class Calendars implements CalendarController {
...
+      return {
+        total: 0,
+        items: [],
+      };
+    }
+
+    const activeCalendars = calendars.filter(
+      (calendar) =>
+        calendar.referencingGroupsContents &&
+        calendar.referencingGroupsContents.findIndex(
```
niezgoda, line 71: This wouldn't work?
calendar.referencingGroupsContents?.findIndex 
so you could remove "calendar.referencingGroupsContents &&" part

niezgoda (approved):

apps/asap-server/src/controllers/calendars.ts
peterstarling: good question! I think most of the logic belongs to this class anyway. The controller wraps stuff like declaring queries and getting the data back, the only thing that is extracted (and was) is the parsing logic.

Mind you - we do not have CONTRIBUTING guide that would actually specify which type of logic should belong to controller and what should be moved up, but there is a ticket for it: https://trello.com/c/yCshrAnG/1666-split-be-controllers-into-2-layers

apps/asap-server/src/controllers/calendars.ts
peterstarling: actually it wouldn't - the part before `?` would resolve to undefined which is not equal to `-1` and would therefore pass the condition, but the requirements are that a calendar without a group (ie empty array or no reference to groups at all) should be skipped.

niezgoda: So you could use sth like calendar.referencingGroupsContents?.findIndex(...) ?? false or so :)

PR #1265, opened 2022-01-06
Feat/1787 Add new section on "Discover ASAP"

niezgoda:
## requirements:

on Discover ASAP page https://dev.hub.asap.science/discover, a new section should appear

- create a new field on Discover ASAP **schema** called "Scientific Advisory Board"
 - this field is shown on the CMS after "Members"
 - with the same requirements as "Members" (i.e. it is possible to associate users)
 - nice to have: update the label from "Members" to "ASAP Team"
- return this data to FE
- add a **new section** to Discover ASAP **page** called "Meet the Scientific Advisory Board", **below** "Meet the ASAP team"
 - the section should look and behave exactly the same, only with a different header, namely
 - show a list of users which come from the CMS (Scientific Advisory Board, within Discover ASAP schema)
 - omit if there are 0 users on the list
 - ~~for each user, show an avatar and 3 lines of text (1st line: Name, 2nd: Position, 3rd: Institution)~~ 
 - ~~all lines should truncate if they exceed 1 line of content~~

- header of the new section is "Meet the Scientific Advisory Board"
- update the header of the other section from "Meet the ASAP team" to "Meet the Team"


peterstarling (changes requested):
which test in BE shows the response with the extra `scientificAdvisoryBoard` property? you've added some empty arrays to fixtures but have you tested that once the its returned from squidex it is parsed and return from the controller properly?

FYI I would imagine this is where it would get tested:
https://github.com/yldio/asap-hub/blob/215ed9a14eaccca639891a6e555968222934f6d8/apps/asap-server/test/controllers/discover.test.ts#L107
so your expectation would be here: https://github.com/yldio/asap-hub/blob/215ed9a14eaccca639891a6e555968222934f6d8/apps/asap-server/test/fixtures/discover.fixtures.ts#L176

niezgoda:
@peterstarling this is WIP :)
Unfortunately I have recreated the PR and forget to add the WIP label :(

PR #1268, opened 2022-01-10
Feat/1822 Create Squidex availability metric

niezgoda:
We want to create metric for when squidex responds with: 

- 5XX 
- 4XX
- GraphQL errors

## Requirements

- have a visual representation of squidex downtime (see list above) along with duration and frequency


apps/asap-server/src/autogenerated-gql/graphql.ts
```diff
@@ -4231,17 +4231,13 @@ export type UsersDataSocialInputDto = {
 /** The structure of the Teams nested schema. */
 export type UsersDataTeamsChildDto = {
   id: Maybe<Array<Teams>>;
-  mainResearchInterests: Maybe<Scalars['String']>;
-  responsibilities: Maybe<Scalars['String']>;
   /** Attention: Check if this user needs to be added to Smart Simple */
```
kmaid, line 4234: Why is this in this PR

kmaid (changes requested):

apps/asap-server/src/autogenerated-gql/graphql.ts
niezgoda: ask Piotr :D

peterstarling (changes requested):
As I mentioned to your earlier - I think the dashboard definition should be part of the sls config and should be applied to all environments, it can be easily imported especially since our sls config is a ts file. 

You mentioned some limits though? If there is a limit it can be applied conditionally to dev and prod only

peterstarling:
@niezgoda are we closing this?

PR #1272, opened 2022-01-11
Feat/1824 always show calendars which have no group

niezgoda:
missed requirement from this ticket: https://trello.com/c/HhKZss3v/1781-admins-can-mark-groups-as-inactive-and-omit-them-from-calendars-page .

## requirement

- when a calendar is NOT associated to any group, then default to the behaviour of a calendar associated to an active group (therefore always show the calendar).

## Expected behaviour:
Calendars without a group are treated as a calendar with group=active (therefore are shown in Calendar Page)

## Current behaviour
Calendars without a group are treated as a calendar with group=inactive (therefore are hidden in Calendar Page)

### DIFF
--- a/apps/asap-server/test/controllers/calendars.test.ts
+++ b/apps/asap-server/test/controllers/calendars.test.ts
@@ -153,13 +153,21 @@ describe('Calendars controller', () => {
         });
       });
 
-      test('Should skip the calendars which do not belong to any group', async () => {
+      test('Should show the calendars which do not belong to any group', async () => {
         const squidexGraphqlResponse = getSquidexCalendarsGraphqlResponse();
 
         const calendar1 = getSquidexGraphqlCalendar();
+        calendar1.flatData.googleCalendarId = 'calendar1@google.com';
         calendar1.referencingGroupsContents = null;
         const calendar2 = getSquidexGraphqlCalendar();
+        calendar2.flatData.googleCalendarId = 'calendar1@google.com';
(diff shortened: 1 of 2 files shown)


apps/asap-server/test/controllers/calendars.test.ts
```diff
@@ -153,14 +153,22 @@ describe('Calendars controller', () => {
...
 
-      test('Should skip the calendars which do not belong to any group', async () => {
+      test('Should show the calendars which do not belong to any group', async () => {
         const squidexGraphqlResponse = getSquidexCalendarsGraphqlResponse();
 
         const calendar1 = getSquidexGraphqlCalendar();
+        calendar1.flatData.googleCalendarId = 'calendar1@google.com';
         calendar1.referencingGroupsContents = null;
         const calendar2 = getSquidexGraphqlCalendar();
+        calendar2.flatData.googleCalendarId = 'calendar1@google.com';
```
peterstarling, line 163: i think if you update the googleCalendarId to the same value for each of the calendars then it doesnt really make much sense anyway does it?

peterstarling (approved):
small comment but looks good

apps/asap-server/test/controllers/calendars.test.ts
niezgoda: because there is maybe in the type - if you do not assign the googleCalendarId it shows that there could be an issue :P

PR #1273, opened 2022-01-12
chore/clean up user fixtures

markcsims:
https://trello.com/c/sAKTLxnQ/1693-clean-up-be-fixtures-users

- starts adding mock client
- uses mock client on fetch users
- updates fetchByCode to use client mock
- updates fetchById to use client mock
- updates update to use client mock
- syncOrcidProfile and connectByCode don't use graphql
- updates updateAvatar to use client mock
- removes getGraphqlResponseFetchUser
- renames usersGraphqlClient
- removes unused fixtures
- fixture generators
- refactor response generator


apps/asap-server/test/controllers/users.test.ts
```diff
@@ -32,86 +30,52 @@ describe('Users controller', () => {
...
     describe('with mock-server', () => {
       test('Should fetch the users from squidex graphql', async () => {
-        const result = await usersMockGraphql.fetch({});
+        const result = await usersMockGraphqlServer.fetch({});
 
         expect(result).toMatchObject(getListUserResponse());
       });
     });
 
     describe('with intercepted http layer', () => {
```
peterstarling, line 39: just so you know the reason i split the mock-server and 'intercepted http layer' in other test files was because of nock and having to add identity and nock expectations in beforeAll and afterEach, if `Fetch` method is tested without nock then you might not need this split - checkout  https://github.com/yldio/asap-hub/blob/93ac40b96a40f4a0ea366b17b1800f1661016e31/apps/asap-server/test/controllers/research-outputs.test.ts#L32 as an example

apps/asap-server/test/fixtures/teams.fixtures.ts
```diff
@@ -251,7 +256,13 @@ export const getGraphqlTeam = (
     ],
     tools,
   },
-  referencingUsersContents: [getGraphQLUser(undefined, id)],
+  referencingUsersContents: [
+    getGraphQLUser({
+      flatData: {
+        teams: [getGraphQLUserTeam({ id: [getGraphQLUserTeamId({ id })] })],
+      } as GraphQLUserTeamFlatData,
```
peterstarling, line 263: you shouldnt need the cast?

apps/asap-server/test/fixtures/users.fixtures.ts
```diff
@@ -6,97 +6,91 @@ import {
...
+  generateGraphqlFetchUsersResponse(
+    [...Array(total)]
+      .fill(getGraphQLUser())
+      .map((_, index) => getGraphQLUser({ id: `user-id-${index + 1}` })),
+  );
+export const getSquidexUserGraphqlResponse = (): FetchUserQuery => ({
+  findUsersContent: getGraphQLUser(),
+});
+
+const generateGraphqlFetchUsersResponse = (
```
peterstarling, line 21: I'd stick to the `get....` naming convention

apps/asap-server/test/fixtures/users.fixtures.ts
```diff
@@ -6,97 +6,91 @@ import {
...
-                linkedIn: null,
-                researcherId: null,
-                researchGate: null,
-                twitter: null,
-                website1: null,
-                website2: null,
-              },
-            ],
-            teams: [
+export const getSquidexUsersGraphqlResponse = (
```
peterstarling, line 9: do you need you generators (this one and the ones below) to really accept the input args? Not that it is invalid, i just think it would be nice to keep the fixtures clean, use nested ones but never have to alter them too much. 

What I would normally recommend is that any changes to fixtures for the tests should be done after the data is generated, inside the test, through the update of properties, like this: https://github.com/yldio/asap-hub/blob/93ac40b96a40f4a0ea366b17b1800f1661016e31/apps/asap-server/test/controllers/calendars.test.ts#L50-L51

I find it much easier to read, but it is possible that I am not seeing the exact reason why you have this here

apps/asap-server/test/fixtures/users.fixtures.ts
```diff
@@ -6,97 +6,91 @@ import {
...
+  );
+export const getSquidexUserGraphqlResponse = (): FetchUserQuery => ({
+  findUsersContent: getGraphQLUser(),
+});
+
+const generateGraphqlFetchUsersResponse = (
+  items: NonNullable<FetchUserQuery['findUsersContent']>[],
+): FetchUsersQuery => ({
+  queryUsersContentsWithTotal: {
+    total: items.length,
```
peterstarling, line 25: to add up to the two above comments - I would have a list fixture compose of a single element in an array and a total of 1, like this one: https://github.com/yldio/asap-hub/blob/93ac40b96a40f4a0ea366b17b1800f1661016e31/apps/asap-server/test/fixtures/calendars.fixtures.ts#L159-L161

. Should you need more you could just change it in the test (see the comment at the top of this file).

peterstarling (changes requested):
the review is mostly just about an specific approach which may vary, your way is totally fine but I thought maybe we would like to make the fixtures simpler and align them with the other ones. Feel free to dismiss and I will approve

apps/asap-server/test/controllers/users.test.ts
markcsims: I've removed the describe('with mock-server' and the describe('with intercepted http layer'. It wasn't clear to me, I thought they were to distinguish between mocking the server and the client.

apps/asap-server/test/fixtures/teams.fixtures.ts
markcsims: Yes. getGraphQLUser takes a partial NonNullable<FetchUserQuery['findUsersContent']> - from my understand partial only works on a shallow level. Here we want to pass in a Partial flatData.

apps/asap-server/test/fixtures/users.fixtures.ts
markcsims: Named it differently as it's not an exported function. Its purpose is to be a helper function generating the boilerplate around the structure of the response.

Happy to change it, but wanted it to be different to the 2 methods being called to get the fixtures.

Let me know.

apps/asap-server/test/fixtures/users.fixtures.ts
markcsims: I prefer not to mutate objects/data. I prefer seeing getSquidexUsersGraphqlResponse({ total: 0 }) or getGraphQLUser({ id: 'some-new-id'})


Let's have a call to discuss?

apps/asap-server/test/fixtures/users.fixtures.ts
markcsims: See above. Lets discuss 👍

markcsims: It's the same as this: https://github.com/yldio/asap-hub/blob/d4b4ecbfe2ee2e204cc527965fbcbc2b44ea322a/apps/asap-server/test/fixtures/users.fixtures.ts#L9-L16
But this allows multiple items to be generated, as required by one of the tests.

peterstarling (approved):

PR #1278, opened 2022-01-13
feat: ros listing default sorting

ana-yldio:
https://trello.com/c/hsxrZAtm/1770-2-sort-algolia-listings-when-no-search-has-been-performed

### DIFF
--- a/apps/frontend/src/hooks/algolia.ts
+++ b/apps/frontend/src/hooks/algolia.ts
@@ -11,7 +11,7 @@ import {
 
 export const INDEX = {
   primary: ALGOLIA_INDEX,
-  'asc(addedDate)': ALGOLIA_INDEX_REPLICA,
+  'desc(addedDate)': ALGOLIA_INDEX_REPLICA,
 };
 
 export const useAlgolia = () => {
--- a/apps/frontend/src/hooks/algolia.ts
+++ b/apps/frontend/src/hooks/algolia.ts
@@ -11,7 +11,7 @@ import {
 
 export const INDEX = {
   primary: ALGOLIA_INDEX,
-  'desc(addedDate)': ALGOLIA_INDEX_REPLICA,
+  ROSListing: ALGOLIA_INDEX_REPLICA,
 };
 
 export const useAlgolia = () => {
--- a/apps/frontend/src/hooks/algolia.ts
+++ b/apps/frontend/src/hooks/algolia.ts
@@ -3,7 +3,16 @@ import algoliasearch from 'algoliasearch/lite';
 import { useCurrentUser } from '@asap-hub/react-context';
 import { User } from '@asap-hub/auth';
 
-import { ALGOLIA_APP_ID, ALGOLIA_INDEX } from '../config';
+import {
+  ALGOLIA_APP_ID,
+  ALGOLIA_INDEX,
+  ALGOLIA_INDEX_REPLICA,
+} from '../config';
+
+export const INDEX = {
+  primary: ALGOLIA_INDEX,
+  'asc(addedDate)': ALGOLIA_INDEX_REPLICA,
+};
 
 export const useAlgolia = () => {
   const initAlgolia = (user: User | null) => {
--- a/apps/frontend/src/shared-research/ResearchOutputList.tsx
+++ b/apps/frontend/src/shared-research/ResearchOutputList.tsx
@@ -27,7 +27,7 @@ const ResearchOutputList: React.FC<ResearchOutputListProps> = ({
     filters,
     currentPage,
     pageSize,
-    searchIndex: !searchQuery.length ? INDEX['desc(addedDate)'] : INDEX.primary,
+    searchIndex: !searchQuery.length ? INDEX.ROSListing : INDEX.primary,
   });
 
   const { client } = useAlgolia();
@@ -45,7 +45,7 @@ const ResearchOutputList: React.FC<ResearchOutputListProps> = ({
       (paginationParams) =>
         getResearchOutputs(
           client.initIndex(
-            !searchQuery.length ? INDEX['desc(addedDate)'] : INDEX.primary,
+            !searchQuery.length ? INDEX.ROSListing : INDEX.primary,
           ),
           {
             filters,
--- a/apps/frontend/src/shared-research/ResearchOutputList.tsx
+++ b/apps/frontend/src/shared-research/ResearchOutputList.tsx
@@ -1,10 +1,9 @@
 import { SharedResearchList } from '@asap-hub/react-components';
 import { sharedResearch } from '@asap-hub/routing';
 import { format } from 'date-fns';
-
 import { useResearchOutputs } from './state';
 import { usePaginationParams, usePagination } from '../hooks';
-import { useAlgolia } from '../hooks/algolia';
+import { INDEX, useAlgolia } from '../hooks/algolia';
 import { getResearchOutputs } from './api';
 import {
   createCsvFileStream,
@@ -28,8 +27,10 @@ const ResearchOutputList: React.FC<ResearchOutputListProps> = ({
     filters,
     currentPage,
     pageSize,
+    searchIndex: !searchQuery.length ? INDEX['asc(addedDate)'] : INDEX.primary,
   });
-  const { index } = useAlgolia();
+
+  const { client } = useAlgolia();
 
   const { numberOfPages, renderPageHref } = usePagination(
     result?.total || 0,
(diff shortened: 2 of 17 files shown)


apps/frontend/src/hooks/algolia.ts
```diff
@@ -3,7 +3,16 @@ import algoliasearch from 'algoliasearch/lite';
...
-import { ALGOLIA_APP_ID, ALGOLIA_INDEX } from '../config';
+import {
+  ALGOLIA_APP_ID,
+  ALGOLIA_INDEX,
+  ALGOLIA_INDEX_REPLICA,
+} from '../config';
+
+export const INDEX = {
+  primary: ALGOLIA_INDEX,
+  'desc(addedDate)': ALGOLIA_INDEX_REPLICA,
```
niezgoda, line 14: I don't like the name of this field: "desc(addedDate)".
I have never seen such convention,

apps/frontend/src/shared-research/ResearchOutputList.tsx
```diff
@@ -28,8 +27,10 @@ const ResearchOutputList: React.FC<ResearchOutputListProps> = ({
     filters,
     currentPage,
     pageSize,
+    searchIndex: !searchQuery.length ? INDEX['desc(addedDate)'] : INDEX.primary,
```
niezgoda, line 30: you should have one place - one function which is going to return proper index based on search query.
Now you have multiple (two?) locations to make such changes. It should always be just one :)

niezgoda (changes requested):

apps/frontend/src/shared-research/ResearchOutputList.tsx
ana-yldio: It is necessary in this case because the algolia object is initialized in two different places.

https://github.com/yldio/asap-hub/blob/d3d3653cb75b040b9167ae377a7b70ed03cb5ef8/apps/frontend/src/shared-research/state.ts#L108

https://github.com/yldio/asap-hub/blob/d3d3653cb75b040b9167ae377a7b70ed03cb5ef8/apps/frontend/src/shared-research/ResearchOutputList.tsx#L33

ci/algolia-schema.json
```diff
@@ -20,7 +20,7 @@
   "unretrievableAttributes": null,
   "optionalWords": null,
   "queryLanguages": ["en"],
-  "replicas": [],
+  "replicas": ["ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA"],
```
peterstarling, line 23: if we have this setting here, do we still need to manually create and upload data to the replica? what is the purpose of this then?

peterstarling (changes requested):

ci/algolia-schema.json
ana-yldio: The purpose is to add the name of the replica to the primary index. 
`ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA` is a placeholder, that will be replaced with the value of the var `$ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA`
https://github.com/yldio/asap-hub/blob/70b4f295a89ecacf55137e26baa3007b3862afab/package.json#L19
https://github.com/yldio/asap-hub/blob/70b4f295a89ecacf55137e26baa3007b3862afab/.github/workflows/env-sync.yml#L23
https://github.com/yldio/asap-hub/blob/70b4f295a89ecacf55137e26baa3007b3862afab/.github/workflows/master-sync.yml#L25

kmaid: Shouldn't we use a better name. Its clear that its a replica what isn't clear is why the replica is different to the normal result set

ana-yldio: What name would you suggest?

kmaid: My first thought was:
ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA_<fieldname>_<asc/desc>

However Algolia is redundant, Research output no longer reflects the contents of the index, Should we be calling a replica an index? 

What do we think about 
ASAP_INDEX_REPLICA_<fieldname>_<asc/desc>?

peterstarling: I understand the placeholder, what i dont understand is this - we have a file `algolia-schema.json`, yeah? what is it being used for? before your changes?

ana-yldio: I don't think it would work. This naming convention you are proposing works if there's only one attribute. 
i.e. `ASAP_INDEX_REPLICA_addedData_desc`

But, the replica can be ranked by one or more fields. Then we can end up with names like `ASAP_INDEX_REPLICA_addedData_desc_title_asc`  or `ASAP_INDEX_REPLICA_addedDataDesc_titleAsc` 

Perhaps something like `ASAP_INDEX_REPLICA_section_REPLICA` works better. 
i.e. `ASAP_INDEX_REPLICA_ROSLISTING_REPLICA`

What do you think @kmaid ?

ana-yldio: > I understand the placeholder, what i dont understand is this - we have a file `algolia-schema.json`, yeah? what is it being used for? before your changes?

@peterstarling, sorry I misunderstood your comment.  
`algolia-schema.json` defines the config for the primary index. Before this PR, the config file looks like this
https://github.com/yldio/asap-hub/blob/master/ci/algolia-schema.json#L23

This PR adds a replica to the primary index
https://github.com/yldio/asap-hub/blob/70b4f295a89ecacf55137e26baa3007b3862afab/ci/algolia-schema.json#L23

And it also defines a different config for the replica. We cannot use the same config because the ranking is different
https://github.com/yldio/asap-hub/blob/70b4f295a89ecacf55137e26baa3007b3862afab/ci/algolia-replica-schema.json#L2

peterstarling: ok my questions is this - do we need this here? and if so, with just this config doesnt it automatically create the replica? this is what i am getting at - i would imagine that when you sync it would create the replica too but i might be wrong

ana-yldio: That's a good point! I'll chat with you offline

ci/config-algolia-replicas.sh
```diff
@@ -0,0 +1,9 @@
+# Uses dir passed as argument or default one
+backupPath=${1:-"./ci"}
+
+# Replace PLACEHOLDERS with variable available on the CI
+find $backupPath \
```
kmaid, line 5: What does this do? It seems like it backs up replica data

package.json
```diff
@@ -16,7 +16,9 @@
   },
   "scripts": {
     "algolia:download": "algolia getsettings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RESEARCH_OUTPUT_INDEX | prettier --parser json > ./ci/algolia-schema.json",
+    "prealgolia:upload": "sh ./ci/config-algolia-replicas.sh",
     "algolia:upload": "algolia setsettings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RESEARCH_OUTPUT_INDEX -s ./ci/algolia-schema.json",
```
kmaid, line 20: Can we change this so we only need to run algolia:upload? All other commands should be started by this command if we are going to need a few for other rankings

ci/config-algolia-replicas.sh
ana-yldio: This script replaces the string `ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA` with the value of the var `$ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA`

package.json
ana-yldio: I think we need the three commands. But happy to hear suggestions:) 
Before running `algolia:upload` we need to inject the value of the var `$ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA` into `algolia-schema.json`.  
And then, we need to load the config of the replica.

ci/config-algolia-replicas.sh
kmaid: So we are mutating this file as part of the build process?

package.json
```diff
@@ -16,7 +16,9 @@
...
+    "algolia:upload:replica": "algolia setsettings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA -s ./ci/algolia-replica-schema.json",
```
kmaid, line 21:
```suggestion
 "algolia:upload": "yarn algolia:upload:index && yarn algolia:upload:replica ",
    "algolia:upload:index": "algolia setsettings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RESEARCH_OUTPUT_INDEX -s ./ci/algolia-schema.json",
    "algolia:upload:replica": "sh ./ci/config-algolia-replicas.sh && algolia setsettings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RESEARCH_OUTPUT_INDEX_REPLICA -s ./ci/algolia-replica-schema.json",

```

kmaid: Look at build for an example. I think this is what is required

ci/config-algolia-replicas.sh
kmaid: Can't we just specify the index name on the algolia CLI using -n? I would prefer not to generate new config as part of the build process

kmaid (changes requested):

ana-yldio:
https://github.com/yldio/asap-hub/pull/1286

PR #1279, opened 2022-01-14
updating user profile for staff role

jhonny-m:
https://trello.com/c/PU7bgepN/1788-update-template-for-users-with-staff-role


packages/react-components/src/organisms/UserProfileRole.tsx
```diff
@@ -131,6 +137,21 @@ const UserProfileRole: React.FC<UserProfileRoleProps> = ({
           )}
         </div>
       )}
+      {(reachOut || isOwnProfile) && role === 'Staff' && (
```
ana-yldio, line 140: Can we remove `reachOut` from this condition as it later checks for `reachOut` on line 145?

```suggestion
      {isOwnProfile && role === 'Staff' && (
```

jhonny-m: It wouldn't show `reachOut` if  `isOwnProfile` is false

packages/react-components/src/organisms/UserProfileRole.tsx
```diff
@@ -11,10 +11,14 @@ import { getUniqueCommaStringWithSuffix } from '../utils';
...
-> & {
-  labs?: Lab[];
-};
+  | 'firstName'
+  | 'teams'
+  | 'researchInterests'
+  | 'responsibilities'
+  | 'role'
+  | 'reachOut'
+  | 'labs'
```
kmaid, line 20: 👍

PR #1281, opened 2022-01-14
Feat/1813 Get Algolia ready to preform two different searches

niezgoda:
We will soon expand our `shared-research` index in Algolia to contain `users` too. This will create problems in `/shared-research`. This task will tackle that problem before it emerges

## requirements 

- make Algolia capable of having different types of entities in a single index
- `/shared-research` search should only return results of the Shared Outputs entity

---

## test scenario:

- create an item which is NOT part of Shared Outputs entity 
- that item should not be returned in the search results in `/shared-research`

---

## notes:

- make sure to re-index production Algolia
- specify type on `/shared-research` to only search among `shared outputs` (and ignore `users`, if there are any)


peterstarling (changes requested):
because the type for Algolia ROs is not declared separately nor is it actually being used in either saving
https://github.com/yldio/asap-hub/blob/c248b2c63ab1797c4c04c499af30397686a036c7/apps/asap-server/src/handlers/research-output/index-handler.ts#L41
or restoring
https://github.com/yldio/asap-hub/blob/c248b2c63ab1797c4c04c499af30397686a036c7/apps/frontend/src/shared-research/api.ts#L27
you have now added a new field to research-outputs in Algolia and you forgot to update the indexers (the field is irrelevant when fetching the data though)

I think we should declare a new, `ResearchOutputAlgoliaResponse` type and make sure it is being used when saving https://github.com/yldio/asap-hub/blob/c248b2c63ab1797c4c04c499af30397686a036c7/apps/asap-server/src/handlers/research-output/index-handler.ts#L41 so if the type changes we remember to update it

packages/algolia/src/indexes/research-output.ts
```diff
@@ -0,0 +1,31 @@
+import { WaitablePromise } from '@algolia/client-common';
+import { DeleteResponse, SaveObjectResponse } from '@algolia/client-search';
+
+import { ResearchOutputResponse } from '@asap-hub/model';
+import { SearchIndex } from './search-index';
+import { SearchResponse } from '../types/response';
+
+export type ResearchOutputSearchResponse = SearchResponse<
```
peterstarling, line 8: is any of this tested?

apps/asap-server/test/handlers/research-output/index-handler.test.ts
```diff
@@ -15,6 +16,15 @@ import {
 } from '../../mocks/algolia-client.mock';
 import { researchOutputControllerMock } from '../../mocks/research-outputs-controller.mock';
 
+const getResearchOutputResponse = (): ResearchOutputSearchResponse => {
```
peterstarling, line 19: I'd create a fixture file for that

packages/algolia/src/indexes/research-output.ts
```diff
@@ -0,0 +1,31 @@
...
+  ResearchOutputResponse,
+  'research-output'
+>;
+
+export class ResearchOutputSearchIndex extends SearchIndex<
```
peterstarling, line 13: is it possible to create a `fetch` method here and use the package to fetch the data in FE? otherwise this does not need to be a package as it is only used by asap-server

At the moment the FE API is detached from the types:
https://github.com/yldio/asap-hub/pull/1281/files#diff-c307e5f0df5297cb25db5b6ef5d0683f149ba7ce5279271666870227c27fde4dL72

peterstarling (changes requested):

apps/asap-server/test/handlers/research-output/index-handler.test.ts
niezgoda: ok

apps/frontend/src/hooks/algolia.ts
```diff
@@ -1,26 +1,49 @@
...
     if (!user) {
       throw new Error('Algolia unavailable while not logged in');
     }
-    const client = algoliasearch(ALGOLIA_APP_ID, user.algoliaApiKey);
-    const index = client.initIndex(ALGOLIA_INDEX);
+
+    const client = algoliasearchLite(
+      ALGOLIA_APP_ID,
+      user.algoliaApiKey,
+    ) as SearchClient;
```
peterstarling, line 29: whats up with the casting?

peterstarling (approved):
looks about right, one minor comment about casting. I would ask someone to review the FE part

apps/frontend/src/hooks/algolia.ts
niezgoda: my bad

apps/frontend/src/hooks/algolia.ts
```diff
@@ -1,26 +1,46 @@
...
-  const initAlgolia = (user: User | null) => {
+  const initAlgolia = (user: User | null): AlgoliaHook => {
     if (!user) {
       throw new Error('Algolia unavailable while not logged in');
     }
-    const client = algoliasearch(ALGOLIA_APP_ID, user.algoliaApiKey);
-    const index = client.initIndex(ALGOLIA_INDEX);
+
+    const client = algoliasearchLite(ALGOLIA_APP_ID, user.algoliaApiKey);
+    const researchOutput = new ResearchOutputSearchIndex(
```
jhonny-m, line 27: I am a bit confused on what does this do over just having the `client.initIndex(ALGOLIA_INDEX)` ? is it for the types?

peterstarling: pretty much :D but i think type safety is quite important so

jhonny-m: we can pass the response type to the client search method, which was what we were doing. the only thing saying which response type we get is the facet filter we send, so I don't see this giving us more type safety. We just would have a  few clients instantiated pointing to the same index.

packages/algolia/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  SearchIndex as AlgoliaSearchIndex,
+  SearchClient as AlgoliaSearchClient,
+} from 'algoliasearch';
+import algoliasearchLiteFn, {
+  SearchClient as AlgoliaSearchClientLite,
+} from 'algoliasearch/lite';
+
+export type SearchClient = AlgoliaSearchClient;
+export type SearchClientLite = AlgoliaSearchClientLite;
+export type SearchIndex = AlgoliaSearchIndex;
```
jhonny-m, line 11: Whats the advantage of doing this?

packages/algolia/src/indexes/research-output.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.saveObject(researchOutput.id, researchOutput);
+  }
+
+  public async search(
+    query: string,
+    requestOptions?: SearchOptions,
+  ): Promise<ResearchOutputSearchResponse> {
+    return this.index.search<ResearchOutputSearchResponseEntity>(
+      query,
+      requestOptions,
```
jhonny-m, line 41: I can be wrong, but if we are doing this shouldn't we add the filter for the research-outputs entity here?
else nothing is stopping us from using this index to search for other entity types

jhonny-m: Yeah I don't think there anything that guarantees that we are searching for the right entity type

niezgoda: we do

packages/algolia/src/index.ts
niezgoda: You have no dependency on raw algoliasearch.
You can substitute AlgoliaSearchIndex with something else that has same interface :)

apps/frontend/src/hooks/algolia.ts
```diff
@@ -1,26 +1,46 @@
...
+      client.initIndex(ALGOLIA_INDEX) as SearchIndex,
```
kmaid, line 28: Why are we casting here?

apps/frontend/src/__fixtures__/algolia.ts
```diff
@@ -0,0 +1,56 @@
+import {
```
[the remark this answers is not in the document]
niezgoda: it would result in circular dependency :(

apps/frontend/src/hooks/algolia.ts
niezgoda: due to it's taken from algoliasearch/lite and not algoliasearch.
Sth wrong with types on this algoliasearch lib

PR #1282, opened 2022-01-14
Feat: Group inactive

kmaid:
https://trello.com/c/88gDXWFR/1799-omit-subscribe-from-group-and-event-pages-when-group-is-inactive

### DIFF
--- a/packages/react-components/src/templates/GroupProfileHeader.tsx
+++ b/packages/react-components/src/templates/GroupProfileHeader.tsx
@@ -38,14 +38,15 @@ type GroupProfileHeaderProps = {
   readonly numberOfTeams: number;
   readonly lastModifiedDate: string;
   readonly groupTeamsHref: string;
+  readonly active: boolean;
 } & ComponentProps<typeof EventSearch>;
 const GroupProfileHeader: React.FC<GroupProfileHeaderProps> = ({
   id,
+  active,
   name,
   numberOfTeams,
   lastModifiedDate,
   groupTeamsHref,
-
   searchQuery,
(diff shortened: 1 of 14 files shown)


packages/react-components/src/templates/GroupProfileHeader.tsx
```diff
@@ -89,10 +90,15 @@ const GroupProfileHeader: React.FC<GroupProfileHeaderProps> = ({
         </div>
         <TabNav>
           <TabLink href={route.about({}).$}>About</TabLink>
-          <TabLink href={route.calendar({}).$}>Calendar</TabLink>
-          <TabLink href={route.upcoming({}).$ + queryParamString(searchQuery)}>
-            Upcoming Events
-          </TabLink>
+          {active && <TabLink href={route.calendar({}).$}>Calendar</TabLink>}
+          {active && (
```
ana-yldio, line 94: Very minor. What do you think about this?
```
    {active && (
      <TabLink href={route.calendar({}).$}>Calendar</TabLink>
      <TabLink href={route.upcoming({}).$ + queryParamString(searchQuery)}>
         Upcoming Events
       </TabLink>
     )}
```

kmaid: It would need to be within a fragment. I did similar originally but the tab nav inspects the first level of nodes and thus breaks when it finds a fragment. Doing the conditional statement twice seemed a pragmatic solution than trying to find tab links in a tree the other side.

peterstarling (approved):

PR #1290, opened 2022-01-18
feat: Sync users with algolia

peterstarling:
https://trello.com/c/57EivHKQ/1795-prepare-sync-of-users-between-squidex-and-algolia


apps/asap-server/scripts/cli.ts
```diff
@@ -0,0 +1,31 @@
...
+          choices: ['users', 'research-outputs'],
+          demandOption: true,
+        })
+        .option('filename', {
+          alias: 'f',
+          type: 'string',
+          description: 'The output file name',
+        }),
+    handler: async ({ entity, filename }) =>
+      exportEntity(entity as 'users' | 'research-outputs', filename),
```
niezgoda, line 25: Could you move this type somewhere? You are using it again in export-entity.ts

apps/asap-server/scripts/export-entity.ts
```diff
@@ -0,0 +1,46 @@
...
+  const file = await fs.open(filename || `${entity}.json`, 'w');
+
+  let recordCount = 0;
+  let total: number;
+  let records: ListResponse<unknown>;
+  let page = 1;
+
+  await file.write('[\n');
+
+  do {
```
niezgoda, line 25: this loop could be in some helper :)

niezgoda (changes requested):

apps/asap-server/scripts/cli.ts
kmaid: Could you potentially type it in line 8?

peterstarling:
i love it how nobody noticed that this is introducing two different indices where we clearly said it would be just one

peterstarling:
apologies -  @kmaid noticed

kmaid:
image

PR #1292, opened 2022-01-18
Create a new Shared Output from the FE

ana-yldio:
https://trello.com/c/PPw0WII1/1839-create-a-new-shared-output-from-the-fe
https://trello.com/c/EBW4gVhg/1842-record-team-when-creating-a-new-output


packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import TeamCreateOutputForm from '../TeamCreateOutputForm';
+
+describe('TeamCreateOutputForm', () => {
+  test('click on button calls callback method', () => {
+    const spyOnCreate = jest.fn();
```
kmaid, line 7: This is a mock not a spy

packages/react-components/src/templates/__tests__/TeamCreateOutputPage.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { render, screen } from '@testing-library/react';
+import { ResearchOutput } from '@asap-hub/model';
+import { createResearchOutputResponse } from '@asap-hub/fixtures';
+import userEvent from '@testing-library/user-event';
+import TeamCreateOutputPage from '../TeamCreateOutputPage';
+
+describe('TeamCreateOutputPage', () => {
+  test('renders the research output type in the header', () => {
```
kmaid, line 8: The rest of the FE uses it rather than test

kmaid:
I was hoping we could refactor the modal save functionality into something generic we could use on this form. It does things like deal with disabling inputs and buttons while we are saving and asking the users if they are sure they want to navigate away without saving changes

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
markcsims: We're using this as a spy not as a full mock.

Mock functions are also known as "spies", because they let you spy on the behavior of a function that is called indirectly by some other code, rather than only testing the output.

packages/react-components/src/templates/__tests__/TeamCreateOutputPage.test.tsx
markcsims: Thanks! Done.

ana-yldio: thanks, done

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
kmaid: but a spy is a wrapper that will just intercept inputs and ouputs of a real function. This is just a jest.fn with no logic inside 😅

markcsims: https://jestjs.io/docs/mock-function-api

packages/react-components/src/templates/TeamCreateOutputPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+  onCreate: () => void;
+  researchOutput: ResearchOutput;
+};
+
+const TeamCreateOutputPage: React.FC<TeamCreateOutputPageProps> = ({
+  researchOutput,
+  onCreate,
+}) => (
+  <>
+    <TeamCreateOutputHeader type={researchOutput.type} />
```
kmaid, line 16: > Generally we pass in the entire research output like {...researchOutput} this means we don't have to remap as many props and sometimes we can get away with requiring additional attributes

markcsims: > I was hoping we could refactor the modal save functionality into something generic we could use on this form. It does things like deal with disabling inputs and buttons while we are saving and asking the users if they are sure they want to navigate away without saving changes

image

markcsims: I think this is part of another ticket coming up soon.

kmaid: Not sure you have replied to the comment above

kmaid (changes requested):

packages/react-components/src/templates/TeamCreateOutputPage.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { ResearchOutput } from '@asap-hub/model';
+
+import { TeamCreateOutputHeader, TeamCreateOutputForm } from '../organisms';
+
+type TeamCreateOutputPageProps = {
```
kmaid, line 6: Ideally we compose props of child components using : 
`type TeamCreateOutputHeaderProps = ComponentProps<typeof ChildComponent>` this allows us to avoid redefining them and having to change them every time a new prop is added

kmaid (changes requested):

packages/react-components/src/templates/TeamCreateOutputPage.tsx
kmaid: Think the above belongs in another comment

packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { ResearchOutputType } from '@asap-hub/model';
+
+import { Display, Paragraph } from '../atoms';
+import { perRem } from '../pixels';
+import { steel, paper } from '../colors';
+import { contentSidePaddingWithNavigation } from '../layout';
+
+type TeamCreateOutputHeaderProps = {
+  type: ResearchOutputType;
+};
```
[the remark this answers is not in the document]
markcsims: old as we are now passing a ResearchOutput

packages/react-components/src/organisms/__tests__/TeamCreateOutputHeader.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { render, screen } from '@testing-library/react';
+import TeamCreateOutputHeader from '../TeamCreateOutputHeader';
+
+describe('TeamCreateOutputHeader', () => {
+  it('renders the research output type in the header', () => {
+    render(<TeamCreateOutputHeader type="Bioinformatics" />);
```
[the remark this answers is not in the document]
markcsims: We were following best practices from Kent C Dodds blog here: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

apps/frontend/src/network/teams/__tests__/TeamOutput.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      screen.queryByRole('heading', { name: /Share bioinformatics/i }),
+    ).not.toBeInTheDocument();
+    expect(
+      screen.getByRole('heading', {
+        name: /Sorry! We can’t seem to find that page/i,
+      }),
+    ).toBeInTheDocument();
+  });
+
+  const renderPage = async (teamId: string) => {
```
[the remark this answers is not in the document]
markcsims: We moved it to the bottom of the spec, as we thought the important parts of the spec should be visible first. i.e. the tests. 

renderPage is a helper method. The linting seems happy.

apps/frontend/src/network/teams/state.ts
```diff
@@ -127,3 +128,9 @@ export const usePatchTeamById = (id: string) => {
     setPatchedTeam(await patchTeam(id, patch, authorization));
   };
 };
+export const usePostTeamResearchOutput = (teamId: string) => {
+  const authorization = useRecoilValue(authorizationState);
+  return async (payload: Partial<ResearchOutput>) => {
+    await createTeamResearchOutput(teamId, payload, authorization);
```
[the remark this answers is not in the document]
markcsims: We decided not to set anything in state, as we are not using the return value. (which is an id).

If and when we might use this, we could then decide to set in state.

packages/react-components/src/templates/TeamCreateOutputPage.tsx
markcsims: Fabulous to know! Done!

apps/frontend/src/network/teams/state.ts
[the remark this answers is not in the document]
kmaid: I think we might want to look at the return type. For patch requests we return the new entity this should probably be consistent

apps/frontend/src/network/teams/__tests__/TeamOutput.test.tsx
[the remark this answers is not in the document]
kmaid: Usually we have such helper methods at the top as multiple tests use them potentially with different parameters or mocks. Its minor happy to let it go

apps/frontend/src/network/teams/state.ts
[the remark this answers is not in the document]
markcsims: We wanted to do this, but we can't. The Research Output is on draft.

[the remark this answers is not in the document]
markcsims: Any suggestions welcome :)

apps/asap-server/src/controllers/teams.ts
```diff
@@ -67,6 +68,22 @@ export default class Teams implements TeamController {
     return this.fetchById(id);
   }
 
+  async merge(id: string, outputs: string[]): Promise<TeamResponse> {
```
peterstarling, line 71: I wonder if our team controller should really expose such method - what if we change the underlying data store and the relationship can be embedded on the user input? I think I would simply add this logic to the `create` method instead, so that it's all done (or could be done) atomically

apps/asap-server/src/routes/research-outputs.route.ts
```diff
@@ -49,6 +51,33 @@ export const researchOutputRouteFactory = (
     },
   );
 
+  researchOutputRoutes.post('/research-outputs', async (req, res) => {
+    const {
```
peterstarling, line 55: surely we would like to add some validation here, no?

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -609,6 +614,50 @@ describe('ResearchOutputs controller', () => {
       });
     });
   });
+
+  describe('create', () => {
+    afterEach(() => {
+      expect(nock.isDone()).toBe(true);
```
peterstarling, line 620: is there a reason why you decided to use nock? I think the idea is that we remove nock entirely, or at least for the controller tests, and use the client mock instead for assertions. 

The client could be tested in isolation and that's where we could use nock

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -609,6 +614,50 @@ describe('ResearchOutputs controller', () => {
...
+    });
+
+    afterEach(() => {
+      nock.cleanAll();
+    });
+    test('creating the research output should return the id from squidex rest', async () => {
```
peterstarling, line 626: I'm sure that's what squidex REST does but I thought maybe you considered returning the entire new object instead?

peterstarling: btw there is a lot more going on here than just testing the return - i think it's good to create another test where you validate the input that is send to the client, for instance i can see that although our input only contains the type the payload has two extra fields (which im sure is intended)

apps/asap-server/src/controllers/teams.ts
peterstarling: btw you could achieve the 'atomicity' (thats a word apparently) then by inserting the RO, then try/catching the team merging and if it fails - removing the RO

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -609,6 +614,50 @@ describe('ResearchOutputs controller', () => {
...
+      expect(id).toEqual(90210);
+    });
+    test('should throw when fails to create the research output - 400', async () => {
+      const researchOutput = { type: 1 } as unknown as ResearchOutput;
+      nock(config.baseUrl)
+        .post(`/api/content/${config.appName}/research-outputs?publish=false`)
+        .reply(400);
+
+      await expect(researchOutputs.create(researchOutput)).rejects.toThrow(
+        'Bad Request',
```
peterstarling, line 646: maybe we should be more expllicit with errors in our controllers? for instance maybe we should capture the HTTP errors and rethrow with a more meaningful message like 'Research Output failed to create', you could attach the original error to the stack im sure

peterstarling (changes requested):
there is a lot going on here and i think thats partly because two tickets were done at once - I am submitting some comments for now as I think you have a lot of work to do to address it - I will do a more thorough review again when its done

apps/asap-server/test/controllers/research-outputs.test.ts
markcsims: > I'm sure that's what squidex REST does but I thought maybe you considered returning the entire new object instead?

We wanted to return the whole object, but we can't. The Research Output is on draft.
 
Any suggestions welcome :)
But we thought this could wait, until we start editing the RO

apps/asap-server/src/controllers/teams.ts
markcsims: We are trying to do 2 updates/inserts to squidex.

- Create a RO.
- update a team with the RO relationship.

Ideally we'd like to do this in a transaction, but we can't. 

Say we insert an RO, and then to create the team relationship. If the latter fails, what should happen? We could follow something like the Saga pattern, and try to roll back the previous create (i.e. delete the RO). 

But what happens if this delete/removal fails too? We'd need to build in some retry logic etc. We decided in Refinement to log this error (in another ticket) and we'd have manual intervention for now.

markcsims: > I wonder if our team controller should really expose such method - what if we change the underlying data store and the relationship can be embedded on the user input?

Are you suggesting we move the RO relationship to a team, away from the team record and put it on the RO? Sounds heavy?

apps/asap-server/test/controllers/research-outputs.test.ts
markcsims: We were trying to follow the pattern from the other controllers (to be more consistent). We didn't want to introduce a new more explicit error handling pattern.

apps/asap-server/test/controllers/research-outputs.test.ts
markcsims: Looking at the other controller tests, all the REST requests are using nock. I think this will be the first time we will try to use the client mock for REST calls.

Just checking but have we used the client mock for REST calls?

apps/asap-server/src/routes/research-outputs.route.ts
markcsims: Doing this now, we had this but can't find it :)

apps/asap-server/src/controllers/teams.ts
peterstarling: with regards to the transaction - sure that can be done in a separate ticket so dont worry about it now.


But I still stand by what i said -  think logically it would make more sense to create a RO through a single controller method, i dont think we should expose a 'merge' method from the team controller. To me all the controllers (or alter - the clients/repositories if we shift the db logic further down) should only expose methods like create/update/delete.

apps/asap-server/test/controllers/research-outputs.test.ts
peterstarling: ahhh i see, yeah i guess that makes sense, best to use the same method then change - maybe we should create a ticket do address that too? not sure, its not the same as with graphql cause we dont have another way of testing like we do with the schema, but i think there is no value in testing REST at the http level anyway.

apps/asap-server/src/routes/research-outputs.route.ts
markcsims: Done

apps/asap-server/test/controllers/research-outputs.test.ts
peterstarling: you can actually fetch the drafts from squidex too. The only thing i would say is that you could do it now so you dont have to change the interface later. Maybe return an object with an ID, rather than just ID? so that later you can add the rest of the fields?

apps/asap-server/test/controllers/research-outputs.test.ts
peterstarling: Right, i actually think some of them throw more specific errors, or maybe its the parsers or something. Sure thats fine

apps/frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -7,6 +7,7 @@ import { v4 as uuid } from 'uuid';
 
 import { useTeamById } from './state';
 import Frame, { SearchFrame } from '../../structure/Frame';
+import TeamOutput from './TeamOutput';
```
jhonny-m, line 10: we can lazy load this one

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -120,6 +135,48 @@ export default class ResearchOutputs implements ResearchOutputController {
       ),
     };
   }
+  async create(
+    researchOutputData: ResearchOutput,
```
peterstarling, line 139: Do you know what - i think we should separate these input types from the models, what are models anyway? squidex data models right? and here we want to build something detached from the implementation. I don't mind extending `ResearchOutput` type but i think we could define the controller input type down below and have the `teamId` be part of the payload?

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -120,6 +135,48 @@ export default class ResearchOutputs implements ResearchOutputController {
...
+    teamId: string,
+  ): Promise<Partial<ResearchOutputResponse>> {
+    const { id: researchOutputId } = await this.createResearchOutput(
```
peterstarling, line 142: is there any reason why you extracted these methods to separate, private methods? its not like we are reusing them and not like we would want to test private methods either. It's not a huge deal but im usually against anything that we just dont need right now (YAGNE all the way)

peterstarling: unless its just for visibility - so like - when you open the method you know it creates the RO then associates the team, but i dont know..

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -609,6 +617,61 @@ describe('ResearchOutputs controller', () => {
...
+  describe('create', () => {
+    afterEach(() => {
+      expect(nock.isDone()).toBe(true);
+    });
+
+    afterEach(() => {
+      nock.cleanAll();
+    });
+    test('creating the research output should return the id from squidex rest', async () => {
+      const researchOutput = { type: 'test' } as unknown as ResearchOutput;
```
peterstarling, line 630: it would be nice to use an actual fixture with real data here

apps/asap-server/test/controllers/research-outputs.test.ts
```diff
@@ -609,6 +617,61 @@ describe('ResearchOutputs controller', () => {
...
+      nock(config.baseUrl)
+        .post(`/api/content/${config.appName}/research-outputs?publish=false`)
+        .reply(400);
+
+      await expect(
+        researchOutputs.create(researchOutput, teamId),
+      ).rejects.toThrow('Bad Request');
+    });
+
+    test('should throw when fails to create the research output - 500', async () => {
```
peterstarling, line 663: add a test to show what happens when team association fails maybe?

peterstarling: we are going to need it later anyway when we add some sort of level of atomicity

peterstarling (changes requested):

apps/asap-server/src/controllers/research-outputs.ts
markcsims: It's just for visibility. 

We could have had them as functions inside the main function, but then they need to be declared first. If they are standalone function we need to pass in the client.

apps/asap-server/test/controllers/research-outputs.test.ts
markcsims: done

apps/frontend/src/network/teams/__mocks__/api.ts
```diff
@@ -27,3 +27,5 @@ export const getTeams = jest.fn(
   async ({ pageSize }: GetListOptions): Promise<ListTeamResponse> =>
     createListTeamResponse(pageSize ?? 10),
 );
+
+export const createTeamResearchOutput = jest.fn();
```
kmaid, line 31: Could we please type this mock with the response output? I think we now return the draft RO right?

apps/frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -35,8 +41,11 @@ const TeamProfile: FC<Record<string, never>> = () => {
   if (team) {
     return (
       <Frame title={team.displayName}>
-        <TeamProfilePage teamListElementId={teamListElementId} {...team}>
-          <Switch>
+        <Switch>
+          <Route path={path + route({ teamId }).createOutput.template}>
```
kmaid, line 45: This should probablly be wrapped in a Frame. Frames allow you to set the page title and provide the highest level error boundary which should be in scope imo

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
```diff
@@ -0,0 +1,27 @@
...
+  alignItems: 'end',
+  justifyContent: 'end',
+  columnGap: `${12 / perRem}em`,
+  padding: `${36 / perRem}em ${contentSidePaddingWithNavigation(8)} ${
+    48 / perRem
+  }em `,
+});
+
+const TeamCreateOutputForm: React.FC<{ onCreate: () => void }> = ({
+  onCreate,
```
kmaid, line 18: Usually we default functions like this to noop (you can import it). This just means you don't neeed floating empty functions everywhere like your storybook etc.

packages/react-components/src/templates/TeamCreateOutputPage.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ComponentProps } from 'react';
+
+import { TeamCreateOutputHeader, TeamCreateOutputForm } from '../organisms';
+
+type TeamCreateOutputPageProps = ComponentProps<typeof TeamCreateOutputHeader> &
+  ComponentProps<typeof TeamCreateOutputForm> &
+  ComponentProps<typeof TeamCreateOutputHeader>;
```
kmaid, line 7: You have this in there twice.

apps/frontend/src/network/teams/__tests__/TeamOutput.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useFlags } from '@asap-hub/react-context';
+import { render, screen, waitFor } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+import { Suspense } from 'react';
+import { StaticRouter } from 'react-router-dom';
+import { RecoilRoot } from 'recoil';
+import { refreshTeamState } from '../state';
+import TeamOutput from '../TeamOutput';
+
+describe('TeamOutput', () => {
```
kmaid, line 14: Duplication

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -131,4 +188,12 @@ export interface ResearchOutputController {
   }) => Promise<ListResearchOutputResponse>;
 
   fetchById: (id: string) => Promise<ResearchOutputResponse>;
+  create: (
+    researchOutputRequest: ResearchOutputRequest,
+  ) => Promise<Partial<ResearchOutputResponse>>;
 }
+
+export type ResearchOutputRequest = Pick<
```
peterstarling, line 196: hate to be picky but i think we should name the controller input data differently - maybe `....InputData` or something, basically i would try to move away from naming everything `request` and `response` - like i said earlier the controllers should not return API specific responses anyway. 

Unless you think we should for now align this with the response (in which case `request` is appropriate) and change later

peterstarling (approved):

packages/react-components/src/templates/__tests__/TeamCreateOutputPage.test.tsx
```diff
@@ -0,0 +1,48 @@
+import { render, screen } from '@testing-library/react';
+import { createResearchOutputResponse } from '@asap-hub/fixtures';
+import { useFlags } from '@asap-hub/react-context';
+import { renderHook } from '@testing-library/react-hooks';
+import userEvent from '@testing-library/user-event';
+import TeamCreateOutputPage from '../TeamCreateOutputPage';
+
+describe('TeamCreateOutputPage', () => {
```
kmaid, line 8: Dup

kmaid (changes requested):

apps/asap-server/src/controllers/research-outputs.ts
markcsims: We've rename it. to ResearchOutputInputData

But the previous name was used as it was complementary to the output.

The output type is: ResearchOutputResponse, therefore we thought the input should be a Request i.e. ResearchOutputRequest

kmaid (approved):

PR #1314, opened 2022-02-03
Clean up BE fixtures: calendars

ana-yldio:


apps/asap-server/src/routes/calendars.route.ts
```diff
@@ -15,10 +16,7 @@ export const calendarRouteFactory = (
       'query',
       parameters,
       querySchema,
-    ) as unknown as {
-      take: number;
-      skip: number;
-    };
+    ) as FetchPaginationOptions;
```
kmaid, line 19: Does this have to be casted?

ana-yldio: thanks, done

niezgoda (approved):

PR #1315, opened 2022-02-03
feat: CRN-652 add tag selector to the Shared Outputs form

marot:


packages/react-components/src/organisms/TeamCreateOutputExtraInformationCard.tsx
```diff
@@ -0,0 +1,735 @@
...
+import { FormCard, LabeledMultiSelect } from '../molecules';
+import { noop } from '../utils';
+
+type TeamCreateOutputExtraInformationProps = {
+  values: string[];
+  onChange?: (values: string[]) => void;
+};
+
+// The list is auto generated // TODO: Where should we put this?
+const tags = [
```
jhonny-m, line 11: we already have these in the code at `expertiseAndResourceSuggestions`

marot: Ah thank you - I didn't think of searching for them.

The list seems to be updated:
- ~300 new items
- There are many renames
- Some deletions

I'd propose to reuse them and update them in a separate issue.

marot: Follow up: The list should be separate. I will move it next to `expertiseAndResourceSuggestions` with a note that the duplication is intentional

apps/storybook/src/FormCard.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import { FormCard } from '@asap-hub/react-components';
+
+export default {
+  title: 'Molecules / FormCard',
```
kmaid, line 4:
```suggestion
  title: 'Molecules / Form Card',
```

apps/storybook/src/FormCard.stories.tsx
```diff
@@ -0,0 +1,10 @@
...
+  component: FormCard,
+};
+
+export const Normal = () => (
+  <FormCard title={'What are you sharing?'}>Card content</FormCard>
```
kmaid, line 9: Use the text Knob so designers can play around with this value

kmaid (approved):

PR #1317, opened 2022-02-04
CRN-652 adapt form card header

marot:
Part of https://asaphub.atlassian.net/browse/CRN-652

### DIFF
--- a/packages/react-components/src/atoms/Headline3.tsx
+++ b/packages/react-components/src/atoms/Headline3.tsx
@@ -1,16 +1,31 @@
-import { AllowedChildren, layoutStyles, headlineStyles } from '../text';
+import { css } from '@emotion/react';
+import { AllowedChildren, headlineStyles, layoutStyles } from '../text';
 
 interface Headline3Props {
   readonly children: AllowedChildren;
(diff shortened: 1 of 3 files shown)


packages/react-components/src/atoms/Headline3.tsx
```diff
@@ -1,16 +1,31 @@
...
 const Headline3: React.FC<Headline3Props> = ({
   children,
   styleAsHeading = 3,
   id,
+  noMargin = false,
 }) => (
-  <h3 css={[layoutStyles, headlineStyles[styleAsHeading]]} id={id}>
+  <h3
+    css={[
+      layoutStyles,
```
marot, line 23: This adds a 12px margin - added to the parent card padding this is too much. I'm not sure what the best approach to overwrite this behaviour is. Any ideas?
- Pass additional `SerializedStyles` to make it possible to overwrite any styles?
- Duplicate the component?

TBD: Why isn't this an issue in other areas? I need to check the paddings of other cards/headers.

Also, should I make the change for the other heading components too?

kmaid (approved):

PR #1318, opened 2022-02-07
CRN-651 allow users to create outputs of different types

marot:
https://asaphub.atlassian.net/browse/CRN-651

### DIFF
--- a/packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
+++ b/packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { css } from '@emotion/react';
 import { ResearchOutput } from '@asap-hub/model';
 
+import { news } from '@asap-hub/routing';
 import { Display, Link, Paragraph } from '../atoms';
 import { perRem } from '../pixels';
 import { paper, steel } from '../colors';
@@ -21,7 +22,7 @@ const visualHeaderStyles = css({
   maxWidth: `${720 / perRem}em`,
 });
 
-const headerCopy = ({ outputType }: { outputType: ResearchOutput['type'] }) => {
+const headerCopy = (outputType: ResearchOutput['type']) => {
   switch (outputType) {
     case 'Protocol':
       return 'Share a protocol';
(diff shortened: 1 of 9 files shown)


packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
```diff
@@ -21,18 +21,97 @@ const visualHeaderStyles = css({
   maxWidth: `${720 / perRem}em`,
 });
 
+const headerCopy = ({ outputType }: { outputType: ResearchOutput['type'] }) => {
```
niezgoda, line 24: Can we do this in a way

const outputTypeMap = {
  'Protocol': 'Share a protocol'
...
}

return outputTypeMap[outputType] ?? 'Share a resource';

niezgoda (changes requested):

packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
kmaid: I think its been done because the URL can have anything technically and there needs to be a default case.

kmaid: Might make sense to have it case insensitive though

kmaid (approved):

packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
marot: @niezgoda I think both approaches are okay and there isn't a clear winner. So I'd opt with going what we already have.
@kmaid In this case the outputType is typed, so it can only contain specific values. The url parameter is also typed, so the only way for it to be wrong is if the user manually modifies it. I think we are fine without checking the url parameter for case

marot: @kmaid What part do you mean? It is already using `ResearchOutput['type']`. Do you mean the entire parameter type `{ outputType: ResearchOutput['type'] }`? Where would you put shared types like that?

niezgoda (approved):

packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
kmaid: Ah, nevermind. Not sure why you are taking an object

marot: I usually default to objects because it makes refactoring easier when adding parameters. I can remove it if you want 👍

packages/react-components/src/organisms/TeamCreateOutputHeader.tsx
```diff
@@ -21,18 +21,97 @@ const visualHeaderStyles = css({
...
+}) => {
+  switch (outputType) {
+    case 'Protocol':
+      return (
+        <>
+          Add your protocol to a repository (e.g. protocols.io) before sharing
+          on the Hub. Find out{' '}
+          <Link
+            href={
+              'https://hub.asap.science/news/7d3dd1ec-14ef-441c-8a2c-19a23d6264f3'
```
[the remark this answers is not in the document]
marot: Mhm... the content only exists on production, using the routing library would mean on dev/PR envs it points to nothing. I'd go with this because we are unlikely to change the domain/routes - and then we only have two broken links.

But I have no hard feelings about this...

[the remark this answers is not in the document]
kmaid: This link exists in development https://dev.hub.asap.science/news/7d3dd1ec-14ef-441c-8a2c-19a23d6264f3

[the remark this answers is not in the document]
marot: Oh, I didn't expect that. Will change 👍

PR #1320, opened 2022-02-07
feat: CRN-650 Sync users with Algolia

peterstarling:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/6?modal=detail&selectedIssue=CRN-650

### DIFF
--- a/serverless.ts
+++ b/serverless.ts
@@ -204,6 +204,7 @@ const serverlessConfig: AWS = {
       environment: {
         AUTH0_CLIENT_ID: `\${ssm:auth0-client-id-${envAlias}}`,
         AUTH0_SHARED_SECRET: `\${ssm:auth0-shared-secret-${envAlias}}`,
+        ALGOLIA_SEARCH_API_KEY: `\${ssm:algolia-search-api-key-${envAlias}}`,
       },
     },
     auth0ConnectByCode: {


niezgoda (approved):

niezgoda (dismissed):

.github/workflows/env-setup.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       - name: Download the files and restore the index
         run: |
           aws s3 cp s3://$BUCKET_NAME/$MOST_RECENT_BACKUP_DIR $MOST_RECENT_BACKUP_DIR --recursive
-          export ALGOLIA_INDEX=asap-hub_research_outputs_CI-$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
           export ALGOLIA_BACKUP=$(readlink -f $MOST_RECENT_BACKUP_DIR)
+          export ALGOLIA_RO_INDEX=asap-hub_research_outputs_CI-$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
```
markcsims, line 26: I know this was existing code, but I think we can get the PR number from the event: github.event.number

apps/asap-server/test/mocks/algolia-client.mock.ts
```diff
@@ -1,10 +1,7 @@
...
-} as unknown as jest.Mocked<SearchIndex>;
-
-export const algoliaClientMock = {
-  initIndex: jest.fn().mockReturnValue(algoliaIndexMock),
-} as unknown as jest.Mocked<SearchClient>;
+export const algoliaSearchClientMock = {
+  save: jest.fn(),
+  remove: jest.fn(),
+  searchEntity: jest.fn(),
+} as unknown as jest.Mocked<AlgoliaSearchClient>;
```
markcsims, line 7: We have to do the cast because of the constructor? Have we tried using the mocking built into jest?

apps/frontend/src/__fixtures__/algolia.ts
```diff
@@ -1,20 +1,10 @@
...
-  TEntityName extends string,
->(
-  data: TEntity[],
-  overrides: Partial<SearchResponse<TEntity, TEntityName>> = {},
-  type: string = 'research-output',
-): SearchResponse<TEntity, TEntityName> => ({
+export const createAlgoliaResponse = (
+  data: EntityRecord<'research-output'>[],
+  overrides: Partial<SearchResponse<EntityRecord<'research-output'>>> = {},
+): SearchResponse<EntityRecord<'research-output'>> => ({
```
markcsims, line 7: 'research-output' could be a constant or part of an enumeration?

Seems like it's from: 

```
export type EntityResponses = {
  'research-output': ResearchOutputResponse;
  user: UserResponse;
};```

File: apps/frontend/src/config.ts [hunk 4]
```diff
@@ -1,8 +1,6 @@
 export const GTM_CONTAINER_ID = process.env.REACT_APP_GTM_CONTAINER_ID;
 export const API_BASE_URL =
   process.env.REACT_APP_API_BASE_URL || 'http://localhost:3333';
-export const ALGOLIA_APP_ID =
-  process.env.REACT_APP_ALGOLIA_APP_ID || 'LVYWOPQ0A9';
 export const ALGOLIA_INDEX =
   process.env.REACT_APP_ALGOLIA_INDEX || 'asap-hub_research_outputs_dev';
```
markcsims, line 5: is asap-hub_research_outputs_dev the correct name until we fully migrate over?

apps/frontend/src/hooks/__tests__/algolia.test.tsx
```diff
@@ -1,31 +1,22 @@
...
+import { ALGOLIA_INDEX } from '@asap-hub/frontend/src/config';
 import { renderHook } from '@testing-library/react-hooks';
 import { RecoilRoot } from 'recoil';
+import { algoliaSearchClientFactory } from '@asap-hub/algolia';
 
 import { useAlgolia } from '../algolia';
 import { Auth0Provider, WhenReady } from '../../auth/test-utils';
 
-var mockInitIndex: jest.MockedFunction<SearchClient['initIndex']>;
+var mockAlgoliaSearchClientFactory: jest.MockedFunction<
```
markcsims, line 9: Can this be a let?

apps/frontend/src/network/teams/Outputs.tsx
```diff
@@ -56,17 +56,17 @@ const OutputsList: React.FC<OutputsListProps> = ({
     result.total,
     pageSize,
   );
-  const { index } = useAlgolia();
+  const { client } = useAlgolia();
   const exportResults = () =>
-    algoliaResultsToStream(
+    algoliaResultsToStream<'research-output'>(
```
markcsims, line 61: Magic 'research-output' string. Same as above, should this be a constant?

apps/frontend/src/network/users/Outputs.tsx
```diff
@@ -59,17 +59,17 @@ const OutputsList: React.FC<OutputsListProps> = ({
     result.total,
     pageSize,
   );
-  const { index } = useAlgolia();
+  const { client } = useAlgolia();
   const exportResults = () =>
-    algoliaResultsToStream(
+    algoliaResultsToStream<'research-output'>(
```
markcsims, line 64: Magic 'research-output' string. Same as above, should this be a constant?

apps/frontend/src/shared-research/ResearchOutputList.tsx
```diff
@@ -29,20 +29,20 @@ const ResearchOutputList: React.FC<ResearchOutputListProps> = ({
...
-  const { index } = useAlgolia();
+  const { client } = useAlgolia();
 
   const { numberOfPages, renderPageHref } = usePagination(
     result?.total || 0,
     pageSize,
   );
   const exportResults = () =>
-    algoliaResultsToStream(
+    algoliaResultsToStream<'research-output'>(
```
markcsims, line 39: Magic 'research-output' string. Same as above, should this be a constant?

apps/frontend/src/shared-research/__mocks__/api.ts
```diff
@@ -12,6 +12,6 @@ export const getResearchOutput = jest.fn(
 );
 
 export const getResearchOutputs = jest.fn(
-  async (): Promise<Partial<ResearchOutputSearchResponse>> =>
+  async (): Promise<Partial<SearchResponse<EntityRecord<'research-output'>>>> =>
```
markcsims, line 15: Magic 'research-output' string. Same as above, should this be a constant?

apps/frontend/src/shared-research/__tests__/export.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
markcsims, line 1: In this file we're allowing "any" - is there a way to avoid this?

.github/workflows/env-setup.yml
```diff
@@ -22,6 +22,8 @@ jobs:
...
+          yarn workspace @asap-hub/asap-server run algolia import -s $ALGOLIA_BACKUP -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_RO_INDEX
+          export ALGOLIA_INDEX=asap-hub_CI-$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
```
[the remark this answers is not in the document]
peterstarling: done

apps/asap-server/test/fixtures/users.fixtures.ts
```diff
@@ -449,3 +451,70 @@ export const userPublishedEvent: WebhookPayload<User> = {
...
+
+export const getUserWebhookPayload = (
+  id: string,
+  type:
+    | 'UsersCreated'
+    | 'UsersPublished'
+    | 'UsersUpdated'
+    | 'UsersUnpublished'
+    | 'UsersDeleted',
+): WebhookPayload<User> => ({
```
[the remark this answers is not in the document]
peterstarling: i dont think so, its a fixture/

apps/asap-server/test/handlers/user/index-handler.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    algoliaSearchClientMock,
+  );
+
+  afterEach(() => jest.clearAllMocks());
+
+  test('Should fetch the user and create a record in Algolia when the user is created', async () => {
+    const userResponse = getUserResponse();
+    userControllerMock.fetchById.mockResolvedValueOnce(userResponse);
+
+    await indexHandler(createEvent('user-1234'));
```
marot, line 22: Maybe make the id of `createEvent` optional so it is not necessary here? It looks like this exact value is required, while it can be anything.

When looking at this, I thought the id in `userResponse` is `user-1234`.

apps/asap-server/test/handlers/user/index-handler.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+describe('User index handler', () => {
+  const indexHandler = indexUserHandler(
+    userControllerMock,
+    algoliaSearchClientMock,
+  );
+
+  afterEach(() => jest.clearAllMocks());
+
+  test('Should fetch the user and create a record in Algolia when the user is created', async () => {
```
marot, line 18: Maybe also add a test that `fetchById` is called with the correct `userId`?

apps/asap-server/test/mocks/algolia-client.mock.ts
peterstarling: yes and no :D

apps/frontend/src/config.ts
peterstarling: yes

apps/frontend/src/hooks/__tests__/algolia.test.tsx
peterstarling: actually it cannot, jest is a bit funny about hoisting

apps/asap-server/src/handlers/user/index-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+      const user = await userController.fetchById(event.detail.payload.id);
+
+      logger.debug(`Fetched user ${user.id}`);
+
+      await algoliaClient.save(user);
+
+      logger.debug(`Saved user ${user.id}`);
+    } catch (e) {
+      if (e?.output?.statusCode === 404) {
+        await algoliaClient.remove(event.detail.payload.id);
```
marot, line 31: When we receive a delete event, can we be 100% sure that `userController.fetchById` returns 404?

packages/algolia/src/client.ts
```diff
@@ -0,0 +1,58 @@
...
+    objectID: string;
+    __meta: {
+      type: T;
+    };
+  };
+
+export const getEntityType = (
+  entity: EntityResponses[keyof EntityResponses],
+): keyof EntityResponses => {
+  if ('title' in entity) {
```
marot, line 21: This feels like a weak property. Maybe we add a title to user at some point...

peterstarling: hmmm.. i could use more properties to distinguish but I think there is not much else to tell them apart, but yeah agreed - just the title isn't going to cut it

apps/asap-server/test/handlers/user/index-handler.test.ts
peterstarling: done mate

apps/asap-server/src/handlers/user/index-handler.ts
peterstarling: so actually the way this handler works is that no matter what event type it receives it will look up the record in Squidex and if 404 is received it assumes that the record was deleted. 

So regardless of the event type we assume that a 404 means record was deleted - are we sure though that this is the only case? yeah i suppose for the 404 we are sure we can remove it from algolia.

marot: my concern would be that we receive the delete event but the endpoint still returns a value for the user (e.g. due to cache)
in that case the index would be inconsistent

peterstarling: yeah it could happen but on the flip side if we receive a delete event and remove the record without checking that may lead to an even worse outcome

apps/asap-server/src/utils/types.ts
```diff
@@ -43,3 +43,7 @@ declare global {
   type Instant = string & { __type: 'SquidexDate' };
   type JsonScalar = Record<string, unknown> & { __type: 'JsonScalar' };
 }
+
+export type EventBridgeHandler<TDetailType extends string, TDetail> = (
```
kmaid, line 47: Surprised we cant import this from some aws package?

markcsims (approved):

PR #1324, opened 2022-02-08
Feat: Basic RO Data

kmaid:


packages/react-components/src/organisms/TeamCreateOutputExtraInformationCard.tsx
```diff
@@ -1,26 +1,32 @@
...
-        suggestions={suggestions}
+        values={tags}
+        suggestions={tagSuggestions}
         placeholder="Add a keyword (E.g. Cell Biology)"
-        values={values}
         onChange={onChange}
       />
-      <Link href="mailto:techsupport@asap.science?subject=New+keyword">
+
+      <Link href={mailToSupport({ subject: 'New keyword' }).toString()}>
```
kmaid, line 29: @marot Minor improvement FYI

marot: Thanks, good to know!

packages/react-components/src/organisms/Form.tsx
```diff
@@ -0,0 +1,76 @@
...
+    onSave: () => void | Promise<void>;
+  }) => ReactNode;
+};
+const Form: React.FC<FormProps> = ({
+  dirty,
+  children,
+  validate = () => true,
+  onSave = noop,
+}) => {
+  const formRef = useRef<HTMLFormElement>(null);
```
marot, line 31: What's the policy on using libraries, like e.g. formik for this?

apps/frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -34,7 +34,8 @@ const TeamProfile: FC<Record<string, never>> = () => {
   useEffect(() => {
     loadAbout()
       .then(team?.tools ? loadWorkspace : undefined)
-      .then(loadOutputs);
+      .then(loadOutputs)
+      .then(loadTeamOutput);
```
kmaid, line 38: @ana-yldio This tells webpack to preload the lazy chuncks

packages/react-components/src/organisms/Form.tsx
kmaid: I intend to refactor that implementation to reuse this. I just wanted to get it tested first. Would appreciate any feedback.

kmaid: I am not sure I like the tripple wrap

apps/asap-server/src/routes/research-outputs.route.ts
```diff
@@ -52,35 +53,19 @@ export const researchOutputRouteFactory = (
...
-    });
+    const researchOutput = await researchOutputController.create(createRequest);
 
     res.status(201).json(researchOutput);
   });
 
   return researchOutputRoutes;
 };
-const createSchema = Joi.object({
+const createSchema = Joi.object<ResearchOutputPostRequest>({
```
kmaid, line 65: @ana-yldio @markcsims @peterstarling Thought this was neat as it ensures we keep Joi up to date with the typed request

peterstarling (dismissed):

packages/react-components/src/organisms/__tests__/Form.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+      });
+
+      it('resets to initial state once a saved form matches the default state', async () => {
+        const { getByText, rerender } = result;
+
+        userEvent.click(getByText(/^save/i));
+        await waitFor(() =>
+          expect(getByText(/^save/i).closest('button')).toBeEnabled(),
+        );
+        rerender(
```
marot, line 178: Does this test anything? Isn't this the same as 're-enables the save button'?

apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -138,15 +138,17 @@ export default class ResearchOutputs implements ResearchOutputController {
   async create({
     teamId,
     ...researchOutputData
-  }: ResearchOutputInputData): Promise<Partial<ResearchOutputResponse>> {
+  }: ResearchOutputPostRequest): Promise<Partial<ResearchOutputResponse>> {
```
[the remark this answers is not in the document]
kmaid: This is the contract between the backend and frontend. It might be remapped after this point but we need to take in a request validate it and then you remap it or whatever.

I find it strange that you treating your controller methods as like the API methods shouldn't there be some separation here?

packages/react-components/src/organisms/__tests__/Form.test.tsx
kmaid: This is unfortunately for coverage right now. I will refactor this out once I modify EditModal to reuse this component. I just want to be able to merge to not block new form work

apps/asap-server/src/routes/research-outputs.route.ts
kmaid: ResearchOutputPostRequest is the contract between the BE and FE and this function validates that its compliant. I don't understand why you would retype this? The API is restful what other resources would you create using this route? Can you provide some examples scenarios?

apps/asap-server/src/controllers/research-outputs.ts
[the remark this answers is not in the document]
peterstarling: so contract is not at controller level, it is at the request level. This is exactly what i am telling you - to use the `ResearchOutputPostRequest` type like you did inside the RO route but then remap it to something else for the controller. I simply dont want the controller to be tied to the API request contract

packages/react-components/src/organisms/__tests__/Form.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+it('renders a form with given children', () => {
+  const { getByText } = render(<Form {...props}>{() => 'Content'}</Form>, {
+    wrapper: MemoryRouter,
+  });
+  expect(getByText('Content')).toBeVisible();
+});
+
+it('initially does not prompt when trying to leave', () => {
+  const { getByText } = render(
```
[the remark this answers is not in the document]
kmaid: Yep, it won't or if the fields are restored to empty

apps/frontend/src/network/teams/state.ts
```diff
@@ -127,10 +127,9 @@ export const usePatchTeamById = (id: string) => {
...
 };
-export const usePostTeamResearchOutput = (teamId: string) => {
+export const usePostTeamResearchOutput = () => {
   const authorization = useRecoilValue(authorizationState);
-  return async (payload: Partial<ResearchOutput>) => {
+  return (payload: ResearchOutputPostRequest) =>
     // TODO: Store the response in the state
-    await createTeamResearchOutput(teamId, payload, authorization);
-  };
+    createTeamResearchOutput(payload, authorization);
```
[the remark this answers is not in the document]
kmaid: We can't. Normally I would expect the newly created entity to come back from this API response like it does for updates on profile. This way we can then send the user to the new RO (assuming it wasn't in draft state) and have things like its object Id.

apps/asap-server/src/routes/research-outputs.route.ts
kmaid: Long story short. Joi doesn't allow the use of generics on the validation function so the result is untyped and lame. Raised https://asaphub.atlassian.net/browse/CRN-714 to move away from Joi as it seems like a fairly important feature of validation library.

jhonny-m (approved):

PR #1325, opened 2022-02-08
CRN-648 Feature flags not working as expected in production

marot:
In 87312e7eeb54abe64e8c36d82fbdfc85c47d033a `yarn run build` was removed from the `deploy:sls:production` job. Running `yarn run build` would override the build artifact created in the `build:ts:prod` step.

This change adds NODE_ENV production to the `build:ts` step.

Note that the frontend would build in production by default because of react-scripts. It is important to run `yarn run build:babel` using `NODE_ENV=production`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -53,6 +53,8 @@ stages:
     paths:
       - 'packages/*/build*'
       - 'apps/*/build*'
+  variables:
+    NODE_ENV: 'production'
   before_script:
     - source ci/env-setup.sh
     - export REACT_APP_ALGOLIA_APP_ID=$(aws ssm get-parameter --name "algolia-app-id-dev" --query Parameter.Value --output text)
--- a/packages/flags/src/__tests__/index.test.ts
+++ b/packages/flags/src/__tests__/index.test.ts
@@ -7,30 +7,30 @@ import {
   enable,
 } from '..';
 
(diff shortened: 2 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -53,6 +53,8 @@ stages:
     paths:
       - 'packages/*/build*'
       - 'apps/*/build*'
+  variables:
+    NODE_ENV: 'production'
```
jhonny-m, line 57: will this not set the `NODE_ENV` as `production` for all environments?

marot: Isn't this something that we want? Usually `development` is only set for local developer machines.
We want dev/PR environments to be as close to the prod environment as possible, right?

I did experience differences between development/production builds in the past - and feature flag would be one example where there are also differences.

jhonny-m: the only thing we check with it is the Feature Flags and to register a service worker. The default behavior for feature flags, should be On on Dev and PR's and off on Prod

kmaid: We do want to keep feature flags enabled by default on development. Perhaps we have overloaded NODE_ENV

marot: @kmaid @jhonny-m how about we use REACT_APP_ENVIRONMENT to enable feature flags by default? It is only set to production in production.

kmaid (approved):

PR #1330, opened 2022-02-10
Feature/CRN-598 keep user lab relationships up to date in Algolia & F…

niezgoda:
CRN-598
CRN-597

requirement

Create event listener for Lab changes

when a User's relationship to a Lab changes, that change should be reflected in Algolia

when a Lab which is related to a User changes , those Lab changes should be reflected on the User information in Algolia

when a Lab information is changed, reflect that on all users assigned to the Lab

information should be sent on Event Bus

AC testing

relationship changes:

new relationship

relationship removed

role on Lab has changed

lab changes:

lab name changed

lab removed

lab moved to draft

lab moved to published

TRELLO LINK* https://trello.com/c/MT0KBans/1792-keep-user-lab-relationships-up-to-date-in-algolia


apps/asap-server/src/handlers/lab/users-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+  algoliaSearchClientFactory,
+  BatchRequest,
+} from '@asap-hub/algolia';
+import Users, { UserController } from '../../controllers/users';
+import { LabEventType } from '../webhooks/webhook-lab';
+import { algoliaApiKey, algoliaAppId, algoliaIndex } from '../../config';
+import { loopOverCustomCollection } from '../../utils/migrations';
+import logger from '../../utils/logger';
+
+export const indexLabUsersHandler =
```
peterstarling, line 15: I would refactor this. 

Perhaps you would like to take a look at: 
https://github.com/yldio/asap-hub/blob/622ce9b622db69cec0253b94060c9acfd2c73e55/apps/asap-server/src/handlers/teams/index-handler.ts#L14

It works on a similar basis - for every team that is updated we want to update ROs that are associated with them. However - unlike the team->RO relationship - you will not receive a list of users associated to the lab, so instead of grabbing that from the payload you need to fetch the list of users by the lab - something that you are already doing.

apps/asap-server/src/handlers/lab/users-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+          ),
+        );
+
+        logger.info(
+          `Updated ${foundUsers.total} users with algolia response: ${algoliaResponse}`,
+        );
+      };
+
+      await loopOverCustomCollection(fetchFunction, processingFunction);
+    } catch (error) {
```
peterstarling, line 57: what are we catching here anyway? If anything I would probably try catch inside the loop to skip to make sure it continues on failure.

apps/asap-server/src/handlers/teams/users-handler.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): ((
+    event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
+  ) => Promise<void>) =>
+  async (
+    event: EventBridgeEvent<TeamsEventType, SquidexWebhookTeamPayload>,
+  ): Promise<void> => {
+    logger.debug(`Event ${event['detail-type']}`);
+
+    try {
+      const fetchFunction = (
```
peterstarling, line 31: actually for this handler I dont think you need to fetch-users-by-team-relationship. Why don't use simply fetch the team by ID via `TeamController` and use the `members` list to update the users?

serverless.ts
```diff
@@ -338,6 +338,44 @@ const serverlessConfig: AWS = {
...
+      events: [
+        {
+          eventBridge: {
+            eventBus: 'asap-events-${self:provider.stage}',
+            pattern: {
+              source: ['asap.lab'],
+              'detail-type': [
+                'LabPublished',
+                'LabUpdated',
+                'LabCreated',
```
peterstarling, line 367: i don't think we need to listen to `LabCreated` event, what do you think? since there should be no users to update anyway

peterstarling (changes requested):

apps/asap-server/src/utils/migrations.ts
```diff
@@ -31,3 +32,19 @@ export const applyToAllItemsInCollection = async <
...
 };
+
+export const loopOverCustomCollection = async <TResponse>(
+  fetcher: (skip: number) => Promise<ListResponse<TResponse>>,
+  processEntities: (result: ListResponse<TResponse>) => Promise<void>,
+): Promise<void> => {
+  let pointer = 0;
+  let results: ListResponse<TResponse>;
+
+  do {
```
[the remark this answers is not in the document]
niezgoda: Not necessary - the API has it's limits too :)

apps/asap-server/src/utils/migrations.ts
```diff
@@ -31,3 +32,19 @@ export const applyToAllItemsInCollection = async <
     pointer += 10;
   } while (pointer < result.total);
 };
+
+export const loopOverCustomCollection = async <TResponse>(
```
peterstarling, line 36: i think this method should not really belong to `migrations`

peterstarling (changes requested):

apps/asap-server/src/handlers/lab/users-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      throw error;
+    }
+  };
+
+export type SquidexWebhookLabPayload = {
+  type: 'LabsPublished' | 'LabsUpdated' | 'LabsUnpublished' | 'LabsDeleted';
+  payload: {
+    $type: 'EnrichedContentEvent';
+    type: 'Published' | 'Updated' | 'Unpublished' | 'Deleted';
```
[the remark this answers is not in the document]
niezgoda: I don't think so.
This is a separate entity. We do not need to share everything - as you can see this objects differ :)

apps/asap-server/src/utils/migrations.ts
niezgoda: It has to be somewhere.
I find this place as it has some loop-over collection.
Where do you want to place it?

serverless.ts
niezgoda: Good point ;)
However - we can get this event anytime - maybe 1h after creation due to network conditions.

apps/asap-server/src/handlers/teams/users-handler.ts
```diff
@@ -0,0 +1,77 @@
...
+import Users, { UserController } from '../../controllers/users';
+import {
+  TeamsEventType,
+  SquidexWebhookTeamPayload,
+} from '../webhooks/webhook-teams';
+import { loopOverCustomCollection } from '../../utils/migrations';
+import logger from '../../utils/logger';
+import { algoliaApiKey, algoliaAppId, algoliaIndex } from '../../config';
+
+export const indexTeamsUsersHandler =
```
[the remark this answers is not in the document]
niezgoda: I would like to - however - this might vary in the future as we are going to add more entities to search by.

apps/asap-server/src/controllers/users.ts
```diff
@@ -226,6 +233,49 @@ export default class Users implements UserController {
     return parseGraphQLUser(items[0]);
   }
 
+  async fetchByRelationship(
```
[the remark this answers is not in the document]
niezgoda: I think that refactoring of fetch method would be painful :(

apps/asap-server/src/utils/migrations.ts
peterstarling: migration should imho be part of squidex package except then we wouldn't be able to use the controllers (maybe we shouldn't, dont know). Can you move it to `utils`?

apps/asap-server/src/controllers/users.ts
```diff
@@ -31,6 +31,7 @@ export interface UserController {
   fetch(options: FetchOptions): Promise<ListUserResponse>;
   fetchById(id: string): Promise<UserResponse>;
   fetchByCode(code: string): Promise<UserResponse>;
+  fetchByLabId(labId: string, options: FetchOptions): Promise<ListUserResponse>;
```
kmaid, line 34: This seems like a strange typing. Are we returning a ListUserResponse for Labs?

peterstarling (changes requested):

apps/asap-server/src/controllers/users.ts
```diff
@@ -226,6 +227,41 @@ export default class Users implements UserController {
     return parseGraphQLUser(items[0]);
   }
 
+  async fetchByLabId(
```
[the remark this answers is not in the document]
niezgoda: @kmaid @marot what do you think?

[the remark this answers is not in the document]
kmaid: I think having one endpoint that can handle complex queries would be better. We might be able to reuse different combinations of filters etc to solve future requirements rather than having to create a brand new endpoint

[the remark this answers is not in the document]
kmaid: I also think we really should have documentation for our endpoints and what filters they take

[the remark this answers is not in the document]
marot: It looks like `fetchByLabId` takes the `FetchOptions` parameter but only uses it partially. I think it makes sense
to integrate this lab filter into the fetch function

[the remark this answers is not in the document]
peterstarling: yeah totally agree (which might be obvious since i was the one to suggest it) but with one remark - you guys do realise that  controllers != endpoints though right? What i think we want to document are http endpoints since this is the interface the API exposes. But yeah eventually this comes down to using whatever the controllers offer

apps/asap-server/test/handlers/lab/index-users-by-lab-handler.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      await indexHandler(event);
+
+      expect(algoliaSearchClientMock.batch).toHaveBeenCalledWith(
+        usersBatchResponse,
+      );
+    },
+  );
+
+  describe('Should process the events, handle race conditions and not rely on the order of the events', () => {
+    test.concurrent.each(possibleRacingConditionEvents)(
```
peterstarling, line 96: to be fair although there does seem to be a pattern i dont think we should use `each` for the race conditions. I personally find this completely unreadable. If there are really that many repetitions then maybe only test fewer cases?

apps/asap-server/test/handlers/teams/index-users-by-team-handler.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  ['deleted and created', deleteEvent, createEvent],
+
+  ['updated and deleted', updateEvent, deleteEvent],
+  ['deleted and updated', deleteEvent, updateEvent],
+
+  ['updated and unpublished', updateEvent, unpublishedEvent],
+  ['unpublished and updated', unpublishedEvent, updateEvent],
+];
+
+describe('Index Users on Team event handler', () => {
```
peterstarling, line 57: with the concurrent tests i no longer know whats going on so i need to ask - is there a test that shows what happens when only some of the `Promise.allSettled` promises actually settle?

packages/algolia/src/client.ts
```diff
@@ -10,6 +15,11 @@ export type EntityResponses = {
   [USER_ENTITY_TYPE]: UserResponse;
 };
 
+export type BatchRequest = {
+  action: BatchActionType;
```
peterstarling, line 19: im not sure we want to expose Algolia's implementation like that here? After all the purpose of this client is to abstract away. Perhaps we want this to become a `saveMany` method or something like that instead?

packages/algolia/test/client.test.ts
```diff
@@ -15,7 +17,33 @@ describe('Algolia Search Client', () => {
   const algoliaSearchIndex = getAlgoliaSearchIndexMock();
   const algoliaSearchClient = new AlgoliaSearchClient(algoliaSearchIndex);
 
-  test('Should save the Research Output', async () => {
+  it('Should do batch on entities', async () => {
+    await algoliaSearchClient.batch([
```
peterstarling, line 21: since it's a batch request i would test it with an actual batch (say 2 or 3 elements), not just one element in that array

packages/algolia/test/client.test.ts
```diff
@@ -27,7 +55,7 @@ describe('Algolia Search Client', () => {
     });
   });
 
-  test('Should save the User', async () => {
+  it('Should save the User', async () => {
```
peterstarling, line 58: test

apps/asap-server/src/handlers/teams/index-users-by-team-handler.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      logger.info(`Fetched ${batchRequests.length} users.`);
+
+      const algoliaResponse = await algoliaClient.batch(batchRequests);
+
+      logger.info(
+        `Updated ${batchRequests.length} users with algolia response: ${algoliaResponse}`,
+      );
+    } catch (error) {
+      if (error?.output?.statusCode === 404) {
```
peterstarling, line 57: what exactly are you catching here? this would be a 404 of what?

serverless.ts
```diff
@@ -454,7 +493,27 @@ const serverlessConfig: AWS = {
       },
     },
     indexTeamResearchOutputs: {
-      handler: 'apps/asap-server/src/handlers/teams/index-handler.handler',
+      handler:
+        'apps/asap-server/src/handlers/teams/research-outputs-handler.handler',
```
peterstarling, line 497: should be index-team-reasearch-outputs maybe?

serverless.ts
```diff
@@ -454,7 +493,27 @@ const serverlessConfig: AWS = {
...
+        },
+      ],
+      environment: {
+        ALGOLIA_API_KEY: `\${ssm:algolia-index-api-key-${envAlias}}`,
+        ALGOLIA_INDEX: `asap-hub_research_outputs_${envRef}`,
+      },
+    },
+    indexTeamUsers: {
+      handler:
+        'apps/asap-server/src/handlers/teams/index-users-by-team-handler.handler',
```
peterstarling, line 516: should be index-team-users maybe?

peterstarling (changes requested):

apps/asap-server/test/controllers/users.test.ts
```diff
@@ -495,7 +496,7 @@ describe('Users controller', () => {
       expect(result.contactEmail).not.toBeDefined();
     });
 
-    test('Should update social and questions', async () => {
+    it('Should update social and questions', async () => {
```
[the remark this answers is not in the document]
niezgoda: :) okey

apps/asap-server/test/controllers/users.test.ts
```diff
@@ -846,4 +847,75 @@ describe('Users controller', () => {
...
+      mockResponse.queryUsersContentsWithTotal!.total = 0;
+      squidexGraphqlClientMock.request.mockResolvedValueOnce(mockResponse);
+
+      const result = await usersMockGraphqlClient.fetch({
+        filter: { labId: ['not-found-lab-id'] },
+      });
+      expect(result).toEqual({ items: [], total: 0 });
+    });
+
+    it('Should query with labId filter and return the users', async () => {
```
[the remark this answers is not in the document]
niezgoda: it is in the fetch test already

apps/asap-server/src/utils/migrations.ts
niezgoda: but it is in utils :D

packages/algolia/src/client.ts
niezgoda: There is such feature in Algolia - like save many, (as far as I remember).

peterstarling: i dont think you understand what i mean - at the moment what we want is an ability to save multiple objects yeah?

We built an abstraction to hide some of the implementation details to make the interface more dev friendly for us. So we shouldn't have to expose an `action` for a `batch` request and instead have a method that enables us to save multiple objects. Underneath it should call the batch method, imho

packages/algolia/test/client.test.ts
```diff
@@ -15,7 +17,37 @@ describe('Algolia Search Client', () => {
   const algoliaSearchIndex = getAlgoliaSearchIndexMock();
   const algoliaSearchClient = new AlgoliaSearchClient(algoliaSearchIndex);
 
-  test('Should save the Research Output', async () => {
+  it('Should do batch on entities', async () => {
```
peterstarling, line 20: `test`

but also - this test already suggests that maybe we dont need a `batch` method, just a `saveMany` method? This seems quite generic in comparison to what we have built for the algolia client.

serverless.ts
```diff
@@ -468,7 +503,26 @@ const serverlessConfig: AWS = {
       ],
       environment: {
         ALGOLIA_API_KEY: `\${ssm:algolia-index-api-key-${envAlias}}`,
-        ALGOLIA_INDEX: `asap-hub_research_outputs_${envRef}`,
+        ALGOLIA_INDEX: `asap-hub_${envRef}`,
```
peterstarling, line 506: well thats not quite right. You are switching to the new index which means we will no longer sync the old one, i think this should be done once we move over @marot might want to take a look at this

peterstarling (approved):

peterstarling (approved):

PR #1331, opened 2022-02-10
Build Docker Images

markcsims:
https://asaphub.atlassian.net/browse/CRN-655


.github/workflows/build-images.yml
```diff
@@ -0,0 +1,42 @@
...
+          docker tag $BASE_IMAGE_NAME:$IMAGE_TAG $BASE_IMAGE_NAME:latest
+          docker push --all-tags $BASE_IMAGE_NAME
+      - name: Build and push integration
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: asap-hub/node-python-sq
+          IMAGE_TAG: ${{ github.sha }}
+        run: |
+          BASE_IMAGE_NAME=$ECR_REGISTRY/$ECR_REPOSITORY
+          docker build -t $BASE_IMAGE_NAME:$IMAGE_TAG . -f ci/Dockerfile --target integration
```
marot, line 40: Can we rename $BASE_IMAGE_NAME to $IMAGE_NAME or $INTEGRAITON_IMAGE_NAME or something similar? I was a little bit confused because the other image has `--target base` and I thought $BASE_IMAGE_NAME is related to that.

.github/workflows/dev-backup.yml
```diff
@@ -4,10 +4,28 @@ on:
     - cron: '0 7 * * MON-FRI' # At 07:00 UTC on every day-of-week from Monday through Friday.
   workflow_dispatch:
 jobs:
+  setup:
```
peterstarling, line 7: so does this mean that every step that uses the ECR image needs to have a setup? and the purpose of the setup isn't just to output the image path but to log docker into ECR or something like that, right?

do we need a step to output the image path btw - like can we not just hardcode it like we did with gitlab or dockerhub?

.gitlab-ci.yml
```diff
@@ -23,9 +23,8 @@ include:
 variables:
   ASAP_HOSTNAME: hub.asap.science
   PLAYWRIGHT_BROWSERS_PATH: 0
-  BASE_NODE_IMAGE: $CI_REGISTRY/$CI_PROJECT_PATH/node:14.17-alpine
 
-image: $BASE_NODE_IMAGE
+image: node:14.17-alpine
```
peterstarling, line 27: we might run into docker limits again with this, no?

.gitlab-ci.yml
```diff
@@ -115,6 +114,22 @@ prepare:squidex:branch:
     - mkdir -p first_run
     - touch first_run/squidex_app_create
 
+sync:test:pr:
```
peterstarling, line 117: is this a new job? why is it here if i may ask?

peterstarling (changes requested):
this is not so much a change request but more just a list of questions

.github/workflows/build-images.yml
markcsims: Done

.gitlab-ci.yml
markcsims: Thanks! I've removed, we added this together as a test.

.github/actions/ecr-registry/action.yml
```diff
@@ -0,0 +1,48 @@
+name: 'ECR Repository Name'
+description: 'Outputs the name of the image'
```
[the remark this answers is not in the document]
markcsims: It's only purpose is to output the name of the image.

.github/workflows/dev-backup.yml
markcsims: The images in gitlab are hardcoded using environment variables. So it's configurable in 1 place.

We tried using secrets in GitHub i.e. 
```
    container:
      image: ${{ secrets.IMAGE_SQUIDEX_UTILS }}
```
But get get:

```
The workflow is not valid. .github/workflows/dev-backup.yml (Line: 10, Col: 14): Unrecognized named-value: 'secrets'. Located at position 1 within expression
```

The setup step was an attempt to avoid pasting the location of the images in multiple places. Downside, it will be needed in each step that uses it.

.gitlab-ci.yml
markcsims: Now reading this from ECR.

.github/workflows/dev-backup.yml
markcsims: Instead of requesting the account id via the AWS credentials, we're now hardcoding the account id in the action.

This reduces the runtime from 15 to 8 seconds.

PR #1334, opened 2022-02-14
CRN-645 move user search from squidex to algolia

marot:


.github/workflows/dev-backup.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       AWS_DEFAULT_REGION: 'us-east-1'
-      ALGOLIA_INDEX: 'asap-hub_research_outputs_dev'
+      ALGOLIA_INDEX: 'asap-hub_dev'
```
niezgoda, line 53: this should be also changed with feature flag

apps/asap-server/src/autogenerated-gql/graphql.ts
```diff
@@ -1375,7 +1377,7 @@ export type Asset = {
   fileType: Scalars['String'];
   /** The version of the file. */
   fileVersion: Scalars['Int'];
-  /** The ID of the object. */
+  /** The ID of the object (usually GUID). */
   id: Scalars['String'];
```
niezgoda, line 1381: I would not say that.
After migration to other CMS this might change :)

apps/frontend/src/config.ts
```diff
@@ -4,7 +4,7 @@ export const API_BASE_URL =
 export const ALGOLIA_APP_ID =
   process.env.REACT_APP_ALGOLIA_APP_ID || 'LVYWOPQ0A9';
 export const ALGOLIA_INDEX =
-  process.env.REACT_APP_ALGOLIA_INDEX || 'asap-hub_research_outputs_dev';
+  process.env.REACT_APP_ALGOLIA_INDEX || 'asap-hub_dev';
```
niezgoda, line 7: This should be behind feature flag

niezgoda (changes requested):

.github/workflows/dev-backup.yml
marot: @niezgoda thats only for the creation of the dev backup. There is no way to change the CI script with a feature flag, right?

apps/asap-server/src/autogenerated-gql/graphql.ts
marot: should be reverted now. `schema:update` did weird things...

apps/frontend/src/config.ts
marot: The default value is only for local development - so this should be fine, no?

.github/workflows/dev-backup.yml
niezgoda: nope, you could use SSM.
@peterstarling - shall we change it or use SSM?

PR #1336, opened 2022-02-15
CRN.627 add labs to form

jhonny-m:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/8?modal=detail&selectedIssue=CRN-627


apps/frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -60,6 +67,11 @@ const TeamOutput: React.FC<TeamOutputProps> = ({ teamId }) => {
     description: 'example',
     tags: [],
   };
+  const authorization = useRecoilValue(authorizationState);
```
kmaid, line 70: Are we loading all labs into memory?

jhonny-m: yeah

niezgoda:
@jhonny-m please add Jira ticket ID to the title and some description would be great !

kmaid:
Do we need the Jira ID anymore now we are linking ticket ids in branch names? @niezgoda

niezgoda:
> Do we need the Jira ID anymore now we are linking ticket ids in branch names? @niezgoda

@kmaid I would like to have consistent naming convention :)

packages/react-components/src/atoms/AsyncMultiSelect.tsx
```diff
@@ -0,0 +1,77 @@
...
+  id,
+  enabled = true,
+  placeholder = '',
+  noOptionsMessage,
+  values = [],
+  onChange = noop,
+}) => {
+  const [inputValues, setInputValues] = useState<OptionValue[]>(values);
+  const [validationMsg, setValidationMsg] = useState('');
+  let inputRef: AsyncSelect<OptionTypeBase, true> | null;
```
marot, line 41: Maybe use ``useRef`` here?

packages/react-components/src/molecules/LabeledMultiSelect.tsx
```diff
@@ -16,15 +17,40 @@ type LabeledMultiSelectProps = {
   readonly title: ReactNode;
   readonly subtitle?: React.ReactNode;
   readonly description?: ReactNode;
-} & Exclude<ComponentProps<typeof MultiSelect>, 'id'>;
-const LabeledMultiSelect: React.FC<LabeledMultiSelectProps> = ({
+};
+export type SyncLabeledMultiSelectProps = {
```
marot, line 21: We can add a helper method `isSync(data: Sync... | Async...): data is Sync { return data?.async === false }` and use that in the if in `forContent`. Then we can avoid the type casts (I think)

marot (approved):

packages/react-components/src/molecules/LabeledMultiSelect.tsx
jhonny-m: We are going to have another PR to refactor this 👍 
cc @kmaid

PR #1343, opened 2022-02-16
CRN-645 move user search from squidex to algolia

marot:


apps/frontend/src/network/users/__tests__/UserList.test.tsx
```diff
@@ -18,8 +20,22 @@ jest.mock('../../teams/api');
 jest.mock('../../groups/api');
 
 const mockGetUsers = getUsers as jest.MockedFunction<typeof getUsers>;
+const mockGetUsersWithAlgolia = getUsersWithAlgolia as jest.MockedFunction<
+  typeof getUsersWithAlgolia
+>;
+
+const renderUserList = async (opts: { algoliaEnabled: boolean }) => {
```
kmaid, line 27: I don't really like this because when the feature flag is removed we will have to change every test. Ideally they should not be changed and there should only be tests & code removed

marot: @kmaid That is two tests and we only need to remove the options passed in the `renderUserList` - I don't see the difference between removing setup code in the test vs removing setup options.

peterstarling (changes requested):
you're gonna want to update the serverless configuration - unfortunately the indexes are hardcoded in there - but basically we want to handlers to start syncing to the new index. Or you can duplicate the handler to maintain both. Or have the handlers support multiple indexes, your call.

apps/frontend/src/network/users/__tests__/api.test.ts
```diff
@@ -63,6 +69,117 @@ describe('getUsers', () => {
...
+    currentPage: null,
+    filters: new Set(),
+  };
+
+  beforeEach(() => {
+    searchEntity.mockReset();
+    searchEntity.mockResolvedValue({
+      hits: [],
+      nbHits: 0,
+    } as unknown as SearchEntityResponse<'user'>);
```
kmaid, line 92: Can we use createAlgoliaResponse

apps/frontend/src/network/users/api.ts
```diff
@@ -30,6 +31,23 @@ export const getUser = async (
   return resp.json();
 };
 
+export const getUsersWithAlgolia = async (
```
kmaid, line 34: can we call this getUsers and the legacy function getUsersLegacy ?

apps/frontend/src/network/users/__tests__/api.test.ts
marot: Yes!

apps/frontend/src/network/users/api.ts
marot: Yes I will change it 👍

marot:
> you're gonna want to update the serverless configuration - unfortunately the indexes are hardcoded in there - but basically we want to handlers to start syncing to the new index. Or you can duplicate the handler to maintain both. Or have the handlers support multiple indexes, your call.

@peterstarling Looking into this atm:
- I'll remove dev-algolia-sync.yml and prod-algolia-sync.yml
- Remove all references to ALGOLIA_RESEARCH_OUTPUT_INDEX and replace them with ALGOLIA_INDEX
- Verify that search of research outputs is still working (We don't have any other usae of algolia, right?)
- Maybe refactor prod-algolia-sync to use a common action?

Does that sound good to you?

apps/frontend/src/network/users/__mocks__/api.ts
```diff
@@ -35,10 +35,11 @@ export const postUserAvatar = jest.fn(
   },
 );
 
-export const getUsers = jest.fn(
+export const getUsersLegacy = jest.fn(
   async ({ pageSize }: GetListOptions): Promise<ListUserResponse> =>
     createListUserResponse(pageSize ?? 10),
 );
+export const getUsers = jest.fn(() => createListUserResponse(10));
```
kmaid, line 42: Please type this

kmaid (approved):

apps/frontend/src/network/users/__mocks__/api.ts
marot: yes 👍

niezgoda (approved):

niezgoda (approved):

PR #1345, opened 2022-02-17
Feat/CRN-715 Manually sync all external authors in Algolia

niezgoda:
Requirements:
- all items from the external author collection should be accurately shown in the Author Selector on the ROMS form
- make it possible to sync all external authors manually (expand GitHub actions)


apps/asap-server/src/queries/externa-authors.queries.ts
```diff
@@ -0,0 +1,39 @@
+import { gql } from 'graphql-tag';
```
peterstarling, line 1: invalid file name - it should be `external`

apps/asap-server/test/controllers/external-authors.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    test('Should fetch the users from squidex graphql', async () => {
+      const result = await usersMockGraphqlServer.fetch({});
+
+      expect(result).toMatchObject({
+        total: 8,
+        items: [getExternalAuthorResponse(), getExternalAuthorResponse()],
+      });
+    });
+
+    test('Should return an empty result', async () => {
```
peterstarling, line 33: you should add the one about sending the correct parameters (ie take and skip)

peterstarling (changes requested):

marot:
@niezgoda You can't satisfy the first requirement because there is no authors section in the ROMS form yet, right?
I think we should have tackled https://asaphub.atlassian.net/browse/CRN-611 before this - should we split this requirement into a separate issue?

niezgoda:
> @niezgoda You can't satisfy the first requirement because there is no authors section in the ROMS form yet, right? I think we should have tackled https://asaphub.atlassian.net/browse/CRN-611 before this - should we split this requirement into a separate issue?

This is only manual sync between Squidex and Algolia.
There is no front part in this ticket.

peterstarling:
> @niezgoda You can't satisfy the first requirement because there is no authors section in the ROMS form yet, right? I think we should have tackled https://asaphub.atlassian.net/browse/CRN-611 before this - should we split this requirement into a separate issue?

I'm with @marot on this one - although I don't mind merging this, doing the tests ourselves (by looking at algolia index) and then waiting for the other ticket to unblock it so it could go through AC.

peterstarling (approved):

peterstarling (approved):

PR #1349, opened 2022-02-18
feat: update Events UI

ana-yldio:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/8?modal=detail&selectedIssue=CRN-721

### DIFF
--- a/packages/react-components/src/templates/EventPage.tsx
+++ b/packages/react-components/src/templates/EventPage.tsx
@@ -33,6 +33,7 @@ type EventPageProps = ComponentProps<typeof EventInfo> &
     | 'videoRecording'
     | 'presentation'
     | 'meetingMaterials'
+    | 'hideMeetingLink'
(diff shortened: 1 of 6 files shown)


niezgoda (approved):

packages/react-components/src/templates/EventPage.tsx
```diff
@@ -42,6 +43,7 @@ const EventPage: React.FC<EventPageProps> = ({
   backHref,
   lastModifiedDate,
   calendar,
+  hideMeetingLink,
```
marot, line 46: Personally, I'd prefer to pass `JoinEvent: ReactNode` here - and then pass `null` if we don't want to render this.

The way this component is set up, it takes the props of all sub-components. If that is the case, I'd rather switch to passing the sub-components directly.

But I think that might be out of scope for this issue.

marot (approved):

PR #1350, opened 2022-02-18
CRN-620 remove old research output index

marot:
The CI-1350 search index contains users and research outputs
The PR environment only lists shared outputs in the shared outputs search, as intended.

We can't verify dev/prod envs before merging

Please also check if I missed any usage of research_outputs index


peterstarling (approved):
@marot you might want to consider renaming the -v2 prod and dev sync (simply drop the `v2` suffx). 

Also I would recommend that you clone the current prod and dev index. Once you do that we should run the sync on both and make sure it still works as normal. If something goes wrong we can rename the indexes until we revert so its not a big deal

marot:
@peterstarling To be sure, I should duplicate/clone the indexes, merge the PR, run the migration, and if there is a problem delete the new prod/dev indexes and rename the duplicated on back to prod/dev, right?

marot:
@peterstarling after thinking a little about it, I'd prefer to keep the v2 suffix for a little bit. (So that it is clear what workflow is being triggered in the github actions) Otherwise the bahaviour of the workflow might differ depending on what branch you select - when triggering actions manually)

.github/workflows/algolia-sync-v2.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Copy index without data to a temporary index (ALL)
+        if: ${{ inputs.entity == 'all'}}
+        run: yarn workspace @asap-hub/algolia run algolia transferindexconfig -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_INDEX -d $ALGOLIA_APP_ID -y $ALGOLIA_API_KEY -i $ALGOLIA_INDEX_TEMP
+      - name: Copy index with data to a temporary index (ENTITY)
+        if: ${{ inputs.entity != 'all'}}
+        run: yarn workspace @asap-hub/algolia run algolia transferindex -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_INDEX -d $ALGOLIA_APP_ID -y $ALGOLIA_API_KEY -i $ALGOLIA_INDEX_TEMP
+      - name: Remove the entity data (ENTITY)
+        if: ${{ inputs.entity != 'all'}}
+        env:
+          ENTITY_TYPE: ${{ (inputs.entity == 'users' && 'user') || (inputs.entity == 'research-outputs' && 'research-output') || 'external-author' }}
```
niezgoda, line 64: `${{ (inputs.entity == 'users' && 'user') || (inputs.entity == 'research-outputs' && 'research-output') || (inputs.entity == 'external-authors' && 'external-author') }}`

niezgoda (changes requested):

PR #1353, opened 2022-02-21
Feature/CRN-714-update-validation-library

peterstarling:
## Notable changes
AJV assumes that once a property is marked as optional it should also accept null. This means we have to make every optional parameter nullable as well which in turn affects the types. For this purpose I made our user patch endpoint accept nulls but decided to filter them out much like the optional properties were previously


apps/asap-server/src/entities/user.ts
```diff
@@ -30,7 +30,7 @@ export const userUpdateSchema = Joi.object({
   lastName: Joi.string().allow(''),
   jobTitle: Joi.string().allow(''),
   degree: Joi.string()
-    .allow('BA', 'BSc', 'MSc', 'PhD', 'MD', 'PhD, MD')
+    .valid('BA', 'BSc', 'MSc', 'PhD', 'MD', 'PhD, MD')
```
kmaid, line 33: It be nice if we could take these values from the model. Ive been turning the enums into arrays or sets and then dynamically generating a type from it for the model and using the values where appropirate

apps/asap-server/src/entities/user.ts
```diff
@@ -40,7 +40,7 @@ export const userUpdateSchema = Joi.object({
   expertiseAndResourceDescription: Joi.string().allow(''),
   researchInterests: Joi.string().allow(''),
   responsibilities: Joi.string().allow(''),
-  reachOut: Joi.string().allow('').max(250),
+  reachOut: Joi.string().allow('').max(250), //
```
kmaid, line 43: Random //

apps/asap-server/src/validation/user.validation.ts
```diff
@@ -0,0 +1,90 @@
...
+  properties: {
+    jobTitle: { type: 'string', nullable: true },
+    onboarded: { type: 'boolean', nullable: true },
+    contactEmail: { type: 'string', nullable: true },
+    firstName: { type: 'string', nullable: true },
+    lastName: { type: 'string', nullable: true },
+    degree: {
+      type: 'string',
+      enum: [
+        'BA',
```
kmaid, line 16: Same again here if we could just use the model array of these in a spread with the extra options so we can change it one place and have it filter through to where it needs to be

apps/asap-server/src/validation/research-output.validation.ts
```diff
@@ -0,0 +1,80 @@
...
+      usedInPublication: { type: 'boolean', nullable: true },
+      addedDate: { type: 'string' },
+      teamId: { type: 'string' },
+      labs: {
+        type: 'array',
+        items: { type: 'string' },
+        nullable: true,
+      },
+    },
+    required: [
```
niezgoda, line 62: Can we cross-check it with squidex schema requirements? - some test would be useful.

niezgoda (changes requested):

packages/model/src/user.ts
```diff
@@ -139,15 +139,13 @@ export type UserMetadataResponse = Omit<UserResponse, 'labs'> & {
...
-  displayName?: string;
   contactEmail?: string;
   firstName?: string;
   lastName?: string;
   biography?: string;
   jobTitle?: string;
-  orcid?: string;
   institution?: string;
-  degree?: UserDegree;
+  degree?: UserDegree | '';
```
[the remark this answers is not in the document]
peterstarling: we don't but that's what we have been accepting so far, according to the validation, the types weren't aligned with it though

apps/asap-server/src/validation/research-output.validation.ts
marot: As discussed in person, this would be OOS here

packages/model/src/user.ts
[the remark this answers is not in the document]
marot: bump @niezgoda is the answer acceptable?

niezgoda (approved):

PR #1356, opened 2022-02-22
Feat/CRN-683-keep-external-authors-in-sync-in-algolia

niezgoda:
requirements

any changes made to any external author on the CMS should be “immediately” reflected on the ROMS form:
- name changes
- publication state changes (to draft, published, archived)
- is deleted or created

OOS:
- make the webhook universal (refactor)


apps/asap-server/src/handlers/external-author/index-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+import {
+  ExternalAuthorEventType,
+  SquidexWebhookExternalAuthorPayload,
+} from '../webhooks/webhook-external-author';
+import { EventBridgeHandler } from '../../utils/types';
+import { algoliaApiKey, algoliaAppId, algoliaIndex } from '../../config';
+
+export const indexExternalAuthorHandler =
+  (
+    externalauthorController: ExternalAuthorsController,
```
marot, line 20: NIT: I'd prefer externalAuthorController

apps/asap-server/src/handlers/external-author/index-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+      await algoliaClient.saveEntity({
+        id: event.detail.payload.id,
+        type: 'external-author',
+        data: externalauthor,
+      });
+
+      logger.debug(`Saved external author  ${externalauthor.displayName}`);
+    } catch (e) {
+      if (e?.output?.statusCode === 404) {
+        await algoliaClient.remove(event.detail.payload.id);
```
marot, line 45: Is it guaranteed that ids are unique across users, research outputs and external authors?

packages/algolia/src/client.ts
```diff
@@ -39,6 +48,14 @@ export class AlgoliaSearchClient {
     await this.index.saveObject(AlgoliaSearchClient.getAlgoliaObject(payload));
   }
 
+  async saveEntity({ id, type, data }: Entity): Promise<void> {
```
marot, line 51: Why can't we use save?

niezgoda: I could - this needs to be refactored.
We shouldn't check entity by field "existence".
We should provide it explicitly.

apps/asap-server/src/handlers/external-author/index-handler.ts
niezgoda: That is good question.
Maybe @peterstarling @kmaid @markcsims know the answers. I do not

apps/asap-server/test/controllers/external-authors.test.ts
```diff
@@ -76,4 +77,30 @@ describe('External Authors controller', () => {
       );
     });
   });
+
+  describe('FetchById', () => {
+    test('Should fetch the users from squidex graphql', async () => {
```
peterstarling, line 82:
```suggestion
    test('Should fetch the user from squidex graphql', async () => {
```

packages/algolia/src/client.ts
peterstarling: and why is that? with this method now am i able to save an entity 'research-output' with any arbitrary data? looks like it - judging by the type https://github.com/yldio/asap-hub/blob/56f9d41f8bfc9ca8aea35fee861c859ac60bfc11/packages/algolia/src/client.ts#L39
I think we shouldnt be allowing for any random entity data to be stored and remove type-safety like that - if we are going to we might as well use algolia client directly

peterstarling (changes requested):

apps/asap-server/src/handlers/external-author/index-handler.ts
kmaid: Squidex IDs are GUIDs so I guess so?

apps/asap-server/src/handlers/webhooks/webhook-user.ts
```diff
@@ -59,5 +66,19 @@ const getEventType = (customType: string): UserEventType | undefined => {
...
+  type: UserSquidexEventType;
+  timestamp: string;
+  payload: {
+    $type: 'EnrichedContentEvent';
+    type: 'Published' | 'Updated' | 'Unpublished' | 'Deleted' | 'Created';
+    id: string;
+    created: string;
+    lastModified: string;
+    version: number;
+    data: { [x: string]: { iv: unknown } | null };
```
kmaid, line 79: Could we not use Partial<UserResponse> or something?

niezgoda:
reopened

packages/algolia/src/client.ts
```diff
@@ -21,29 +30,32 @@ export type EntityRecord<T extends keyof EntityResponses> =
...
 
-export const getEntityType = (
-  entity: EntityResponses[keyof EntityResponses],
-): keyof EntityResponses => {
-  if ('title' in entity && 'sharingStatus' in entity) {
-    return 'research-output';
-  }
-
-  return 'user';
+export type Payload<T extends EntityResponses[keyof EntityResponses]> = {
```
peterstarling, line 33: yeah that wont work - i shouldnt be able to do this:
```typescript
const client = new AlgoliaSearchClient('asdas' as any);

client.save({ data: createResearchOutputResponse(), type: 'external-author' });
```

yet the definition for the `save` method allows it. If we need to be explicit we need to tie the given `type` to the `data`.

packages/algolia/src/client.ts
```diff
@@ -1,13 +1,22 @@
...
+  ExternalAuthorResponse,
+  LabResponse,
+  ResearchOutputResponse,
+  UserResponse,
+} from '@asap-hub/model';
 
 export const RESEARCH_OUTPUT_ENTITY_TYPE = 'research-output';
 export const USER_ENTITY_TYPE = 'user';
+export const EXTERNAL_AUTHOR_ENTITY_TYPE = 'external-author';
+export const LAB_ENTITY_TYPE = 'Lab';
```
peterstarling, line 13: if anything it should be
```suggestion
export const LAB_ENTITY_TYPE = 'lab';
```

apps/asap-server/test/mocks/algolia-client.mock.ts
```diff
@@ -2,6 +2,7 @@ import { AlgoliaSearchClient } from '@asap-hub/algolia';
 
 export const algoliaSearchClientMock = {
   save: jest.fn(),
+  saveEntity: jest.fn(),
```
peterstarling, line 5: yeah i dont think thats necessary anymore?

peterstarling (changes requested):

packages/algolia/src/client.ts
niezgoda: updated

marot (commented):
The changes since my last review LGTM. I'll let @markcsims approve since it is a backend PR

PR #1371, opened 2022-02-25
CRN-732-refactor-multi-select-component

kmaid:


packages/react-components/src/organisms/__tests__/TeamCreateOutputContributorCard.test.tsx
```diff
@@ -5,73 +5,71 @@ import { ComponentProps } from 'react';
...
+    const loadOptions = jest.fn();
+    loadOptions.mockReturnValue(new Promise((resolve) => resolve([])));
+    const { getByLabelText, queryByText } = render(
+      <TeamCreateOutputContributorsCard
+        {...props}
+        labSuggestions={loadOptions}
+      />,
+    );
+    userEvent.click(getByLabelText(/Labs/i));
+    await waitFor(() =>
```
markcsims, line 70: Is it possible to use the waitForElementToBeRemoved method here? Is there any advantage in doing so?

markcsims (approved):
Looks good to me.  I like how the MultiSelect looks.

One general comment about RTL and async. You'd know better than me.

We're using waitFor, could we use waitForElementToBeRemoved or find for example?

PR #1389, opened 2022-03-02
feat/Deploy for PRs

markcsims:
https://asaphub.atlassian.net/browse/CRN-571

- [x] yarn sls deploy
- [x] e2e branch
- [x] verify branch
- [x] verify schema
- [x] define pipeline dependancies
- [x] Teardown sls 
- [x] Teardown environments left behind
- [x] Test Calendar
- [x] Test Squidex Data
- [x] Check env with no gitlab
- [x] Remove AWS_ECR_PASSWORD secret


apps/asap-cli/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "composite": true,
```
peterstarling, line 4: why do we need it?

apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
```diff
@@ -197,6 +197,14 @@ describe('Fetch-user-by-code handler', () => {
       });
 
       test('should ask for a key with expiration which is 601 minutes (10 hours 1 minute) ahead of now', async () => {
+        const dateReference = '2021-07-06T09:21:23.000Z';
```
peterstarling, line 200: strange that this is part of this PR no?

.github/workflows/build-pipeline.yml
```diff
@@ -347,41 +449,176 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-      - uses: ./.github/actions/slack/
-        with:
-          message: 'Build Failed on GitHub: ${{ needs.setup.outputs.pr-number || github.head_ref }}'
-          webhook: ${{ secrets.SLACK_WEBHOOK }}
-          status: failure
+        uses: actions/checkout@v3
+      # - uses: ./.github/actions/slack/
```
peterstarling, line 593: pretty sure we want this to happen no?

peterstarling (dismissed):

apps/asap-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
markcsims: Flaky test, decided to fix it.

.github/workflows/build-pipeline.yml
markcsims: I got asked to comment it out while we are developing, there was too much noise in the channel.

apps/asap-cli/tsconfig.json
markcsims: The "build-typecheck" step, generates .d.ts files for non-composite projects.

These types are then used in the "build sls package" and "build frontend".

asap-cli for example, doesn't need it's types build but it does need to be "typecheck-ed". Which is now done in the "typecheck" step.

.github/workflows/build-pipeline.yml
markcsims: I'll add it back, but we may need to re-think how and when we notify Slack.

peterstarling: ahh i see, why is there so much noise though, i wonder? we do not notify in slack when the PR pipeline fails in gitlab, do we? in which case you probably want to remove it entirely, no?

markcsims: Removed!

kmaid (approved):

PR #1391, opened 2022-03-02
CRN-615-add-team-selector-to-the-form

kmaid:


apps/asap-server/src/controllers/research-outputs.ts
```diff
@@ -136,20 +136,24 @@ export default class ResearchOutputs implements ResearchOutputController {
...
   async create({
-    teamId,
+    teams,
     ...researchOutputData
   }: ResearchOutputInputData): Promise<Partial<ResearchOutputResponse>> {
     const { id: researchOutputId } = await this.createResearchOutput(
       researchOutputData,
     );
-    await this.associateResearchOutputToTeam(teamId, researchOutputId);
+    await Promise.all(
```
niezgoda, line 145: What about retries?
Could we add here code to retry this assignment on failure?

apps/frontend/src/network/teams/state.ts
```diff
@@ -142,11 +142,25 @@ export const usePostTeamResearchOutput = () => {
...
-      await getLabs(
-        { searchQuery, filters: new Set(), currentPage: null, pageSize: null },
-        authorization,
-      )
-    ).items.map(({ id, name }) => ({ label: `${name} Lab`, value: id }));
+  return (searchQuery: string) =>
+    getLabs(
+      { searchQuery, filters: new Set(), currentPage: null, pageSize: null },
+      authorization,
+    ).then(({ items }) =>
```
niezgoda, line 149: good catch :)

niezgoda (changes requested):

apps/asap-server/src/controllers/research-outputs.ts
markcsims: I'm not sure there's much we can do here. The RO needs to be associated with a team, and this much occurs after the RO is created. 

In terms of retrying, how long do we retry? How many times? The user is waiting for this to return before they can continue, should we take the create event asynchronous and return straight away? What happens if the create event continues to fail, should we DLQ?

There are no transactions in Squidex, (as far as I know) so we can't simply rollback.

We could "manually" rollback, i.e. delete the RO we've just created. But what happens if that delete fails? Do we have to retry, for how long? Do we DLQ or put the event somewhere?

packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -20,27 +20,36 @@ const containerStyles = css({
...
   | null;
 
 type MultiSelectProps = {
   readonly customValidationMessage?: string;
   readonly id?: string;
   readonly enabled?: boolean;
   readonly placeholder?: string;
-  readonly onChange?: (newValues: OptionsType<OptionTypeBase>) => void;
-  readonly values?: OptionsType<OptionTypeBase>;
+  readonly onChange?: (newValues: OptionsType<MultiSelectOptionsType>) => void;
```
marot, line 39: Personally, I would prefer to expose `onChange` and `components` and let clients overwrite the behaviour they want, instead of adding the fixed behaviour to all intances of this component.

marot (approved):

PR #1393, opened 2022-03-02
Start adding validation tests

marot:
Add tests to verify that refactoring in issue https://asaphub.atlassian.net/browse/CRN-714 works


apps/asap-server/test/routes/user.route.test.ts
```diff
@@ -95,6 +95,16 @@ describe('/users/ route', () => {
     });
 
     describe('Parameter validation', () => {
+      test('Should return a 400 error when additional properties exist', async () => {
+        const response = await supertest(appWithMockedAuth)
```
niezgoda, line 99: maybe you could use test.each?

Same code as on line 293-301

apps/asap-server/test/routes/news.route.test.ts
```diff
@@ -53,6 +53,13 @@ describe('/news/ route', () => {
     });
 
     describe('Parameter validation', () => {
+      test('Should return a validation error when additional fields exist', async () => {
+        const response = await supertest(app).get(`/news`).query({
```
niezgoda, line 57: You can create new set of tests and put there all of the routers.
Then use test.each to validate invalid parameters.

niezgoda (changes requested):

apps/asap-server/test/routes/user.route.test.ts
marot: I understand the concern but in this case I'd prefer to be consistent with the existing code.
Let's discuss at a RTM if we should refactor authentication/validation tests and group them together.

apps/asap-server/test/routes/news.route.test.ts
marot: see other thread

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
```diff
@@ -42,11 +42,39 @@ describe('Calendar handler', () => {
...
   afterEach(() => {
-    subscribe.mockClear();
-    unsubscribe.mockClear();
+    subscribe.mockReset();
+    unsubscribe.mockReset();
+    calendarControllerMock.update.mockReset();
   });
 
   describe('Validation', () => {
+    test.each([
```
peterstarling, line 51: to me using `.each` like this (ie for only two cases and with such complicated setup in the array) seems unnecessary - it always makes it a little harder to read

peterstarling (approved):
looks good, just a minor comment + I would expect some description and a reference to a ticket?

niezgoda (approved):

apps/asap-server/test/handlers/calendar/subscribe-handler.test.ts
marot: @peterstarling yes I agree. I had more test cases before but deleted them after discussion with @kmaid. I'll remove the `.each`

PR #1396, opened 2022-03-03
CRN-714 update validation library (new environment)

marot:


apps/asap-server/src/utils/sync-google-event.ts
```diff
@@ -1,10 +1,61 @@
...
 import { EventStatus } from '@asap-hub/model';
 import { calendar_v3 as calendarV3 } from 'googleapis';
+import { JSONSchemaType } from 'ajv';
 import { EventController } from '../controllers/events';
 import logger from './logger';
-import { RequiredAndNonNullable } from './squidex';
+import { validateInput } from '../validation';
+import { NullableOptionalProperties } from './types';
+
+type GoogleEvent = NullableOptionalProperties<{
```
marot, line 10: @peterstarling should we move all validation schemas into a `.validation.ts` file? Even if they are used only once?

peterstarling: I guess we could? why not

peterstarling (approved):
as you suggested i would probably move validation to the validation directory - i think a pattern might emerge even from the handlers so we might be able to somehow extend existing schemas or something and reuse them in the handlers and routes

apps/asap-server/src/utils/sync-google-event.ts
marot: will do (in the other PR)

PR #1397, opened 2022-03-03
CRN-683 Keep external authors in sync in algolia

niezgoda:
### Requirements

any changes made to any external author on the CMS should be “immediately” reflected on the ROMS form:
- name changes
- publication state changes (to draft, published, archived)
- is deleted or created

### OOS:
- make the webhook universal (refactor)


niezgoda:
Previous PRs:
- https://github.com/yldio/asap-hub/pull/1356
- https://github.com/yldio/asap-hub/pull/1380

packages/algolia/src/client.ts
```diff
@@ -66,13 +86,20 @@ export class AlgoliaSearchClient {
     return this.index.search<EntityRecord<T>>(query, options);
   }
 
+  public static toPayload(
```
peterstarling, line 89: I'm not sure I understand this function, is it just for tests? if so - this is a static function, surely you could just make it a helper function inside the test directory?

peterstarling (changes requested):

PR #1407, opened 2022-03-07
chore: adds recoil to pages

ana-yldio:


apps/frontend/src/App.tsx
```diff
@@ -78,61 +79,63 @@ const App: FC<Record<string, never>> = () => {
...
-                <Route path={logout.template}>
-                  <Frame title="Logout">
-                    <Logout />
-                  </Frame>
-                </Route>
-                <Route exact path={staticPages({}).terms.template}>
-                  <BasicLayout>
-                    <Frame title={null}>
-                      <Content pageId="terms-and-conditions" />
+    <RecoilRoot>
```
kmaid, line 82: Have we removed RecoilRoot from the rest of the application (excluding tests) as it will now be available everywhere?

apps/frontend/src/api/__tests__/use-safe-fetch.test.tsx
```diff
@@ -1,178 +0,0 @@
-import nock, { ReplyFnContext, ReplyFnResult } from 'nock';
```
kmaid, line 1: good riddance!

kmaid: Have we removed useHttp from Package.json?

apps/frontend/src/App.tsx
ana-yldio: I think so @kmaid. I removed it from `AuthenticatedApp` component
See this commit https://github.com/yldio/asap-hub/pull/1407/commits/443b1d184098d02ab1d4beb2853e02f96879a1e9

apps/frontend/src/api/__tests__/use-safe-fetch.test.tsx
ana-yldio: thanks!

kmaid (approved):

PR #1409, opened 2022-03-08
CRN-716 make url optional for lab resources

marot:
https://asaphub.atlassian.net/browse/CRN-716

### DIFF
--- a/apps/asap-server/src/routes/research-outputs.route.ts
+++ b/apps/asap-server/src/routes/research-outputs.route.ts
@@ -72,7 +72,7 @@ const createSchema = Joi.object<ResearchOutputPostRequest>({
     .items(Joi.string().valid(...researchOutputSubtypes)),
   description: Joi.string().required(),
   tags: Joi.array().required(),
-  link: Joi.string().required(),
+  link: Joi.string(),
   title: Joi.string().required(),
   asapFunded: Joi.boolean(),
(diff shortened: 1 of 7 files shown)


niezgoda (approved):

kmaid (changes requested):
What about backend validation? Also I think we have some defaults in TeamOutput and a typed api response that might need to be updated

niezgoda:
I have assumed that on the backend this field is optional @kmaid

marot:
@kmaid The field is not optional, but we are sending an empty string. There is no validation on the backend that the URL starts with http or anything similar - it just needs a string (and it can be empty afair)

marot:
@kmaid The defaults in TeamOutput.tsx have no effect because they are only applied for parts that are not implemented.

kmaid (approved):

niezgoda:
I have approved this, but.
We shouldn't send empty string.
This field should be optional.
The migration shall be written to replace all empty strings with undefined / null.

niezgoda (changes requested):
I have approved this, but.
We shouldn't send empty string.
This field should be optional.
The migration shall be written to replace all empty strings with undefined / null.

marot:
@niezgoda can you check again?

marot:
I will re-create the PR because the gitlab env is broken

PR #1422, opened 2022-03-10
CRN-716 make url optional for lab resources new env

marot:


marot:
@niezgoda please re-review here

apps/asap-server/src/migrations/1646992836-replace-empty-string-with-null-in-research-output-link.ts
```diff
@@ -0,0 +1,22 @@
...
+import { applyToAllItemsInCollection } from '../utils/migrations';
+
+export default class MapResearchOutputDeprecatedSubtype extends Migration {
+  up = async (): Promise<void> => {
+    await applyToAllItemsInCollection<RestResearchOutput>(
+      'research-outputs',
+      async (researchOutput, squidexClient) => {
+        if (researchOutput?.data?.link?.iv === '') {
+          await squidexClient.patch(researchOutput.id, {
+            link: undefined,
```
niezgoda, line 13: @peterstarling is it going to work?
If I remember correctly squidex had some problem with undefined fields

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
```diff
@@ -100,7 +100,7 @@ const TeamCreateOutputForm: React.FC<TeamCreateOutputFormProps> = ({
       onSave={() =>
         onSave({
           tags,
-          link,
+          link: link === '' ? undefined : link,
```
niezgoda, line 103: String(link).trim()

niezgoda (changes requested):

apps/asap-server/src/migrations/1646992836-replace-empty-string-with-null-in-research-output-link.ts
marot: @niezgoda I went with what the type system allowed me. What other approach would there be?

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
marot: Yeah. To be sure, you mean `String(link).trim() === ''`, right?

apps/asap-server/src/migrations/1646992836-replace-empty-string-with-null-in-research-output-link.ts
marot: @niezgoda Afaik the migration runs when the sls is deployed. Shouldn't this throw an error if it doesn't work? cc @peterstarling

apps/crn-server/src/migrations/1646992836-replace-empty-string-with-null-in-research-output-link.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default class MapResearchOutputDeprecatedSubtype extends Migration {
+  up = async (): Promise<void> => {
+    await applyToAllItemsInCollection<RestResearchOutput>(
+      'research-outputs',
+      async (researchOutput, squidexClient) => {
+        if (researchOutput?.data?.link?.iv === '') {
+          await squidexClient.patch(researchOutput.id, {
+            link: {
+              // @ts-expect-error // The type system does not support null here.
```
peterstarling, line 14: i guess you're right, although for the purpose of this migration you could alter the type on line 8 to allow null for the link and then wouldn't need to ts-expect-error here

but i see your point, the types are correct for the GET endpoints in squidex but it's different with the patch

peterstarling (approved):

apps/crn-server/src/migrations/1646992836-replace-empty-string-with-null-in-research-output-link.ts
marot: I think hard-coding the type is more problematic than the `// @ts-expect-error`, because the expect-error indicates that the types are off, while hardcoding the types make it seem like everything is ok, while in truth we are casting to some specific type.

niezgoda (changes requested):

PR #1423, opened 2022-03-10
CRN-682 search across users and external authors

niezgoda:
## Requirements:
- on the Share an Output form, in the author selector, show items which can be either users or external authors 
- display  external authors differently from users:
- external authors 's names should be followed by “(Non CRN)”
- be able to select from at least 3 existing external authors 
- search what is typed on the author selector against these fields only:
  - display name of  external authors 
  - first name of  users 
  - last name of users 
- when user clicks "Share", all users and external authors   selected must be added to the "Authors" field on the CMS
- the order of users and external authors must be remembered
- when user clicks "Share", this field must become disabled (as the page is saving)
- testing strategy: on the FE, it will be possible to find and select those  external authors 

## OOS:
- keep external authors in sync (manually or automatically)

### DIFF
--- a/packages/algolia/src/client.ts
+++ b/packages/algolia/src/client.ts
@@ -72,15 +72,19 @@ export class AlgoliaSearchClient {
   }
 
   async searchEntity<T extends keyof EntityResponses>(
-    entityType: T,
+    entityTypes: T[],
     query: string,
     requestOptions?: SearchOptions,
   ): Promise<SearchEntityResponse<T>> {
+    const entityTypesFilter = entityTypes
+      .map((entityType) => `__meta.type:"${entityType}"`)
+      .join(' OR ');
+
     const options: SearchOptions = {
       ...requestOptions,
       filters: requestOptions?.filters
-        ? `${requestOptions.filters} AND __meta.type:"${entityType}"`
-        : `__meta.type:"${entityType}"`,
+        ? `${requestOptions.filters} AND (${entityTypesFilter})`
+        : entityTypesFilter,
     };
 
     return this.index.search<EntityRecord<T>>(query, options);
--- a/packages/algolia/src/client.ts
+++ b/packages/algolia/src/client.ts
@@ -1,5 +1,8 @@
 import { SearchIndex } from 'algoliasearch';
-import { SearchOptions, SearchResponse } from '@algolia/client-search';
+import {
+  SearchOptions,
+  SearchResponse as AlgoliaSearchResponse,
+} from '@algolia/client-search';
 import {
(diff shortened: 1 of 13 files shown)


packages/algolia/src/client.ts
```diff
@@ -72,15 +72,19 @@ export class AlgoliaSearchClient {
...
+    const entityTypesFilter = entityTypes
+      .map((entityType) => `__meta.type:"${entityType}"`)
+      .join(' OR ');
+
     const options: SearchOptions = {
       ...requestOptions,
       filters: requestOptions?.filters
-        ? `${requestOptions.filters} AND __meta.type:"${entityType}"`
-        : `__meta.type:"${entityType}"`,
+        ? `${requestOptions.filters} AND (${entityTypesFilter})`
```
peterstarling, line 86: actually i think we need brackets (we did previously too)
```suggestion
        ? (`${requestOptions.filters}) AND (${entityTypesFilter})`
```
and there should be a test for that - imagine there is an OR clause in the first bit

packages/algolia/test/client.test.ts
```diff
@@ -138,20 +138,38 @@ describe('Algolia Search Client', () => {
     expect(algoliaSearchIndex.search).toBeCalledWith('query', {
       hitsPerPage: 10,
       page: 0,
-      filters: 'some-filters AND __meta.type:"research-output"',
+      filters: 'some-filters AND (__meta.type:"research-output")',
```
peterstarling, line 141: its worth adding a test where `some-filters` is a list with an `OR` - that will show that we need brackets for that part too

packages/algolia/src/client.ts
```diff
@@ -72,15 +72,19 @@ export class AlgoliaSearchClient {
   }
 
   async searchEntity<T extends keyof EntityResponses>(
```
peterstarling, line 74: to be fair I think the idea was that this method is used to search just a single entity. If you want to change it thats fine but in that case its probably best to rename to either `searchEntities` or just `search`

peterstarling: I noticed that when we search across multiple entities

```typescript
const result = await client.searchEntity(
    ['external-author', 'research-output'],
    'asdas',
  );

  for (const record of result.hits) {
    if (record.__meta.type === "external-author") {
      record
    }
  }
```
the result contains hits which are a union of the selected entities, which is great, but we are also unable to easily tell them apart using the __meta.type. In fact I am able to change it to anything as long as its a key EntityResponses. Perhaps now its time to improve it?

peterstarling (changes requested):

PR #1427, opened 2022-03-11
feat: Set up GP2 back-end

peterstarling:
Rename the asap-server app to crn-server and frontend to crn-frontend. This will then enable us to name the upcoming new apps using a common pattern


.github/workflows/algolia-sync-v2.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
       - name: Build
-        run: yarn exec "yarn workspaces foreach -vptR --from '{@asap-hub/asap-cli,@asap-hub/asap-server}' run build:babel"
+        run: yarn exec "yarn workspaces foreach -vptR --from '{@asap-hub/asap-cli,@asap-hub/crn-server}' run build:babel"
```
marot, line 52: What about renaming asap-cli? Is it specific to crn?

package.json
```diff
@@ -32,7 +32,7 @@
     "release:frontend": "yarn workspace @asap-hub/frontend release:frontend",
     "sls": "babel-node --extensions=.js,.jsx,.ts,.tsx --config-file=./babel-cjs.config.js serverless/link-plugins.ts && TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' yarn pnpify serverless",
     "start": "yarn workspaces foreach -vpi -j 1000 run start",
-    "start:backend": "yarn workspace @asap-hub/asap-server start",
+    "start:backend": "yarn workspace @asap-hub/crn-server start",
     "start:frontend": "yarn workspace @asap-hub/frontend start",
```
marot, line 36: crn-frontend

package.json
```diff
@@ -32,7 +32,7 @@
     "release:frontend": "yarn workspace @asap-hub/frontend release:frontend",
```
marot, line 32: crn-frontend?

marot (commented):
Nothing suspicious was found. However, I only did a code review - should I test the changes too?

.github/workflows/algolia-sync-v2.yml
peterstarling: its not

niezgoda (commented):
I would go with naming
/apps/crn/frontend instead of /apps/crn-frontend

niezgoda (changes requested):
I would go with naming
/apps/crn/frontend instead of /apps/crn-frontend

jhonny-m:
@niezgoda why is that?

niezgoda:
@jhonny-m - to easily divide the app between crn and gp2.
It would be just easier to go to particular project and do some search in just one project :)

jhonny-m:
if thats the reason it shouldnt really matter too much as we can just filter just as well.
I just want to keep frontend and server the same

niezgoda:
@kmaid what do you think about the naming convention.
Should it be /apps/crn-server or /apps/crn/server ?

niezgoda (approved):

PR #1429, opened 2022-03-14
CRN-682 Search across users and external authors

niezgoda:
## Requirements:
- on the Share an Output form, in the author selector, show items which can be either users or external authors 
- display  external authors differently from users:
- external authors 's names should be followed by “(Non CRN)”
- be able to select from at least 3 existing external authors 
- search what is typed on the author selector against these fields only:
  - display name of  external authors 
  - first name of  users 
  - last name of users 
- when user clicks "Share", all users and external authors   selected must be added to the "Authors" field on the CMS
- the order of users and external authors must be remembered
- when user clicks "Share", this field must become disabled (as the page is saving)
- testing strategy: on the FE, it will be possible to find and select those  external authors 

## OOS:
- keep external authors in sync (manually or automatically)

Previous PR: https://github.com/yldio/asap-hub/pull/1423

### DIFF
--- a/apps/crn-frontend/src/__fixtures__/algolia.ts
+++ b/apps/crn-frontend/src/__fixtures__/algolia.ts
@@ -8,11 +8,10 @@ import {
 } from '@asap-hub/algolia';
 
 export const createAlgoliaResponse = <EntityType extends keyof EntityResponses>(
-  type: EntityType,
-  data: EntityResponses[EntityType][],
+  hits: EntityHit<EntityType>[],
   overrides: Partial<SearchEntityResponse<EntityType>> = {},
 ): SearchEntityResponse<EntityType> => ({
-  nbHits: data.length,
+  nbHits: hits.length,
   page: 0,
   nbPages: 1,
   hitsPerPage: 10,
@@ -21,15 +20,8 @@ export const createAlgoliaResponse = <EntityType extends keyof EntityResponses>(
   params: 'page=0&hitsPerPage=10&validUntil=1629454922296',
   renderingContent: {},
   processingTimeMS: 1,
+  hits,
   ...overrides,
-  hits: data.map(
-    (item, i) =>
-      ({
-        ...item,
-        objectID: `${i}`,
-        __meta: { type },
-      } as EntityHit<EntityType>),
-  ),
 });
 
 export const createResearchOutputAlgoliaRecord = (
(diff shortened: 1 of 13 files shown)


peterstarling (approved):

apps/crn-frontend/src/__fixtures__/algolia.ts
```diff
@@ -21,11 +22,14 @@ export const createAlgoliaResponse = <EntityType extends keyof EntityResponses>(
...
-    objectID: `${i}`,
-    __meta: { type },
-  })),
+  hits: data.map(
+    (item, i) =>
+      ({
+        ...item,
+        objectID: `${i}`,
+        __meta: { type },
+      } as EntityHit<EntityType>),
```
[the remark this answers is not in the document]
niezgoda: We have to. Otherwise TS is screaming on us.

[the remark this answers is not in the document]
peterstarling: to be fair i noticed this and was going to suggest setting the return type for the map as well but these are just fixtures so i dont mind the cast as much anyway

apps/crn-frontend/src/network/users/api.ts
```diff
@@ -48,6 +50,24 @@ export const getUsers = async (
   return { items: result.hits, total: result.nbHits };
 };
 
+export const getUsersOrExternalAuthors = async (
```
kmaid, line 53: Shouldn't this be getUsersAndExternalAuthours?

apps/crn-frontend/src/__fixtures__/algolia.ts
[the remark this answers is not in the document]
niezgoda: @peterstarling that is not going to work :(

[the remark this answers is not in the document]
niezgoda: @kmaid - what solution are you suggesting?
We have trick there to enforce that data and __meta.type is same type.
Eg. ...items is RO and __meta.type must be "research-output"

apps/crn-frontend/src/__fixtures__/algolia.ts
```diff
@@ -21,11 +22,14 @@ export const createAlgoliaResponse = <EntityType extends keyof EntityResponses>(
   renderingContent: {},
   processingTimeMS: 1,
   ...overrides,
-  hits: data.map((item, i) => ({
-    ...item,
-    objectID: `${i}`,
-    __meta: { type },
-  })),
+  hits: data.map(
+    (item, i) =>
```
kmaid, line 26: > (item, i): EntityHit<EntityType> =>

peterstarling: yeah that would only work if the input types were dont differently - ie the `data` and `type` were actually tied together. Technically at the moment you can give it a type that does not correspond to the data which is why typescript moans. Like I said its just a fixture so I wouldn't get too hung up on it and simply leave the cast

niezgoda: That's not going to work

niezgoda: @kmaid If you have better solution on your mind please push it into this branch/PR :)

peterstarling (approved):

kmaid (approved):

PR #1434, opened 2022-03-15
CRN-693 Add access instructions

niezgoda:
As a user sharing an output that is hard to access
I want to explain to others how to access that output 
so that I comply with ASAP guidelines

## Requirements
- add a new input field on the Share an Output form, on the second section, after additional keywords and after identifiers (if already done)
- Label: Access instructions
- This field is optional
- This is a text input field similar to “Description”
- Save changes to accessInstructions on the CMS
- Please include a place holder as per the designs

## Note:
- this is currently RTF on the CMS
- it would be nice to accept hyperlinks in the future

### DIFF
--- a/apps/crn-server/src/validation/research-output.validation.ts
+++ b/apps/crn-server/src/validation/research-output.validation.ts
@@ -63,6 +63,7 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
         nullable: true,
       },
       teams: { type: 'array', items: { type: 'string' }, minItems: 1 },
+      accessInstructions: { type: 'string', nullable: true },
     },
     required: [
       'type',
--- a/packages/react-components/src/organisms/TeamCreateOutputForm.tsx
+++ b/packages/react-components/src/organisms/TeamCreateOutputForm.tsx
@@ -84,6 +84,8 @@ const TeamCreateOutputForm: React.FC<TeamCreateOutputFormProps> = ({
   const [description, setDescription] =
     useState<ResearchOutputPostRequest['description']>('');
   const [link, setLink] = useState<ResearchOutputPostRequest['link']>('');
+  const [accessInstructions, setAccessInstructions] =
+    useState<ResearchOutputPostRequest['accessInstructions']>('');
 
   return (
     <Form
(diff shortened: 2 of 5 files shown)


packages/react-components/src/organisms/__tests__/TeamOutputSharingCard.test.tsx
```diff
@@ -28,6 +28,7 @@ it('renders the card with provided values', () => {
   expect(getByDisplayValue('http://example.com')).toBeVisible();
   expect(getByDisplayValue('title')).toBeVisible();
   expect(getByDisplayValue('Preprint')).toBeVisible();
+  expect(getByDisplayValue('description')).toBeVisible();
```
kmaid, line 31: This looks like a mistake

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
```diff
@@ -107,6 +109,10 @@ const TeamCreateOutputForm: React.FC<TeamCreateOutputFormProps> = ({
           authors: authors.map(({ value }) => value),
           labs: labs.map(({ value }) => value),
           teams: teams.map(({ value }) => value),
+          accessInstructions:
+            String(accessInstructions).trim() !== ''
```
kmaid, line 113: Not sure what this is doing. We require a value but this means if you put in " " you can send empty values?

kmaid (changes requested):

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
```diff
@@ -107,6 +109,10 @@ const TeamCreateOutputForm: React.FC<TeamCreateOutputFormProps> = ({
...
+              ? accessInstructions
+              : undefined,
```
kmaid, line 115: Here we are passing undefined.

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -63,6 +63,7 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
         nullable: true,
       },
       teams: { type: 'array', items: { type: 'string' }, minItems: 1 },
+      accessInstructions: { type: 'string', nullable: true },
```
kmaid, line 66: Here we take null?

kmaid (approved):

PR #1435, opened 2022-03-15
CRN-682-search-across-users-and-external-authors-crn2

kmaid:


niezgoda:
This is not working as expected.
Expected is current behaviour which looks like this:
image

Your solution otherwise is working like this:
image

PR #1439, opened 2022-03-18
CRN-619 Improve and refine User search results

niezgoda:


apps/crn-server/src/handlers/lab/index-lab-users-handler.ts
```diff
@@ -48,10 +48,12 @@ export const indexLabUsersHandler =
       );
 
       await algoliaClient.saveMany(
-        foundUsers.items.map((data) => ({
-          data,
-          type: 'user',
-        })),
+        foundUsers.items
```
peterstarling, line 51: is this tested?

apps/crn-server/src/handlers/teams/index-team-users-handler.ts
```diff
@@ -51,10 +51,12 @@ export const indexTeamUsersHandler =
       );
 
       await algoliaClient.saveMany(
-        foundUsers.items.map((data) => ({
-          data,
-          type: 'user',
-        })),
+        foundUsers.items
```
peterstarling, line 54: is this tested?

peterstarling (changes requested):

PR #1441, opened 2022-03-18
CRN-612-show-snippet-of-notes-on-past-events

kmaid:


packages/react-components/src/molecules/Collapsible.tsx
```diff
@@ -27,7 +47,7 @@ const Collapsible: React.FC<CollapsibleProps> = ({
           >
             {expanded ? chevronCircleUpIcon : chevronCircleDownIcon}
           </span>
-          {expanded ? 'Hide' : 'Show'}
+          {expanded ? 'Hide' : 'Show'} more
```
niezgoda, line 50: image

niezgoda: image

niezgoda: I guess that it will be hard to make it "smarter"

niezgoda (approved):

packages/react-components/src/molecules/Collapsible.tsx
kmaid: Yeah, I don't really like it which is why i got beatriz to acknowledge and accept the problems in refinement. I think it be better to show the text in full or ask for an additional summary field.

PR #1458, opened 2022-03-22
feat: CRN-686 Search RO by URL

peterstarling:
Modifies the search endpoint to provide FE with the ability to check the uniqueness of the URL field of research outputs


markcsims (approved):
Other than the unpolished bit, looks good. 

There's an empty file been added though,

packages/squidex/test/upsert.test.ts
```diff
@@ -47,6 +50,66 @@ describe('squidex wrapper - upsert', () => {
...
+        traceId: '00-ba8100d975b2cb551a023702a7d0d5b7-891e647127349001-01',
+        type: 'https://tools.ietf.org/html/rfc7231#section-6.5.1',
+        details: ['link.iv: Another content with the same value exists.'],
+        statusCode: 400,
+      });
+
+    await expect(() =>
+      client.upsert(
+        contentId,
+        JSON.parse(
```
niezgoda, line 68: what is the purpose of this JSON.parse(JSON.stringify(x))?

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -174,7 +174,41 @@ describe('/research-outputs/ route', () => {
...
+        Boom.badRequest('error', {
+          message: 'Validation error',
+          details: ['link.iv: Another content with the same value exists.'],
+          statusCode: 400,
+        }),
+      );
+
+      const response = await supertest(app)
+        .post('/research-outputs')
+        .send({ ...researchOutput })
```
niezgoda, line 189: why do we have here this destructuring?

niezgoda (changes requested):

niezgoda (approved):

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -44,9 +45,39 @@ export const researchOutputRouteFactory = (
...
-    res.status(201).json(researchOutput);
+      res.status(201).json(researchOutput);
+    } catch (error) {
+      // TODO: move this logic to the controller and catch in the error-handler
+      // https://asaphub.atlassian.net/browse/CRN-777
+      if (Boom.isBoom(error)) {
+        if (
+          error.data?.message === 'Validation error' &&
+          Array.isArray(error.data?.details) &&
+          error.data.details[0].includes(
```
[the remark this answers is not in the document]
peterstarling: don't need it, we've got an Array.isArray above so we know its not undefined

packages/squidex/test/upsert.test.ts
peterstarling: you should probably ask the committers of the rest of these tests :D normally this would be to avoid mutation but since we are actually creating the object inline i dont really know what its doing here :D

apps/crn-server/test/routes/research-outputs.route.test.ts
peterstarling: again - im just coping and pasting whatever it is that people used here cause unless im refactoring the whole thing i figured i'd stick to the same patterns.  It looks like the developers wanted to create a shallow clone, dont ask me why

apps/crn-server/src/middleware/error-handler.ts
```diff
@@ -1,8 +1,9 @@
 import { ErrorRequestHandler } from 'express';
 import { isBoom } from '@hapi/boom';
+import { ErrorResponse } from '@asap-hub/model/src/error';
```
kmaid, line 3: Seems imported weirdly

markcsims (approved):

PR #1463, opened 2022-03-23
feat: add  published date selector to the RO form

ana-yldio:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/8?modal=detail&selectedIssue=CRN-692


packages/react-components/src/molecules/LabeledDateField.tsx
```diff
@@ -31,11 +47,14 @@ const LabeledDateField: React.FC<LabeledDateFieldProps> = ({
           id={id}
           value={value ? formatISO(value, { representation: 'date' }) : ''}
           onChange={(newDate) => onChange(parseISO(newDate))}
+          max={max ? format(max, 'yyyy-MM-dd') : undefined}
         />
       )}
     >
       <Paragraph>
         <strong>{title}</strong>
+        <span css={subtitleStyles}>{subtitle}</span> <br />
```
marot, line 56: Since the fields are added, are there no regressions in other usages of the component?

packages/react-components/src/molecules/__tests__/LabeledDateField.test.tsx
```diff
@@ -1,31 +1,57 @@
...
+  expect(screen.getByText(/required/i)).toBeVisible();
+  expect(screen.getByText(/please enter a date before today/i)).toBeVisible();
+});
+
+it('supports max date attribute', () => {
+  render(
+    <LabeledDateField
+      title="Date"
+      value={new Date('2034-02-01')}
+      max={new Date()}
```
marot, line 37: In 12 years this test will fail 😅 
How about specifying the year?

packages/react-components/src/organisms/TeamCreateOutputFormSharingCard.tsx
```diff
@@ -142,6 +147,21 @@ const TeamCreateOutputFormSharingCard: React.FC<TeamCreateOutputFormSharingCardP
...
+            title={'Date Published'}
+            subtitle={'(optional)'}
+            description={
+              'This should be the date your output was shared publicly on it’s repository.'
+            }
+            onChange={onChangePublishDate}
+            value={publishDate}
+            max={new Date()}
+            getValidationMessage={() =>
+              'Publish date cannot be greater than today'
```
marot, line 161: Can the user somehow enter other invalid data? In that case the validation message would be wrong

marot (approved):

packages/react-components/src/molecules/LabeledDateField.tsx
ana-yldio: This is the first time we are using this component

packages/react-components/src/organisms/TeamCreateOutputFormSharingCard.tsx
ana-yldio: I don't think so, the input type is date

packages/react-components/src/molecules/__tests__/LabeledDateField.test.tsx
ana-yldio: I updated the tests. If we set the max date to tomorrow and the current value to today, it'll never fail

PR #1464, opened 2022-03-23
CRN-596 add identifier selector to form

marot:


apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -152,10 +153,35 @@ export default class ResearchOutputs implements ResearchOutputController {
   }
   private async createResearchOutput({
     subTypes,
+    identifierType,
+    identifier: rawIdentifier,
     ...researchOutputData
   }: Omit<ResearchOutputPostRequest, 'teams'>) {
+    let identifier:
```
peterstarling, line 160: maybe extract this to a function? this will make the code a little cleaner (ie no switch, let, etc)

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -711,6 +716,48 @@ describe('ResearchOutputs controller', () => {
...
+      { identifierType: ResearchOutputIdentifierType.RRID, field: 'rrid' },
+      {
+        identifierType: ResearchOutputIdentifierType.LabCatalogNumber,
+        field: 'labCatalogNumber',
+      },
+      {
+        identifierType: ResearchOutputIdentifierType.AcessionNumber,
+        field: 'accession',
+      },
+    ])('should set the field properly', async ({ identifierType, field }) => {
```
peterstarling, line 730: maybe something like `should set the ${field} when the ${identifierType} type is selected`?

also I find this test hard to read - that's only cause of the .each with type declaration and an array definition typed inline, maybe extract them above?

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -64,6 +65,8 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
       },
       teams: { type: 'array', items: { type: 'string' }, minItems: 1 },
       accessInstructions: { type: 'string', nullable: true },
+      identifierType: { type: 'string', enum: researchOutputIdentifierTypes },
+      identifier: { type: 'string', nullable: true },
```
peterstarling, line 69: im not so sure about this, shouldn't we be more explicit about the identifiers? like.. it should set the `doi` property when the `doi` type is selected. In fact - I don't think the BE should receive the `identifierType` at all? it's mostly for the FE anyway (if i remember correctly). 

What if we want to provide many at the some time at some point (Which we are able to do in squidex)? Also what if we want validation in the BE - at the moment it accepts any arbitrary string. 

Finally - should we be accepting any random string? after all we do have regexes for each one of them

peterstarling (changes requested):

apps/crn-server/src/validation/research-output.validation.ts
marot: Sounds good to me. I'll see if using a regex is high effort

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -232,6 +232,63 @@ describe('/research-outputs/ route', () => {
         expect(response.status).toBe(400);
       });
 
+      test('Should validate doi based on a regex', async () => {
```
peterstarling, line 235: this is ok although maybe we could just have a single payload with all the valid fields and test the errors instead? or you can keep these tests but make sure to test the opposite scenario (with error)

peterstarling (changes requested):
im not too fussed about the regex bit, just the test

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -65,6 +65,18 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
       },
       teams: { type: 'array', items: { type: 'string' }, minItems: 1 },
       accessInstructions: { type: 'string', nullable: true },
+      doi: {
+        type: 'string',
+        nullable: true,
+        pattern: '^(doi:)?\\d{2}\\.\\d{4}.*$',
```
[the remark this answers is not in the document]
marot: @peterstarling I tried to extract the regex from there but I think it does not work.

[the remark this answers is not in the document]
marot: @peterstarling Please have a look at this error: https://github.com/yldio/asap-hub/runs/5752112031?check_suite_focus=true

I'm not sure how to best approach this. Special case it, or revert to the hardcoded version?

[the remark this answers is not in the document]
peterstarling: meh, just revert to hardcoded one

apps/crn-server/test/routes/research-outputs.route.test.ts
peterstarling: To be fair its currently possible to have both doi and rrid for instance and maybe we should actually have it documented, and if we want we can already make it impossible. 

Anyway its ok if you keep it this way just make sure to test the errors too

marot: @peterstarling Error cases are tested too (response2). Should I split it into separate tests?

peterstarling: ahhh I see - I missed it. I wouldn't mind it that much although I think the general rule of thumb is that you should have only one expectation and also i think we usually dont do two requests in a single test so yes please - split

peterstarling (changes requested):

PR #1478, opened 2022-03-29
CRN-767 Show Avatars in the user selector

niezgoda:
## Story Description
As a Product Owner
I want to see the appropriate avatar displayed on author selector function of the shared outputs form
So that I am able to visually identify tag labels

## Assumptions
- This behaviour should apply to the ‘Author’ selector on the Shared Outputs form; all other instances of this component should remain unchanged.
- The Avatar should be 24x24 px in a circle - it should be spaced 8px before the tag label 

## Requirement
- Update the component multi-select to display an avatar image with each option (including: those tags that are selected AND those tags that are to be selected)
- for each option on the results list of the user search (i.e. drop-down), show an avatar to the left:
- for each selected option (i.e. in the input text area), show the avatar inside the tag pill on the left
- The avatar should vary depending on the attribute of the option
- for users with an avatar, show their picture (when available)
- for users without an avatar picture, use the fallback avatar for users (Initials + color)  
- external authors should have a placeholder avatar -

### DIFF
--- a/packages/react-components/src/organisms/__tests__/AuthorSelect.test.tsx
+++ b/packages/react-components/src/organisms/__tests__/AuthorSelect.test.tsx
@@ -63,7 +63,7 @@ it('renders a author multi select, passing through props for user with avatar', 
 });
 
 it('renders a author multi select, passing through props for external author', () => {
-  const { getByText, getByLabelText } = render(
+  const { getByText, getByTitle, getByLabelText } = render(
     <AuthorSelect
       title="Title"
       subtitle="Subtitle"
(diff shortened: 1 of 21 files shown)


packages/react-components/src/organisms/__tests__/AuthorSelect.test.tsx
```diff
@@ -86,6 +86,6 @@ it('renders a author multi select, passing through props for external author', (
   expect(getByLabelText(/Title/i)).toBeVisible();
   expect(getByLabelText(/Subtitle/i)).toBeVisible();
   expect(getByLabelText(/Description/i)).toBeVisible();
-  expect(getByText('Andy Smith')).toBeVisible();
-  expect(getByText('AS')).toBeVisible();
+  expect(getByText('Andy Smith (Non CRN)')).toBeVisible();
+  expect(container.querySelector('svg')).toBeVisible();
```
kmaid, line 90: Please use getByTitle

kmaid (changes requested):

kmaid (approved):

PR #1480, opened 2022-03-30
feat: CRN-747 Check for unique titles

peterstarling:
* adds new filter to RO controller of object type
* enables draft flag in RO controller


apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -82,13 +82,23 @@ export default class ResearchOutputs implements ResearchOutputController {
...
       .join(' or ');
 
-    const filterQ = (filter || [])
-      .reduce(
-        (acc: string[], word: string) =>
-          acc.concat([`data/type/iv eq '${word}'`]),
-        [],
-      )
-      .join(' or ');
+    let filterQ: string;
```
niezgoda, line 85: the code below could be function :)

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -187,12 +223,18 @@ export default class ResearchOutputs implements ResearchOutputController {
   }
 }
 
+type ResearchOutputFilter =
```
niezgoda, line 226: I guess that in the future we should refactor this so that only this new "object" filter will stay in the code.

niezgoda (approved):

apps/crn-server/src/controllers/research-outputs.ts
peterstarling: yeah i know it could but im so annoyed when i see all these utility functions in this repo just so the code is more 'functional' that i almost did this on purpose. I guess i will refactor though

apps/crn-server/src/controllers/research-outputs.ts
peterstarling: yeah i guess? cause this string filter makes no sense to me, like i dont know who thought that just 'filter' would be a name thats good enough for something that only refers to filtering types

apps/crn-server/src/controllers/research-outputs.ts
markcsims: I think it helps break it down into single purpose functions

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -201,3 +227,23 @@ export interface ResearchOutputController {
...
 export type ResearchOutputInputData = ResearchOutputPostRequest;
+
+const makeODataFilter = (filter?: ResearchOutputFilter): string => {
+  if (Array.isArray(filter)) {
+    return filter
+      .reduce(
+        (acc: string[], word: string) =>
+          acc.concat([`data/type/iv eq '${word}'`]),
+        [],
+      )
```
markcsims, line 238: Can this be:

```.map(([key, val]) => `data/${key}/iv eq '${val}'`)```

Then maybe there's a way to combine these 2 if clauses?

niezgoda (approved):

PR #1484, opened 2022-03-31
chore: CRN-686 Add validation error response type

peterstarling:


packages/model/package.json
```diff
@@ -13,6 +13,7 @@
     "watch:babel": "../../scripts/build-babel.sh watch"
   },
   "dependencies": {
-    "@babel/runtime-corejs3": "^7.14.0"
+    "@babel/runtime-corejs3": "^7.14.0",
+    "ajv": "^8.10.0"
```
kmaid, line 17: I think this should be a dev dependency

kmaid (approved):

packages/model/package.json
peterstarling: nah, you cant import dev dependencies into your code, even if its just the types (linter moans and i think its correct anyway)

PR #1485, opened 2022-03-31
CRN-753 Create generic event emitter format

niezgoda:
Currently the naming pattern as well as directory structure for webhooks is completely unorganised. It requires some clean up

## Requirements
Choose a unified way for naming webhooks in serverless and within the directory.

1. Every Squidex event should hit the same endpoint and that should be automatically put on the event bus
2. Should an event be unsupported - the infrastructure should capture that (ie. it should not be picked up by any handlers)
3. Hook Events should be mirrored into Event bus (including publishing and unpublishing)


serverless.ts
```diff
@@ -392,12 +377,12 @@ const serverlessConfig: AWS = {
           eventBridge: {
             eventBus: 'asap-events-${self:provider.stage}',
             pattern: {
-              source: ['asap.lab'],
+              source: ['asap.entity-updated'],
```
peterstarling, line 380: maybe declare `eventSource` somewhere at the top and use here and at #576? same goes for the event bus

peterstarling (approved):
looks good, have you tested it manually?

PR #1486, opened 2022-04-01
CRN-738 make identifier required new env

marot:
This PR contains
https://asaphub.atlassian.net/browse/CRN-596
https://asaphub.atlassian.net/browse/CRN-738
https://asaphub.atlassian.net/browse/CRN-736

### DIFF
--- a/apps/crn-server/src/validation/research-output.validation.ts
+++ b/apps/crn-server/src/validation/research-output.validation.ts
@@ -65,6 +65,18 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
       },
       teams: { type: 'array', items: { type: 'string' }, minItems: 1 },
       accessInstructions: { type: 'string', nullable: true },
+      doi: {
+        type: 'string',
+        nullable: true,
+        pattern: '^(doi:)?\\d{2}\\.\\d{4}.*$',
+      },
+      accession: {
+        type: 'string',
+        nullable: true,
+        pattern: '^(\\w+\\d+(\\.\\d+)?)|(NP_\\d+)$',
+      },
+      labCatalogNumber: { type: 'string', nullable: true },
+      rrid: { type: 'string', nullable: true, pattern: 'RRID:[a-zA-Z]+.+$' },
     },
     required: [
       'type',
--- a/apps/crn-server/src/validation/research-output.validation.ts
+++ b/apps/crn-server/src/validation/research-output.validation.ts
@@ -101,27 +101,35 @@ export const validateResearchOutputPostRequestParameters = validateInput(
   },
 );
 
-export const validateResearchOutputPostRequestParametersIdentifiers = (data: ResearchOutputPostRequest) => {
-  const types = researchOutputToIdentifierType[data.type]
+export const validateResearchOutputPostRequestParametersIdentifiers = (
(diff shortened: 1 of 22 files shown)


apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -85,3 +100,50 @@ export const validateResearchOutputPostRequestParameters = validateInput(
     coerce: false,
   },
 );
+
+export const validateResearchOutputPostRequestParametersIdentifiers = (
```
peterstarling, line 104: im sure you have looked into it but doesn't ajv support something similar? we are saying that at least one and only one of the identifiers is required for when asapFunded and usedInPublication are present, correct? 

I guess it's a little more complex with the other condition since we have a list of types that support each of these indetifiers, correct?

peterstarling (approved):
good stuff although one comment about the validation - i thought maybe there is an easy way to implement it with ajv, especially since you've got all tested nicely

packages/react-components/src/organisms/__tests__/TeamCreateOutputExtraInformationCard.test.tsx
```diff
@@ -2,12 +2,20 @@ import { ComponentProps } from 'react';
...
+import { noop } from '../../utils';
 
 const props: ComponentProps<typeof TeamCreateOutputExtraInformationCard> = {
   isSaving: false,
   tagSuggestions: [],
   tags: [],
+  identifier: '',
+  setIdentifier: noop,
+  identifierType: ResearchOutputIdentifierType.None,
+  setIdentifierType: noop,
```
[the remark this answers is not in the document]
marot: Yes!

apps/crn-server/src/validation/research-output.validation.ts
marot: Yes ajv theoretically supports it. But the setup is so complex, with nested levels of field attributes that I decided this approach is easier to test and to understand.

packages/react-components/src/organisms/TeamCreateOutputIdentifier.tsx
```diff
@@ -0,0 +1,139 @@
...
+    value: identifier,
+    label: identifier,
+  }));
+};
+
+const identifierMap = {
+  [ResearchOutputIdentifierType.DOI]: {
+    helpText: 'Your DOI must start with 1 and it cannot be a URL',
+    placeholder: 'DOI number e.g. 10.5555/YFRU1371',
+    regex: '^(doi:)?\\d{2}\\.\\d{4}.*$',
```
kmaid, line 35: I think these regex's should be in the validation package so the BE can potentially reuse them

kmaid: Why are they a string instead of a regular regex?

marot: @kmaid because the ajv and html requires a string instead of a regex.
Unfortunately, I won't have time to adapt this anymore.

apps/crn-server/src/validation/research-output.validation.ts
peterstarling: fair, im cool with that

PR #1499, opened 2022-04-06
feat: create RO handles external authors creation

ana-yldio:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/8?modal=detail&selectedIssue=CRN-794

https://cloud.squidex.io/app/asap-hub-dev/content/research-outputs/6b3bbb16-1dfc-42da-852a-64698c2f17fc/history
Screenshot 2022-04-11 at 10 02 20

### DIFF
--- a/apps/crn-server/src/controllers/research-outputs.ts
+++ b/apps/crn-server/src/controllers/research-outputs.ts
@@ -228,20 +228,15 @@ export default class ResearchOutputs implements ResearchOutputController {
   }
 
   private async associateResearchOutputToAuthors({
-    id,
-    name,
+    externalAuthorId,
+    externalAuthorName,
   }: ExternalAuthorPostRequest) {
-    if (id) return id;
+    if (externalAuthorId) return externalAuthorId;
 
-    if (name) {
-      const { id: authorId } =
-        await this.externalAuthorSquidexRestClient.create({
-          name: { iv: name },
-        });
-      return authorId;
-    }
-
-    throw Boom.badData('Validation error');
+    const { id } = await this.externalAuthorSquidexRestClient.create({
+      name: { iv: externalAuthorName },
+    });
+    return id;
   }
 }
 
--- a/apps/crn-server/src/controllers/research-outputs.ts
+++ b/apps/crn-server/src/controllers/research-outputs.ts
@@ -1,5 +1,5 @@
 import {
-  ExternalAuthorPostRequest,
+  ExternalAuthorInput,
   ListResearchOutputResponse,
   ResearchOutputPostRequest,
   ResearchOutputResponse,
@@ -227,23 +227,14 @@ export default class ResearchOutputs implements ResearchOutputController {
     });
   }
 
-  private async associateResearchOutputToAuthors({
-    userId,
-    externalAuthorId,
-    externalAuthorName,
-  }: ExternalAuthorPostRequest) {
-    if (userId) return userId;
+  private async associateResearchOutputToAuthors(data: ExternalAuthorInput) {
+    if ('userId' in data) return data.userId;
+    if ('externalAuthorId' in data) return data.externalAuthorId;
 
-    if (externalAuthorId) return externalAuthorId;
-
-    if (externalAuthorName) {
-      const { id } = await this.externalAuthorSquidexRestClient.create({
-        name: { iv: externalAuthorName },
-      });
-      return id;
-    }
-
-    throw Boom.badData('Validation error');
+    const { id } = await this.externalAuthorSquidexRestClient.create({
+      name: { iv: data.externalAuthorName },
+    });
+    return id;
   }
 }
 
--- a/apps/crn-server/src/controllers/research-outputs.ts
+++ b/apps/crn-server/src/controllers/research-outputs.ts
@@ -236,10 +236,14 @@ export default class ResearchOutputs implements ResearchOutputController {
 
     if (externalAuthorId) return externalAuthorId;
 
-    const { id } = await this.externalAuthorSquidexRestClient.create({
-      name: { iv: externalAuthorName },
-    });
-    return id;
+    if (externalAuthorName) {
+      const { id } = await this.externalAuthorSquidexRestClient.create({
+        name: { iv: externalAuthorName },
+      });
+      return id;
+    }
+
+    throw Boom.badData('Validation error');
   }
 }
 
--- a/apps/crn-server/src/controllers/research-outputs.ts
+++ b/apps/crn-server/src/controllers/research-outputs.ts
@@ -228,9 +228,12 @@ export default class ResearchOutputs implements ResearchOutputController {
   }
 
   private async associateResearchOutputToAuthors({
+    userId,
     externalAuthorId,
     externalAuthorName,
   }: ExternalAuthorPostRequest) {
+    if (userId) return userId;
+
     if (externalAuthorId) return externalAuthorId;
 
     const { id } = await this.externalAuthorSquidexRestClient.create({
--- a/apps/crn-server/src/controllers/research-outputs.ts
+++ b/apps/crn-server/src/controllers/research-outputs.ts
@@ -1,9 +1,11 @@
 import {
+  ExternalAuthorPostRequest,
   ListResearchOutputResponse,
   ResearchOutputPostRequest,
   ResearchOutputResponse,
 } from '@asap-hub/model';
 import {
+  RestExternalAuthor,
   RestResearchOutput,
   RestTeam,
   SquidexGraphqlClient,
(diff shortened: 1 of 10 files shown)


apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -212,6 +226,23 @@ export default class ResearchOutputs implements ResearchOutputController {
...
+  private async associateResearchOutputToAuthors({
+    id,
+    name,
+  }: ExternalAuthorPostRequest) {
+    if (id) return id;
+
+    if (name) {
+      const { id: authorId } =
+        await this.externalAuthorSquidexRestClient.create(
+          parseToSquidex({ name }) as RestExternalAuthor['data'],
```
peterstarling, line 239: any reason why you have to cast it?

peterstarling: in fact - i know why - its because of parseToSquidex which is poorly typed. I would make sure to type it in such a way that the output depends on the input, that way we can keep the type safety

packages/model/src/research-output.ts
```diff
@@ -170,7 +173,7 @@ export type ResearchOutputPostRequest = {
   link?: string;
 
   labs?: string[];
-  authors?: string[];
+  authors?: ExternalAuthorPostRequest[];
```
peterstarling, line 176: quick question - how do we associate `Users` as authors? as you know the `ResearchOutputResponse` `authors` field contains a union of two types - shouldn't our input reflect it? I understand that at the moment it is impossible to link authors, we just link external authors?

https://github.com/yldio/asap-hub/blob/4d76557d1ab0de189274b2d651dfb03c3180ee2f/packages/model/src/research-output.ts#L160

peterstarling (changes requested):

apps/crn-server/src/controllers/research-outputs.ts
ana-yldio: Fair point, thanks. Actually, we don't need `parseToSquidex` for this. It's only handling one prop `name`
https://github.com/yldio/asap-hub/blob/09c69b028726d8bbd6d6dcb63e1250fca0b7a7b1/apps/crn-server/src/controllers/research-outputs.ts#L238-L240

peterstarling: ok basically what i mean is that the input type should enable posting:
1) new external authors 
2) references to external authors
3) references to authors

What I just realised is that actually we don't have to specify which of the two types of resources (ie external author or user) we are giving to squidex - and I now wonder if it should stay that way? I was going to suggest that since we change the interface we should change it to something like:

```typescript
{
(...)
authors?: Array<
    | { externalAuthorId: string }
    | { externalAuthorName: string }
    | { userId: string }
  >;
}
```
Technically we could instead just do:
```typescript
{
(...)
authors?: Array<
    | { id: string }
    | { userId: string }
  >;
}
```
And that's pretty much what you suggested (but not quite!) but I'm tempted to make the clients give us the type of resource they are associating here. What do you think? 

Also CC @markcsims @kmaid

markcsims: So an object like?

`{ id?: string, name: string, type: string}[]`

ana-yldio: @peterstarling @markcsims ok, so we'll change the type of the authors' field for something like this
```
{
     id?: string       // Optional because new external authors don't have an id
     name: string
     type: NewExternalAuthor | ExternalAuthor | InternalUser
}
```

peterstarling: I'm still for this:

(...)
authors?: Array<
    | { externalAuthorId: string }
    | { externalAuthorName: string }
    | { userId: string }
  >;

ana-yldio:
https://cloud.squidex.io/app/asap-hub-dev/content/research-outputs/c62f6700-7d5e-4b88-9980-5d8952f4e06c/history

Screenshot 2022-04-08 at 15 33 40

packages/model/src/external-author.ts
```diff
@@ -7,3 +7,9 @@ export type ExternalAuthorResponse = Pick<
 >;
 
 export type ListExternalAuthorResponse = ListResponse<ExternalAuthorResponse>;
+
+export type ExternalAuthorPostRequest = {
```
peterstarling, line 11: I don't think I understand this interface. Like I said earlier it should be either a reference to a user or an external author or a new external author name. In here we seem to always ask for an external author name but also optionally require userId or externalAuthorId?

peterstarling (changes requested):

packages/model/src/external-author.ts
ana-yldio: `externalAuthorId` needs to be optional. When selecting an internal user, there's no available an `externalAuthorId`. Below is the interface you suggested. How's this different @peterstarling?

```
authors?: Array<
| { externalAuthorId: string }
| { externalAuthorName: string }
| { userId: string }
```

peterstarling: your interface is still invalid. The difference between three optional fields and what I suggested is enormous, it's no surprise that its not visible here though as you don't have tests for the validation either.

You interface cannot allow sending across, for instance, both userId and externalAuthorName in the same object. What would you do in that case? Your code seems to first pick the userId. It should instead throw an error.

So you should:
1. Create a valid interface using unions, as suggested
2. Update your validation to only ever allow ONE property
3. Test the validation

peterstarling (changes requested):

peterstarling (changes requested):

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -272,37 +272,92 @@ describe('/research-outputs/ route', () => {
...
-          },
-        ],
+          ],
+        });
       });
+
+      test.each(fields)(
+        'Should return a validation error when passing more than one required fields (%s)',
+        async (field) => {
+          researchOutputControllerMock.create.mockRejectedValueOnce(
```
peterstarling, line 323: why do you need this?

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -272,37 +272,92 @@ describe('/research-outputs/ route', () => {
...
 
-      const response = await supertest(app)
-        .post('/research-outputs')
-        .send({
-          ...researchOutput,
-          authors: [{ userId: '222', externalAuthorId: '333' }],
-        })
-        .set('Accept', 'application/json');
+      test('Should return a validation error when required field is missing', async () => {
+        researchOutputControllerMock.create.mockRejectedValueOnce(
```
peterstarling, line 280: why do you need this?

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -272,37 +272,92 @@ describe('/research-outputs/ route', () => {
...
-            instancePath: '/authors/0',
-            schemaPath: '#/properties/authors/items/oneOf',
-            keyword: 'oneOf',
-            params: {
-              passingSchemas: [0, 1],
+        const response = await supertest(app)
+          .post('/research-outputs')
+          .send({
+            ...researchOutput,
+            authors: [{ name: 'random-value' }],
```
peterstarling, line 288: name sounds a bit like externalAuthorName, i would suggest using a completely fictional property like 'randomFied'

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -272,37 +272,92 @@ describe('/research-outputs/ route', () => {
...
             },
-            message: 'must match exactly one schema in oneOf',
-          },
-        ],
+          ],
+        });
       });
+
+      test.each(fields)(
+        'Should return a validation error when passing more than one required fields (%s)',
```
peterstarling, line 321: i dont understand this test, you seem to override one of each of the property values in the author object with 'id-1'.

What I asked previously was to make sure that you can only send:
```typescript
{
   userId: 'user-id'
}
```
OR
```typescript
{
   externalAuthorId: 'external-author-uid'
}
```
OR
```typescript
{
   externalAuthorName: 'external-author-name'
}
```
So you should not be able to send
```typescript
{
   somethingRandom: 'external-author-name'
}
```
which you tested correctly
OR
```typescript
{
   externalAuthorName: 'external-author-name',
   externalAuthorId: 'author-1',
}
```
which you have not, similarily you should not be able to send:
```typescript
{
   userId: 'external-author-name',
   externalAuthorId: 'author-1',
}
```
OR
```typescript
{
   userId: 'external-author-name',
   externalAuthorName: 'author-1',
}
```
OR
```typescript
{
   userId: 'external-author-name',
externalAuthorId: 'author-1',
   externalAuthorName: 'author-1',
}
```

does it make sense?

peterstarling (changes requested):

peterstarling (approved):

PR #1505, opened 2022-04-08
fix: Validation error typo

peterstarling:


packages/model/src/error.ts
```diff
@@ -2,7 +2,7 @@ import { ErrorObject } from 'ajv';
 
 export type ValidationErrorResponse = {
   error: 'Bad Request';
-  message: 'Validation Error';
+  message: 'Validation error';
```
niezgoda, line 5: it would be easier to have validation error message const

niezgoda (approved):

packages/model/src/error.ts
peterstarling: to be fair i would have done but i think we want to change the way we throw these errors (see: https://asaphub.atlassian.net/browse/CRN-777) and i think at that point the const will become obsolete - i would imagine we should just use specific error instances? what do you thnink

peterstarling: actually i've made the change :D

niezgoda (approved):

peterstarling:
@kmaid have you tested FE with this change? are we merging it?

packages/model/tests/error.test.ts
```diff
@@ -1,12 +1,15 @@
-import { isValidationErrorResponse } from '../src/error';
+import {
+  isValidationErrorResponse,
+  validationErrorMessage,
```
kmaid, line 3: ???

PR #1519, opened 2022-04-19
CRN-799 Migrate subtype into type - use type field instead of subtype

niezgoda:
Use type field from squidex instead of subtype


apps/crn-server/src/autogenerated-gql/graphql.ts
```diff
@@ -3696,12 +3696,12 @@ export type ResearchOutputsDataSharingStatusInputDto = {
   iv: Maybe<Scalars['String']>;
 };
 
-/** The structure of the Subtype field of the Research Outputs content type. */
+/** The structure of the Subtype (deprecated) field of the Research Outputs content type. */
```
markcsims, line 3699: Can we remove the subtype from the schema?

packages/squidex/schema/schemas/research-outputs.json
```diff
@@ -212,7 +212,7 @@
           "inlineEditable": true,
           "contentType": "Unspecified",
           "editor": "Dropdown",
-          "label": "Subtype",
+          "label": "Subtype (deprecated)",
```
markcsims, line 215: Can this be removed?

markcsims (changes requested):

apps/crn-server/src/autogenerated-gql/graphql.ts
niezgoda: Maybe in next PR when we are going to be sure that everything is working correctly?

markcsims (approved):

PR #1542, opened 2022-04-27
feat: CRN-807 Add methods to ASAP API

peterstarling:


packages/squidex/schema/schemas/research-tags.json
```diff
@@ -61,29 +61,41 @@
         "properties": {
           "fieldType": "Tags",
           "allowedValues": [
-            "Activity Assay",
+            "Preprint",
```
peterstarling, line 64: made a mistake previously with the allowed values, the `Types` field should take the list of research-output types

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -157,9 +160,41 @@ export default class ResearchOutputs implements ResearchOutputController {
...
+      await this.squidexGraphqlClient.request<
+        FetchResearchTagsQuery,
+        FetchResearchTagsQueryVariables
+      >(FETCH_RESEARCH_TAGS, {
+        top: 20,
+        skip: 0,
+      });
+
+    const researchOutputMethods = researchOutputData.methods.map((method) => {
+      const methodTags = queryResearchTagsContentsWithTotal?.items?.filter(
```
niezgoda, line 173: find?

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -157,9 +160,41 @@ export default class ResearchOutputs implements ResearchOutputController {
...
+        (tag) => tag.flatData.name === method,
+      );
+
+      if (methodTags && methodTags.length > 0 && methodTags[0]) {
+        return methodTags[0].id;
+      }
+
+      throw Boom.badRequest('Validation error', [
```
niezgoda, line 181: it would be better to tell which method does not exists

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -157,9 +160,41 @@ export default class ResearchOutputs implements ResearchOutputController {
       authors.map((author) => this.associateResearchOutputToAuthors(author)),
     );
 
+    const { queryResearchTagsContentsWithTotal } =
+      await this.squidexGraphqlClient.request<
+        FetchResearchTagsQuery,
+        FetchResearchTagsQueryVariables
+      >(FETCH_RESEARCH_TAGS, {
+        top: 20,
```
niezgoda: why the hard limit is 20?
what if there will be 21 tags - which is possible for users and RO

niezgoda (changes requested):

apps/crn-server/src/controllers/research-tags.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default class ResearchTags implements ResearchTagController {
+  squidexGraphqlClient: SquidexGraphqlClient;
+
+  constructor(squidexGraphqlClient: SquidexGraphqlClient) {
+    this.squidexGraphqlClient = squidexGraphqlClient;
+  }
+
+  async fetch(options: {
```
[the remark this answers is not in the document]
peterstarling: we would, but it's a lot to do in one PR. At the moment we only have one category for only one entity so I figured we would add them when we actually need them, what do you think?

apps/crn-server/src/controllers/research-outputs.ts
peterstarling: as per requirement atm there are only 23 tags - i have set the limit to a hard 100.

niezgoda (approved):

PR #1543, opened 2022-04-27
Publishing a shared output

markcsims:


apps/crn-frontend/src/network/teams/state.ts
```diff
@@ -165,3 +176,18 @@ export const useAuthorSuggestions = () => {
...
+  const authorization = useRecoilValue(authorizationState);
+  const setRefreshListing = useRefreshResearchOutputListing();
+  const setResearchOutputItem = useSetResearchOutputItem();
+  return async (payload: ResearchOutputPostRequest) => {
+    const researchOutput = await createTeamResearchOutput(
+      payload,
+      authorization,
+    );
+    setResearchOutputItem(researchOutput);
+    setRefreshListing();
```
kmaid, line 190: I think this is broken. Shouldn't it increase the integer?

apps/crn-frontend/src/shared-research/state.ts
```diff
@@ -156,17 +154,13 @@ const setResearchOutput = selector<ResearchOutputResponse | undefined>({
   },
 });
 
-export const usePostTeamResearchOutput = () => {
-  const authorization = useRecoilValue(authorizationState);
+export const useRefreshResearchOutputListing = () => {
```
kmaid, line 157: I think if we increase the number from the other place we can remove this hook as it should never be called in isolation

apps/crn-frontend/src/shared-research/state.ts
```diff
@@ -1,30 +1,42 @@
...
 const researchOutputIndexState = atomFamily<
   { ids: ReadonlyArray<string>; total: number } | Error | undefined,
-  ResearchOutputListOptions
+  RefreshResearchOutputListOptions
 >({
   key: 'researchOutputIndex',
   default: undefined,
 });
+
+const refreshResearchOutputListing = atom<number>({
```
kmaid, line 35: Perhaps this should be called refreshResearchOutputIndex?

packages/react-components/src/organisms/Form.tsx
```diff
@@ -53,11 +51,12 @@ const Form = <T extends void | Record<string, unknown>>({
     if (formRef.current!.reportValidity() && parentValidation) {
       setStatus('isSaving');
       try {
-        const result = await onSave();
-        if (formRef.current) {
-          setStatus('hasSaved');
-        }
-        return result;
+        return await onSave().then((result) => {
```
kmaid, line 54: I think we can remove this promise chain to improve readability

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
```diff
@@ -160,9 +161,10 @@ const TeamCreateOutputForm: React.FC<TeamCreateOutputFormProps> = ({
         authors.length !== 0 ||
         identifierType !== ResearchOutputIdentifierType.None ||
         identifier !== '' ||
+        labCatalogNumber !== '' ||
         teams.length !== 1 // Original team
       }
-      onSave={() => {
+      onSave={async () => {
```
kmaid, line 167: This isn't async, Do we want to remove this?

packages/react-components/src/organisms/__tests__/Form.test.tsx
```diff
@@ -17,6 +17,7 @@ import { Button } from '../../atoms';
 const props: ComponentProps<typeof Form> = {
   dirty: false,
   children: () => null,
+  onSave: jest.fn(),
```
kmaid, line 20: I think this should be a onSave: ()=>Promise.Resolve()

apps/crn-frontend/src/network/teams/state.ts
markcsims: That's what setRefreshListing  is doing. Although we could add that call to setResearchOutputItem?

apps/crn-frontend/src/shared-research/state.ts
markcsims: Done

apps/crn-frontend/src/network/teams/state.ts
markcsims: Done as below

apps/crn-frontend/src/shared-research/state.ts
markcsims: Done

packages/react-components/src/organisms/Form.tsx
markcsims: Done

packages/react-components/src/organisms/TeamCreateOutputForm.tsx
markcsims: done

packages/react-components/src/organisms/__tests__/Form.test.tsx
markcsims: done

kmaid (approved):

PR #1544, opened 2022-04-27
Add additional lab catalog field

kmaid:
Remove default research output code

### DIFF
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from 'react';
 import {
   ValidationErrorResponse,
-  ResearchOutputPostRequest,
   ResearchOutputDocumentType,
   isValidationErrorResponse,
 } from '@asap-hub/model';
@@ -69,23 +68,8 @@ const TeamOutput: React.FC<TeamOutputProps> = ({ teamId }) => {
 
   const createResearchOutput = usePostTeamResearchOutput();
 
-  const defaultOutput: ResearchOutputPostRequest = {
-    documentType,
-    title: 'Output created through the ROMS form',
-    asapFunded: undefined,
-    sharingStatus: 'Network Only',
-    usedInPublication: undefined,
-    description: 'example',
-    type: 'Software',
-    addedDate: new Date().toISOString(),
-    tags: [],
-    teams: [teamId],
-    publishDate: undefined,
-  };
-
   const getLabSuggestions = useLabSuggestions();
   const getAuthorSuggestions = useAuthorSuggestions();
-
   const getTeamSuggestions = useTeamSuggestions();
 
   const showCreateOutputPage = isEnabled('ROMS_FORM');
@@ -116,11 +100,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({ teamId }) => {
             setErrors(clearAjvErrorForPath(errors, instancePath))
           }
           onSave={(output) =>
-            createResearchOutput({
-              ...defaultOutput,
-              ...output,
-              addedDate: new Date().toISOString(),
-            }).catch((error) => {
+            createResearchOutput(output).catch((error) => {
               if (error instanceof BackendError) {
                 const { response } = error;
                 if (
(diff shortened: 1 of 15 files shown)


apps/crn-frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -116,11 +100,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({ teamId }) => {
             setErrors(clearAjvErrorForPath(errors, instancePath))
           }
           onSave={(output) =>
-            createResearchOutput({
-              ...defaultOutput,
-              ...output,
-              addedDate: new Date().toISOString(),
-            }).catch((error) => {
+            createResearchOutput(output).catch((error) => {
```
markcsims, line 103: Should this be a try...catch rather than a promise chain?

packages/react-components/src/organisms/Form.tsx
```diff
@@ -53,11 +51,12 @@ const Form = <T extends void | Record<string, unknown>>({
...
-        if (formRef.current) {
-          setStatus('hasSaved');
-        }
-        return result;
+        return await onSave().then((result) => {
+          if (formRef.current) {
+            setStatus('hasSaved');
+          }
+          return result;
+        });
```
markcsims, line 59: Mixing try...catch with a promise chain

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -57,184 +50,242 @@ it('renders the form', async () => {
...
+  const { getByText } = render(
+    <StaticRouter>
+      <TeamCreateOutputForm {...props} getLabSuggestions={getLabSuggestions} />
+    </StaticRouter>,
+  );
+  userEvent.click(screen.getByLabelText(/Labs/i));
+  await waitFor(() =>
+    expect(screen.queryByText(/loading/i)).not.toBeInTheDocument(),
+  );
+  expect(getByText(/Sorry, no labs match/i)).toBeVisible();
```
markcsims, line 83: We're using screen on line 81, and not on line 83. Should we be using it consistently?

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -57,184 +50,242 @@ it('renders the form', async () => {
...
+    setupForm();
 
-  userEvent.click(screen.getByLabelText(/Authors/i));
-  await waitFor(() =>
-    expect(screen.queryByText(/loading/i)).not.toBeInTheDocument(),
-  );
+    userEvent.click(screen.getByLabelText(/Authors/i));
+    await waitFor(() =>
+      expect(screen.queryByText(/loading/i)).not.toBeInTheDocument(),
+    );
```
markcsims, line 219: This is repeated, should it be moved to a helper function?

packages/react-components/src/organisms/Form.tsx
kmaid: This was driving me crazy. Without it it seems to set status after the result has been returned so i get the batcher error. I might convert it all to be a promise chain though.

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
kmaid: Its only repeated twice. The third time you have to type in an external author in order to create the new external author

apps/crn-frontend/src/network/teams/TeamOutput.tsx
kmaid: image

This is what it would look like if I converted it into async. You have to do the weird return await in order to catch the error from the parent promise. You also have to return a Promise.Resolve as the caught error is now handled which is ugly too. Tempted to leave this as is.

packages/react-components/src/organisms/Form.tsx
kmaid: I have discussed this with @markcsims as it seems like a weird race condition forcing me to do this in the admittedly weird way. He is going to give it a go and see if I am missing something as whats going on doesn't make sense to me.

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
kmaid: Fixed

markcsims (approved):

kmaid:
This has been merged into https://github.com/yldio/asap-hub/pull/1543

PR #1548, opened 2022-04-28
Add additional lab catalog field

kmaid:
Remove default research output code

### DIFF
--- a/packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
+++ b/packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
@@ -105,7 +105,8 @@ describe('on submit', () => {
   });
   const getLabSuggestions = jest.fn();
   const getAuthorSuggestions = jest.fn();
-  const saveFn = jest.fn();
+  const promise = Promise.resolve();
+  const saveFn = jest.fn(() => promise);
   const expectedRequest: ResearchOutputPostRequest = {
     documentType: 'Article',
     tags: [],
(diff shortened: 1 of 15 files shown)


packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -165,11 +162,12 @@ describe('on submit', () => {
     });
   };
   const submitForm = async () => {
-    const button = screen.getByRole('button', { name: /Share/i });
-    userEvent.click(button);
+    userEvent.click(screen.getByRole('button', { name: /Share/i }));
 
-    expect(screen.getByRole('button', { name: /Share/i })).not.toBeEnabled();
-    expect(screen.getByRole('button', { name: /Cancel/i })).not.toBeEnabled();
+    waitFor(() => {
```
kmaid, line 167: I belive this returns a promise so without an away I am not sure this works as intended.

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -178,6 +176,13 @@ describe('on submit', () => {
     );
   };
 
+  it('can submit a form with minimum data', async () => {
+    setupForm();
+    await submitForm();
+    expect(saveFn).toHaveBeenLastCalledWith(expectedRequest);
+    await act(() => promise);
```
kmaid, line 183: I think this should be moved into submitForm function

kmaid:
🖕

PR #1551, opened 2022-04-29
CRN-843 Create an API to send speakers details from the CMS to FE application

niezgoda:
### Requirement
- Create an API to send speakers details from the CMS to the FE. 
-- The API should include the team name, user name and the user role (user object, inc. avatar)
- This response should be sent to the FE in the order that is specified in the CMS
- Maximum of 20 speakers to be provided in the response

### Note
- Can only be tested when the FE is completed - dev tests in first instance


niezgoda:
image

peterstarling (approved):
I would add some more tests around parsing of the events - I know you capture like 3 different cases in your fixtures but I think i would split them, maybe have a fixture of just one valid approach and then test the other cases individually by modifying the existing fixture. But it's your call

peterstarling (approved):

PR #1553, opened 2022-05-03
CRN-833 Enable PMs and ASAP Staff to create shared output

niezgoda:
### Requirements
1. ASAP Staff role should see the share an output button when they navigate to /network/teams/about OR /network/teams/outputs - (users with the role ASAP staff who are viewing any team page)
2. PM role should see the share an output button for the team with which they are associated - (users with team role = Project Manager who are viewing their own team page)
3. All other users should not see the ‘Share an Output’ button when they navigate to the /network/teams/about OR /network/teams/outputs
4. If the user tries to access the page directly (copy and paste link into browser) they should receive an appropriate error message, for example ‘page does not exist’
5. For the users that do not have the right permissions the API should respond with a 403 error message (unauthorised)

### Notes
- “creation” permission should be decided on a team page level


apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -48,14 +49,25 @@ export const researchOutputRouteFactory = (
   researchOutputRoutes.post('/research-outputs', async (req, res) => {
     const { body, loggedInUser } = req;
 
+    // Middleware makes sure that the user is logged in at this stage
+    /* istanbul ignore next */
```
peterstarling, line 53: I think in such a case maybe its better to cast `req` as req with the loggedInUser? in fact you don't need to 'cast', just give the req to the `post` method which accepts a generic

packages/validation/test/permissions/has-create-research-output-permission.test.ts
```diff
@@ -0,0 +1,73 @@
+import { hasCreateResearchOutputPermissions } from '../../src/permissions/has-create-research-output-permission';
+import { createUserResponse } from '@asap-hub/fixtures';
+
+describe('hasCreateResearchOutputPermission', () => {
+  test('Should not has permission when there is no teams assigned to user', () => {
```
peterstarling, line 5:
```suggestion
  test('Should not have permissions when there are no teams assigned to the user', () => {
```

packages/validation/src/permissions/has-create-research-output-permission.ts
```diff
@@ -0,0 +1,14 @@
+import { User } from '@asap-hub/auth';
+import { ResearchOutputPostRequest } from '@asap-hub/model';
+
+export const hasCreateResearchOutputPermissions = (
+  user: Omit<User, 'algoliaApiKey'>,
+  teams: ResearchOutputPostRequest['teams'],
```
peterstarling, line 6: I wonder if this is the right interface, I feel like such validation method should take the user and the whole research output input, even though right now only teams are needed. Normally resource based permissions would be validated on per user and resource ID basis which is why I think that is what the input should reflect here

packages/validation/test/permissions/has-create-research-output-permission.test.ts
```diff
@@ -0,0 +1,73 @@
...
+              proposal: 'proposalId1',
+            },
+          ],
+        },
+        ['team-1', 'team-2'],
+      ),
+    ).toEqual(true);
+  });
+
+  test('Should has permission when the user is Staff in any team', () => {
```
peterstarling, line 55:
```suggestion
  test('Should have the permissions when the user has a Staff role in any team', () => {
```

packages/validation/test/permissions/has-create-research-output-permission.test.ts
```diff
@@ -0,0 +1,73 @@
...
+              proposal: 'proposalId1',
+            },
+          ],
+        },
+        ['team-1', 'team-2'],
+      ),
+    ).toEqual(false);
+  });
+
+  test('Should has permission when the user is Project Manager in particular team', () => {
```
peterstarling, line 36:
```suggestion
  test('Should have the permission when the user has a Project Manager role inside the team that this resource belongs to', () => {
```

packages/validation/test/permissions/has-create-research-output-permission.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        {
+          ...createUserResponse(),
+          teams: [],
+        },
+        ['team-1', 'team-2'],
+      ),
+    ).toEqual(false);
+  });
+
+  test('Should not has permission when the user is not Project Manager in particular team', () => {
```
peterstarling, line 17:
```suggestion
  test('Should not have the permissions to create when the user does not have a Project Manager role inside the team that this resource belongs to', () => {
```

peterstarling (changes requested):

packages/validation/src/permissions/has-create-research-output-permission.ts
niezgoda: On the FE we have only the team ID when we want to check if the user has a permission to show the create RO button

apps/crn-frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -63,16 +63,16 @@ const TeamOutput: React.FC<TeamOutputProps> = ({ teamId }) => {
   );
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
-  const { isEnabled } = useFlags();
+
+  const { canCreate } = useContext(ResearchOutputPermissionsContext);
```
markcsims, line 67: Why do we need to wrap the components with a Context? I can understand it if it's deeply nested but in this case we can access state directly.

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -46,16 +53,21 @@ export const researchOutputRouteFactory = (
   );
 
   researchOutputRoutes.post('/research-outputs', async (req, res) => {
-    const { body, loggedInUser } = req;
+    const { body, loggedInUser } = req as ResearchOutputPostRequest;
```
markcsims, line 56: Can we avoid the "as", for example doing something like:

researchOutputRoutes.post<ResearchOutputPostRequest>(

markcsims (changes requested):

apps/crn-frontend/src/network/teams/TeamOutput.tsx
niezgoda: That was @kmaid suggestion :)

markcsims (approved):

peterstarling (approved):

PR #1555, opened 2022-05-05
Bug/Aloglia pagination bugfix

niezgoda:
Increase limit from 1000 to 10000 as currently there is an error shown when the user selects page above 100 for search results.

### DIFF
--- a/packages/algolia/algolia-schema.json
+++ b/packages/algolia/algolia-schema.json
@@ -46,7 +46,7 @@
   ],
   "attributesToSnippet": null,
   "attributesToHighlight": null,
-  "paginationLimitedTo": 1000,
+  "paginationLimitedTo": 10000,
   "attributeForDistinct": null,
   "exactOnSingleWordQuery": "word",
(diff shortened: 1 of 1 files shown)


packages/algolia/algolia-schema.json
```diff
@@ -46,7 +46,7 @@
   ],
   "attributesToSnippet": null,
   "attributesToHighlight": null,
-  "paginationLimitedTo": 1000,
+  "paginationLimitedTo": 3000,
```
jhonny-m, line 49: on the export function the page size limit is set to 10000, I think that it will cause an error once again when the number of RO increases, this is set in this variable `MAX_ALGOLIA_RESULTS`

jhonny-m: https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/shared-research/export.ts#L10

markcsims (approved):

peterstarling (approved):

PR #1557, opened 2022-05-05
CRN-802 identifier field defaults

markcsims:
https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/8?modal=detail&selectedIssue=CRN-802

1/When I land on the RO form, and navigate to the identifier field I do not want to see a default option of ‘None’ populated in the field. This should have a place holder ‘Choose an identifier’.

2/If I select ‘ASAP Fund' and ‘Used in a publication ' (radio buttons), then I want to remove ‘None’ from the drop down options AND I do not want to populate the field with a default of 'DOI’. DOI and Accession number should be available as options in the list (NO DEFAULTS) pre-populated

image

### DIFF
--- a/packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
+++ b/packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
@@ -11,7 +11,7 @@ import {
   waitForElementToBeRemoved,
   within,
 } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
+import userEvent, { specialChars } from '@testing-library/user-event';
 import { createMemoryHistory, History } from 'history';
 import { ComponentProps } from 'react';
 import { Router, StaticRouter } from 'react-router-dom';
@@ -190,10 +190,13 @@ describe('on submit', () => {
       screen.getByRole('textbox', { name: /description/i }),
       data.description,
     );
-    userEvent.type(
-      screen.getByRole('textbox', { name: /Select the option/i }),
-      data.type,
-    );
+
+    const typeDropdown = screen.getByRole('textbox', {
+      name: /Select the option/i,
+    });
+    userEvent.type(typeDropdown, data.type);
+    userEvent.type(typeDropdown, specialChars.enter);
+
     const identifier = screen.getByRole('textbox', { name: /identifier/i });
     userEvent.type(identifier, 'DOI');
     identifier.blur();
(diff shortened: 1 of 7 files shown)


packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -76,23 +81,47 @@ it('displays proper message when no lab is found', async () => {
...
         {...props}
         team={{ ...createTeamResponse(), displayName: 'example team' }}
       />
     </StaticRouter>,
   );
-  expect(getByText('example team')).toBeVisible();
+  expect(screen.getByText('example team')).toBeVisible();
+});
+
+it('if funded and publication are both yes, then the None option is removed.', async () => {
```
kmaid, line 101:
```suggestion
it('is funded and publication are both yes, then the none option is removed.', async () => {
```

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
```diff
@@ -154,19 +184,25 @@ describe('on submit', () => {
...
-      target: { value: data.title },
-    });
-    fireEvent.change(screen.getByLabelText(/description/i), {
-      target: { value: data.description },
-    });
-    userEvent.type(screen.getByLabelText(/Select the option/i), data.type);
-    fireEvent.keyDown(screen.getByLabelText(/Select the option/i), {
-      keyCode: ENTER_KEYCODE,
-    });
+    userEvent.type(screen.getByRole('textbox', { name: /url/i }), data.link!);
```
kmaid, line 187: I had this as fire event because its faster than typing AFAIK

kmaid (approved):

packages/react-components/src/organisms/__tests__/TeamCreateOutputForm.test.tsx
markcsims: The code was:

```
userEvent.type(screen.getByLabelText(/Select the option/i), data.type);
    fireEvent.keyDown(screen.getByLabelText(/Select the option/i), {
      keyCode: ENTER_KEYCODE,
    });
```
I removed the fireEvent as you don't need it with out Dropdown.

```
userEvent.type(
      screen.getByRole('textbox', { name: /Select the option/i }),
      data.type,
    );

```

PR #1562, opened 2022-05-06
CRN-848 Validation error to be displayed on the URL field when user does not specify the domain

niezgoda:
Expected Behaviour: User should be directed to the URL field with an error message ‘Please enter a valid URL’. This should  not be displayed at the top of the page.

### DIFF
--- a/apps/crn-server/test/controllers/research-outputs.test.ts
+++ b/apps/crn-server/test/controllers/research-outputs.test.ts
@@ -871,7 +871,7 @@ describe('ResearchOutputs controller', () => {
         expect(squidexGraphqlClientMock.request).toHaveBeenCalledWith(
           expect.anything(),
           expect.objectContaining({
-            filter: `(data/link/iv eq '${researchOutputRequest.link}')`,
+            filter: `(data/link/iv eq 'https:%2F%2Fhub.asap.science%2F')`,
           }),
           {
             includeDrafts: true,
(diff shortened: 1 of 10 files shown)


packages/react-components/src/templates/ContactInfoModal.tsx
```diff
@@ -113,7 +113,7 @@ const ContactInfoModal: React.FC<ContactInfoModalProps> = ({
             <LabeledTextField
               title="Website 1"
               subtitle="(Optional)"
-              pattern={USER_SOCIAL_WEBSITE.source}
+              pattern={UrlExpression}
```
markcsims, line 116: Should this be tested?

packages/react-components/src/templates/ContactInfoModal.tsx
```diff
@@ -126,7 +126,7 @@ const ContactInfoModal: React.FC<ContactInfoModalProps> = ({
             <LabeledTextField
               title="Website 2"
               subtitle="(Optional)"
-              pattern={USER_SOCIAL_WEBSITE.source}
```
markcsims, line 129: Should this be tested?

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -1038,7 +1044,9 @@ describe('ResearchOutputs controller', () => {
         expect(squidexGraphqlClientMock.request).toHaveBeenCalledWith(
           expect.anything(),
           expect.objectContaining({
-            filter: `(data/link/iv eq '${researchOutputRequest.link}')`,
+            filter: `(data/link/iv eq '${sanitiseForSquidex(
+              researchOutputRequest.link || '',
+            )}')`,
```
markcsims, line 1049: I don't think we should be asserting using sanitiseForSquidex. It's from the production code, and it's part of the method under test.

markcsims (changes requested):

packages/react-components/src/templates/ContactInfoModal.tsx
niezgoda: image

markcsims (approved):

PR #1565, opened 2022-05-09
Moving frontend functions into packages

jhonny-m:
Moving some frontend functions and component into a shared package so it can be used by gp2-frontend without duplication


packages/frontend-utils/src/api-util/api-util.ts
```diff
@@ -10,10 +9,9 @@ export type GetListOptions = {
 };
 
 export const createListApiUrl = (
-  endpoint: string,
+  url: URL,
```
peterstarling, line 12: I think by shifting the constructing of the URL to the client of this util function you are removing some of it's functionality too, I realise it's because the package does not have access to the config - perhaps you considered making this a factory and then importing in FE, injecting and exporting again with the base url already in it?

peterstarling (approved):

PR #1589, opened 2022-05-13
feat: CRN-853 Add PUT endpoint for RO

peterstarling:


niezgoda (approved):

packages/squidex/src/utils.ts
```diff
@@ -29,12 +29,17 @@ export type RequiredAndNonNullable<T> = Required<
...
-}): SquidexEntity =>
-  Object.entries(object).reduce((acc, [key, value]) => {
-    acc[key] = { iv: value };
-    return acc;
-  }, {} as SquidexEntity);
+export const parseToSquidex = <T>(object: T): SquidexEntityObject<T> =>
+  Object.entries(object).reduce(
+    (acc, [key, value]) => ({
+      ...acc,
+      [key]: typeof value === 'undefined' ? undefined : { iv: value },
```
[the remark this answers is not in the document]
peterstarling: I have added some tests to make sure we know what the desired behaviour is, I have also compared against what squidex input should look like. Basically undefined should be removed from the var whereas nulls should be given as iv: null which is what it does now

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -139,6 +140,9 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
       'sharingStatus',
       'addedDate',
       'teams',
+      'methods',
```
niezgoda, line 143: there is two tasks on the fe part - so when released prior to them - adding RO is going to stop working

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -62,42 +59,35 @@ export const researchOutputRouteFactory = (
...
   });
 
+  researchOutputRoutes.put(
+    '/research-outputs/:researchOutputId',
+    async (req, res) => {
+      const { body, params, loggedInUser } = req;
+      const { researchOutputId } = validateResearchOutputParameters(params);
+      const updateRequest = validateResearchOutputPutRequestParameters(body);
+      validateResearchOutputPostRequestParametersIdentifiers(body);
+
```
niezgoda, line 78: you are not checking the  hasCreateUpdateResearchOutputPermissions

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -147,79 +152,82 @@ export default class ResearchOutputs implements ResearchOutputController {
...
-      subtype,
-    });
+      researchOutputId = result.id;
 
-    await Promise.all(
-      teams.map((teamId) =>
-        this.associateResearchOutputToTeam(teamId, researchOutputId),
-      ),
-    );
+      await Promise.all(
```
markcsims, line 225: I know this isn't from this PR, but maybe one to discuss. 

We're using a Promise.all, and which will reject upon first failure.

niezgoda (approved):

PR #1596, opened 2022-05-18
CRN-694 Edit organisms

niezgoda:
Edit organisms


peterstarling (approved):
Looks good at least from BE perspective

PR #1613, opened 2022-05-24
feat: CRN-847 Add external users to event speakers in the CMS

peterstarling:


niezgoda:
Also CR the script in events schema: https://cloud.squidex.io/app/1613/schemas/events?tab=scripts

apps/crn-server/src/entities/event.ts
```diff
@@ -45,28 +72,50 @@ export const parseEventSpeaker = (
...
 
+    if (user?.__typename === 'ExternalAuthors') {
+      return {
+        externalUser: parseEventSpeakerExternalUser(user),
+      };
+    }
+
     if (!team) {
-      throw new Error('Team is required in event speaker');
+      throw new Error('Either team or external author is required');
```
markcsims, line 96: Should this be throwing GenericError?

apps/crn-server/test/entities/events.test.ts
```diff
@@ -1,14 +1,21 @@
+import {
```
markcsims, line 1: Filename doesn't match the production name

packages/react-components/src/organisms/__tests__/SpeakerList.test.tsx
```diff
@@ -66,6 +73,19 @@ describe('When rendering the speaker list', () => {
...
       expect(getByText('Genetics')).toBeVisible();
     });
+
+    it('Renders external author', async () => {
+      const event = {
+        ...createEventResponse(),
+        speakers: [externalSpeaker],
+      };
+
+      const { getByText } = render(<SpeakersList {...event} />);
```
markcsims, line 83: prefer screen over destructuring the return of the render.

markcsims (changes requested):

apps/crn-server/src/entities/event.ts
niezgoda: GenericError is being thrown by APIs

markcsims (approved):

PR #1614, opened 2022-05-24
CRN 617 split be controllers into 2 layers

markcsims:
- fixes DOI/Accession error message
- fixes Not Sure radio selection
- CRN-617 starts moving users to data provider


apps/crn-server/src/data-providers/users.ts
```diff
@@ -0,0 +1,338 @@
...
+  connectByCode(
+    welcomeCode: string,
+    userId: string,
+  ): Promise<UserDataObject | null>;
+  syncOrcidProfile(
+    id: string,
+    cachedUser?: RestUser | undefined,
+  ): Promise<UserDataObject>;
+}
+export default function createUserDataProvider(
```
peterstarling, line 44: I feel like since we use classes for the controllers we should continue to use them for data providers. I don't mind switching to factories but I would prefer to be consistent

apps/crn-server/src/data-providers/users.ts
```diff
@@ -0,0 +1,338 @@
...
+    syncOrcidProfile,
+  };
+}
+const shouldDoFullUpdate = (userToUpdate: UserPatchDataObject) =>
+  userToUpdate.teams?.length ||
+  Object.values(userToUpdate).some(
+    (value) => value.trim && value.trim() === '',
+  );
+
+async function queryForUsers(
```
peterstarling, line 175: shouldnt we just use consts with arrow functions?

apps/crn-server/src/controllers/users.ts
```diff
@@ -52,283 +34,76 @@ export interface UserController {
...
+    const users = await this.userDataProvider.fetch(options);
 
-    const filterTeams = (options?.filter?.teamId || [])
-      .reduce(
-        (acc: string[], teamId: string) =>
-          acc.concat([`data/teams/iv/id eq '${teamId}'`]),
-        [],
-      )
-      .join(' or ');
+    const total = users.length;
```
niezgoda, line 52: is this correct?
I do not think that we are fetching all of the users at once but rather we paginate :)

apps/crn-server/src/data-providers/users.ts
```diff
@@ -0,0 +1,338 @@
...
+  ): Promise<UserDataObject | null>;
+  syncOrcidProfile(
+    id: string,
+    cachedUser?: RestUser | undefined,
+  ): Promise<UserDataObject>;
+}
+export default function createUserDataProvider(
+  squidexGraphlClient: SquidexGraphqlClient,
+): UserDataProvider {
+  const userSquidexRestClient = new SquidexRest<RestUser>('users');
```
niezgoda, line 47: is there a reason that this is not a class?

apps/crn-server/src/data-providers/users.ts
```diff
@@ -0,0 +1,338 @@
...
+  const fetch = async (options: FetchUsersOptions) => {
+    const queryFilter = generateFetchQueryFilter(options);
+    const { take = 8, skip = 0 } = options;
+    const users = await queryForUsers(
+      squidexGraphlClient,
+      queryFilter,
+      take,
+      skip,
+    );
+    return users.map(parseGraphQLUserToDataObject);
```
niezgoda, line 82: you should return here also the total from the query response

apps/crn-server/src/data-providers/users.ts
```diff
@@ -0,0 +1,338 @@
...
+  };
+  const fetchByCode = async (code: string) => {
+    const filter = `data/connections/iv/code eq '${code}'`;
```
niezgoda, line 85: you can use here queryByCode

niezgoda (dismissed):

peterstarling (changes requested):

peterstarling (changes requested):

apps/crn-server/src/app.ts
```diff
@@ -1,56 +1,55 @@
...
-
-import { tracingHandlerFactory } from './middleware/tracing-handler';
-
-import Groups, { GroupController } from './controllers/groups';
-import Users, { UserController } from './controllers/users';
-import Teams, { TeamController } from './controllers/teams';
-import Events, { EventController } from './controllers/events';
-import Dashboard, { DashboardController } from './controllers/dashboard';
 import Calendars, { CalendarController } from './controllers/calendars';
+import Dashboard, { DashboardController } from './controllers/dashboard';
```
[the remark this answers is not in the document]
markcsims: I think we should rename all the controllers at the same time. Maybe in another PR?

apps/crn-server/src/controllers/users.ts
markcsims: We do paginate, this is now done in the data provider.

apps/crn-server/src/controllers/users.ts
```diff
@@ -52,283 +34,76 @@ export interface UserController {
...
-      throw Boom.forbidden();
+    const users = await this.userDataProvider.fetchByCode(code);
+    if (users.length === 0) {
+      throw new NotFoundError(`user with code ${code} not found`);
     }
 
-    const { total, items } = queryUsersContentsWithTotal;
-    if (total !== 1 || !items?.[0]) {
-      throw Boom.forbidden();
+    if (users.length !== 1 || !users[0]) {
```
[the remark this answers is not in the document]
markcsims: It either returns an Error or the User

apps/crn-server/src/data-providers/users.ts
markcsims: The total is added in by the controller.

apps/crn-server/src/data-providers/users.ts
markcsims: Yes, but this ticket is to deal with splitting the controller into 2. 

We are going to create another ticket for the refactoring. 

In the end the controller will probably have only 3 methods available

fetch
fetchById
update

apps/crn-server/src/data-providers/users.ts
markcsims: done

apps/crn-server/scripts/export-entity.ts
```diff
@@ -46,3 +43,18 @@ export const exportEntity = async (
...
+function getController(entity: Entity) {
+  const controllerMap = {
+    users: Users,
+    'research-outputs': ResearchOutputs,
+    'external-authors': ExternalAuthors,
+  };
+  const squidexGraphqlClient = new SquidexGraphql();
+  if (entity === 'users') {
+    const userDataProvider = createUserDataProvider(squidexGraphqlClient);
+    return new controllerMap[entity](userDataProvider);
```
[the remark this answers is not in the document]
markcsims: absolutely - this is what was there before we started the split.

Our intention is to refactor this further in another PR

peterstarling (approved):

apps/crn-server/src/controllers/users.ts
peterstarling: this is a good point actually, how does your data provider return the count?

niezgoda (approved):

PR #1618, opened 2022-05-26
CRN-850 move crn serverless file into the crn server dir

niezgoda:


.github/actions/sls-deployment/action.yml
```diff
@@ -0,0 +1,88 @@
...
+  crn-app-url:
+    description: 'CRN APP Url'
+    value: ${{ steps.setup.outputs.crn-app-url }}
+  gp2-api-url:
+    description: 'GP2 API Url'
+    value: ${{ steps.setup.outputs.gp2-api-url }}
+  gp2-app-url:
+    description: 'GP2 APP Url'
+    value: ${{ steps.setup.outputs.gp2-app-url }}
+
```
markcsims, line 40: Why do we need outputs?

.github/actions/sls-deployment/action.yml
```diff
@@ -0,0 +1,88 @@
...
+description: Reusable sls deploy
+
+inputs:
+  environment-name:
+    description: 'The environment name'
+    required: true
+  app:
+    description: 'The application - crn-server or gp2-server'
+    required: true
+  AWS_ACCESS_KEY_ID:
```
markcsims, line 11: In the other actions we're tried to make the inputs kebab cased. Can we do the same here?

markcsims (changes requested):

.github/actions/sls-deployment/action.yml
niezgoda: for
image

in reusable-deployment

PR #1649, opened 2022-06-06
[CRN-856] changes to event list view

evemontalvao:


packages/react-components/src/molecules/ToastCard.tsx
```diff
@@ -1,73 +1,92 @@
...
 import { paddingStyles } from '../card';
 
 const toastStyles = css({
   display: 'flex',
-  alignItems: 'flex-start',
-  paddingTop: 0,
-  paddingBottom: `${12 / perRem}em`,
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  padding: `${16 / perRem}em ${24 / perRem}em`,
```
oratiu, line 14: This should be either 15 or 18 as we use multiple of 3 for sizes.

packages/react-components/src/molecules/EventInfo.tsx
```diff
@@ -33,6 +33,7 @@ const imageContainerStyle = css({
 const cardStyles = css({
   display: 'flex',
   flexDirection: 'row',
+  paddingBottom: `${32 / perRem}em`,
```
oratiu, line 36: Here as well 33/perRem

oratiu (changes requested):
Look really good.

packages/react-components/src/molecules/EventInfo.tsx
```diff
@@ -36,11 +31,12 @@ const cardStyles = css({
 });
 
 const listItemStyles = css({
-  padding: `${12 / perRem}em 0`,
+  padding: `${7.5 / perRem}em 0`,
```
evemontalvao, line 34: I know it's not multiple of 3 but Manuel specifically requested 7.5

packages/react-components/src/molecules/EventInfo.tsx
```diff
@@ -36,11 +31,12 @@ const cardStyles = css({
...
   color: lead.rgb,
-  whiteSpace: 'nowrap',
+  whiteSpace: 'break-spaces',
   overflow: 'hidden',
   textOverflow: 'ellipsis',
+  fontSize: `${17 / perRem}em`,
```
evemontalvao, line 39: same as above here

apps/storybook/src/Icons.stories.tsx
```diff
@@ -2,7 +2,7 @@ import * as components from '@asap-hub/react-components';
 
 export default { title: 'Atoms / Icons' };
 
-export const Alert = () => components.alertIcon;
+export const Alert = () => <components.AlertIcon />;
```
kmaid, line 5: ??

packages/react-components/src/icons/alert.tsx
```diff
@@ -31,8 +38,8 @@ const alert = (
                     <g id="Group-10">
                       <g id="Group-4"></g>
                       <rect
+                        fill={color || steel.rgb}
```
kmaid, line 41: Can we set this with CSS? What we normally do is move the fill up to <svg and then target it in CSS

packages/react-components/src/molecules/AssociationList.tsx
```diff
@@ -143,11 +144,15 @@ const AssociationList: React.FC<AssociationListProps> = ({
...
+              {inline || (
+                <div css={dividerStyles}>
+                  <Divider />
+                </div>
+              )}
             </li>
           ))}
           {more && (
-            <li>
+            <li data-testid="associationlist-more">
```
kmaid, line 155: We don't use data-test-ids in this project. Let me know if you need help figuring out what selector to use

kmaid (changes requested):

packages/react-components/src/molecules/AssociationList.tsx
evemontalvao: Just used here because the assertion wasn't working using the role, the list is divided in two groups and I need to assert first one and then the other to make sure the list is divided, using getByRole asserts all of them at once. How do you think it's a better solution for this?

packages/react-components/src/icons/alert.tsx
evemontalvao: In this case, moving the fill to the <svg doesn't work, here's how it looks:
Screenshot 2022-06-10 at 10 20 15

And by the way, why did we choose to use the icons like this instead of using as proper customizable react components?

kmaid: I asked the same question when i joined but I think its because if you import them the other way they aren't put into the bundle by webpack rather loaded separately so for performance.

packages/react-components/src/atoms/__tests__/Link.test.tsx
```diff
@@ -62,6 +62,46 @@ describe('when button-styled', () => {
...
+      <Link href="/" buttonStyle margin={false}>
+        text
+      </Link>,
+    );
+
+    const buttonWithMargin = Number(
+      getComputedStyle(getByRole('link')).margin.replace(/em$/, ''),
+    );
+
+    expect(buttonWithMargin).toBe(0);
```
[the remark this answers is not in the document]
markcsims: Agreed, if there's a margin it would not be zero

packages/react-components/src/molecules/__tests__/EventInfo.test.tsx
```diff
@@ -151,7 +151,7 @@ it('displays the teams with number of additional teams', () => {
     />,
   );
 
-  expect(screen.getAllByRole('listitem').length).toEqual(8);
+  expect(screen.getAllByRole('listitem').length).toEqual(9);
   expect(screen.queryByText('+3')).toBeInTheDocument();
```
markcsims, line 155: We shouldn't be using query here. Use either get or find.

packages/react-components/src/molecules/__tests__/EventInfo.test.tsx
```diff
@@ -169,7 +169,7 @@ it('displays the team only once', () => {
     />,
   );
 
-  expect(screen.getAllByRole('listitem').length).toEqual(1);
+  expect(screen.getAllByText(/the team/i).length).toEqual(1);
```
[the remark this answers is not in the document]
evemontalvao: I'll move this test from here then, cause in the EventInfo component we have more <li> and the test keeps failing because the assertion always results in a number greater than 1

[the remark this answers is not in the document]
evemontalvao: The problem here is that I need to find a specific way to match this <li since there's other 2 <li in the page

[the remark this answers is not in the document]
evemontalvao: I could use ByRole and then use a .find, but it would be pretty much the same thing but with more code

oratiu (approved):

PR #1650, opened 2022-06-07
CRN-819 Add the total number of a user shared outputs

niezgoda:


apps/crn-frontend/src/network/users/UserProfile.tsx
```diff
@@ -50,6 +51,21 @@ const User: FC<Record<string, never>> = () => {
...
+    filters: new Set(),
+    pageSize: 1,
+    searchQuery: '',
+    userId,
+  });
+
+  useEffect(() => {
+    setSharedOutputsCount(result.total || undefined);
+  }, [result.total]);
+
```
markcsims, line 68: I don't think we need to use State here (and no need for the useEffect).

i.e. 

const sharedOutputsCount = result.total || undefined;

or just inline it

apps/crn-frontend/src/network/users/__tests__/UserProfile.test.tsx
```diff
@@ -396,3 +396,12 @@ describe('a header edit button', () => {
     });
   });
 });
+
+it('renders number of shared outputs', async () => {
+  mockGetResearchOutputs.mockResolvedValue({
+    ...createResearchOutputListAlgoliaResponse(5),
+  });
+  const { findByText } = await renderUserProfile(createUserResponse());
```
markcsims, line 404: Prefer screen!

markcsims (changes requested):

markcsims (approved):

PR #1694, opened 2022-06-13
CRN-881-user-profile-events-page-add-upcoming-events-page 

markcsims:


apps/crn-frontend/src/events/__tests__/api.test.ts
```diff
@@ -156,6 +156,20 @@ describe('getEventsFromAlgolia', () => {
...
 
+  it('calls for upcoming events with a certain speaker id', async () => {
+    search.mockResolvedValueOnce(createAlgoliaResponse<'event'>([]));
+
+    await getEventsFromAlgolia(algoliaSearchClient, {
+      ...getEventListOptions(new Date('2021-01-01T12:00:00'), false),
+      userId: 'user-1',
+    });
+    expect(search).toBeCalledWith(['event'], '', {
+      filters: 'endDateTimestamp > 1609498800 AND speakers.user.id: user-1',
```
niezgoda, line 167: the userId should be in ""

apps/crn-frontend/src/events/__tests__/api.test.ts
```diff
@@ -188,3 +202,19 @@ describe('getEventsFromAlgolia', () => {
...
+    expect(getFilters([])).toEqual(undefined);
+  });
+
+  it('OR between elements represented as a string', async () => {
+    expect(getFilters(['filter1', 'filter2'])).toEqual('filter1 OR filter2');
+  });
+
+  it('OR between element and "AND" at the end when userId is present', async () => {
+    expect(getFilters(['filter1', 'filter2'], 'userId')).toEqual(
+      'filter1 OR filter2 AND speakers.user.id: userId',
```
niezgoda, line 217: filters should be in ()

niezgoda: because now it is like filter1 or (filter2 and speakers.user.id: userId)

apps/crn-frontend/src/events/api.ts
```diff
@@ -1,14 +1,29 @@
...
-import { EventResponse, ListEventResponse } from '@asap-hub/model';
 import { createSentryHeaders } from '@asap-hub/frontend-utils';
-
+import { EventResponse, ListEventResponse } from '@asap-hub/model';
 import { API_BASE_URL } from '../config';
-import { GetEventListOptions } from './options';
 import createListApiUrl from '../CreateListApiUrl';
+import { GetEventListOptions } from './options';
+
+export function getFilters(filters: string[], userId?: string) {
```
kmaid, line 8: I am not sure i like this. I dont think we should be constructing queries that aren't tokenized incase someone puts a ) or " or something. Shouldn't this be in the algolia library @peterstarling ?

apps/crn-frontend/src/events/api.ts
```diff
@@ -1,14 +1,29 @@
...
+  if (filters.length === 0) return undefined;
```
kmaid, line 9: This doesn't seem to cover if just a user id is specified which seems odd.

apps/crn-frontend/src/network/users/UserProfile.tsx
```diff
@@ -64,6 +59,23 @@ const User: FC<Record<string, never>> = () => {
...
+  const [currentTime] = useState(startOfMinute(new Date()));
+
+  const upcomingEventsResult = useEvents(
+    getEventListOptions(
+      currentTime,
+      false,
+      {
+        searchQuery: '',
+        currentPage: 0,
+        pageSize: 1,
```
kmaid, line 71: Is it not possible to have parameters match what we do for upcoming events without a search query? It would save us an algolia query in the most common case.

apps/crn-frontend/src/events/api.ts
markcsims: This has moved the the algolia package, and now should filter by userId

apps/crn-frontend/src/events/api.ts
markcsims: I can't find anything in the Algolia library to suggest this: https://www.algolia.com/doc/api-reference/api-parameters/filters/

kmaid (approved):
I really think we shouldn't be putting more logic that belongs into the algolia package into the FE but ive approved

PR #1696, opened 2022-06-14
Renamed access instructions to usage notes on shared output page

oratiu:


niezgoda:
During refinement we were talking about migrating this fields in squidex

markcsims (approved):

PR #1699, opened 2022-06-14
CRN-926 Remove Presentation & Grant Doc from FE edit menu

niezgoda:


packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -68,11 +68,12 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
   ];
 
   const { canCreateUpdate } = useContext(ResearchOutputPermissionsContext);
+
   return (
     <div css={containerStyles}>
       <div css={buttonsContainer}>
         <BackLink href={backHref} />
-        {canCreateUpdate && (
+        {canCreateUpdate && !isGrantDocument && (
```
markcsims, line 76: isGrantDocument doesn't seem the right name. 

How about isEditable?

markcsims (changes requested):

markcsims (approved):

PR #1713, opened 2022-06-20
CRN-948 Identifier type server error

niezgoda:
Remove BE restriction on adding ‘None’ as an identifier type when ASAP Funded and Used in a publication is set to ‘TRUE’


peterstarling (changes requested):
please add tests - this should have been tested somewhere already, surely these tests would now break?

peterstarling (approved):

PR #1721, opened 2022-06-22
CRN-887 groups events page no upcoming no past events

niezgoda:


packages/react-components/src/organisms/GroupNoEvents.tsx
```diff
@@ -0,0 +1,38 @@
...
+      try exploring other {past ? ' past ' : ' upcoming '} events on the Hub.
+    </span>
+    <br />
+    <a href={link} css={exploreEventsStyles}>
+      Explore {past ? ' Past ' : ' Upcoming '} Events
+    </a>
+  </main>
+);
+
+export default GroupNoEvents;
```
evemontalvao, line 38: What do you think about renaming this to NoEvents only and then receive the message as a parameter? Cause it looks a lot similar to the component I would have to create for the team's page so I could just re-use this one.
If you agree I can make this change myself after you merge it

apps/crn-frontend/src/events/options.ts
```diff
@@ -17,13 +17,15 @@ export type GetEventListOptions = GetListOptions &
...
     userId?: string;
+    groupId?: string;
   };
 
 export const getEventListOptions = (
   currentTime: Date,
   past: boolean,
   { searchQuery = '', currentPage = 0, pageSize = CARD_VIEW_PAGE_SIZE } = {},
   userId?: string,
+  groupId?: string,
```
kmaid, line 28: I am not sure about this interface as we add more options it will become less usable. Should this be an object? Potentially extending the search query one?

packages/react-components/src/icons/cross-in-circle.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const crossInCircle = (
+  <svg
+    width="46"
+    height="46"
+    viewBox="0 0 46 46"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <circle cx="23" cy="23" r="22.5" stroke="#4D646B" />
```
kmaid, line 11: This should be in the icons story and also

packages/react-components/src/organisms/GroupNoEvents.tsx
```diff
@@ -0,0 +1,38 @@
+import { css } from '@emotion/react';
+import { getButtonStyles } from '../button';
+import { crossInCircleIcon } from '../icons';
+import { perRem } from '../pixels';
+
+const titleStyles = css({
+  padding: `20px`,
```
kmaid, line 7: perRem

packages/react-components/src/templates/GroupProfileHeader.tsx
```diff
@@ -95,12 +99,12 @@ const GroupProfileHeader: React.FC<GroupProfileHeaderProps> = ({
             <TabLink
               href={route.upcoming({}).$ + queryParamString(searchQuery)}
             >
-              Upcoming Events
+              Upcoming Events{` (${upcomingEventsCount})`}
```
kmaid, line 102: Whats going on here? cant this just be {upcomingEventsCount}

kmaid (changes requested):
Ive done a partial review. I need to have another look in the morning

packages/react-components/src/templates/__tests__/GroupProfilePage.test.tsx
```diff
@@ -9,6 +9,8 @@ const props: ComponentProps<typeof GroupProfilePage> = {
   groupTeamsHref: '#teams',
   lastModifiedDate: '2021-01-01',
   numberOfTeams: 1,
+  pastEventsCount: 0,
+  upcomingEventsCount: 0,
```
kmaid, line 13: I kind of feel by injecting the no events component we should be testing the integration of both components in this file. If it was deleted/omitted I don't think any tests would fail

packages/react-components/src/templates/GroupProfileHeader.tsx
niezgoda: this is due to the space that needs to be displayed

kmaid (approved):

PR #1724, opened 2022-06-23
stop error when user is non onboarded

markcsims:


apps/crn-frontend/src/network/users/__tests__/UserProfile.test.tsx
```diff
@@ -423,16 +426,33 @@ it('renders number of shared outputs', async () => {
...
 it('renders number of upcoming events', async () => {
   const response = createListEventResponse(7);
-  mockUserEvents.mockResolvedValue(response);
+  mockUserEventsFromAlgolia.mockResolvedValue(response);
   await renderUserProfile(createUserResponse());
 
   expect(await screen.findByText(/Upcoming Events \(7\)/i)).toBeVisible();
 });
 
+it('onboarded users do not call the events api', async () => {
```
peterstarling, line 435: I think you meant "do call"

peterstarling (changes requested):

PR #1728, opened 2022-06-27
CRN-901

niezgoda:


apps/crn-server/src/migrations/16563279171200-move-access-instructions-to-usage-notes.ts
```diff
@@ -0,0 +1,22 @@
...
+          usageNotes: {
+            iv: researchOutput.data.accessInstructions?.iv || '',
+          },
+        });
+      },
+    );
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  down = async (): Promise<void> => {};
```
peterstarling, line 21: i would add a `down` method implementation to move the data back in case in screws up some pipelines for us or something

peterstarling (changes requested):

apps/crn-server/src/migrations/16563279171200-move-access-instructions-to-usage-notes.ts
niezgoda: but the data is not being removed

peterstarling: yeah i guess you're right, i thought maybe we should remove it from the new field in the 'down' method but i think having that extra data on the document doesnt actually do anything

peterstarling (approved):

PR #1729, opened 2022-06-27
feat: CRN-935 Split group controller

peterstarling:


apps/crn-server/src/controllers/groups.ts
```diff
@@ -19,117 +12,79 @@ export interface GroupController {
...
-            `(${[
-              [`contains(data/name/iv, '${word}')`],
-              [`contains(data/description/iv, '${word}')`],
-              [`contains(data/tags/iv, '${word}')`],
-            ].join(' or ')})`,
-          ),
-        [],
-      )
-      .join(' and ');
+    const items = total > 0 ? users : [];
```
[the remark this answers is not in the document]
peterstarling: i guess you're right, i've removed these from this controller

apps/crn-server/scripts/export-entity.ts
```diff
@@ -15,8 +15,8 @@ import Events from '../src/controllers/events';
 import ExternalAuthors from '../src/controllers/external-authors';
 import ResearchOutputs from '../src/controllers/research-outputs';
 import Users from '../src/controllers/users';
-import AssetDataProvider from '../src/data-providers/assets.data-provider';
-import UserDataProvider from '../src/data-providers/users.data-provider';
+import { AssetSquidexDataProvider } from '../src/data-providers/assets.data-provider';
+import { UserSquidexDataProvider } from '../src/data-providers/users.data-provider';
```
markcsims, line 19: I can understand this change, but this is an interface. I would have thought the interface would be generic whilst the implementation would be Squidex specific.

apps/crn-server/src/app.ts
```diff
@@ -43,8 +43,18 @@ import ResearchTags, {
...
+  AssetDataProvider,
+  AssetSquidexDataProvider,
+} from './data-providers/assets.data-provider';
+import {
+  GroupDataProvider,
+  GroupSquidexDataProvider,
+} from './data-providers/groups.data-provider';
+import {
+  UserDataProvider,
+  UserSquidexDataProvider,
```
markcsims, line 56: Why is there a SquidexDataProvider and a general data provider interface?

apps/crn-server/src/routes/user.route.ts
```diff
@@ -87,18 +87,11 @@ export const userRouteFactory = (
   userRoutes.get<{ userId: string }>(
     '/users/:userId/groups',
     async (req, res) => {
-      const { query, params } = req;
+      const { params } = req;
```
markcsims, line 90: we used query before, don't we need this to maintain current functionality?

apps/crn-server/scripts/export-entity.ts
peterstarling: it's not though, this is the implentation, the interface remains generic:
https://github.com/yldio/asap-hub/blob/a5636133d05264002fa6b080cfedd43d727b9fdd/apps/crn-server/src/data-providers/assets.data-provider.ts#L6

apps/crn-server/src/app.ts
peterstarling: pretty sure that's exactly what you said you could understand in the comment bellow: `I would have thought the interface would be generic whilst the implementation would be Squidex specific.`?

apps/crn-server/src/routes/user.route.ts
peterstarling: actually it turns out the pagination was broken on this so I decided to remove it from this endpoint entirely, which meant the query string was no longer needed. 

In FE the endpoint is only ever used on the user profile page without pagination anyway. We could revisit once there is a need to paginate.

markcsims (approved):

PR #1737, opened 2022-06-27
CRN 933 add sort by oldest event date for past events in calendar events page VIII

markcsims:
- get algolia settings
- set algolia schema cli
- move scripts out of algolia package
- deploys replica
- yarn fix:format
- bumps prettier
- fixes missing dependancies
- tries different index name
- remove set algolia in build
- format settings files
- fixes algolia sync
- wip
- fix hook


packages/algolia/src/client.ts
```diff
@@ -88,6 +91,7 @@ export class AlgoliaSearchClient {
     entityTypes: T[],
     query: string,
     requestOptions?: SearchOptions,
+    descending?: boolean,
```
kmaid, line 94: I think this should be a parameter like sort as you cannot have two sort methods with algolia

PR #1738, opened 2022-06-27
Fix error card icon

ricardocarvalh0:
Replaces warning icon by error icon in ErrorCard component.


kmaid (changes requested):
This icon should be included in the icon storybook file!

kmaid (approved):

PR #1742, opened 2022-06-29
Crn 933 add sort by oldest event date for past events in calendar events page viii

oratiu:


packages/algolia/test/client.test.ts
```diff
@@ -153,6 +159,22 @@ describe('Algolia Search Client', () => {
...
+  test('Should search event entity', async () => {
+    reverseAlgoliaSearchIndex.search.mockResolvedValueOnce(searchEventResponse);
+
+    const response = await algoliaSearchClient.search(
+      ['event'],
+      'query',
+      {},
+      true,
+    );
+    console.log(response.hits[0]);
```
kmaid, line 171: This should be thrown by eslint. I think this stage I want the prettier commits taken out of this branch and es lint and its fixed put into the new branch to be merged separately before this

packages/algolia/test/client.test.ts
```diff
@@ -153,6 +159,22 @@ describe('Algolia Search Client', () => {
     });
   });
 
+  test('Should search event entity', async () => {
```
kmaid, line 162: I think this test title should probally say if its supposed to be in the past or future

packages/algolia/src/client.ts
```diff
@@ -88,6 +91,7 @@ export class AlgoliaSearchClient {
     entityTypes: T[],
     query: string,
     requestOptions?: SearchOptions,
+    descending?: boolean,
```
peterstarling, line 94: what if we need another replica for another different sort type? Are these replicas there just so we can search in reverse order? I think other combinations might be required later, maybe we need some sort of map here for sort order? also technically at the moment this replica should only ever be used for events - shouldnt our type or implementation reflect that?

peterstarling (changes requested):

schema/algolia-reverse-timestamp-schema.json
```diff
@@ -0,0 +1,52 @@
+{
```
niezgoda, line 1: maybe it is worth to put attributesForFaceting and attributesToIndex into one place?

PR #1752, opened 2022-07-04
Crn 933 add sort by oldest event date for past events in calendar events page v2

oratiu:


apps/asap-cli/src/scripts/algolia/delete-index.ts
```diff
@@ -0,0 +1,31 @@
...
+  indexName,
+}: DeleteAlgoliaIndex): Promise<void> => {
+  const client = algoliasearch(algoliaAppId, algoliaCiApiKey);
+  const unlinkIndex = async (name: string) => {
+    const index = client.initIndex(name);
+    return index.setSettings({ replicas: [] }).wait();
+  };
+  const indices = await client.listIndices();
+  indices.items
+    .filter(({ name }) => name.startsWith(indexName))
```
oratiu, line 22: Piotr: isn't this a little dangerous? we are removing all indexes that start with a certain name. I guess the name in our case it usually the name of the branch but I would think we should be a little more specific? I guess it's because the replica and the primary index are both named with the same prefix?

apps/asap-cli/src/scripts/algolia/delete-index.ts
```diff
@@ -0,0 +1,31 @@
...
+    .forEach(async ({ name }) => {
```
oratiu, line 23: Piotr: Having the comment above in mind I think we should simply get the primary index in question, get all of its replicas, the unlike them all and remove each one?

packages/algolia/src/client.ts
```diff
@@ -88,6 +91,7 @@ export class AlgoliaSearchClient {
     entityTypes: T[],
     query: string,
     requestOptions?: SearchOptions,
+    descending?: boolean,
```
oratiu, line 94: Piotr: what if we need another replica for another different sort type? Are these replicas there just so we can search in reverse order? I think other combinations might be required later, maybe we need some sort of map here for sort order? also technically at the moment this replica should only ever be used for events - shouldnt our type or implementation reflect that?

schema/algolia-reverse-timestamp-schema.json
```diff
@@ -0,0 +1,52 @@
+{
```
oratiu, line 1: Mateusz: maybe it is worth to put attributesForFaceting and attributesToIndex into one place?

oratiu: @niezgoda not sure I follow here.

packages/algolia/src/client.ts
oratiu: @peterstarling Yeah the replicas are the only way to sort, if we do need other combinations later, from what I gather that would mean having multiple replica indexes and we will have to refactor the search function a bit I think.

peterstarling: i think what i was trying to say here was that the interface seems a little poor. On one hand I get it - for now all we need is a parameter to tell us if its descending or not but in reality it should be a little more sophisticated, type-wise. For instance we should not be able to select just any entity type here, just the events, as this is what has been currently sorted in a descending order, correct?

peterstarling: i think we said we should abstract away the interface for search anyway here so I would suggest creating a ticket to do that where we also make the order parameter in a less concrete way

kmaid (approved):

apps/asap-cli/src/scripts/algolia/delete-index.ts
peterstarling: i guess its fine, we actually had a similar thing before
```sh
yarn workspace @asap-hub/crn-server run algolia deleteindicespattern -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -r "^$ALGOLIA_INDEX$" -x false
```

peterstarling (approved):

PR #1763, opened 2022-07-08
Added past events tab on user profile page

oratiu:


oratiu:
Codecov complains about a file I didn't change

apps/crn-frontend/src/network/users/UserProfile.tsx
```diff
@@ -78,6 +78,18 @@ const UserProfile: FC<UserProfileProps> = ({ currentTime }) => {
     currentUser,
   );
 
+  const pastEventsResult = useEvents(
```
jhonny-m, line 81: same as the team profile, can we have `useEvents` return all the events? is it bad for performance?

jhonny-m (approved):

PR #1777, opened 2022-07-14
[CRN-994] Update tags list

evemontalvao:


oratiu (approved):
Looks good, checked "Tags for YLD" xls

PR #1778, opened 2022-07-15
fix CRN-986 Do not fetch research-tag without entity

peterstarling:
* Exports option types for generating typed query strings in FE.
* Adds qs to enable easy query string construction (especially for nested queries).


oratiu (approved):
Looks good

PR #1785, opened 2022-07-19
user card component

jhonny-m:


oratiu (approved):

packages/gp2-components/src/molecules/IconWithLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const { rem } = pixels;
+
+type IconWithLabelProps = {
+  readonly icon: JSX.Element;
+};
+const styles = css({
+  display: 'flex',
+  gap: rem(9),
+  margin: `${rem(8)} 0`,
```
ricardocarvalh0, line 12: I've seen some values has to be multiple of 3, is it one of the cases? For which css rules is it applied?

packages/gp2-components/src/molecules/IconWithLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+});
+const iconStyles = css({
+  display: 'inline-flex',
+});
+
+const IconWithLabel: React.FC<IconWithLabelProps> = ({ icon, children }) => (
```
ricardocarvalh0, line 18: Not sure if it's important to formattiing but `const`s are separated by an empty line here but not in the above.

packages/gp2-components/src/molecules/UserCardInfo.tsx
```diff
@@ -0,0 +1,105 @@
...
+});
+
+const workingGroupsStyles = css({
+  flexWrap: 'wrap',
+  [crossQuery]: {
+    gap: '0.25rem',
+    flexDirection: 'row',
+  },
+});
+const workingGroupsLinkStyles = css({
```
ricardocarvalh0, line 49: Same here for empty lines between `const`s

packages/gp2-components/src/molecules/UserCardInfo.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  <span css={workingGroupsLinkStyles}>{name}</span>
+                </Link>
+              </>
+            ))
+          ) : (
+            <span css={subduedText}>
+              This member isn’t part of any working groups
+            </span>
+          )}
+        </div>
```
ricardocarvalh0, line 86: Maybe extract this `div` to a `<UserWorkingGroups />` ?

packages/gp2-components/src/organism/UserCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+  'region' | 'projects' | 'role' | 'workingGroups'
+> & {
+  id: string;
+  displayName: string;
+  firstName: string;
+  lastName: string;
+  avatarUrl?: string;
+  degree: string[];
+  tags: string[];
+};
```
ricardocarvalh0, line 58: types usually declared before styles?

packages/gp2-components/src/organism/UserCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+  region,
+  workingGroups,
+  projects,
+  tags,
+}) => {
+  const userHref = network({}).users({}).user({ userId: id }).$;
+
+  return (
+    <Card>
+      <div css={[containerStyles]}>
```
ricardocarvalh0, line 76: Why does it has to be an array with one element here?

packages/gp2-components/src/organism/UserCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+          />
+        </Anchor>
+
+        <div css={textContainerStyles}>
+          <div>
+            <Anchor href={userHref}>
+              <h2 css={titleStyles}>
+                {displayName}
+                {degree && !!degree.length && `, ${degree.join(', ')}`}
+              </h2>
```
ricardocarvalh0, line 92: Shouldn't this be a `<Headline2>`?

packages/gp2-components/src/molecules/UserCardInfo.tsx
```diff
@@ -0,0 +1,105 @@
...
+const UserCardInfo: React.FC<UserCardInfoProps> = ({
+  role,
+  region,
+  workingGroups,
+  projects,
+}) => (
+  <div css={containerStyles}>
+    <div css={rowStyles}>
+      <IconWithLabel icon={roleIcon}>{role}</IconWithLabel>
+      <UserRegion region={region}></UserRegion>
```
ricardocarvalh0, line 67: `<UserRegion region={region} />`?

ricardocarvalh0 (approved):
Approved as most of the comments are more like questions to refine my review skills

packages/gp2-components/src/icons/project-icon.tsx
```diff
@@ -12,7 +12,7 @@ const projectIcon = (
...
       fill="#4D646B"
       stroke="#4D646B"
-      stroke-width="0.1"
+      strokeWidth="0.1"
```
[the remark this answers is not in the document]
jhonny-m: I was just fixing an issue with the property name, but this is how I've got the sgv

packages/gp2-components/src/molecules/IconWithLabel.tsx
jhonny-m: I don't think we have a rule for it

packages/gp2-components/src/organism/UserCard.tsx
jhonny-m: Just did a double take and I actually need an `h3` the `Headline` component has some differences both in size and margins with what it's on the designs for gp2, we might need to discuss this later

packages/gp2-components/src/icons/project-icon.tsx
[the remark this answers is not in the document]
kmaid: I always stick my SVGs through https://react-svgr.com/playground/ to convert property names and minify etc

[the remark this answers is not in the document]
kmaid: If the SVG needs to change colour the stroke or fill prop needs to be in the SVG tag so we can select it in CSS. I wouldn't bother messing with it by default though

packages/gp2-components/src/organism/UserCard.tsx
kmaid: Maybe we should be using themeing to vary the padding etc between the projects? I nearly put in a noPadding flag in CRN because of the designers changing the rules 🙄

jhonny-m: maybe we could bring this up in rtm?

PR #1788, opened 2022-07-21
[CRN-996] Remove all references to EVENTS_SEARCH feature flag

ricardocarvalh0:


apps/crn-frontend/src/events/__tests__/EventList.test.tsx
```diff
@@ -73,71 +58,7 @@ const renderEventsListPage = async (
...
-          sortBy: 'endDate',
-          sortOrder: 'desc',
-        },
-      }),
-      expect.anything(),
-    );
-  });
-});
-
 describe('Algolia', () => {
```
kmaid, line 61: This can also be removed as we only get it from algolia

apps/crn-frontend/src/events/__tests__/Events.test.tsx
```diff
@@ -91,25 +68,6 @@ describe('Events', () => {
     );
   });
 
-  it('Defaults to the upcoming events page', async () => {
```
kmaid, line 94: Why was this test removed. It doesn't seem like a legacy test at first glance

apps/crn-frontend/src/events/__tests__/Events.test.tsx
```diff
@@ -133,62 +91,7 @@ describe('Events', () => {
...
-            expect.objectContaining({ searchQuery: 'searchterm' }),
-            expect.anything(),
-          ),
-        );
-      });
-    },
-  );
-});
-
 describe('Algolia', () => {
```
kmaid, line 94: This can be removed

kmaid:
Do we still need the api method getEvents? Can this and the test be removed? Can we rename getEventsFromAlgolia to just getEvents as its the new way to do it?

apps/crn-frontend/src/events/__tests__/Events.test.tsx
ricardocarvalh0: It is legacy, the `getEvents` was used in `apps/crn-frontend/src/events/state.ts` when the flag was disabled.

kmaid (approved):

PR #1820, opened 2022-07-25
GP2 User Controller

markcsims:
Adds tests to:

- user controller
- user data provider


peterstarling (approved):
to be fair this is just temp, throw away code anyway, for as long as we don't know what the user model will look like. We could try to make this reusable and move it to server-common but it just might be too soon

PR #1834, opened 2022-07-27
feat: CRN-950 Reminders: Shared Research

peterstarling:


packages/model/src/reminder.ts
```diff
@@ -0,0 +1,48 @@
...
+  type: 'ResearchOutputShare';
+}
+
+export type ReminderResponse =
+  | ResearchOutputPublishedReminder
+  | ResearchOutputShareReminder;
+
+export type ListReminderResponse = ListResponse<ReminderResponse>;
+
+export const getReminderCard = (reminder: ReminderResponse) => {
```
kmaid, line 38: Not sure about including code in the model? Shouldn't we just have a type discriminator and handle this somewhere else?

packages/model/src/reminder.ts
```diff
@@ -0,0 +1,48 @@
...
+interface Reminder {
+  area: ReminderArea;
+  type: ReminderType;
+  data: unknown;
+}
+
+export interface ResearchOutputPublishedReminder extends Reminder {
+  area: 'Research Output';
+  type: 'ResearchOutputPublished';
+  data: {
```
kmaid, line 22: Do we need to wrap this in "data"?  Maybe it could be called source?

packages/model/src/reminder.ts
```diff
@@ -0,0 +1,48 @@
...
+export type ReminderType = typeof reminderTypes[number];
+
+interface Reminder {
+  area: ReminderArea;
+  type: ReminderType;
+  data: unknown;
+}
+
+export interface ResearchOutputPublishedReminder extends Reminder {
+  area: 'Research Output';
```
kmaid, line 20: Maybe we could use Entity?

packages/model/src/reminder.ts
```diff
@@ -0,0 +1,48 @@
...
+  type: 'ResearchOutputPublished';
+  data: {
+    researchOutput: ResearchOutputResponse;
+  };
+}
+
+export interface ResearchOutputShareReminder extends Reminder {
+  area: 'Research Output';
+  type: 'ResearchOutputShare';
```
kmaid, line 29: Why are we including the "area" in the type? Surely we can just use both to discriminate?

packages/model/src/reminder.ts
peterstarling: sure, i was using whatever they used for designs, but sure

packages/model/src/reminder.ts
peterstarling: im not sure, i dont think having all the extra props flat out is a great idea, feels cleaner this way

packages/model/src/reminder.ts
peterstarling: this is just an example code to show how to recognise which reminder we are dealing with. This way once you recognise the type you can access the data easily

packages/model/src/reminder.ts
kmaid: I am hoping we can just include the source entity flat against the object and maybe a JQL query when required to identify exactly what was changed if required

packages/model/src/reminder.ts
```diff
@@ -0,0 +1,72 @@
...
+  type: ReminderType;
+  data: unknown;
+}
+
+interface ResearchOutputReminder extends Reminder {
+  entity: 'Research Output';
+  type: ResearchOutputReminderType;
+}
+
+interface EventReminder extends Reminder {
```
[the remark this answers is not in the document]
markcsims: depends who you ask. If you know you are going to use it, I'd keep it there.

markcsims (approved):

PR #1858, opened 2022-07-28
fix: Data Provider integration tests

peterstarling:
* adds the `role` required property to the user PATCH request payload
* modify's the integration tests to use data providers public interface only

### DIFF
--- a/apps/crn-server/src/data-providers/users.data-provider.ts
+++ b/apps/crn-server/src/data-providers/users.data-provider.ts
@@ -6,6 +6,7 @@ import {
   OrcidWorkType,
   orcidWorkType,
   Role,
+  UserCreateDataObject,
   UserDataObject,
   UserDegree,
   userDegree,
@@ -15,6 +16,7 @@ import {
   UserUpdateDataObject,
 } from '@asap-hub/model';
 import {
+  InputUser,
   parseDate,
   RestUser,
   sanitiseForSquidex,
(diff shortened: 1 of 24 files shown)


apps/crn-server/src/data-providers/users.data-provider.ts
```diff
@@ -59,6 +56,55 @@ export class UserSquidexDataProvider implements UserDataProvider {
...
+        iv: input.degree || null,
+      },
+      questions: {
+        iv: input.questions?.map((question) => ({ question })) || null,
+      },
+      social: {
+        iv: (input.social && [{ ...input.social }]) || [],
+      },
+    };
+    const { id } = await this.userSquidexRestClient.create(inputUser);
```
markcsims, line 103: Should this return only the id or the newly created user?

I might be wrong but in other places we've returned the newly created entity.

markcsims (approved):

apps/crn-server/src/data-providers/users.data-provider.ts
peterstarling: Pretty sure we return just the string everywhere else?
https://github.com/yldio/asap-hub/blob/b424a7df38b188a6b189bd7a5cb2c6dd16a42f70/apps/crn-server/src/data-providers/assets.data-provider.ts#L7
https://github.com/yldio/asap-hub/blob/b424a7df38b188a6b189bd7a5cb2c6dd16a42f70/apps/crn-server/src/data-providers/external-authors.data-provider.ts#L5
https://github.com/yldio/asap-hub/blob/b424a7df38b188a6b189bd7a5cb2c6dd16a42f70/apps/crn-server/src/data-providers/research-outputs.data-provider.ts#L38

PR #1868, opened 2022-08-02
CRN-969 - Initial work to add user preferences

m-doughty:
See [CRN-969](https://asaphub.atlassian.net/browse/CRN-969) for details.


m-doughty:
Unneeded.

PR #1882, opened 2022-08-04
Initial commit of About Asap page

kmaid:


evemontalvao (approved):

packages/model/src/discover.ts
```diff
@@ -5,6 +5,7 @@ import { NewsResponse } from './news';
 export interface DiscoverResponse {
   aboutUs: string;
   members: ReadonlyArray<UserResponse>;
+  membersHref?: string;
```
peterstarling, line 8: I would propose to return an ID of the ASAP team instead
```suggestion
  asapTeamId?: string;
```

peterstarling (changes requested):

packages/react-components/src/templates/DiscoverAbout.tsx
```diff
@@ -18,18 +18,18 @@ const aboutUsStyles = css({
...
   scientificAdvisoryBoard,
-  membersHref,
+  membersTeamId,
   aboutUs,
 }) => {
-  const membersHrefData = membersHref
+  const membersTeamIdData = membersTeamId
     ? {
-        href: membersHref,
+        href: membersTeamId,
```
kmaid, line 32: This wont work. It needs to use the routing library

PR #1885, opened 2022-08-04
[CRN-981] Title attribute hover

ricardocarvalh0:


apps/storybook/package.json
```diff
@@ -6,7 +6,7 @@
     "node": "16.x"
   },
   "scripts": {
-    "start": "start-storybook --quiet -c config -p 3001",
+    "start": "start-storybook --quiet --no-open -c config -p 3001",
```
kmaid, line 9: ?

kmaid: You should add STORYBOOK_NO_OPEN=true to your .env file

packages/react-components/src/atoms/__tests__/Link.test.tsx
```diff
@@ -199,3 +199,56 @@ describe('when button-styled', () => {
     expect(getByText('text').closest('a')).not.toHaveAttribute('href');
   });
 });
+
+describe('for elipsed links', () => {
+  // !! Unfortunately jest does not respect ellipsed texts (overflow: hidden)
+  // https://github.com/testing-library/react-testing-library/issues/751#issuecomment-691069646
+  // so we cannot test if the text is shown breaked as it should
+  // but we can test other properties and see it working in the StoryBook
```
kmaid, line 207: We do have playwright to test things like this

packages/react-components/src/atoms/__tests__/Link.test.tsx
```diff
@@ -199,3 +199,56 @@ describe('when button-styled', () => {
...
+        'A very long text showing that the elipsis feature is working well, trying to provide as much text here as I can to validate this feature.';
+      const { getByTitle } = render(
+        <div css={{ maxWidth: '10px' }}>
+          <Link href="/" ellipsed>
+            {longText}
+          </Link>
+        </div>,
+      );
+
+      expect(getByTitle(longText)).toBeInTheDocument();
```
kmaid, line 221: This is already tested. You should be using .closest('a').toHaveAttribute('title', 'expected string')

apps/storybook/package.json
ricardocarvalh0: Sorry for that, I forgot to remove before commit.

packages/react-components/src/atoms/__tests__/Link.test.tsx
ricardocarvalh0: For now we don't have any page using this Link component. I guess the components using it are going to be implemented in the future and then we'll be able to introduce the `e2e` tests. But anyways I can remove these comments.

ricardocarvalh0: I've talked with João and he has provided me a page where we can use it for now. Will do it.

packages/react-components/src/atoms/Link.tsx
```diff
@@ -47,17 +48,23 @@ const iconThemeStyles: Record<ThemeVariant, SerializedStyles> = {
   }),
 };
 
+const elipsedTextStyle = css({
```
kmaid, line 51: Could these styles be imported from the Ellipsis component?

kmaid (approved):

PR #1888, opened 2022-08-05
CRN-583 | Add Sentry to all Handlers

m-doughty:
# Add Sentry to all Handlers

## Ticket

[CRN-583](https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/6?modal=detail&selectedIssue=CRN-583)

## Changes

- Created `sentryWrapper` module with a function to easily wrap AWS Lambda handlers using Sentry
- Applied this wrapper to all handlers


apps/crn-server/serverless.ts
```diff
@@ -240,7 +240,8 @@ const serverlessConfig: AWS = {
       },
     },
     auth0ConnectByCode: {
-      handler: './src/handlers/webhooks/webhook-connect-by-code.handler',
+      handler:
+        './src/handlers/webhooks/webhook-connect-by-code.sentryEnabledHandler',
```
peterstarling, line 244: any chance you could make this more consistent ie make sure that we either always export a `sentryEnabledHandler` or just `handler`? right now some of them are called `handlers` even though they are wrapped in sentry wrappers

apps/crn-server/src/handlers/user/invite-handler.ts
```diff
@@ -1,37 +1,22 @@
...
-  dsn: sentryDsn,
-  tracesSampleRate: 1.0,
-  environment,
-  release: currentRevision,
-});
-
 const userRestClient = new SquidexRest<RestUser>(getAuthToken, 'users', {
   appName,
   baseUrl,
 });
```
peterstarling, line 19: /* istanbul ignore next */

peterstarling (approved):

apps/crn-server/test/handlers/webhooks/webhook-connect-by-code.test.ts
```diff
@@ -45,7 +45,7 @@ const user: RestUser = {
 
 describe('POST /webhook/users/connections - validations', () => {
   test('returns 400 when code is not defined', async () => {
-    const res = (await handler(
+    const res = (await unloggedHandler(
```
kmaid, line 48: Why are we using an unlogged handler? If we just don't provide a DSN it wont report. This is how we deal with it in the FE

m-doughty: Unlogged handler isn't used, @peterstarling suggested leaving the 'old' handlers in place so that the tests still run and check the underlying functionality.

kmaid: So test one handler and use a different one in production?

PR #1890, opened 2022-08-05
CRN-970 Get started with asap Modal

kmaid:


oratiu (approved):

apps/storybook/src/Accordion.stories.tsx
```diff
@@ -38,3 +39,38 @@ export const Normal = () => (
...
+          'All you need to know about the Network, the Hub, sharing, meetings, communications, publishing and more.',
+        hrefText: 'Open the packet',
+        href: 'http://google.com',
+      },
+      {
+        icon: confidentialIcon,
+        title: 'Confidentiality Rules',
+        description: 'View all confidentiality rules related to the Hub.',
+        hrefText: 'Read more',
+        href: 'http://google.com',
```
ricardocarvalh0, line 67: are we merging it with these placeholder links?

packages/react-components/src/atoms/__tests__/Button.test.tsx
```diff
@@ -125,12 +125,6 @@ it('renders a stretched small button', () => {
   expect(getComputedStyle(screen.getByRole('button')).flexGrow).toBe('1');
 });
 
-it('renders a non-stretched small button', () => {
-  render(<Button small stretch={false} />);
-
-  expect(getComputedStyle(screen.getByRole('button')).flexGrow).toBe('');
-});
-
```
ricardocarvalh0, line 133: Not sure if I understand it correct but I guess we'll need tests for the new `noMargin` prop?

packages/react-components/src/icons/info-info.tsx
```diff
@@ -0,0 +1,20 @@
...
+    height="24"
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Info</title>
+    <path
+      d="M12 0C5.3728 0 0 5.3728 0 12C0 18.6272 5.3728 24 12 24C18.6272 24 24 18.6272 24 12C24 5.3728 18.6272 0 12 0ZM13.5 19.5H10.5V10.5H13.5V19.5ZM12 8.1C11.4696 8.1 10.9609 7.88929 10.5858 7.51421C10.2107 7.13914 10 6.63043 10 6.1C10 5.56957 10.2107 5.06086 10.5858 4.68579C10.9609 4.31071 11.4696 4.1 12 4.1C12.5304 4.1 13.0391 4.31071 13.4142 4.68579C13.7893 5.06086 14 5.56957 14 6.1C14 6.63043 13.7893 7.13914 13.4142 7.51421C13.0391 7.88929 12.5304 8.1 12 8.1Z"
+      fill="#0C8DC3"
+      stroke="none"
```
ricardocarvalh0, line 15: Shouldn't stroke/fill be in the `<svg>` tag level?

packages/react-components/src/molecules/Accordion.tsx
```diff
@@ -79,39 +108,54 @@ type AccordionProps = {
...
-            >
-              <Paragraph accent={'lead'}>{description}</Paragraph>
-              <div css={{ width: 'fit-content' }}>
-                <Link buttonStyle small primary href={href}>
-                  {hrefText} {!isInternalLink(href)[0] && externalLinkIcon}
-                </Link>
+            <div css={[hiddenStyles, index === opened && openStyles]}>
+              <div css={bodyStyles}>
+                <Paragraph accent={'lead'}>{description}</Paragraph>
+                <div css={{ width: 'fit-content' }}>
```
ricardocarvalh0, line 138: Is it ok to use "inline" css directly in the html tag? I assume it would be fine for this case as it doesn't receive any variables + it's only one css prop.

apps/storybook/src/Accordion.stories.tsx
ricardocarvalh0: Nevermind, just realized it's for storybook

ricardocarvalh0 (approved):
Approved. Left some questions in order to improve my review skills.

packages/react-components/src/atoms/__tests__/Button.test.tsx
kmaid: This was a test for the noStrech prop. It was unnecessary because you can make it not stretch by changing the parent elements display to inline-block. There is a form of technical debt called prop creep and the Link and Button components are struggling with it so I am trying to reduce it where i can

packages/react-components/src/icons/info-info.tsx
kmaid: It should if we need to colour change. There was no requirement to do so so I have not.

packages/react-components/src/molecules/Accordion.tsx
kmaid: I think its okay to do for like one css rule but mostly we prefer to separate it where possible.

kmaid:
@ricardocarvalh0 I hope ive answered your questions and I appreciate the feedback! These were good questions to ask :)

PR #1895, opened 2022-08-08
Added hover effect to required elements by creating new component

oratiu:


packages/react-components/src/molecules/ImageLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+          ':hover': {
+            color: fern,
+            opacity: '64%',
+          },
+        },
+      }
+    : containerPropStyle;
+
+  return (
+    <a css={css({ ...wrapperStyle })} href={link}>
```
kmaid, line 34: You should be able to use Achor. It does some extxra stuff like opening external links in a new window

packages/react-components/src/organisms/NewsCard.tsx
```diff
@@ -90,21 +84,17 @@ const NewsCard: React.FC<NewsCardProps> = ({
...
-              alt={`"${title}"'s thumbnail`}
-              src={thumbnail}
-              css={[imageStyle]}
-            />
-          ) : (
-            placeholders[type]
-          )}
-        </div>
+        <ImageLink
+          containerPropStyle={imageContainerStyle}
```
kmaid, line 92: I really don't like injected styles. It means if someone changes <ImageLink> it will be affected here but it isn't easy to find usages of. It should be avoided wherever possible

packages/react-components/src/organisms/__tests__/NewsCard.test.tsx
```diff
@@ -64,12 +64,12 @@ describe('News', () => {
...
   it('links to detail page when text is present', () => {
-    const { rerender, getByRole, queryByRole } = render(
+    const { rerender, getAllByRole, queryByRole } = render(
       <NewsCard {...newsCardProps} />,
     );
     expect(queryByRole('link')).not.toBeInTheDocument();
 
     rerender(<NewsCard {...newsCardProps} text={'<h1>title</h1>'} />);
-    expect(getByRole('link').textContent).toContain(newsCardProps.title);
+    expect(getAllByRole('link')[1].textContent).toContain(newsCardProps.title);
```
kmaid, line 73: Please don't use magic numbers in tests. Inject some data into news card and then get by text or something

PR #1904, opened 2022-08-10
feat: CRN-989 Add team-id to Discover response

peterstarling:


kmaid (approved):

apps/crn-server/src/autogenerated-gql/graphql.ts
```diff
@@ -2025,22 +2048,33 @@ export type DiscoverDataDto = {
...
 
-/** The structure of the ASAP Team field of the Discover ASAP content input type. */
+/** The structure of the ASAP Team Members field of the Discover ASAP content input type. */
 export type DiscoverDataMembersInputDto = {
   iv: Maybe<Array<Scalars['String']>>;
 };
 
+/** The structure of the ASAP Team field of the Discover ASAP content type. */
+export type DiscoverDataMembersTeamDto = {
+  iv: Maybe<Array<Teams>>;
```
m-doughty, line 2070: What's `iv` mean here? Just `input value`?

m-doughty (approved):

apps/crn-server/src/autogenerated-gql/graphql.ts
kmaid: So this comes from squidex and I assume so. These types are generated from the squidex GraphQL schema. I wouldn't try and review this!

PR #1907, opened 2022-08-10
CRN-583 | Add sentry to all handlers

m-doughty:
# CRN-583 | Add Sentry to all Handlers

## Ticket

[CRN-583](https://asaphub.atlassian.net/browse/CRN-583)

## Brief Summary

- Adds a wrapper utility to wrap a handler with Sentry
- Wraps all handlers in CRN & GP2 with this handler
- Correctly sets up all Lambdas with sentry DSNs


apps/crn-server/src/handlers/webhooks/cronjob-sync-orcid.ts
```diff
@@ -56,3 +57,7 @@ export const handler = async (): Promise<lambda.Response> => {
     statusCode: 200,
   };
 };
+
+export const unloggedHandler = rawHandler;
```
kmaid, line 61: I really don't understand why we are exposing the raw handler over just the sentry wrapped one. It is whats used in production so it improves the integration tests quality. We can omit DSN when we run the tests so nothing is logged into sentry like we do for FE.

kmaid: By doing this if someone changes the sentry wrapper in a way that doesn't work for event handlers it will be caught automatically which seems important for something that is hard to test locally

m-doughty: Our handlers are inconsistent, some take just `event`, and some take `event, context`, and sometimes these shapes are custom formats or pulled in from different libraries. 

In order to make all of the tests work against the new handlers, we would need to rewrite all of the handler tests and create proper `context` fixtures for each one. I spoke to @peterstarling about doing this, and we agreed it needs to be done but it's out of scope for this ticket.

We agreed that we'd expose the old handlers so that the tests can continue to run and pass code coverage, knowing that we're testing everything except the wrapper, and build new tests for the wrapper to ensure that it calls the correct functions (we don't want to test the underlying `Sentry.AWSLambda` module.

I agree with you that we should get rid of these but only after we've:

- Created a proper `context` fixture;
- Properly rewritten all of our handlers to take AWS `event` and `context`;
- Rewritten all of the tests to provide `event` and `context`.

When we do that, we can mock the calls to `Sentry` to do nothing in the tests, or provide no DSN.

peterstarling (approved):
to me this looks good, but:
1. Although we have tested sentryWrapper on its own we do not test the integration with the actual handler anyway, or do we?
2. I realise it was my recommendation to not test the wrapped handlers - and in fact what is important to note here is that not only are we missing that sentry integration here we are also testing handlers with injected dependencies (we only test with mocks). Perhaps we should rethink how we test these handlers in general?
@kmaid

packages/server-common/src/utils/sentry-wrapper.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Handler } from 'aws-lambda';
+
+export interface SentryConfig {
+  currentRevision: string | undefined;
+  environment: string | undefined;
+  sentryDsn: string | undefined;
+  sentryTraceSampleRate: number | undefined;
+}
+
+export const sentryWrapperImpl = (
```
peterstarling, line 11: i would suggest that you make this a factory:
```suggestion
export const sentryWrapperFactory = (config: SentryConfig) => (handler: Handler) => .....
```

and then in the crn/gp2 utils just do
```typescript
export const sentryWrapper = sentryWrapperFactory({
    currentRevision,
    environment,
    sentryDsn,
    sentryTraceSampleRate,
  });
```

peterstarling (changes requested):

peterstarling (approved):

peterstarling (approved):

markcsims (approved):
LGTM

PR #1925, opened 2022-08-15
feat: GP2-52 Add working-groups to gp2 API

peterstarling:


apps/gp2-server/src/app.ts
```diff
@@ -80,6 +91,8 @@ export const appFactory = (libs: Libs = {}): Express => {
    * Public routes --->
    */
   const userController = libs.userController || new Users(userDataProvider);
+  const workingGroupController =
+    libs.workingGroupController || new WorkingGroup(workingGroupDataProvider);
```
markcsims, line 95: The name should be plural to be consistent with other controllers. Same for the filename, and the data provider.

apps/gp2-server/src/controllers/working-group.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface WorkingGroupController {
+  fetch(): Promise<gp2.ListWorkingGroupResponse>;
+}
+
+export default class WorkingGroup implements WorkingGroupController {
+  constructor(private workingGroupDataProvider: WorkingGroupDataProvider) {}
+
+  async fetch(): Promise<gp2.ListWorkingGroupResponse> {
+    const result = await this.workingGroupDataProvider.fetch();
+    return result;
```
markcsims, line 13: `return this.workingGroupDataProvider.fetch();`?

apps/gp2-server/test/controllers/working-group.controller.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  getListWorkingGroupResponse,
+} from '../fixtures/working-group.fixtures';
+import { workingGroupDataProviderMock } from '../mocks/working-group-data-provider.mock';
+
+describe('Working Group controller', () => {
+  const workingGroupController = new WorkingGroup(workingGroupDataProviderMock);
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
```
markcsims, line 13: Don't think we need this.

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const squidexGraphqlClientMockServer = getSquidexGraphqlClientMockServer();
+  const workingGroupDataProviderMockGraphqlServer =
+    new WorkingGroupSquidexDataProvider(squidexGraphqlClientMockServer);
+
+  beforeAll(() => {
+    identity();
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
```
markcsims, line 26: We should resetAllMocks, i.e. reset the mock to it's initial state

apps/gp2-server/test/mocks/squidex-graphql-client-with-server.mock.ts
```diff
@@ -28,10 +29,12 @@ export const getSquidexGraphqlClientMockServer = (): SquidexGraphqlClient => {
     Int: () => 8,
     Instant: () => '2021-10-12T15:42:05Z',
     JsonScalar: () => {},
-    Users: () => getGraphQLUser(),
-    UsersResultDto: resultDto,
     Dashboard: () => getSquidexGraphqlDashboard(),
     DashboardResultDto: resultDto,
+    Users: () => getGraphQLUser(),
+    UsersResultDto: resultDto,
+    WorkingGroups: () => getGraphQLWorkingGroup(),
```
markcsims, line 36: inconsistent naming and casing: Here it's getGraphQLWorkingGroup, above it's getSquidexGraphqlDashboard.

QL vs ql
with Squidex in the name or not.

apps/gp2-server/test/routes/working-group.route.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import { workingGroupControllerMock } from '../mocks/working-group-controller.mock';
+
+describe('/working-groups/ route', () => {
+  const app = appFactory({
+    workingGroupController: workingGroupControllerMock,
+    authHandler: authHandlerMock,
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
markcsims, line 14: same as the resetAllMocks comment above.

markcsims (changes requested):

apps/gp2-server/src/app.ts
peterstarling: so the name of the class should be in plural to be consistent with others but the file name and data provider should not, as seen in the example above. It might be inconsistent with CRN maybe? Either way i think we should sort out the naming strategies in this repo

apps/gp2-server/test/mocks/squidex-graphql-client-with-server.mock.ts
peterstarling: well its consistent with Users  (both are getGraphQl .... ) so i guess its a problem we need to solve globally. probably not in this PR?

apps/gp2-server/test/routes/working-group.route.test.ts
peterstarling: in the one above it was used twice so i removed it - here its only applied once, and we do need it

apps/gp2-server/src/app.ts
markcsims: Agreed, let's create a ticket.

apps/gp2-server/test/mocks/squidex-graphql-client-with-server.mock.ts
markcsims: agreed

PR #1949, opened 2022-08-18
[CRN-1007] Shared Outputs amount

evemontalvao:


oratiu (approved):
Looks good

packages/react-components/src/templates/TeamProfileHeader.tsx
```diff
@@ -247,7 +249,9 @@ const TeamProfileHeader: React.FC<TeamProfileHeaderProps> = ({
         {tools && (
           <TabLink href={route.workspace({}).$}>Team Workspace</TabLink>
         )}
-        <TabLink href={route.outputs({}).$}>Team Outputs</TabLink>
+        <TabLink href={route.outputs({}).$}>
+          Shared Outputs ({sharedOutputsCount})
```
kmaid, line 253: Shared Outputs Count should default to zero as the value is optional?

evemontalvao: yes, you're right! changing it now

apps/crn-frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -74,6 +74,8 @@ const TeamProfile: FC<TeamProfileProps> = ({ currentTime }) => {
   );
 
   if (team) {
+    const sharedOutputsCount = team.outputs?.length || 0;
```
kmaid, line 77: Could we default this inside the component? It be cleaner

packages/react-components/src/templates/TeamProfileHeader.tsx
```diff
@@ -145,6 +146,7 @@ const TeamProfileHeader: React.FC<TeamProfileHeaderProps> = ({
   teamListElementId,
   labCount,
   upcomingEventsCount,
+  sharedOutputsCount,
```
kmaid, line 149: here

packages/react-components/src/templates/__tests__/TeamProfileHeader.test.tsx
```diff
@@ -17,6 +17,7 @@ const boilerplateProps: ComponentProps<typeof TeamProfileHeader> = {
   labCount: 15,
   upcomingEventsCount: 0,
   pastEventsCount: 0,
+  sharedOutputsCount: 0,
```
kmaid, line 20: If these are optional props we shouldn't include them in the default props

apps/crn-frontend/src/network/teams/TeamProfile.tsx
evemontalvao: Sure, I thought you would prefer this way to keep it like the other two counts, where there's no need to use a default value, but changing it

PR #1955, opened 2022-08-19
feat/add gp2 working group users 2

markcsims:
https://asaphub.atlassian.net/browse/GP2-55


apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+        const expected = getWorkingGroupDataObject();
+        expect(workingGroupDataObject).toEqual(expected);
+      });
+
+      test('the members in working group are parsed', () => {
+        const workingGroup = getGraphQLWorkingGroup();
+        const workingGroupWithMembers = {
+          ...workingGroup,
+          flatData: {
+            ...workingGroup.flatData,
```
peterstarling, line 96: you're trying to avoid mutations of fixtures in tests which imho seems unnecessary (we have created fixture factories specifically for this scenario) and now you are making it difficult for yourself cause you have to spread the object at different nesting levels, which also makes the fixture less readable. 

I would just mutate the members property of the `workingGroup` object

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+                  },
+                ],
+              },
+            ],
+          },
+        };
+        const workingGroupDataObject = parseWorkingGroupToDataObject(
+          workingGroupWithMembers,
+        );
+        const expected = getWorkingGroupDataObject({
```
peterstarling, line 121: you don't really need to compare the entire result, you're just interested in members here, right? well in that case you could either extract the members for the expectation or use .toMatchObject and use a partial comparison

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+            {
+              userId: '42',
+              role: 'Chair',
+              firstName: 'Tony',
+              lastName: 'Stark',
+              avatarUrl: `${baseUrl}/api/assets/${appName}/avatar-id`,
+            },
+          ],
+        });
+        expect(workingGroupDataObject).toEqual(expected);
```
peterstarling, line 175: same as above - not only is it difficult to understand what the input really is supposed to be - you are comparing the result to a large expectations where in fact you are just interested in the avatar

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+      test('invalid role', () => {
+        const workingGroup = getGraphQLWorkingGroup();
+        const workingGroupWithMembers = {
+          ...workingGroup,
+          flatData: {
+            ...workingGroup.flatData,
+            members: [
+              {
+                id: 'member-id',
+                role: 'something else',
```
peterstarling, line 187: same as above - i bet your basic fixture already has members in which case you could just mutate the .role property here

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+          },
+        };
+        expect(() =>
+          parseWorkingGroupToDataObject(workingGroupWithMembers),
+        ).toThrowErrorMatchingInlineSnapshot(
+          '"Invalid working group role on members : something else"',
+        );
+      });
+
+      test('user in member is not defined', () => {
```
peterstarling, line 212: wut? do you mean - should skip the user from the result if the user property is undefined?

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -77,5 +78,160 @@ describe('Working Group Data Provider', () => {
...
+        const workingGroup = getGraphQLWorkingGroup();
+        const workingGroupWithMembers = {
+          ...workingGroup,
+          flatData: {
+            ...workingGroup.flatData,
+            members: [
+              {
+                id: 'member-id',
+                role: 'Chair',
+                user: undefined,
```
peterstarling, line 222: same, id just mutate this property

apps/gp2-server/test/fixtures/working-group.fixtures.ts
```diff
@@ -1,12 +1,15 @@
 import type { gp2 } from '@asap-hub/model';
 import { FetchWorkingGroupsQuery } from '../../src/autogenerated-gql/graphql';
 
-export const getWorkingGroupDataObject = (): gp2.WorkingGroupDataObject => ({
-  id: 'working-group-id-1',
+export const getWorkingGroupDataObject = (
+  overrides: Partial<gp2.WorkingGroupDataObject> = {},
```
peterstarling, line 5: not a huge fan of those overrides, i'd just have a simple fixture with a single member and then i would modify the output of this factory when necessary

apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -32,15 +56,53 @@ export class WorkingGroupSquidexDataProvider
...
-          members: [],
-        }),
+        parseWorkingGroupToDataObject,
       ),
     };
   }
 }
+type Member = NonNullable<GraphQLWorkingGroup['flatData']['members']>[number];
+
+const parseWorkingGroupMembers = (
```
peterstarling, line 66: this function's signature feels a little odd. It's been written this way only to work with reduce. It would make so much more sense if it accepted an input, which should be a `Member` (it should actually be something else but I'll get to that) and and returned a single item of type `gp2.WorkingGroupMember`.

In order to do that I would keep the reduce elsewhere and have this mapper/parser used inside the reduce.

apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -32,15 +56,53 @@ export class WorkingGroupSquidexDataProvider
...
+) => {
+  if (!isWorkingGroupMemberRole(member.role)) {
+    throw new Error(`Invalid working group role on members : ${member.role}`);
+  }
+  const user = member.user && member.user[0];
+  if (!user) {
+    return membersList;
+  }
+
+  const flatAvatar = user.flatData.avatar || [];
```
peterstarling, line 78: once you do what I wrote above you can use that mapper from this point (this is where filtering stops and mapping starts).

apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -32,15 +56,53 @@ export class WorkingGroupSquidexDataProvider
...
-          shortDescription: workingGroup.flatData.shortDescription || '',
-          leadingMembers: workingGroup.flatData.leadingMembers || '',
-          members: [],
-        }),
+        parseWorkingGroupToDataObject,
       ),
     };
   }
 }
+type Member = NonNullable<GraphQLWorkingGroup['flatData']['members']>[number];
```
peterstarling, line 64: I think if you use this snippet (and another one below) you can get rid of all this type craziness at the top:
```suggestion
type Member = NonNullable<
  NonNullable<
    NonNullable<
      NonNullable<
        NonNullable<FetchWorkingGroupsQuery>['queryWorkingGroupsContentsWithTotal']
      >['items']
    >[number]['flatData']
  >['members']
>[number];
```

apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -32,15 +56,53 @@ export class WorkingGroupSquidexDataProvider
...
+      role: member.role as gp2.WorkingGroupMemberRole,
+      firstName: user.flatData.firstName || '',
+      lastName: user.flatData.lastName || '',
+      avatarUrl,
+    },
+  ];
+};
+
+export function parseWorkingGroupToDataObject(
+  workingGroup: GraphQLWorkingGroup,
```
peterstarling, line 96:
```suggestion
  workingGroup: NonNullable<
    NonNullable<
      NonNullable<FetchWorkingGroupsQuery>['queryWorkingGroupsContentsWithTotal']
    >['items']
  >[number],
```

peterstarling (changes requested):

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: Done

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: done

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: done

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: No, the fixture doesn't have a members property

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: done

apps/gp2-server/test/fixtures/working-group.fixtures.ts
markcsims: done

apps/gp2-server/src/data-providers/working-group.data-provider.ts
markcsims: done

apps/gp2-server/src/data-providers/working-group.data-provider.ts
markcsims: done

apps/gp2-server/src/data-providers/working-group.data-provider.ts
markcsims: done

apps/gp2-server/src/data-providers/working-group.data-provider.ts
markcsims: done

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: created a member fixture.

peterstarling (approved):

PR #1956, opened 2022-08-19
1 concurrent integration test

markcsims:


.github/workflows/reusable-test.yml
```diff
@@ -62,6 +62,10 @@ jobs:
 
   integration:
     if: ${{ inputs.environment-name=='Branch' }}
+    # limiting to one instance at a time as running these in parallel can cause
```
peterstarling, line 65: why do we think running them in parallel can cause issues? we run every test against a different application?

peterstarling, line 65: why do we think running them in parallel can cause issues? we run every test against a different application?

peterstarling (changes requested):

.github/workflows/reusable-test.yml
markcsims: When renovate raises 5 PRs in the morning, they all start running their deployment pipelines in parallel. What we have observed is when this occurs, 3-4 of the PRs fail with the integration test step.  

We do run every test against a completely new application, so they should be all separate. But we'd like to rule out the possibility that running in parallel has some influence on the failure occurrences.

peterstarling (approved):

PR #1960, opened 2022-08-19
feat: CRN-952 Reminders: event happening today

peterstarling:


m-doughty (approved):
I really like this approach to testing.

PR #1962, opened 2022-08-22
Working group detail

jhonny-m:


packages/frontend-utils/package.json
```diff
@@ -14,7 +14,8 @@
   },
   "peerDependencies": {
     "react": "^17.0.2",
-    "react-router-dom": "^5.2.0"
+    "react-router-dom": "^5.2.0",
```
ricardocarvalh0, line 17: is this a dependency we don't want to pin?

packages/gp2-components/src/molecules/CardWithBackground.tsx
```diff
@@ -0,0 +1,27 @@
...
+type CardWithBackgroundProps = {
+  image: string;
+};
+
+const containerStyles = (image: string) =>
+  css({
+    backgroundImage: `url(${image})`,
+    borderRadius: `8px`,
+    backgroundSize: '100%',
+    padding: '16px',
```
ricardocarvalh0, line 13: shouldn't padding be `perRem` ?

ricardocarvalh0 (approved):

packages/frontend-utils/package.json
jhonny-m: to pin? what do you mean?

ricardocarvalh0: We have this change recently where it removes `^` from versions. https://github.com/yldio/asap-hub/commit/c4e964bf771f6423f4c692088b864c9764dc8927
I wonder if it should be the same for react-router-dom

oratiu (approved):

PR #1972, opened 2022-08-25
Added new Guides tab elements

oratiu:


packages/react-components/src/icons/budget.tsx
```diff
@@ -0,0 +1,43 @@
...
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Budget Icon</title>
+    <path
+      d="M12.23 11.3889C10.892 11.0611 10.4618 10.7222 10.4618 10.1944C10.4618 9.58887 11.0571 9.16664 12.0531 9.16664C13.1023 9.16664 13.4913 9.63887 13.5266 10.3333H14.8292C14.7879 9.37775 14.1691 8.49998 12.9372 8.21664V6.99998H11.1691V8.19998C10.0256 8.43331 9.10619 9.13331 9.10619 10.2055C9.10619 11.4889 10.2319 12.1278 11.8763 12.5C13.3498 12.8333 13.6445 13.3222 13.6445 13.8389C13.6445 14.2222 13.3557 14.8333 12.0531 14.8333C10.839 14.8333 10.3616 14.3222 10.2968 13.6666H9.0001C9.07082 14.8833 10.0374 15.5666 11.1691 15.7944V17H12.9372V15.8055C14.0865 15.6 15.0001 14.9722 15.0001 13.8333C15.0001 12.2555 13.5679 11.7166 12.23 11.3889Z"
+      fill="#00202C"
+    />
+    <path
+      fill-rule="evenodd"
```
jhonny-m, line 17: we should run the svg's through svgr to clean up with like `fill-rule` into `fillRule`

packages/react-components/src/templates/DiscoverGuides.tsx
```diff
@@ -0,0 +1,106 @@
...
+  budgetIcon,
+  reportIcon,
+  shareIcon,
+  moneyIcon,
+  toolsIcon,
+} from '../icons';
+import policyIcon from '../icons/policy';
+
+const styles = css({
+  display: 'grid',
```
jhonny-m, line 19: do we need a `grid`? we are not setting up row and columns, a simple flex should do the job

packages/react-components/src/templates/DiscoverGuides.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Paragraph accent="lead">
+        Explore our guides, templates and forms to help you with your daily
+        activities.
+      </Paragraph>
+    </div>
+    <Accordion
+      items={[
+        {
+          icon: giftIcon,
+          title: 'Grant Welcome Packet',
```
jhonny-m, line 37: Not for this ticket, but this looks like something that should be in the CMS and not hard codded in the code

jhonny-m (changes requested):

packages/react-components/src/templates/DiscoverGuides.tsx
oratiu: Yeah, the initial plan is to have it static then we move it later in the CMS

jhonny-m (approved):

PR #1973, opened 2022-08-26
Fixed bug that sent a new addedDate each time a RO was edited

oratiu:


packages/react-components/src/organisms/ResearchOutputForm.tsx
```diff
@@ -400,7 +400,7 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
             documentType === 'Lab Resource' && labCatalogNumber !== ''
               ? labCatalogNumber
               : undefined,
-          addedDate: new Date().toISOString(),
+          addedDate: researchOutputData?.addedDate || new Date().toISOString(),
```
peterstarling, line 403: no tests??

peterstarling (changes requested):

packages/react-components/src/organisms/ResearchOutputForm.tsx
```diff
@@ -400,7 +404,7 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
             documentType === 'Lab Resource' && labCatalogNumber !== ''
               ? labCatalogNumber
               : undefined,
-          addedDate: new Date().toISOString(),
+          addedDate: getAddedDate(researchOutputData?.addedDate),
```
peterstarling, line 407: thats fine that you extracted it but i think it was unnecessary. It's worrying that we have no tests to assert this payload. As far as I am away right now you could just ignore the getAddedDate function and pass 'foo' to addedDate and it would still be green - thats not right. FE tests should be a little more thorough. To me the issue is here:
https://github.com/yldio/asap-hub/blob/b043e864a31d3d6e2ec2f04c2541236c2a7f2aeb/packages/react-components/src/organisms/__tests__/ResearchOutputForm.test.tsx#L214
You can keep the expect.anything but then make sure to have a couple of tests that assert just that one field based on whether this is a new or an existing RO

peterstarling (changes requested):

PR #1977, opened 2022-08-26
Use odata-query lib for buuilding OData queries

ricardocarvalh0:


apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -247,24 +261,23 @@ export type FetchResearchOutputOptions = FetchOptions<ResearchOutputFilter> & {
   includeDrafts?: boolean;
 };
 
-const makeODataFilter = (filter?: ResearchOutputFilter): string => {
+const makeODataFilter = (filter?: ResearchOutputFilter): Filter | null => {
```
peterstarling, line 264: although im generally against that i think it would be helpful to export this function and write some unit tests for it. It's just a little hard to read right now so I'm interested to see what the input and output should look like

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -76,24 +77,37 @@ export class ResearchOutputSquidexDataProvider
...
+            },
+          },
+        ],
+        [] as Filter[],
+      );
 
-    const filterQ = makeODataFilter(filter);
+    const searchQ = containsFilters.length
+      ? containsFilters.length === 1
+        ? containsFilters[0]
```
peterstarling, line 102: there is some funky nested ternary fun happening here and as codecov suggests it is not fully tested. Perhaps you would like to add a test for when this should evaluate to { or ... }?

apps/crn-server/src/utils/urls.ts
```diff
@@ -1,4 +1,26 @@
+import buildQuery, { Filter } from 'odata-query';
 import { createUrlFactory } from '@asap-hub/squidex';
 import { appName, baseUrl } from '../config';
 
 export const createUrl = createUrlFactory({ appName, baseUrl });
+
+export const buildODataFilter = (filterObj: Filter): string =>
+  buildQuery({ filter: filterObj }).replace('?$filter=', '');
+
+export const buildEqFilterForWords = (
```
peterstarling, line 10: can you unit test this function?

apps/crn-server/test/data-providers/users.data-provider.test.ts
```diff
@@ -567,11 +567,11 @@ describe('User data provider', () => {
       await userDataProvider.fetch(fetchOptions);
 
       const expectedFilter =
-        "data/onboarded/iv eq true and data/role/iv ne 'Hidden' and" +
-        " ((contains(data/firstName/iv, '%22')" +
-        " or contains(data/lastName/iv, '%22')" +
-        " or contains(data/institution/iv, '%22')" +
-        " or contains(data/expertiseAndResourceTags/iv, '%22')))";
+        "data/onboarded/iv eq true and not(data/role/iv eq 'Hidden') and" +
+        " ((contains(data/firstName/iv,'\"'))" +
```
peterstarling, line 571: as you can see the test says `Should sanitise double quotation mark by encoding to hex` - clearly after your changes it no longer encodes to hex but escapes it with a backslash instead. I would not only make sure to change the tests - i would actually make sure that this still works as normal in squidex

peterstarling (changes requested):

apps/crn-server/src/types.ts
```diff
@@ -0,0 +1,5 @@
+export type ResearchOutputFilter = {
```
peterstarling, line 1: this type has no reason to sit in a separate file and should be part of `apps/crn-server/src/utils/odata.ts`

apps/crn-server/src/utils/odata.ts
```diff
@@ -0,0 +1,48 @@
...
+    return '';
+  }
+
+  const filterKey = `data/${field}/iv${subField ? `/${subField}` : ''}`;
+  const filters = words.reduce(
+    (acc: Filter[], word: string) => acc.concat({ [filterKey]: word }),
+    [],
+  );
+
+  return getFirstOrAll(filters) as Filter;
```
peterstarling, line 47: you shouldn't have to cast

apps/crn-server/src/utils/urls.ts
```diff
@@ -1,4 +1,8 @@
+import buildQuery, { Filter } from 'odata-query';
 import { createUrlFactory } from '@asap-hub/squidex';
 import { appName, baseUrl } from '../config';
 
 export const createUrl = createUrlFactory({ appName, baseUrl });
+
+export const buildODataFilter = (filterObj: Filter): string =>
```
peterstarling, line 7: isnt this the same as https://github.com/yldio/asap-hub/blob/18bc42017ae19b6f777075404e3fc0e0de5fb192/apps/crn-server/src/utils/odata.ts#L29

apps/crn-server/src/utils/odata.ts
```diff
@@ -0,0 +1,48 @@
+import buildQuery, { Filter } from 'odata-query';
+import { ResearchOutputFilter } from '../types';
+import { getFirstOrAll } from './arrays';
+
+const FILTER_URL_QUERY_PARAM = '?$filter=';
+
+export const makeODataFilter = (
+  filter?: ResearchOutputFilter,
```
peterstarling, line 8: this clearly isn't a util if it's taking a very specific type as input. You may want to move this next to the data provider

apps/crn-server/src/utils/arrays.ts
```diff
@@ -0,0 +1,8 @@
+export const getFirstOrAll = <T>(
```
peterstarling, line 1: this util is not tested, nor does it seem particularly helpful, the output type seems very much connected to the filter tool itself. I'm not sure what the purpose of it is but i would at least move it to the odata filter util

apps/crn-server/test/utils/odata.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      expect(makeODataFilter()).toBeNull();
+      expect(makeODataFilter(undefined)).toBeNull();
+    });
+  });
+});
+
+describe('buildEqFilterForWords', () => {
+  test('builds the query correctly', () => {
+    expect(
+      buildEqFilterForWords('some field', ['a', 'b', 'c'], 'subField'),
```
peterstarling, line 35: i have to say using a space in the field name caused a lot of confusion for me. Took me a while to understand it was adding the subfield at the end. Your fixtures should be somewhat reflective of a real life use case - you could even use an example of a team and a nested ID field.

peterstarling (changes requested):

apps/crn-server/src/utils/odata.ts
peterstarling: still outstanding

apps/crn-server/src/utils/urls.ts
peterstarling: still outstanding

peterstarling (changes requested):

peterstarling (approved):

PR #1982, opened 2022-08-29
feat: GP2-57 Create a working group members section

peterstarling:


packages/gp2-components/src/templates/WorkingGroupOverview.tsx
```diff
@@ -0,0 +1,36 @@
...
+  gridRowGap: `${36 / 17}em`,
+});
+
+type WorkingGroupOverviewProps = Pick<WorkingGroupResponse, 'members'>;
+
+const WorkingGroupOverview: React.FC<WorkingGroupOverviewProps> = ({
+  members,
+}) => (
+  <div css={styles}>
+    {members.length ? (
```
jhonny-m, line 16: I would have done it like 
```
!!members.length && <section> ... 
```

jhonny-m (changes requested):

packages/gp2-components/src/templates/WorkingGroupOverview.tsx
peterstarling: im only copying other engineer's approach https://github.com/yldio/asap-hub/blob/ffa7f88b22d6b919f90a90ca8193e665c5527f0f/packages/react-components/src/templates/TeamProfileAbout.tsx#L50

PR #1984, opened 2022-08-30
Omit the added date on edits

kmaid:


oratiu (approved):

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -132,12 +132,17 @@ describe('ResearchOutputs controller', () => {
...
       );
 
-      expect(
+      await expect(
         researchOutputs.create(getResearchOutputCreateData()),
       ).rejects.toThrow(GenericError);
     });
 
     test('Should create the new research output and return it', async () => {
+      const mockDate = new Date('2010-01-01');
```
peterstarling, line 141: can you use jest timer mocks with modern timers https://jestjs.io/docs/26.x/timer-mocks  ?

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -832,6 +832,7 @@ describe('ResearchOutputs data provider', () => {
 
   describe('Create and update', () => {
     afterEach(() => {
+      // eslint-disable-next-line jest/no-standalone-expect
```
peterstarling, line 835: is this really needed?

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -150,9 +155,11 @@ describe('ResearchOutputs controller', () => {
 
       const researchOutputCreateDataObject =
         getResearchOutputCreateDataObject();
-      expect(researchOutputDataProviderMock.create).toBeCalledWith(
-        researchOutputCreateDataObject,
-      );
+      expect(researchOutputDataProviderMock.create).toBeCalledWith({
+        ...researchOutputCreateDataObject,
+        addedDate: mockDate.toISOString(),
```
peterstarling, line 160: as per the types the addedDate is an optional parameter which CAN be defaulted to current time - I think this test should continue to use the data from the fixture.

We should have another tests that asserts that when the addedDate is missing it defaults it to the current time

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -445,7 +452,7 @@ describe('ResearchOutputs controller', () => {
           expect.objectContaining({
             authors: [
               {
-                externalAuthorId: externalAuthorId,
+                externalAuthorId,
```
peterstarling, line 455: how was this not picked up by the linter, i wonder?

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -162,6 +163,7 @@ export class ResearchOutputSquidexDataProvider
       environments: environmentIds,
       organisms: organismIds,
       subtype: (subtypeId && [subtypeId]) || [],
+      addedDate: addedDate || new Date(Date.now()).toISOString(),
```
peterstarling, line 166: is this the same functionality that we added to the controller? https://github.com/yldio/asap-hub/blob/e1286c235c1d078a82e434bc1756a2b9ff31fe3d/apps/crn-server/src/controllers/research-outputs.ts#L91?

This defaulting behaviour should imho only sit in the controller. The data-provider input type should therefore require that the `addedDate` is obligatory.

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -844,12 +845,13 @@ describe('ResearchOutputs data provider', () => {
         const researchOutputRequest = getResearchOutputCreateDataObject();
         const teamId = researchOutputRequest.teamIds[0];
         const researchOutputId = 'created-output-id';
+        const now = Date.now().toString();
```
peterstarling, line 848: i dont think we want any extra functionality inside our data providers - they should just accept data as is and it should be the controllers responsibility to default the `addedDate` to current time

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -282,73 +283,82 @@ export const getResearchOutputCreateDataObject =
...
-    rrid: { iv: 'RRID:AB_90755' },
-    labCatalogNumber: {
-      iv: 'http://example.com',
-    },
-    usageNotes: {
-      iv: 'some access instructions',
-    },
-  });
+export const getRestResearchOutputCreateData = (
+  addedDate?: string,
```
peterstarling, line 299: let's not use different patterns now and continue to use fixture generators without extra parameters. Should the change be required it can easily be done after assigning the result of this function to a variable.

packages/model/src/research-output.ts
```diff
@@ -221,9 +221,13 @@ export type ResearchOutputCreateDataObject = ResearchOutputCoreObject & {
   organismIds: string[];
   subtypeId?: string;
   teamIds: string[];
+  addedDate?: string;
```
peterstarling, line 224: i wonder if we really want this to be optional? do we want the clients to have the option to change it if necessary? (im just asking, its possible that we do)

peterstarling (changes requested):

apps/crn-server/test/controllers/research-outputs.test.ts
ricardocarvalh0: I did try that but tests started to fail with that "took more than 5 secs to run" error. I will double check it but it seems we'd need to update Jest in order to get it working correctly. Will keep you updated.

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
ricardocarvalh0: Lint was failing there. And adding it to that error that wasn't picked up by the linter I'd say it's not being ran correctly everywhere, will also gather more info about it.

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
ricardocarvalh0: Yes I was asking myself where would be the best place for it. I ended up putting it at both provider and controller considering they could be called by different places that wouldn't pass the date. Will apply it for the controller only.

packages/model/src/research-output.ts
ricardocarvalh0: That's an important decision to make as this type is used in many places/tests (data-providers and controllers, for its inputs/outputs/queries/responses)

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -304,6 +306,7 @@ export interface ResearchOutputController {
 }
 export type ResearchOutputCreateData = ResearchOutputPostRequest & {
   createdBy: string;
+  addedDate?: string;
```
peterstarling, line 309: the `ResearchOutputCreateData` is the type that the controller accepts as input, and here:
https://github.com/yldio/asap-hub/blob/e5eac73a96c54e8d705c26cd8cf699fa39d2e9d4/apps/crn-server/src/controllers/research-outputs.ts#L92
you are telling the controller to fill the date for  data-provider when it's not present. The thing is though - as you removed this property from the payload it will never be available anyway. I would say that you should remove this property from the `ResearchOutputCreateData` and only keep it in `ResearchOutputCreateDataObject`

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -341,14 +341,16 @@ export const getResearchOutputUpdateData = (): ResearchOutputUpdateData => ({
...
-      getRestResearchOutputCreateData();
-    return {
-      doi: { iv: null },
-      accession: { iv: null },
-      rrid: { iv: null },
-      ...restResearchOutputData,
-    };
+export const getRestResearchOutputUpdateData = (): Omit<
+  InputResearchOutput['data'],
+  'addedDate'
```
peterstarling, line 346: `InputResearchOutput` is a type that we use for squidex payload and i think that the addedDate should be present in the fixtures still?

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -149,7 +151,12 @@ describe('/research-outputs/ route', () => {
     });
 
     test('Should return 403 when user is not permitted to create research output', async () => {
-      const researchOutput = getResearchOutputPostRequest();
+      const {
+        // @ts-ignore
+        addedDate: _addedDate,
+        ...researchOutput
+      } = getResearchOutputPostRequest();
```
peterstarling, line 158: if we are removing addedDate from the `ResearchOutputPostRequest` this fixture should not really contain an addedDate, and its probably what typescript was telling you until you added the ignore. Update the fixture to not contain the no longer supported property value

packages/model/src/research-output.ts
peterstarling: ok i think i got it wrong, this is the type for the data-provider. However I can see that the addedDate is in the ResearchOutputCoreObject and it should stay required as is. 

All you are doing in this PR is removing the `addedDate` from the Post request (and therefore type) and making sure that the controller fills in the blank

peterstarling (changes requested):

apps/crn-server/test/controllers/research-outputs.test.ts
ricardocarvalh0: So I've tried again with no success. It seems that jest 26 doesn't work well with fakerTimers('modern'). Either I get the timeout issue or the controller doesn't use the fake time set

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -282,8 +283,11 @@ export const getResearchOutputCreateDataObject =
 
 export const getResearchOutputUpdateDataObject =
   (): ResearchOutputUpdateDataObject => {
-    const { createdBy: _, ...researchOutputCreateDataObject } =
-      getResearchOutputCreateDataObject();
+    const {
+      createdBy: _,
+      addedDate: __,
```
peterstarling, line 288: similar to my comments above i think this is incorrect - both create and update should allow for modification of addedDate

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -149,7 +147,8 @@ describe('/research-outputs/ route', () => {
     });
 
     test('Should return 403 when user is not permitted to create research output', async () => {
-      const researchOutput = getResearchOutputPostRequest();
+      const { ...researchOutput } = getResearchOutputPostRequest();
```
peterstarling, line 150: this is not necessary

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -126,9 +126,7 @@ describe('/research-outputs/ route', () => {
     const researchOutputResponse = getResearchOutputResponse();
 
     test('Should return a 201 when is hit', async () => {
-      const createResearchOutputRequest = {
-        ...getResearchOutputPostRequest(),
-      };
+      const { ...createResearchOutputRequest } = getResearchOutputPostRequest();
```
peterstarling, line 129: i realise it was like that previously, but the spread syntax is unnecessary here. You can just do 
```suggestion
      const createResearchOutputRequest = getResearchOutputPostRequest();
```
as we dont need a shallow clone

packages/model/src/research-output.ts
```diff
@@ -221,9 +221,13 @@ export type ResearchOutputCreateDataObject = ResearchOutputCoreObject & {
   organismIds: string[];
   subtypeId?: string;
   teamIds: string[];
+  addedDate: string;
 };
 
-export type ResearchOutputUpdateDataObject = ResearchOutputCoreObject & {
+export type ResearchOutputUpdateDataObject = Omit<
+  ResearchOutputCoreObject,
+  'addedDate'
```
peterstarling, line 229: ok so this is what i was getting at in some comments above - i think the update should allow the modification of addedDate

peterstarling (changes requested):

apps/crn-server/test/controllers/research-outputs.test.ts
ricardocarvalh0: I thought the controller wouldn't accept the date in its params. It would always set it to current time. Considering that, this test is actually doing the second scenario. It's checking if provider was called with current time. If the controller needs to also accept the parameter from outside then we need to add the first scnearion

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -987,11 +985,13 @@ describe('ResearchOutputs data provider', () => {
       test('Should update the existing research-output and return its ID', async () => {
         const researchOutputUpdateData = getResearchOutputUpdateDataObject();
 
+        const { addedDate: _, ...restResearchOutputUpdateData } =
```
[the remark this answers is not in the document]
ricardocarvalh0: Now I'm confused, isn't it the purpose of the fix? The ticket says it expects that "When a user edits an existing Research Output the ‘addedDate’ field should not be modified"

apps/crn-server/test/fixtures/research-output.fixtures.ts
ricardocarvalh0: I thought both create and update should not allow anything regarding addedDate.

create because it will set it as current time
update because it cannot be changed

as per ticket AC:
When a user creates a new Research Output the ‘addedDate’ field is populated with the current date, **as expected**

ricardocarvalh0:
Closing this as an attempt to fix pipeline failing step

PR #1992, opened 2022-08-31
feat: Projects page and route added

csbarbosa:
Ticket (https://asaphub.atlassian.net/browse/GP2-67)
Created route and page with a simple header for the Projects tab


apps/gp2-frontend/src/AuthenticatedApp.tsx
```diff
@@ -7,9 +7,10 @@ import { FC, lazy, useEffect } from 'react';
 import { Route, Switch, useRouteMatch } from 'react-router-dom';
 import { useRecoilState, useResetRecoilState } from 'recoil';
 import { auth0State } from './auth/state';
+import Projects from './projects/Projects';
```
jhonny-m, line 10: we should lazy load this as well, see line 17

jhonny-m (changes requested):

jhonny-m (approved):

PR #2007, opened 2022-09-08
Crn 1026 integrate the new component for past events in the dashboard

lctrt:


packages/react-components/src/templates/__tests__/DashboardPageBody.test.tsx
```diff
@@ -72,6 +74,31 @@ it('hides add links to your work space section when user is not a member of a te
   expect(screen.queryByText(/Add important links/i)).toBeNull();
 });
 
+describe('the past events card', () => {
+  const events = [
+    createEventResponse({}, 0),
```
oratiu, line 79: Can't you pass createListEventResponse(3) there ?

lctrt: didn't know it existed, will do that!

apps/crn-frontend/src/dashboard/Body.tsx
```diff
@@ -1,5 +1,25 @@
...
+  const { items: events } = useEvents(
+    getEventListOptions(date, {
+      past: true,
+      pageSize: 3,
+      currentPage: 1,
+    }),
+    user,
+  );
+
+  return <DashboardPageBody events={events} {...props} />;
```
oratiu, line 22: I think we should rename it to pastEvents to be more verbose

oratiu (approved):
Looks good.

PR #2028, opened 2022-09-14
feat/gp2 user api

markcsims:
adds to users route

- fetch users
- fetch user by id

### DIFF
--- a/apps/gp2-server/test/routes/user.route.test.ts
+++ b/apps/gp2-server/test/routes/user.route.test.ts
@@ -1,11 +1,19 @@
+import { FetchOptions } from '@asap-hub/model';
 import Boom from '@hapi/boom';
 import supertest from 'supertest';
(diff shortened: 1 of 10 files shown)


apps/gp2-server/test/routes/user.route.test.ts
```diff
@@ -15,6 +23,122 @@ describe('/users/ route', () => {
...
+          });
+
+        expect(response.status).toBe(400);
+      });
+
+      test('Should return the results correctly when a filter is used', async () => {
+        userControllerMock.fetch.mockResolvedValueOnce(fetchExpectation);
+
+        const response = await supertest(appWithMockedAuth).get(
+          '/users?filter=Project+Manager',
```
peterstarling, line 97: try using query as in line #86

peterstarling (changes requested):

peterstarling (approved):

PR #2030, opened 2022-09-14
fix gp2 login

markcsims:
https://asaphub.atlassian.net/browse/GP2-95

### DIFF
--- a/apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
+++ b/apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
@@ -1,4 +1,4 @@
-import { UserMetadataResponse } from '@asap-hub/model';
+import { gp2, UserMetadataResponse } from '@asap-hub/model';
 import nock from 'nock';
 import addUserMetadata from '../add-user-metadata';
 import * as handleError from '../handle-error';
@@ -97,6 +97,17 @@ const apiUser: UserMetadataResponse = {
   algoliaApiKey: 'test-api-key',
 };
 
+const gp2ApiUser: gp2.UserResponse = {
+  displayName: 'Joao Tiago',
+  firstName: 'Joao',
+  lastName: 'Tiago',
+  email: 'joao.tiago@yld.io',
+  id: 'myRandomId123',
+  createdDate: '2020-08-21T14:23:31.924Z',
+  role: 'Trainee',
+  region: 'Europe',
+};
+
 describe('Auth0 Rule - Add User Metadata', () => {
   const apiURL = 'https://api.hub.asap.science';
   const appDomain = 'hub.asap.science';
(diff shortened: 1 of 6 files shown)


apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
```diff
@@ -187,6 +198,47 @@ describe('Auth0 Rule - Add User Metadata', () => {
     });
   });
 
+  it('adds the user metadata on successful fetch for gp2', async () => {
```
peterstarling, line 201: you may wanna update the other test's description with 'for crn'

apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
```diff
@@ -187,6 +198,47 @@ describe('Auth0 Rule - Add User Metadata', () => {
...
+      },
+      cb,
+    );
+
+    expect(cb).toHaveBeenCalled();
+    const [err, resUser, resContext] = cb.mock.calls[0];
+    expect(err).toBeFalsy();
+    expect(resUser).not.toBeNull();
+    expect(resContext).not.toBeNull();
+    expect(resContext.idToken['https://gp2.asap.science/user']).toStrictEqual({
```
peterstarling, line 230: it might be a good idea to extract the expectation to a variable and type it - it's a good practice because that way any time the type changes we know which expectations need to be updated. You would think jest would just make sure that whatever is in .toEqual(...) should adhere to the type inside expect() but i think it does not have to

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -1,10 +1,63 @@
...
-import type { UserMetadataResponse } from '@asap-hub/model';
+import type { gp2, UserMetadataResponse, UserResponse } from '@asap-hub/model';
 import got from 'got';
 import { URL, URLSearchParams } from 'url';
 import { handleError } from './handle-error';
 import type { Rule } from './types';
 
+type Auth0UserResponse = UserMetadataResponse | gp2.UserResponse;
+
+function isUserResponse(
```
peterstarling, line 10: surely this could be an arrow function

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -1,10 +1,63 @@
...
+  response: UserMetadataResponse | gp2.UserResponse,
+): response is UserResponse & { algoliaApiKey: string } {
```
peterstarling, line 12:
```suggestion
): response is UserMetadataResponse {
```

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -1,10 +1,63 @@
...
+  return 'onboarded' in (response as UserMetadataResponse);
```
peterstarling, line 13: the cast does not seem necessary, nor does it look right. Are you sure its needed?

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -1,10 +1,63 @@
...
+        role: team.role,
+      })),
+      algoliaApiKey,
+    };
+  }
+  const { id, email, firstName, lastName, avatarUrl } = response;
+
+  return {
+    id,
+    displayName: '',
```
peterstarling, line 50: actually i think we have displayName in the gp2 user response?

apps/auth0-rules/src/add-user-metadata.ts
```diff
@@ -1,10 +1,63 @@
...
 
+type Auth0UserResponse = UserMetadataResponse | gp2.UserResponse;
+
+function isUserResponse(
+  response: UserMetadataResponse | gp2.UserResponse,
+): response is UserResponse & { algoliaApiKey: string } {
+  return 'onboarded' in (response as UserMetadataResponse);
+}
+
+const extractUser = (response: Auth0UserResponse) => {
```
peterstarling, line 16: for helper methods like this i found that adding both input and output type make it easier to understand what they actually do, especially with a name as vague as this one
```suggestion
const extractUser = (response: Auth0UserResponse): User => {
```

peterstarling (changes requested):

apps/auth0-rules/src/add-user-metadata.ts
markcsims: done

apps/auth0-rules/src/add-user-metadata.ts
markcsims: done

apps/auth0-rules/src/add-user-metadata.ts
markcsims: done

apps/auth0-rules/src/add-user-metadata.ts
markcsims: done. The cast was from the Typescript documentation.

apps/auth0-rules/src/__tests__/add-user-metadata.test.ts
markcsims: Done.

peterstarling (approved):
you may wanna add some description to the PR

PR #2033, opened 2022-09-15
CRN-1028 test new schema

lctrt:


peterstarling:
@lctrt can we close this?

PR #2042, opened 2022-09-16
feature/gp2 78 create user entity integrate

markcsims:
https://asaphub.atlassian.net/browse/GP2-78

Integrates Users index with API


oratiu (approved):
Looks good, there are a few tests that rely on fixture data that should be modified but those were written before.

csbarbosa (approved):

PR #2048, opened 2022-09-19
adding frequency field to news schema

gabiayako:
Following the `Squidex schema changes and graphql` section of this repo README https://github.com/yldio/asap-hub#squidex-schema-changes-and-graphql, I've created a new field called `frequency` in the Squidex dev UI

Screen Shot 2022-09-19 at 11 26 49
Screen Shot 2022-09-19 at 11 27 11
Screen Shot 2022-09-19 at 11 27 24

and then I ran `yarn schema:update` that generated the changes committed in this PR.


packages/squidex/src/rest.ts
```diff
@@ -20,18 +20,25 @@ export interface Filter {
     | 'ne'
     | 'contains'
     | 'startsWith'
-    | 'endsWith';
-  value: string[] | number[] | string | number | boolean;
+    | 'endsWith'
+    | 'empty';
```
gabiayako, line 24: This empty is a valid filter :) I've found out using Squidex UI and checking the network tab, but I also found it here https://github.com/Squidex/squidex/blob/6ac635a3b819145e0fec90ed724617b3819fe282/src/Squidex.Infrastructure/Queries/FilterOperator.cs#L13

apps/crn-server/test/controllers/news.test.ts
```diff
@@ -120,6 +120,164 @@ describe('News controller', () => {
...
+                        value: ['CRN Quarterly', 'News Articles'],
+                      },
+                    ],
+                  },
+                ],
+              },
+              sort: [{ order: 'descending', path: 'created' }],
+            }),
+          })
+          .reply(200, newsSquidexApiResponse);
```
gabiayako, line 171: Although this `newsSquidexApiResponse` does not have appropriate data for this test, my understanding was that these tests are checking the query not the functionality itself. That's why I didn't change the `newsSquidexApiResponse` throughout the tests. But I am not sure if that's the best thing to do. What do you all think? :)

apps/crn-server/test/controllers/news.test.ts
```diff
@@ -120,6 +120,164 @@ describe('News controller', () => {
...
+
+        const result = await news.fetch({
+          take: 8,
+          skip: 5,
+          filter: {
+            frequency: ['CRN Quarterly'],
```
gabiayako, line 210: I've changed one record in squidex to have frequency CRN Quaterly then started the app and call http://localhost:3333/news?take=10&skip=0&filter=CRN+Quarterly passing an auth token I've got from the network tab in the hub. It worked 🎉 

Screen Shot 2022-09-20 at 08 35 04

packages/squidex/src/rest.ts
```diff
@@ -20,18 +20,25 @@ export interface Filter {
...
+  value: string[] | number[] | string | number | boolean | null;
+}
+
+export interface LogicalFilter {
+  and?: Filter[];
+  not?: Filter;
+  or?: Filter[];
```
gabiayako, line 31: I've made this query in the Squidex UI

Screen Shot 2022-09-21 at 06 19 44

that generated this query in the network tab. So I could see that this and blocks can be nested :)
Screen Shot 2022-09-21 at 06 21 27

apps/crn-server/test/controllers/news.test.ts
peterstarling: thats fine, news is a very old one which still has those 'test-it-all' kind of tests so we're going to address it later when we split the controller into two layers

apps/crn-server/test/controllers/news.test.ts
```diff
@@ -120,6 +120,164 @@ describe('News controller', () => {
...
+        })
+        .reply(200, newsSquidexApiResponse);
+
+      const result = await news.fetch();
+
+      expect(result).toEqual(listNewsResponse);
+    });
+
+    describe('Frequency Filter', () => {
+      test('Should query data properly when News Articles is selected', async () => {
```
peterstarling, line 142: i think you want to say 
```suggestion
      test('Should query data properly when CRN Quarterly and News Articles frequency is selected', async () => {
```

apps/crn-server/test/routes/news.route.test.ts
```diff
@@ -54,6 +54,23 @@ describe('/news/ route', () => {
       expect(newsControllerMock.fetch).toBeCalledWith(expectedParams);
     });
 
+    test('Should call the controller with the right filters', async () => {
```
peterstarling, line 57: should you add the text search one here as well maybe?

packages/squidex/src/rest.ts
peterstarling: oh nice

peterstarling (changes requested):

apps/crn-server/src/routes/news.route.ts
```diff
@@ -9,9 +10,14 @@ export const newsRouteFactory = (newsController: NewsController): Router => {
...
     const { query } = req;
 
-    const options = validateFetchPaginationOptions(query);
+    const options = validateFetchOptions(query);
 
-    const result = await newsController.fetch(options);
+    const result = await newsController.fetch({
+      ...options,
+      filter: options.filter && {
+        frequency: options.filter as NewsFrequency[],
```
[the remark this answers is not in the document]
peterstarling: ok so basically what TS is telling us is that we are trying to assign a string into a union, as far as I can tell.

What I propose you should do is to change the validation to be more specific ie use a dedicated validation function (like we for many other routes) then change the validation type and implementation to accept an enum instead of string.

peterstarling (approved):
this is exactly right, although i would check about any possible near future requirements regarding filtering - with the filter you introduced it is not really future proof in case we wanted to add more filter parameters - we would have to rename or something

PR #2064, opened 2022-09-20
feat: GP2-79 Add pagination to user directory

peterstarling:


apps/gp2-frontend/src/users/__tests__/UserList.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { container } = await renderUserList({
+    total: 2,
+    items: gp2Fixtures.createUsersResponse(2).items.map((user, i) => ({
+      ...user,
+      id: `${i}`,
+      displayName: `Display Name ${i}`,
+    })),
+  });
+  expect(container.textContent).toContain('Display Name 0');
+  expect(container.textContent).toContain('Display Name 1');
```
markcsims, line 68: Can we try to use screen (and GetByRole) rather than return the container.

apps/gp2-frontend/src/users/__tests__/UserList.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+            </MemoryRouter>
+          </WhenReady>
+        </Auth0Provider>
+      </Suspense>
+    </RecoilRoot>,
+  );
+  await waitFor(() =>
+    expect(result.queryByText(/loading/i)).not.toBeInTheDocument(),
+  );
+  return result;
```
markcsims, line 42: waitForElementToBeRemoved?

packages/fixtures/src/gp2/users.ts
```diff
@@ -19,9 +19,7 @@ export const createUserResponse = (
   ...overrides,
 });
 
-export const createUsersResponse = (
-  items = [createUserResponse()],
-): gp2.ListUserResponse => ({
-  items,
-  total: items.length,
+export const createUsersResponse = (items = 1): gp2.ListUserResponse => ({
+  items: Array.from({ length: items }, () => createUserResponse()),
```
markcsims, line 23: Should we set the id to the index of the array?

markcsims (changes requested):

apps/gp2-frontend/src/users/__tests__/UserList.test.tsx
markcsims: done

packages/fixtures/src/gp2/users.ts
markcsims: done

markcsims (approved):

PR #2073, opened 2022-09-21
CRN-1016 add news frequency filter

gabiayako:
Following the `Squidex schema changes and graphql` section of this repo README https://github.com/yldio/asap-hub#squidex-schema-changes-and-graphql, I've created a new field called `frequency` in the Squidex dev UI

Screen Shot 2022-09-19 at 11 26 49
Screen Shot 2022-09-19 at 11 27 11
Screen Shot 2022-09-19 at 11 27 24

and then I ran `yarn schema:update` that generated the changes committed in this PR.


packages/squidex/src/rest.ts
```diff
@@ -20,18 +20,25 @@ export interface Filter {
     | 'ne'
     | 'contains'
     | 'startsWith'
-    | 'endsWith';
-  value: string[] | number[] | string | number | boolean;
+    | 'endsWith'
+    | 'empty';
```
gabiayako, line 24: This empty is a valid filter :) I've found out using Squidex UI and checking the network tab, but I also found it here https://github.com/Squidex/squidex/blob/6ac635a3b819145e0fec90ed724617b3819fe282/src/Squidex.Infrastructure/Queries/FilterOperator.cs#L13

apps/crn-server/test/controllers/news.test.ts
```diff
@@ -120,6 +120,164 @@ describe('News controller', () => {
...
+      nock(baseUrl)
+        .get(`/api/content/${appName}/news-and-events`)
+        .query({
+          q: JSON.stringify({
+            take: 8,
+            filter: { path: 'data.type.iv', op: 'ne', value: 'Tutorial' },
+            sort: [{ order: 'descending', path: 'created' }],
+          }),
+        })
+        .reply(200, newsSquidexApiResponse);
```
gabiayako, line 134: Although this newsSquidexApiResponse does not have appropriate data for this test, my understanding was that these tests are checking the query not the functionality itself. That's why I didn't change the newsSquidexApiResponse throughout the tests. But I am not sure if that's the best thing to do. What do you all think? :)

apps/crn-server/test/controllers/news.test.ts
```diff
@@ -120,6 +120,164 @@ describe('News controller', () => {
...
+              sort: [{ order: 'descending', path: 'created' }],
+            }),
+          })
+          .reply(200, newsSquidexApiResponse);
+
+        const result = await news.fetch({
+          take: 8,
+          skip: 5,
+          filter: {
+            frequency: ['CRN Quarterly'],
```
gabiayako, line 210: I've changed one record in squidex to have frequency CRN Quaterly then started the app and call http://localhost:3333/news?take=10&skip=0&filter=CRN+Quarterly passing an auth token I've got from the network tab in the hub. It worked 🎉 

Screen Shot 2022-09-20 at 08 35 04

packages/squidex/src/rest.ts
```diff
@@ -20,18 +20,25 @@ export interface Filter {
...
+  value: string[] | number[] | string | number | boolean | null;
+}
+
+export interface LogicalFilter {
+  and?: Filter[];
+  not?: Filter;
+  or?: Filter[];
```
gabiayako, line 31: I've made this query in the Squidex UI

Screen Shot 2022-09-21 at 06 19 44

that generated this query in the network tab. So I could see that this `and` and `or` blocks can be nested :)
Screen Shot 2022-09-21 at 06 21 27

PR #2077, opened 2022-09-21
reminder event video

gabiayako:


packages/squidex/schema/crn/schemas/events.json
```diff
@@ -9,7 +9,7 @@
...
...ew Set();\n\nif (\n    ctx.data.speakers\n    && ctx.data.speakers.iv\n) {\n    for (const speaker of ctx.data.speakers.iv) {\n        if (!speaker.team && !speaker.user) {\n            return reject('You must select team or user or both for spekaer');\n        }\n        \n        if (\n            speaker.team.length === 1\n            && speaker.user.length === 1\n        ) {\n            const teamSpeaker = `${speaker.team[0]}-${speaker.user[0]}`;\n            if (teamsSpeakersSet.has(teamSpeaker)) {\n                return reject('Duplicates with same team and speaker are not allowed');\n            }\n            \n            teamsSpeakersSet.add(teamSpeaker);\n            \n            getReference(speaker.user[0], function (users) {\n              if (users && users.length === 1 && users[0].SchemaId.name === 'external-authors') {\n                  return reject('Selecting team for speaker is not allowed when user is external author');\n              }\n            });\n        }     \n    }\n}\n\nif (ctx.data.videoRecording.iv !== ctx.oldData.videoRecording.iv) {\n    ctx.data.videoRecordingUpdatedAt.iv = new Date().toISOString().split('.')[0]+\"Z\";\n    replace();\n}\n"
```
peterstarling, line 12: have you tried using just
```javascript
ctx.data.videoRecordingUpdatedAt.iv = new Date().toISOString();
```
?

packages/model/src/reminder.ts
```diff
@@ -2,11 +2,15 @@ import { ListResponse } from './common';
 import { EventResponse } from './event';
 import { ResearchOutputDataObject } from './research-output';
 
-type ReminderEntity = 'Research Output' | 'Event';
+type ReminderEntity = 'Research Output' | 'Event' | 'Event Material';
```
peterstarling, line 5: I think the point of this union is to list all entities (ie Research Output, Event, User) that the reminder might be relating to. `Event Material` really isn't one anyway and I think it should just be `Event`

packages/model/src/reminder.ts
```diff
@@ -2,11 +2,15 @@ import { ListResponse } from './common';
...
 
 type ResearchOutputReminderType = 'Published';
 type EventReminderType = 'Happening Today' | 'Happening Now';
-type ReminderType = ResearchOutputReminderType | EventReminderType;
+type EventMaterialReminderType = 'Video';
```
peterstarling, line 9: with the above comment in mind I think this should move to `EventReminderType` and be a new element of the union called something like `Video Updated`

apps/crn-server/src/queries/events.queries.ts
```diff
@@ -24,6 +24,7 @@ export const eventContentFragment = gql`
       notes
       videoRecordingPermanentlyUnavailable
       videoRecording
+      videoRecordingUpdatedAt
```
peterstarling, line 27: do we need this field anywhere other than the reminders? If not - I think just adding it to this file should suffice: https://github.com/yldio/asap-hub/pull/2077/files#diff-efb614c8fe87caa1c2764a46bdd6ae5248322d1383f1db077527ba6816223012

apps/crn-server/test/data-providers/reminders.data-provider.test.ts
```diff
@@ -564,6 +565,94 @@ describe('Reminder Data Provider', () => {
...
+        jest.useFakeTimers('modern');
+      });
+
+      afterAll(() => {
+        jest.useRealTimers();
+      });
+
+      beforeEach(async () => {});
+
+      test('Should fetch the reminder if its video was upserted in the last 24 hours', async () => {
```
peterstarling, line 579: actually if the video was `inserted` as part of the Event payload (as in - if the event was created with the video materials already in there) - this field wouldn't be populated (see: squidex create script - it's empty) so I think saying `upserted` here is a little misleading

apps/crn-server/test/data-providers/reminders.data-provider.test.ts
```diff
@@ -564,6 +565,94 @@ describe('Reminder Data Provider', () => {
...
+          videoRecordingUpdatedAt;
+
+        const result = await reminderDataProvider.fetch(fetchRemindersOptions);
+        expect(result).toEqual({
+          total: 1,
+          items: [expectedVideoEventUpdatedReminder],
+        });
+      });
+
+      test('Should not fetch the reminder if its video was upserted after 24 hours', async () => {
```
peterstarling, line 607: 24 hours after what...? I am trying to understand what these timers mean here. 

As far as I can tell we have 2 scenarios to cover:
1. The reminder is available because the video was uploaded anywhere between now and 24 hours ago
2. The reminder is not available because the video was uploaded more than 24 hours ago

There might be a secret 3 scenario in case we dont trust the data which could 
3. The reminder is not available because the video was uploaded in the future (as in - the date of the upload is a future date)

But I don't think we need to cover this. I'm pretty sure all scenarios are covered in this test file somehow but it would be good to make the test descriptions clearer - unless obviously I missed something?

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -204,6 +219,42 @@ const getEventRemindersFromQuery = (
...
+): VideoEventReminder[] => {
+  const eventReminders = (queryEventsContents || []).reduce<
+    VideoEventReminder[]
+  >((events, event) => {
+    const videoRecordingUpdatedAt = DateTime.fromISO(
+      event.flatData.videoRecordingUpdatedAt,
+    );
+
+    const ONE_DAY_IN_SECONDS = 24 * 60 * 60;
+    const secondsFromVideoUpdatedAt =
```
peterstarling, line 233: have you considered using luxon for this a little more? For instance you could use
```javascript
DateTime.now().diff(DateTime.fromISO(event.flatData.videoRecordingUpdatedAt)).seconds
  ```
to get the diff (I would imagine it would give you a positive and negative difference)

**Review by peterstarling — CHANGES_REQUESTED** [review]
I think you should also add integration tests, see this file:
https://github.com/yldio/asap-hub/blob/master/apps/crn-server/test/integration/reminders.integration-test.ts

Not only would that be in line with how previous reminder were built but also it would give us that flexibility that should we want to change the way to store reminder these integration tests should still run as normal. Finally - this would enable you to test the squidex script (for updating video uploaded timestamp) which currently isn't tested

File: apps/crn-server/test/data-providers/reminders.data-provider.test.ts (hunk 6 above)
Comment (gabiayako) [reply apps/crn-server/test/data-providers/reminders.data-provider.test.ts:607]: Nice! Yeah, I think the test descriptions are not that clear 😅  I added this case 3 in here https://github.com/yldio/asap-hub/blob/CRN-958-reminders-published-event-materials-video/apps/crn-server/test/data-providers/reminders.data-provider.test.ts#L630-L631

File: apps/crn-server/src/queries/events.queries.ts (hunk 4 above)
Comment (gabiayako) [reply apps/crn-server/src/queries/events.queries.ts:27]: The `videoRecordingUpdatedAt` is being used in this sort function

```
const getSortDate = (reminder: ReminderDataObject): DateTime => {
  if (reminder.entity === 'Research Output') {
    return DateTime.fromISO(reminder.data.addedDate);
  }

  if (reminder.type === 'Happening Today') {
    return DateTime.fromISO(reminder.data.startDate);
  }

  if (reminder.type === 'Video Updated') {
    return DateTime.fromISO(reminder.data.videoRecordingUpdatedAt);
  }

  return DateTime.fromISO(reminder.data.endDate);
};
```

I believe it's possible to remove the startDate and endDate from the videoReminder though :)

File: packages/squidex/schema/crn/schemas/events.json (hunk 1 above)
Comment (gabiayako) [reply packages/squidex/schema/crn/schemas/events.json]: nice! I haven't tried that before! it works perfectly :)

File: apps/crn-server/src/autogenerated-gql/graphql.ts [hunk 8]
```diff
@@ -4778,16 +4791,6 @@ export type UsersDataAdminNotesInputDto = {
   iv: InputMaybe<Scalars['String']>;
 };
 
-/** The structure of the AlumniSinceDate field of the Users content type. */
-export type UsersDataAlumniSinceDateDto = {
-  iv: Maybe<Scalars['Instant']>;
-};
-
-/** The structure of the AlumniSinceDate field of the Users content input type. */
-export type UsersDataAlumniSinceDateInputDto = {
```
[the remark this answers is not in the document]
m-doughty: It looks like this commit [has only been on `master` for 5 hours](https://github.com/yldio/asap-hub/commit/8dd16debb6fc7f8da567c206d045053377e542ad) so you may need to pull and rebase.

gabiayako:
just rebased to master and force pushed so ci updates PR env with changes in https://github.com/yldio/asap-hub/pull/2093 and then I can run `yarn schema:update:crn` again to update the autogenerated files

gabiayako:
now autogenerated files seems ok after rebasing to master with https://github.com/yldio/asap-hub/pull/2093 merged and running `yarn schema:update:crn`

m-doughty (approved):
Looks good to me.

peterstarling (approved):

PR #2079, opened 2022-09-22
Crn 1028 alumni ci rerun

lctrt:


kmaid (changes requested):

packages/react-components/src/organisms/__tests__/PeopleCard.test.tsx
```diff
@@ -40,3 +41,14 @@ it('links to the profile', () => {
     expect.stringMatching(/42$/),
   );
 });
+
+describe('alumni badge', () => {
+  it('renders the alumni badge when required', () => {
+    render(<PeopleCard {...props} alumniSinceDate="2022-01-01" />);
+    expect(screen.getByTestId('alumni-badge')).toHaveTextContent('Alumni');
```
kmaid, line 48: Test ID

kmaid: Also another approach to this is to have the one test and render and assert twice to toggle this behaviour. That way it doesn't need to be stuck in a describe block for the small bit functionality. Could also assert that the icon is there and not there based on the title but thats super minior

kmaid (approved):

PR #2081, opened 2022-09-22
feat: CRN-1047 Remove non-onboarded users from speakers

peterstarling:


markcsims (approved):

apps/crn-server/src/entities/event.ts
```diff
@@ -101,7 +89,7 @@ export const parseGraphQLSpeakers = (
         ?.filter((t) => t.id && t.id[0]?.id === team.id)
         .filter((s) => s.role)[0]?.role || undefined;
 
-    if (!user || !role) {
+    if (!user || !role || user.flatData.onboarded !== true) {
```
gabiayako, line 92: when there is only a team not the user among the speakers data, I saw that appears this message `Speaker was not announced`
Screen Shot 2022-09-22 at 11 21 28

Do we want to leave it like that? :)

markcsims:
We should update the description with  the Jira ticket etc.

apps/crn-server/src/entities/event.ts
peterstarling: i think we probably do, it's pretty much the same behaviour as if there was no user. What we are saying is that attaching a speak who isn't onboarded (read - not active) is the same as attaching no speakers

gabiayako: Ok! :) just wanted to confirm that 😉

gabiayako (approved):

apps/crn-server/src/entities/event.ts
peterstarling: but i think i might make a note of that in the ticket as it is a very good point

PR #2082, opened 2022-09-23
Added header area and tag for inactive groups

oratiu:


packages/react-components/src/templates/GroupProfilePage.tsx
```diff
@@ -12,10 +13,16 @@ const mainStyles = css({
...
+  active,
   ...props
 }) => (
   <article>
-    <GroupProfileHeader {...props} />
+    {!active && (
+      <Toast accent="info">
+        This group is inactive and might not have all content available.
+      </Toast>
+    )}
```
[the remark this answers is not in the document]
oratiu: Thank you for catching that, idk how it slipped from me.

lctrt (approved):

packages/react-components/src/organisms/Toast.tsx
```diff
@@ -72,21 +80,38 @@ const crossPlaceholderStyles = css(iconStyles, {
 const wrapStyles = css({
   display: 'flex',
   flexWrap: 'wrap',
+  flexFlow: 'row',
 });
 
+export const getIcon = (accent?: 'alert' | 'info'): EmotionJSX.Element => {
```
kmaid, line 86: Wouldn't this be better expressed as a Record and set a default accent in the props?

packages/react-components/src/organisms/Toast.tsx
```diff
@@ -72,21 +80,38 @@ const crossPlaceholderStyles = css(iconStyles, {
...
   children: ReactNode;
   onClose?: () => void;
+  accent?: 'alert' | 'info';
 }
-const Toast: React.FC<ToastProps> = ({ children, onClose }) => (
-  <section css={styles}>
+const Toast: React.FC<ToastProps> = ({
+  children,
+  onClose,
+  accent = 'alert',
```
kmaid, line 105: This means your default case can't ever be reached no?

packages/react-components/src/icons/info-circle-yellow.tsx
```diff
@@ -0,0 +1,26 @@
+/* istanbul ignore file */
+
+const infoCircleYellow = (
```
kmaid, line 3: This should be added to the icons storybook. Surprised we don't already have this

apps/storybook/src/Icons.stories.tsx
```diff
@@ -40,6 +40,7 @@ export const Groups = () => components.groupsIcon;
 export const HidePassword = () => components.hidePasswordIcon;
 export const Info = () => components.infoIcon;
 export const InfoCircle = () => components.infoCircleIcon;
+export const infoCircleYellow = () => components.infoCircleYellow;
```
ricardocarvalh0, line 43: Shouldn't the name start with upper case as the others?

ricardocarvalh0 (approved):

PR #2083, opened 2022-09-23
Upgrade react-scripts

kmaid:
Current work:
By upgrading to React Scripts 5.01 we are upgrading major versions of webpack 5, Jest 27, and ESLint 8. This will unblock our upgrade path for some of our core dependencies most notable React itself!

Deployment:
Changes to webpack have altered our auth0 login page HTML. I think the old HTML will work but due do auth0 having shared envs is not easy to test and can only be done in dev/prod. We will need a maintenance window to deploy these changes.

Testing:
We should try and regression test the entire website as much as possible. Some areas I have had to do more migration work are:
- Teams, user and research output exports
	- Client might need to test this as there were problems with the initial implementation and windows excel. I have tested with mac Excel and it seems okay.
	- I can't find the buttons to export.
- Login pages (these can only be tested properly in dev)
- Rich text format
- GP2
- Avatar upload

Future Work
I will bring this up in RTM today discuss priority and impact  of these issues. I think they are all pretty important.

Upgrade Typescript

Change FE build output to ESM and drop building CJS versions. (https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)
- Once we are outputting ESM we can update Unified to Version 10 and remove the forced Resolution for VFile
- Speed up pipeline by no-longer compiling for CJS
- Remove jest hack for csv-stream ('^csv-stringify/browser/esm': require.resolve('csv-stringify'),)

Upgrade React to 18 

Speed up pipeline with Jest sharding

Fix the problem with Broken timers on Event tests
- I am going to comment this test out and ignore coverage complaints. 

Investigate the bundle file size
- Investigate the main.js bundle size ballooning to 260kb+ (from 160kb which was already bumped from 128 kb max bundle size limit we had set before)
- Auth Frontend looks to have hit 900kb
	- This doesn't look related to react scripts changes it looks like its been a problem for a while


apps/crn-frontend/package.json
```diff
@@ -76,8 +79,9 @@
     "jose": "2.0.6",
     "nock": "13.0.11",
     "pirates": "4.0.5",
-    "react-scripts": "4.0.3",
+    "react-scripts": "5.0.1",
     "react-test-renderer": "17.0.2",
-    "typescript": "4.6.4"
+    "typescript": "4.6.4",
+    "webpack": "5.74.0"
```
kmaid, line 85: Not sure we need this 🤷

apps/crn-frontend/scripts/__tests__/build-unsupported-browser-page.build-output-test.ts
```diff
@@ -41,18 +41,6 @@ describe('the detection script', () => {
     delete mockNavigator.userAgent;
   });
 
-  describe('on a modern browser', () => {
```
kmaid, line 44: Urgh, this is also outstanding.

apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -1,7 +1,8 @@
 import { EntityResponses, SearchEntityResponse } from '@asap-hub/algolia';
 import { ResearchOutputResponse } from '@asap-hub/model';
 import { isInternalUser } from '@asap-hub/validation';
-import { CsvFormatterStream, Row, format } from '@fast-csv/format';
+/* eslint-disable-next-line import/no-unresolved */
+import { stringify, Options, Stringifier } from 'csv-stringify/browser/esm';
```
kmaid, line 5: Eslint cant resolve this for the same reason we had to hack this module in the jest tests.

apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -91,31 +94,33 @@ export const researchOutputToCSV = (
...
+  csvStream: Stringifier,
   getResults: ({
     currentPage,
     pageSize,
   }: Pick<GetListOptions, 'currentPage' | 'pageSize'>) => Readonly<
     Promise<SearchEntityResponse<T>>
   >,
-  transform: (result: EntityResponses[T]) => Row,
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  transform: (result: EntityResponses[T]) => Record<string, any>,
```
kmaid, line 123: This is a hack because of the algolia library work. I have just maintained the type from Fast-CSV (we now use CSV because they include relavant polyfills.)

apps/crn-frontend/src/shared-research/export.ts
kmaid: We do support ESM modules not in the jest environment. I wanted to just import the CJS module that is included but couldn't.

apps/crn-server/webpack.config.js
```diff
@@ -23,6 +23,9 @@ module.exports = {
   },
   resolve: {
     extensions: ['.js', '.jsx', '.ts', '.tsx'],
+    fallback: {
```
kmaid, line 26: @peterstarling seems strange we need this

apps/gp2-auth-frontend/package.json
```diff
@@ -19,10 +19,13 @@
     "@asap-hub/react-components": "workspace:*",
     "@babel/runtime-corejs3": "7.19.1",
     "history": "4.10.1",
+    "path": "npm:path-browserify@1.0.1",
+    "path-browserify": "1.0.1",
```
kmaid, line 23: I don't think we need this.

apps/gp2-frontend/package.json
```diff
@@ -14,41 +14,39 @@
     "release:frontend": "node ./scripts/sentry-release.js"
   },
   "dependencies": {
+    "@asap-hub/algolia": "workspace:*",
     "@asap-hub/auth": "workspace:*",
+    "@asap-hub/fixtures": "workspace:*",
     "@asap-hub/flags": "workspace:*",
     "@asap-hub/frontend-utils": "workspace:*",
     "@asap-hub/gp2-components": "workspace:*",
+    "@asap-hub/model": "workspace:*",
```
kmaid, line 23: We have dictionaries and stuff we use in prod code so its a dependency not a dev dependency. Not sure how i feel about this but ESLint called us out.

apps/gp2-frontend/package.json
```diff
@@ -14,41 +14,39 @@
...
     "history": "4.10.1",
+    "path": "npm:path-browserify@1.0.1",
     "react": "17.0.2",
     "react-app-polyfill": "3.0.0",
     "react-dom": "17.0.2",
     "react-error-boundary": "3.1.4",
     "react-router-dom": "5.3.3",
     "react-router-last-location": "2.0.1",
     "recoil": "patch:recoil@^0.1.2#../../patches/recoil.patch",
-    "streamsaver": "2.0.6",
```
kmaid, line 40: There is not export functionality on GP2 ATM. If there is we should move it to a shared module or maybe into the algolia library.

package.json
```diff
@@ -55,7 +55,8 @@
   },
   "resolutions": {
     "encoding": "0.1.13",
-    "expect-playwright": "0.8.0"
+    "expect-playwright": "0.8.0",
```
kmaid, line 58: We have to force an older version of this dependency because it includes playwright. Later versions of playwright require Jest 28 which isn't in React Scripts. I should write documentation

package.json
```diff
@@ -55,7 +55,8 @@
...
+    "vfile": "4.2.1"
```
kmaid, line 59: This is a peer dependency of Unified and minor version later as it includes the relevant polyfills for Webpack 5 support. Unified 10 is ESM only which we currently cannot support.

package.json
```diff
@@ -91,7 +92,7 @@
     "files": [
       {
         "path": "apps/crn-frontend/build/**/*.js",
-        "maxSize": "160 kB"
+        "maxSize": "300 kB"
```
kmaid, line 95: This is lame and needs looked into. I don't know why this could be as we should be including less polyfills. I need to create a ticket and include the bundle size report so we can compare before and after to see what has been moved into the main bundle.

package.json
```diff
@@ -106,39 +107,46 @@
...
+    "react-test-renderer": "17.0.2",
     "rimraf": "3.0.2",
     "typescript": "4.6.4",
-    "webpack": "5.37.0"
+    "webpack": "5.74.0"
   },
   "dependencies": {
-    "ts-node": "10.9.1"
+    "ts-node": "10.9.1",
+    "uuid": "8.3.2"
```
kmaid, line 149: @peterstarling Do you know why we had to include uuid?

packages/auth-frontend-scripts/src/__tests__/build.test.ts
```diff
@@ -44,29 +44,29 @@ describe('build', () => {
...
       <!doctype html>
         <html lang="en">
-          <body>
-            <script src="/static/js/2.eb5d7501.chunk.js">
-            </script><script src="/static/js/main.f9ee5131.chunk.js"></script>
-          </body>
+        <head>
+          <script
+            defer="defer"
+            src="https://dev.hub.asap.science/.auth/static/js/main.80665dab.js"
```
kmaid, line 50: Our Regex failed because .chunk. is no longer included in chunk files. I also don't know why we have gone from 2 chunks to 1

apps/crn-frontend/package.json
peterstarling: i dont think we do, no, this was for react-app-rewired

jest.config.js
```diff
@@ -36,7 +36,10 @@ const lintConfigs = [
 ].map(([dir, packageOrApp]) => ({
   rootDir: resolve(dir, packageOrApp),
   runner: require.resolve('jest-runner-eslint'),
-  testMatch: ['<rootDir>/src/**/*.{js,jsx,ts,tsx}'],
+  testMatch: [
+    '<rootDir>/src/**/*.{js,jsx,ts,tsx}',
```
kmaid, line 40: Shouldnt this just be *.test.{js....

jest/jest-base.config.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
 
   moduleNameMapper: {
     '^@asap-hub/([^/]+)$': '@asap-hub/$1/src',
+    uuid: require.resolve('uuid'),
```
kmaid, line 29: Can we remove this

peterstarling (approved):

jest.config.js
kmaid: These are lint configs

PR #2092, opened 2022-09-26
[CRN-1069] update latest news sections on dashboard according to new designs

ricardocarvalh0:
1. Add news description
2. Add view all functionality

### DIFF
--- a/apps/storybook/src/DashboardPageBody.stories.tsx
+++ b/apps/storybook/src/DashboardPageBody.stories.tsx
@@ -5,7 +5,7 @@ import {
   createListReminderResponse,
   createListResearchOutputResponse,
 } from '@asap-hub/fixtures';
-import { number } from '@storybook/addon-knobs';
+import { number, text } from '@storybook/addon-knobs';
 
 export default {
   title: 'Templates / Dashboard / Page Body',
@@ -17,7 +17,11 @@ const props = (): ComponentProps<typeof DashboardPageBody> => ({
       id: 'uuid-1',
       created: new Date().toISOString(),
       type: 'News' as const,
-      title: "Coordinating different research interests into Parkinson's",
+      title: 'Learn about Protocols.io, an ASAP preferred tool',
+      shortText:
+        'Discover current and planned tools (e.g., animal & cell models, antibodies, vectors, tissues, etc.) in a sortable table. This will be a living reference.',
+      link: text('Link', 'https://picsum.photos/200'),
+      linkText: text('Link Text', 'External Link'),
     },
     {
       id: 'uuid-2',
--- a/apps/storybook/src/NewsSectionDashboard.stories.tsx
+++ b/apps/storybook/src/NewsSectionDashboard.stories.tsx
@@ -1,0 +1,28 @@
+import { NewsSection } from '@asap-hub/react-components';
+import { ComponentProps } from 'react';
+import { text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'Organisms / Dashboard / News Section',
+  component: NewsSection,
+};
+
+const props = (): ComponentProps<typeof NewsSection> => ({
+  title: 'Latest News',
+  subtitle: 'Explore the latest shared research and learn more about them.',
+  news: [
+    {
+      id: 'uuid-1',
+      created: new Date().toISOString(),
+      type: 'News' as const,
+      title: 'Learn about Protocols.io, an ASAP preferred tool',
+      shortText:
+        'Discover current and planned tools (e.g., animal & cell models, antibodies, vectors, tissues, etc.) in a sortable table. This will be a living reference.',
+      link: text('Link', 'https://picsum.photos/200'),
+      linkText: text('Link Text', 'External Link'),
+      noPill: true,
+    },
+  ],
+});
+
+export const Normal = () => <NewsSection {...props()} />;
--- a/packages/gp2-components/src/templates/DashboardPageBody.tsx
+++ b/packages/gp2-components/src/templates/DashboardPageBody.tsx
@@ -2,6 +2,9 @@ import { ComponentProps } from 'react';
 import { css } from '@emotion/react';
 
 import { PagesSection, pixels, NewsSection } from '@asap-hub/react-components';
+import { Link } from '@asap-hub/react-components/build';
+import { news as newsRoute } from '@asap-hub/routing/build';
+import { perRem } from '@asap-hub/react-components/build/pixels';
 
 const { rem } = pixels;
 
@@ -9,6 +12,11 @@ const styles = css({
   display: 'grid',
   gridRowGap: `${rem(72)}`,
   marginBottom: `${rem(24)}`,
+});
+
+const viewAllStyles = css({
+  marginTop: `${24 / perRem}em`,
+  textAlign: 'right',
 });
 
 type DashboardPageBodyProps = Omit<
@@ -29,7 +37,16 @@ const DashboardPageBody: React.FC<DashboardPageBodyProps> = ({
       <PagesSection title={'Not sure where to start?'} pages={pages} />
     ) : null}
     {news.length ? (
-      <NewsSection title="Latest News from GP2" news={news} />
+      <div>
+        <NewsSection
+          title="Latest News"
+          subtitle="Explore the latest shared research and learn more about them."
+          news={[news[0]]}
+        />
+        <p css={viewAllStyles} data-testid="view-news">
+          <Link href={newsRoute({}).$}>View All →</Link>
+        </p>
+      </div>
     ) : null}
   </div>
 );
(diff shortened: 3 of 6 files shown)


m-doughty (approved):

apps/storybook/src/NewsSectionDashboard.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import { NewsSection } from '@asap-hub/react-components';
+import { ComponentProps } from 'react';
+import { text } from '@storybook/addon-knobs';
+
+export default {
```
kmaid, line 5: Why is there a new storybook for an existing component? Like the component is called NewsSection so the convention is that it should be NewsSection.stories.tex

kmaid (changes requested):
See comments

apps/storybook/src/NewsSectionDashboard.stories.tsx
ricardocarvalh0: The first reason was because other dashboard "parts" are on specific stories (PastEventsDashboard and UpcomingEventsDashboard) so I thought on following the same pattern

The second reason is because the news dashboard section is not only the NewsSection but also the View All link which can't be inside NewsSection because the latter is being used in other places like DiscoverNews. That might be a reason to create a new component only for the dashboard like DashboardNewsSection but I also thought would be additional effort for a small case

The third reason is I thought it wouldn't be needed a specific story for NewsSection as it's been used in DiscoverNews and there's no story for it also

packages/gp2-components/src/templates/DashboardPageBody.tsx
```diff
@@ -1,16 +1,27 @@
...
+const { rem, perRem } = pixels;
 
 const styles = css({
   display: 'grid',
   gridRowGap: `${rem(72)}`,
   marginBottom: `${rem(24)}`,
 });
 
+const viewAllStyles = css({
+  marginTop: `${24 / perRem}em`,
```
jhonny-m, line 21: we can use the `rem` funtion and make it: 
```marginTop: rem(24),```

File: apps/storybook/src/NewsSectionDashboard.stories.tsx (hunk 1 above)
Comment (kmaid) [reply apps/storybook/src/NewsSectionDashboard.stories.tsx:5]: PastEventsDashboard and Upcoming Events Dashboard are two separate components.

Generally we put header and description in the relevant page layout as the card content themselves tend to be more reusable. I think you will find all the other components on the dashboard page do not have the Title and Description/Short Text within them/ 

The dashboard was created very quickly by the old project lead in order to meet contractual obligations so it's not the best example of how things should be done.

File: packages/gp2-components/src/templates/DashboardPageBody.tsx (hunk 2 above)
Comment (kmaid) [reply packages/gp2-components/src/templates/DashboardPageBody.tsx:21]: I wonder if that function name should be pxToRem or something.

File: packages/gp2-components/src/templates/__tests__/DashboardPageBody.test.tsx [hunk 3]
```diff
@@ -49,5 +49,12 @@ it('renders news section when there are no pages', () => {
   ).not.toBeInTheDocument();
   expect(
     screen.getAllByRole('heading').map(({ textContent }) => textContent),
-  ).toEqual(expect.arrayContaining(['News Title', 'Tutorial Title']));
+  ).toEqual(expect.arrayContaining(['News Title']));
+  expect(screen.getByTestId('view-news').querySelector('a')).toHaveTextContent(
```
kmaid, line 53: Please do not use test IDs, you should inject text and get by that instead. I know this was done on two other PRs of which one has been corrected. The intention is that we should be mimicking users as much as possible in our tests. They can see text on a page the cannot see a random ID in order to create better tests

packages/react-components/src/templates/__tests__/DashboardPageBody.test.tsx
```diff
@@ -69,7 +63,14 @@ it('renders news section', () => {
 
   expect(
     screen.getAllByRole('heading').map(({ textContent }) => textContent),
-  ).toEqual(expect.arrayContaining(['News Title', 'Tutorial Title']));
+  ).toEqual(expect.arrayContaining(['News Title']));
+  expect(screen.getByTestId('view-news').querySelector('a')).toHaveTextContent(
```
kmaid, line 67: Test Ids

apps/storybook/src/DashboardPageBody.stories.tsx
```diff
@@ -17,7 +17,11 @@ const props = (): ComponentProps<typeof DashboardPageBody> => ({
       id: 'uuid-1',
       created: new Date().toISOString(),
       type: 'News' as const,
-      title: "Coordinating different research interests into Parkinson's",
+      title: 'Learn about Protocols.io, an ASAP preferred tool',
+      shortText:
+        'Discover current and planned tools (e.g., animal & cell models, antibodies, vectors, tissues, etc.) in a sortable table. This will be a living reference.',
+      link: text('Link', 'https://picsum.photos/200'),
```
kmaid, line 23: Where possible we try and use http://example.com as its exactly their use case (https://en.wikipedia.org/wiki/Example.com)

kmaid: Internal links we generally use #wrong

apps/crn-server/src/schema/autogenerated-schema.graphql
```diff
@@ -2731,6 +2743,7 @@ type EventsFlatDataDto {
   videoRecording: String
   """This box is automatically ticked if no output is added after 14 days from the event's end date."""
   videoRecordingPermanentlyUnavailable: Boolean
+  videoRecordingUpdatedAt: Instant
```
peterstarling, line 2746: whats all this doing here?

peterstarling (changes requested):

packages/react-components/src/templates/DashboardPageBody.tsx
```diff
@@ -22,34 +22,34 @@ import {
...
 const listStyles = css({
-  paddingLeft: `${18 / perRem}em`,
+  paddingLeft: `${rem(18)}em`,
 });
 
 const infoStyles = css({
   color: lead.rgb,
-  padding: `${3 / perRem}em 0 ${24 / perRem}em`,
-  lineHeight: `${24 / perRem} em`,
+  padding: `${rem(3)}em 0 ${rem(24)}em`,
```
gabiayako, line 47: Isn't this substitution supposed to be 

```suggestion
  padding: `${rem(3)} 0 ${rem(24)}`,
```


The spacing looks a bit weird in here https://2092.hub.asap.science/, I think it may be due to this change

2092 hub asap science_(iPhone XR)

ricardocarvalh0: good catch, fixed!

apps/crn-server/src/schema/autogenerated-schema.graphql
kmaid: Won't a rebase remove this?

packages/gp2-components/src/templates/DashboardPageBody.tsx
```diff
@@ -11,6 +17,11 @@ const styles = css({
   marginBottom: `${rem(24)}`,
 });
 
+const viewAllStyles = css({
+  marginTop: `${rem(24)}`,
```
kmaid, line 21:
```suggestion
  marginTop: rem(24),
```

packages/react-components/src/templates/__tests__/DashboardPageBody.test.tsx
```diff
@@ -103,7 +97,7 @@ it('displays events cards or placeholder if there are no events', () => {
 });
 
 describe('the past events card', () => {
-  const events = createListEventResponse(3).items;
+  const events = createListEventResponse(4).items;
```
kmaid, line 100: This seems weird. Why are we giving 4 events and checking three are displayed?

kmaid (changes requested):

packages/react-components/src/templates/__tests__/DashboardPageBody.test.tsx
ricardocarvalh0: The 'View All' link for past events wasn't checking for the number os past events, therefore it was showing it no matter how many results were being returned. I added the `pastEvents && pastEvents.length > 3` check so it could behave like the others and then it has to be tested receiving 3+ results in order to make sure the link is being rendered.

ricardocarvalh0: I'll add the fourth one for the first test.

apps/crn-server/src/schema/autogenerated-schema.graphql
ricardocarvalh0: Just tried rebase and it didn't work. maybe reopen the PR?

ricardocarvalh0:
Closing it as an attempt to fix schema issue.

PR #2098, opened 2022-09-27
CRN-959 Reminders: Published Event Materials - Presentation

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-959

It follows the same idea and steps of CRN-958 Reminders: Published Event Materials - Video
 https://github.com/yldio/asap-hub/pull/2077.

1. It was created a field called presentationUpdatedAt in Event that it is disabled for editing
2. This field is populated by the script update in packages/squidex/schema/crn/schemas/events.json
3. This new field presentationUpdatedAt was added to reminders query in apps/crn-server/src/queries/reminders.queries.ts
4. It was added a condition to also get the events which presentationUpdatedAt is greater than 1 day ago in getEventFilter (apps/crn-server/src/data-providers/reminders.data-provider.ts)
5. The function getEventMaterialsRemindersFromQuery (apps/crn-server/src/data-providers/reminders.data-provider.ts) checks if the material was updated in the last 24 hours, if so, it is added to reminders. 

Obs1: Since 3 fields in events are going to be checked to see if they are updated (video, notes, presentation) I am running the same tests for each type of material using jest table https://jestjs.io/docs/api#2-testeachtablename-fn-timeout
Obs2: As it can be read in this thread https://yld.slack.com/archives/CV4UY4411/p1664356426184539 there’s an edge case where the person comes back and erases the material from the event. In this case, the reminder shouldn’t appear. I am fixing this issue in the presentation but not in the video yet.

---

#### Validation

I've updated an event in the squidex PR env https://cloud.squidex.io/app/crn-2098/content/events/929457f1-b004-4bf8-bf27-f3dfd6fb1b07/history and then I could check that it appeared in https://2098.hub.asap.science/

Screen Shot 2022-09-28 at 11 28 22


apps/crn-server/test/data-providers/reminders.data-provider.test.ts
```diff
@@ -565,93 +570,124 @@ describe('Reminder Data Provider', () => {
...
+          );
+          expect(result).toEqual({
+            total: 0,
+            items: [],
+          });
         });
-      });
-    });
+      },
+    );
```
peterstarling, line 690: What if both video and presentation material types are updated in the last 24h? I think with your implementation we will just get both of them but it would be nice to write a test for this

peterstarling (changes requested):

peterstarling (approved):

PR #2100, opened 2022-09-28
add project description contact information and keyword

markcsims:
https://asaphub.atlassian.net/browse/GP2-69


packages/gp2-components/src/templates/ProjectOverview.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const contentStyles = css({
+  marginTop: rem(32),
+});
+
+const columnStyles = css({
+  display: 'grid',
+  columnGap: rem(32),
+  gridRowGap: rem(32),
+  [`@media (min-width: ${smallDesktopScreen.min}px)`]: {
```
jhonny-m, line 29: I usually use `crossQuery`

apps/gp2-server/src/data-providers/project.data-provider.ts
```diff
@@ -104,13 +104,21 @@ export function parseProjectToDataObject(
...
-    projectProposalUrl: project.flatData.projectProposal || undefined,
+    id,
+    title: project.title || '',
+    startDate: project.startDate || '',
+    endDate: project.endDate || undefined,
+    status: project.status,
+    projectProposalUrl: project.projectProposal || undefined,
+    pmEmail: project.pmEmail || undefined,
+    leadEmail: project.leadEmail || undefined,
+    description: project.description || undefined,
```
jhonny-m, line 120: I've defaulted the description to `''` in working groups and have it as a mandatory string. not sure which is the right approach, we should check and make both the same

jhonny-m (approved):

apps/gp2-server/src/data-providers/project.data-provider.ts
markcsims: on this, it's not mandatory

packages/gp2-components/src/templates/ProjectOverview.tsx
markcsims: done

PR #2103, opened 2022-09-29
Crn 1028 api and components

lctrt:


apps/crn-server/test/fixtures/users.fixtures.ts
```diff
@@ -95,7 +95,7 @@ export const getGraphQLUser = (
   created: '2020-09-23T20:45:22Z',
   ...user,
   flatData: {
-    alumniSinceDate: '2020-10-22T15:33:18Z',
+    alumniSinceDate: null, //'2020-10-22T15:33:18Z',
```
lctrt, line 98: (I'm aware of this, comment removed in upcoming commit)

packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -63,6 +63,18 @@ const containerStyles = css({
   },
 });
 
+const alumniBadgeStyles = css({});
```
kmaid, line 66: ?

kmaid (approved):

packages/react-components/src/templates/UserProfileHeader.tsx
lctrt: good catch, removing it now :)

PR #2105, opened 2022-09-29
CRN-1072 adding active/inactiveSince fields to Teams

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-1072

This PR adds the field `active` and `inactiveSince` in Team schema

### Active

Name/Label          |  Validation
:-------------------------:|:-------------------------:
| Screen Shot 2022-09-29 at 12 32 34 | Screen Shot 2022-09-29 at 12 32 40

### Inactive Since

Name/Label          |  Validation
:-------------------------:|:-------------------------:
Screen Shot 2022-09-29 at 12 32 52 | Screen Shot 2022-09-29 at 12 33 00 

The field `inactiveSince` is not being updated yet. This will be done in another PR.

### DIFF
--- a/apps/asap-cli/src/import/users/insert.ts
+++ b/apps/asap-cli/src/import/users/insert.ts
@@ -61,6 +61,9 @@ const insertTeam = async (data: Data, cache: Cache): Promise<RestTeam> => {
     applicationNumber: {
       iv: application,
     },
+    active: {
+      iv: true,
+    },
     projectTitle: {
       iv: projectTitle,
     },
--- a/packages/squidex/schema/crn/schemas/teams.json
+++ b/packages/squidex/schema/crn/schemas/teams.json
@@ -7,9 +7,7 @@
       "label": "Teams",
       "validateOnPublish": false
     },
-    "scripts": {
-      "update": "const wasActive = !!ctx.oldData.active.iv\nconst isActive = !!ctx.data.active.iv\n\nconst becameInactive = wasActive && !isActive\n\nconst becameActive = !wasActive && isActive\n\nif (becameInactive) {\n    ctx.data.inactiveSince.iv = new Date().toISOString()\n    replace();\n} else if (becameActive) {\n    ctx.data.inactiveSince.iv = null\n    replace();\n}\n\n"
-    },
+    "scripts": {},
     "fieldsInReferences": ["displayName"],
     "fieldsInLists": [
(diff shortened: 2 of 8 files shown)


apps/asap-cli/src/import/users/insert.ts
```diff
@@ -61,6 +61,9 @@ const insertTeam = async (data: Data, cache: Cache): Promise<RestTeam> => {
     applicationNumber: {
       iv: application,
     },
+    active: {
```
gabiayako, line 64: I just hard coded this true here, but I am not sure if that's the right thing to do 🤔 I saw that above we have

```
  const team = {
    displayName: {
      iv: lastName,
    },
```

which is kind of weird that the team name is the person's last name. I am assuming that the data here is not important.

peterstarling: meh, we dont really use this import anymore anyway, so

packages/squidex/schema/crn/schemas/teams.json
```diff
@@ -46,6 +46,38 @@
           "isHalfWidth": true
         }
       },
+      {
```
peterstarling, line 49: im not so sure about adding these two fields like this, especially without scripting. So this would mean that the user would have to change the team status manually and remember to record the date in another field when they do that, right? I think we should just use the same mechanism we used to alumnis which is that we only have a single field with a date. We would make an assumption that this would only ever be a past date which would mean that to check if the team is active we would just check the present of any date in that field

peterstarling (changes requested):

packages/squidex/schema/crn/schemas/teams.json
peterstarling: ok now i see you've added the script in this PR https://github.com/yldio/asap-hub/pull/2106

See - i still think it should just be a single field

gabiayako: it makes sense :) I'll change the https://github.com/yldio/asap-hub/pull/2106 to have a single field. I'll close this one and update everything in https://github.com/yldio/asap-hub/pull/2106 then. I was trying to break in small PRs to make them easier to review, but I think it got messier

gabiayako:
Closing this one as the PR changes will be sorted in https://github.com/yldio/asap-hub/pull/2106

PR #2106, opened 2022-09-29
CRN-1072 adding inactiveSince field to Teams

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-1072


packages/model/src/team.ts
```diff
@@ -42,6 +45,8 @@ export interface TeamMember {
 
 export type TeamDataObject = Omit<TeamCreateRequest, 'applicationNumber'> & {
   id: string;
+  active: boolean;
+  inactiveSince?: string | null;
```
peterstarling, line 49: I wonder if we need to surface two fields at all, like I mentioned in the other PR I think this could be a single field in the CMS, definitely in the data object and probably even in the response type, although if we need to add a boolean for simplicity we can.

peterstarling (changes requested):

apps/storybook/src/TeamProfilePage.stories.tsx
```diff
@@ -14,6 +14,7 @@ export default {
 const props = (): Omit<ComponentProps<typeof TeamProfilePage>, 'children'> => ({
   id: '42',
   displayName: 'Ramirez, T',
+  inactiveSince: null,
```
kmaid, line 17: Shouldn't this be undefined

gabiayako: I believe when we create a new team and don't input anything in `inactiveSince` or we don't set `inactiveSince` value for existing records, its value is `null`. And the idea is to always bring this field so we can check if the team is active or not :)

I think I misunderstood the comment when I first read 😛 I read as an affirmation not a question so I added this change here https://github.com/yldio/asap-hub/pull/2106/commits/93b9988fb9de2101f24124d1715422280fc21256 so the default value is not null

kmaid: I don't think we should be returning Null as we don't do it for any other response type. We use the typings to say what could be on a entity rather than the presence of null

gabiayako: Oh ignore my comment above 😝 I'll change that shortly

apps/crn-server/test/fixtures/teams.fixtures.ts
```diff
@@ -211,6 +213,7 @@ export const getInputTeam = (): InputTeam['data'] => ({
   displayName: {
     iv: 'Team A',
   },
+  inactiveSince: { iv: null },
```
gabiayako, line 216: setting to null because this is the way it comes from squidex

apps/crn-server/src/data-providers/teams.data-provider.ts
```diff
@@ -128,6 +128,7 @@ export class TeamSquidexDataProvider implements TeamDataProvider {
     const inputTeam: InputTeam['data'] = {
       applicationNumber: { iv: input.applicationNumber },
       displayName: { iv: input.displayName },
+      inactiveSince: { iv: input.inactiveSince || null },
```
gabiayako, line 131: setting to null when it does not have a value because this is the way it comes from squidex

peterstarling (approved):

PR #2115, opened 2022-10-03
[CRN-1043] Alumni badge shared

ricardocarvalh0:
image


packages/react-components/src/icons/alumni-badge.tsx
```diff
@@ -1,18 +1,18 @@
 const alumniBadge = (
   <svg
-    width="24"
-    height="24"
-    viewBox="0 0 24 24"
+    width="16"
```
oratiu, line 3: All the icons should be 24x24, I don't think you should have changed it based on Figma. 
By changing this, all the places where alumni-badge is used will also be affected.
image

packages/react-components/src/molecules/UsersList.tsx
```diff
@@ -38,7 +38,7 @@ const itemStyles = css({
 const userStyles = css({
   overflow: 'hidden',
   display: 'grid',
-  gridTemplateColumns: 'min-content 1fr',
+  gridTemplateColumns: 'min-content 1fr 0.5fr',
```
oratiu, line 41:
```suggestion
  gridTemplateColumns: 'min-content 1fr min-content',
```
Why did you use a fraction, the first icon used min-content.

packages/react-components/src/molecules/__tests__/UsersList.test.tsx
```diff
@@ -54,6 +54,53 @@ it('does not link external users', () => {
...
+    const { getByText } = render(
+      <UsersList
+        max={1}
+        users={Array.from({ length: 1 }).map(
+          () =>
+            ({
+              ...createUserResponse(),
+              displayName: 'John Doe',
+              alumniSinceDate: new Date(2021, 6, 12, 14, 32).toISOString(),
+            } as ExternalAuthorResponse),
```
oratiu, line 68: Why "as ExternalAuthorResponse" when the user is a CRN one?
And why use Array.from instead of ` users={[
        {
          ...createUserResponse(),
          displayName: 'John',
          id: '42',
         alumniSinceDate: new Date(etc.),
        },
      ]}`

packages/react-components/src/molecules/__tests__/UsersList.test.tsx
```diff
@@ -54,6 +54,53 @@ it('does not link external users', () => {
...
+  });
+  it('does not show alumni badge if user is not alumni', () => {
+    const { queryByText } = render(
+      <UsersList
+        max={1}
+        users={Array.from({ length: 1 }).map(
+          () =>
+            ({
+              displayName: 'John Doe',
+            } as ExternalAuthorResponse),
```
oratiu, line 82: Here as well, why "as ExternalAuthorResponse"

packages/react-components/src/icons/alumni-badge.tsx
ricardocarvalh0: I thought we would have to ignore the whitespace around it since it's adding 8px of margin on it. I tested other places and it looked ok. Will revert it though. Thanks
image

ricardocarvalh0:
Closing this as an attempt to fix schema issue

PR #2116, opened 2022-10-03
feat: CRN-1033 Prevent Alumni from accessing the hub

peterstarling:
- add alumni rule to add-user-metadata webhook
- fix the auth0 rule invalid import
- change the error message for alumni error

### DIFF
--- a/apps/auth0-rules/src/__tests__/connect-user.test.ts
+++ b/apps/auth0-rules/src/__tests__/connect-user.test.ts
@@ -1,6 +1,5 @@
 import nock from 'nock';
 import connectUser from '../connect-user';
-import * as handleError from '../handle-error';
 import type { RuleContext, User } from '../types';
 
 declare global {
@@ -87,6 +86,8 @@ describe('Auth0 Rule - Connect User', () => {
   const apiSharedSecret = 'auth0_shared_secret';
   const invitation_code = 'sampleInvitationCode';
 
+  const cb: jest.MockedFunction<Parameters<typeof connectUser>[2]> = jest.fn();
+
   beforeEach(() => {
     global.configuration = {
       APP_DOMAIN: appDomain,
@@ -96,11 +97,13 @@ describe('Auth0 Rule - Connect User', () => {
     nock.cleanAll();
   });
 
-  it('should callback with same user + context if receives no invitation_code', () => {
-    const cb: jest.MockedFunction<Parameters<typeof connectUser>[2]> =
-      jest.fn();
+  afterEach(() => {
+    cb.mockClear();
+  });
 
+  it('should callback with same user + context if receives no invitation_code', () => {
     connectUser(user, context, cb);
+
     expect(cb).toHaveBeenCalled();
     const [err, resUser, resContext] = cb.mock.calls[0];
     expect(err).toBeFalsy();
@@ -109,7 +112,6 @@ describe('Auth0 Rule - Connect User', () => {
   });
 
   it('should return an error if fails to connect the user', async () => {
-    const spy = jest.spyOn(handleError, 'handleError');
     nock(apiURL, {
       reqheaders: {
         authorization: `Basic ${apiSharedSecret}`,
@@ -121,21 +123,17 @@ describe('Auth0 Rule - Connect User', () => {
       })
       .reply(404);
 
-    const cb: jest.MockedFunction<Parameters<typeof connectUser>[2]> =
-      jest.fn();
-
     await connectUser(
       user,
       { ...context, request: { query: { invitation_code }, body: {} } },
       cb,
     );
 
-    expect(cb).toHaveBeenCalled();
+    expect(cb).toHaveBeenCalledWith(expect.any(Error));
     const [err, resUser, resContext] = cb.mock.calls[0];
     expect(err).not.toBeNull();
     expect(resUser).toBeUndefined();
     expect(resContext).toBeUndefined();
-    expect(spy).toHaveBeenCalled();
   });
 
   it('should connect user if receives an invitation_code', async () => {
@@ -150,9 +148,6 @@ describe('Auth0 Rule - Connect User', () => {
       })
       .reply(202);
 
-    const cb: jest.MockedFunction<Parameters<typeof connectUser>[2]> =
-      jest.fn();
-
     await connectUser(
       user,
       { ...context, request: { query: { invitation_code }, body: {} } },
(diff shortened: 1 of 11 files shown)


apps/auth0-rules/src/__tests__/connect-user.test.ts
```diff
@@ -96,11 +97,13 @@ describe('Auth0 Rule - Connect User', () => {
     nock.cleanAll();
   });
 
-  it('should callback with same user + context if receives no invitation_code', () => {
-    const cb: jest.MockedFunction<Parameters<typeof connectUser>[2]> =
-      jest.fn();
+  afterEach(() => {
+    cb.mockClear();
```
markcsims, line 101: I saw this in apps/auth0-rules/src/__tests__/add-user-metadata.test.ts also but should we be using mockClear or mockReset?

I guess they will both set it back to jest.fn()?

apps/auth0-rules/src/connect-user.ts
```diff
@@ -1,7 +1,13 @@
...
-import { handleError } from './handle-error';
 import type { Rule } from './types';
 
+const handleError = (err: unknown): Error => {
+  if (err instanceof Error) {
+    return err;
+  }
+  return new Error('Unexpected Error');
+};
+
```
markcsims, line 10: Why can't we use the import?

apps/gp2-frontend/src/auth/Signin.tsx
```diff
@@ -36,7 +36,12 @@ const Signin: React.FC<Record<string, never>> = () => {
       <UtilityBar>
         <WelcomePage
           onClick={signin}
-          authFailed={searchParams.has('state') && searchParams.has('error')}
+          authFailed={
+            (searchParams.has('state') &&
+              searchParams.has('error') &&
+              'invalid') ||
+            undefined
```
markcsims, line 43: Is this the same as this ternary

```suggestion
            (searchParams.has('state') &&
              searchParams.has('error')) ? 'invalid' :  undefined
```

packages/react-components/src/templates/__tests__/WelcomePage.test.tsx
```diff
@@ -1,28 +1,33 @@
...
 import { noop } from '../../utils';
 
 it('renders the signin page', () => {
   const handleClick = jest.fn();
-  const { getByRole } = render(<WelcomePage onClick={handleClick} />);
-  expect(getByRole('button').textContent).toMatchInlineSnapshot(`"Sign in"`);
+  render(<WelcomePage onClick={handleClick} />);
+  expect(screen.getByRole('button').textContent).toMatchInlineSnapshot(
+    `"Sign in"`,
+  );
```
markcsims, line 11:
```suggestion
    expect(screen.getByRole('button', { name: /Sign in/ })).toBeVisible();
```

packages/react-components/src/templates/__tests__/WelcomePage.test.tsx
```diff
@@ -1,28 +1,33 @@
...
   const handleClick = jest.fn();
-  const { getByRole } = render(
-    <WelcomePage allowSignup onClick={handleClick} />,
-  );
-  expect(getByRole('button').textContent).toMatchInlineSnapshot(
+  render(<WelcomePage allowSignup onClick={handleClick} />);
+  expect(screen.getByRole('button').textContent).toMatchInlineSnapshot(
     `"Activate account"`,
   );
 });
```
markcsims, line 20:
```suggestion
  expect(
    screen.getByRole('button', { name: /Activate account/ }),
  ).toBeVisible();
```

packages/react-components/src/templates/__tests__/WelcomePage.test.tsx
```diff
@@ -1,28 +1,33 @@
...
 
 it('shows an auth failed error message', () => {
-  const { queryByText, rerender } = render(<WelcomePage onClick={noop} />);
-  expect(queryByText(/problem/i)).not.toBeInTheDocument();
+  render(<WelcomePage onClick={noop} />);
+  expect(screen.queryByText(/problem/i)).not.toBeInTheDocument();
+
+  render(<WelcomePage authFailed={'invalid'} onClick={noop} />);
+  expect(screen.queryByText(/problem/i)).toBeVisible();
```
markcsims, line 27:
```suggestion
  expect(screen.getByText(/problem/i)).toBeVisible();
```

packages/react-components/src/templates/__tests__/WelcomePage.test.tsx
```diff
@@ -1,28 +1,33 @@
...
+});
 
-  rerender(<WelcomePage authFailed onClick={noop} />);
-  expect(queryByText(/problem/i)).toBeVisible();
+it('shows an alumni no-access error message', () => {
+  render(<WelcomePage authFailed={'alumni'} onClick={noop} />);
+  expect(screen.queryByText(/alumni/i)).toBeVisible();
```
markcsims, line 32:
```suggestion
 expect(screen.getByText(/problem/i)).toBeVisible();
```

markcsims (changes requested):

apps/auth0-rules/src/connect-user.ts
peterstarling: cause auth0 does allow us to do that and we dont have a setup for bundling all into one file - we could do it now but we have a ticket to revamp the rules at some point so

apps/auth0-rules/src/__tests__/connect-user.test.ts
peterstarling: here we probably could but in the other one we had other mocks too which we didnt want to clear between tests, i think its better to be more specific here

apps/gp2-frontend/src/auth/Signin.tsx
peterstarling: yeah i think so and i prefer ternary too, i just thought in FE we wanted to go all FP and use these instead of ternaries

packages/react-components/src/templates/__tests__/WelcomePage.test.tsx
peterstarling: i dont think this is what we are looking for? we want to make sure this is an alumni specific error message

markcsims: typo, the point I'm trying to make is to use getByText and not queryByText. queryByText should only be used when checking for non-existence.

```suggestion
  expect(screen.getByText(/alumni/i)).toBeVisible();
```

PR #2128, opened 2022-10-05
CRN-1083 - Change from Contentful Schema to Migrations

m-doughty:
# CRN-1083 - Change from Contentful Schema to Migrations

## This PR handles

- Adding `contentful-migrate`;
- All migrations to mirror what we had in schema;
- Removing the schema;
- Adding Contentful variables to `.env.example`;
- Documenting how to add and run migrations;
- Yarn convenience methods to run the migrations.
- GitHub action to run migrations on merge to `master`.

## Testing

- Set all env variables to point at the `Migration` environment (see README);
- Delete all content types from `Migration` environment on Contentful;
- Run `yarn workspace @asap-hub/contentful ctf-migrate init` to add the `Migrations` back;
- Run `yarn workspace @asap-hub/contentful space:migrate:crn` to run the migrations.

## Additionally

- Merging to master should run the migrations on `Production` and `Development`;
- I have already run `init` on these envs.


package.json
```diff
@@ -125,6 +127,7 @@
     "babel-plugin-transform-inline-environment-variables": "0.5.0-alpha.42",
     "babel-preset-react-app": "10.0.1",
     "bundlewatch": "0.3.3",
+    "contentful-migrate": "^0.17.0",
```
peterstarling, line 130: i wouldnt add it at the root, similar to how we should not have squidex or algolia deps anywhere outside of the squidex/algolia package

package.json
```diff
@@ -27,6 +27,8 @@
     "build:typecheck": "tsc -b -v $(node scripts/get-composite-ts-projects.js)",
     "check:packages": "yarn install --immutable --immutable-cache",
     "clean": "rimraf \"{apps,packages}/*/{build*,*.tsbuildinfo}\" \"coverage\" \".serverless*\"",
+    "contentful:migrate:dryrun": "yarn ctf-migrate up --all --dry-run",
```
peterstarling, line 30: I'd expose these inside the contentful package

packages/contentful/migrations/externalAuthors/20221005103626-create-external-authors.js
```diff
@@ -0,0 +1,89 @@
+module.exports.description = 'Create content model for External Authors';
```
peterstarling, line 1: as specified in the ticket - at this stage we are only supposed to create a POC entity. I think what you did was convert all the existing content types into migrations but you shouldn't have. They were just there as a POC too and should be removed. 

Why don't you just keep one migration checked in for, lets say, news (which we know we are going to create next)?

packages/contentful/package.json
```diff
@@ -16,6 +16,7 @@
     "schema:update:crn": "graphql-codegen --config codegen.yml"
   },
   "dependencies": {
+    "contentful-migrate": "^0.17.0",
```
peterstarling, line 19: this is fine i just wanted to point out that this is a duplicate of what you have in the root package.json, but as mentioned earlier it should be the only place you add it

peterstarling (changes requested):

package.json
m-doughty: Done.

package.json
m-doughty: Done.

packages/contentful/package.json
m-doughty: This is removed from root package.

packages/contentful/package.json
```diff
@@ -11,11 +11,12 @@
   "scripts": {
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
-    "space:export:crn": "contentful space export --skip-content=true --space-id=$CONTENTFUL_SPACE_ID --environment-id=$CONTENTFUL_ENVIRONMENT --content-file=./schema/crn-schema.json",
-    "space:import:crn": "contentful space import --content-model-only=true --space-id=$CONTENTFUL_SPACE_ID --environment-id=$CONTENTFUL_ENVIRONMENT --content-file=./schema/crn-schema.json",
+    "space:migrate:crn": "yarn ctf-migrate up --all",
```
peterstarling, line 14: this is fine but it will cause the github actions to fail on master - see:
https://github.com/yldio/asap-hub/blob/af464a8f1d30ea0ef072ad6683ac7d3690f5f106/.github/workflows/reusable-deployment.yml#L218

.env.example
```diff
@@ -20,4 +20,6 @@ REACT_APP_GP2_AUTH0_CLIENT_ID=sUSMSrBS3FMhb58bxiwQZOTPbZMPWIje
 REACT_APP_GP2_AUTH0_DOMAIN=dev-gp2-hub.us.auth0.com
 SQUIDEX_BASE_URL=https://cloud.squidex.io
 CONTENTFUL_SPACE_ID=5v6w5j61tndm
-CONTENTFUL_ENVIRONMENT=dev
+CONTENTFUL_ENV_ID=Development
+CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=
+CONTENTFUL_MIGRATIONS_DIR=/Users/me/code/asap-hub/packages/contentful/migrations
```
peterstarling, line 25: I don't think we want an absolute path to migrations passed like this? I would add the relative path to the yarn command

peterstarling (changes requested):

.env.example
m-doughty: Yeah I can remove this now that everything is in `contentful` package. Done.

m-doughty:
Sigh.

PR #2141, opened 2022-10-06
feat: CRN-1033 Prevent alumni from accessing the API

peterstarling:
- prevent alumni from accessing the hub
- update onboarding permission check


kmaid (approved):

kmaid (changes requested):
I think we should discuss this with tiffiny

peterstarling:
> I think we should discuss this with tiffiny

@kmaid this behaviour was accepted by Tiffany

PR #2150, opened 2022-10-10
CRN-1049 Migration script to move data to Tutorials

gabiayako:


apps/crn-server/src/migrations/1665394546-move-type-tutorial-news-to-entity-tutorials.ts
```diff
@@ -0,0 +1,64 @@
...
+        } catch (err) {
+          console.log('error: ', err);
+        }
+      }
+
+      pointer += 10;
+    } while (pointer < result.total);
+  };
+
+  down = async (): Promise<void> => {};
```
gabiayako, line 63: do we want a rollback script in this case? also, what happens if the script fails in the middle? do we want to save the migrated news ids somewhere?

peterstarling: the migrated news will be whatever the record where moved into tutorials, to be fair the rollback is there but i dont think we ever used it, but if you wanted to feel safe i'd say just move all records from tutorials back to news and delete them from tutorials. Or leave it empty, whatever, but then make sure to have a backup before the migration. In fact - have it anyway

peterstarling (changes requested):

peterstarling (approved):

PR #2151, opened 2022-10-10
adds members card to gp2 projects

markcsims:
https://asaphub.atlassian.net/browse/GP2-68


apps/gp2-server/test/data-providers/user.data-provider.test.ts
```diff
@@ -122,16 +170,68 @@ describe('User data provider', () => {
...
+      region                      | expected
+      ${'Africa'}                 | ${UsersDataRegionEnum.Africa}
+      ${'Asia'}                   | ${UsersDataRegionEnum.Asia}
+      ${'Australia/Australiasia'} | ${UsersDataRegionEnum.AustraliaAustraliasia}
+      ${'Europe'}                 | ${UsersDataRegionEnum.Europe}
+      ${'North America'}          | ${UsersDataRegionEnum.NorthAmerica}
+      ${'South America'}          | ${UsersDataRegionEnum.SouthAmerica}
+      ${'Latin America'}          | ${UsersDataRegionEnum.LatinAmerica}
+    `(
+      'Should update the region and last name through a clean-update $region => $expected',
```
peterstarling, line 183: is there a reason why you're testing the update of region together with the name? I would test the region and the role individually, then have one test to verify that they can be all updated together too

apps/gp2-server/test/data-providers/user.data-provider.test.ts
```diff
@@ -60,36 +64,80 @@ describe('User data provider', () => {
...
+
+    test.each`
+      role                                         | expected
+      ${UsersDataRoleEnum.WorkingGroupParticipant} | ${'Working Group Participant'}
+      ${UsersDataRoleEnum.NetworkInvestigator}     | ${'Network Investigator'}
+      ${UsersDataRoleEnum.NetworkCollaborator}     | ${'Network Collaborator'}
+      ${UsersDataRoleEnum.Administrator}           | ${'Administrator'}
+      ${UsersDataRoleEnum.Trainee}                 | ${'Trainee'}
+    `(
+      'Should correctly map regions $role => $expected',
```
peterstarling, line 130:
```suggestion
      'Should correctly map roles $role => $expected',
```

apps/gp2-server/test/data-providers/user.data-provider.test.ts
```diff
@@ -60,36 +64,80 @@ describe('User data provider', () => {
...
         userDataProvider.fetchById('user-id'),
-      ).rejects.toThrowErrorMatchingInlineSnapshot('"Invalid user region"');
+      ).rejects.toThrowErrorMatchingInlineSnapshot(
+        '"Region not defined: null"',
+      );
     });
     test('Should correctly map MD, PhD Correctly', async () => {
       const degreeUser = getGraphQLUser();
-      degreeUser.flatData.degree = ['MD_PhD'] as UsersDataDegreeEnum[];
+      degreeUser.flatData.degree = [UsersDataDegreeEnum.MdPhD];
```
peterstarling, line 92: might as well update this test to use an `each` too?

apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -106,13 +115,8 @@ export function parseWorkingGroupToDataObject(
...
         member: GraphQLWorkingGroupMember,
       ) => {
-        if (!isWorkingGroupMemberRole(member.role)) {
-          throw new Error(
-            `Invalid working group role on members : ${member.role}`,
-          );
-        }
         const user = member.user && member.user[0];
-        if (!user) {
+        if (!(user && member.role)) {
```
peterstarling, line 119: i believe this is tested here right:
https://github.com/yldio/asap-hub/pull/2151/files#diff-b7cd257b868f623e71c6e1e34dab897113804b4842635229175e8df46b80287cR143
and what about a missing user, do you know?

Anywy i feel like this line could just be 
```suggestion
        if (!user || !member.role) {
```

Which to make makes more sense - it reads something like `if there is no user or no role`

peterstarling (approved):

apps/gp2-server/test/data-providers/user.data-provider.test.ts
markcsims: I think this is how the tests were copied over.

PR #2152, opened 2022-10-10
CRN-1041 add new alumni filters

lctrt:


packages/react-components/src/organisms/CheckboxGroup.tsx
```diff
@@ -25,16 +26,24 @@ export default function CheckboxGroup<V extends string>({
...
+          </Caption>
+        ) : (
+          <LabeledCheckbox
+            key={`${groupName}-${index}`}
+            groupName={groupName.current}
+            title={option.label}
+            enabled={option.enabled}
+            checked={values.has(option.value)}
+            onSelect={() =>
+              option.value !== undefined && onChange(option.value)
```
kmaid, line 42: I don't think you need the undefined check you have already done it above
```suggestion
onChange(option.value)
```

packages/react-components/src/organisms/CheckboxGroup.tsx
```diff
@@ -3,16 +3,21 @@ import { v4 as uuidV4 } from 'uuid';
...
 
-interface CheckboxGroupProps<V extends string> {
-  readonly options: ReadonlyArray<Option<V>>;
+export interface Title {
+  title: string;
+  label?: undefined;
+}
 
+interface CheckboxGroupProps<V extends string> {
+  readonly options: ReadonlyArray<Option<V> | Title>;
```
kmaid, line 20: I think this type is the one that needs to be exported and the other two shouldn't need to be. Then use it where i mark below

packages/react-components/src/organisms/Filter.tsx
```diff
@@ -54,26 +55,23 @@ const showMenuStyles = css({
 interface FilterProps<V extends string> {
   readonly filters?: Set<V>;
   readonly onChangeFilter?: (filter: V) => void;
-  readonly filterOptions: Option<V>[];
-  readonly filterTitle: string;
+  readonly filterOptions: ReadonlyArray<Option<V> | Title>;
```
kmaid, line 58: here

packages/react-components/src/organisms/Filter.tsx
```diff
@@ -54,26 +55,23 @@ const showMenuStyles = css({
...
   filters = new Set(),
   onChangeFilter = noop,
   filterOptions,
-  filterTitle,
 }: FilterProps<V>): ReturnType<React.FC> {
   const [menuShown, setMenuShown] = useState(false);
   useEffect(() => {
     setMenuShown(false);
-  }, [filterTitle]);
+  }, [filterOptions]);
```
kmaid, line 68: Not sure this actually needs anything. I think on mount we just want to ensure the menu is closed
```suggestion
  }, []);
```

packages/react-components/src/organisms/Filter.tsx
```diff
@@ -54,26 +55,23 @@ const showMenuStyles = css({
...
 
   const [debouncedFilters] = useDebounce(filters, 5000);
   useEffect(() => {
     if (filters.size && filters === debouncedFilters) {
       window.dataLayer?.push({
         [FILTERS_KEY]: [...filters],
```
kmaid, line 74: This is for analytics. Not sure if we should drop this or need to do something else

packages/react-components/src/organisms/SearchAndFilter.tsx
```diff
@@ -16,10 +17,9 @@ const styles = css({
 
 type SearchFieldProps = ComponentProps<typeof SearchField>;
 interface SearchAndFilterProps<V extends string> {
-  readonly filterTitle: ComponentProps<typeof Filter>['filterTitle'];
   readonly filters?: Set<V>;
   readonly onChangeFilter?: (filter: V) => void;
-  readonly filterOptions: Option<V>[];
+  readonly filterOptions: ReadonlyArray<Option<V> | Title>;
```
kmaid, line 22: here too

packages/react-components/src/organisms/__tests__/Filter.test.tsx
```diff
@@ -20,23 +22,15 @@ it('shows and hides the dropdown menu', () => {
   expect(getByText('Filter by Stuff')).not.toBeVisible();
 });
 
-it('hides the dropdown menu when the title changes', () => {
+it('hides the dropdown menu when the options changes', () => {
```
kmaid, line 25: This seems like a weird behaviour xD. I am going to regret looking at blame

packages/react-components/src/organisms/__tests__/CheckboxGroup.test.tsx
```diff
@@ -18,6 +18,13 @@ it('renders a checkbox for each option', () => {
   expect(getByLabelText('Blue')).toBeVisible();
 });
 
+it('renders title', () => {
```
kmaid, line 21: Can we test this a little better. Im looking for a series of titles and options and that it comes out in the right order

kmaid (approved):

PR #2158, opened 2022-10-11
CRN-1049 Removing date field from News

gabiayako:
Part of the ticket: https://asaphub.atlassian.net/browse/CRN-1049

In the ticket, it is written `Remove - Event Date and Type fields from the News schema`, but it's not possible yet to remove type from News because there isn't a way to differentiate `Working Groups` from `News` without this field yet.

It is possible to remove event date though, as this field doesn't even come to the frontend https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/entities/news.ts (`created` is not the same field as `date`).


gabiayako:
rebase to master

PR #2162, opened 2022-10-12
adds properties to users schema

markcsims:
https://asaphub.atlassian.net/browse/GP2-122


packages/squidex/schema/gp2/schemas/users.json
```diff
@@ -202,6 +202,123 @@
...
+          "contentType": "Unspecified",
+          "editor": "Input",
+          "label": "Country",
+          "isRequired": true,
+          "isRequiredOnPublish": false,
+          "isHalfWidth": true
+        }
+      },
+      {
+        "name": "city",
```
peterstarling, line 226: hang on, didnt we say location? Although I would much prefer having it as a city, but if we are going to show it as a Location in FE then maybe we want to have it consistent all across the stack

apps/gp2-server/src/data-providers/user.data-provider.ts
```diff
@@ -198,6 +198,17 @@ export const parseGraphQLUserToDataObject = ({
       return degree;
     });
 
+  const positions =
+    item.positions?.map(({ role, department, institution }) => {
```
peterstarling, line 202: i was just wondering if a forEach wouldnt be enough? TypeScript might moan but technically it is more fit for the validation that does not actually map anything.

peterstarling (changes requested):
Minor comments.

I'd say about 50% of this PR is changing the user fixture from the famous actor to a superhero character

peterstarling (approved):

peterstarling:
sorry im stupid i meant to approve it the first time round

PR #2164, opened 2022-10-12
Crn 1040 Alumni user profile message

lctrt:


apps/crn-server/test/fixtures/users.fixtures.ts
```diff
@@ -96,6 +96,7 @@ export const getGraphQLUser = (
   ...user,
   flatData: {
     alumniSinceDate: null,
+    alumniLocation: null,
```
peterstarling, line 99: if there no tests for this in this PR and the only fixture change is the addition of a null `alumniLocation` to the fixture I think we need a test for when it comes back from squidex as not null, no?

peterstarling (changes requested):

peterstarling (approved):

PR #2170, opened 2022-10-13
GP2 96 Storybook

csbarbosa:
(Ticket:https://asaphub.atlassian.net/browse/GP2-96)
Added GP2 components into storybook


apps/storybook/src/gp2-components/IconWithLabel.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import projectIcon from '@asap-hub/gp2-components/src/icons/project-icon';
+import IconWithLabel from '@asap-hub/gp2-components/src/molecules/IconWithLabel';
+import { text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'GP2 / Molecules / Icon With Label',
+  component: IconWithLabel,
+};
+export const Logo = () => (
```
markcsims, line 9: Should be Normal
```suggestion
export const Normal = () => (
```

apps/storybook/src/gp2-components/HeaderLogo.stories.tsx
```diff
@@ -0,0 +1,12 @@
...
+export default {
+  title: 'GP2 / Molecules / Header Logo',
+  component: HeaderLogo,
+};
+
+const props = {
+  logoHref: '/',
+};
+
+export const Logo = () => <HeaderLogo {...props} />;
```
markcsims, line 12:
```suggestion
export const Normal = () => <HeaderLogo {...props} />;
```

apps/storybook/src/gp2-components/ExpandableText.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+  title: 'GP2 / Molecules / Expandable Text',
+  component: ExpandableText,
+};
+
+const props = {
+  description:
+    'In order to show you the Hub, we will need to make your profile public to the Hub network. Would you like to continue?',
+};
+
+export const Text = () => (
```
markcsims, line 15:
```suggestion
export const Normal = () => (
```

apps/storybook/src/gp2-components/UserCardInfo.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+export default {
+  title: 'GP2 / Molecules / User Card Info',
+  component: UserCardInfo,
+};
+
+type UserCardInfoProps = Pick<ComponentProps<typeof UserRegion>, 'region'> & {
+  workingGroups: { id: string; name: string }[];
+  projects: { id: string; name: string }[];
+  role: string;
+};
```
markcsims, line 20: Looks wrong to be defining the props here. 

maybe we need something like:

```
ComponentProps<typeof UserCardInfo>
```

markcsims (changes requested):

apps/storybook/src/gp2-components/IconWithLabel.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  <IconWithLabel icon={projectIcon}>
+    {text(
+      'description',
```
jhonny-m, line 12: I would rename this to `label` and the text could just be "label", doesn't make too much sense being a big text

apps/storybook/src/gp2-components/NavigationLink.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { text } from '@storybook/addon-knobs';
+
+export default {
+  title: 'GP2 / Molecules / Navigation Link',
+  component: IconWithLabel,
+};
+export const Logo = () => (
+  <NavigationLink href={text('href', 'https://www.google.com')} icon={roleIcon}>
+    {text(
+      'description',
```
jhonny-m, line 13: `text` and maybe a "text" as the value

apps/storybook/src/gp2-components/UserCardInfo.stories.tsx
jhonny-m: we shouldn't really need to type cast it tho

jhonny-m: is there any particular prop that it complains about?

markcsims (changes requested):

PR #2172, opened 2022-10-13
CRN-1070 Adding filter to groups list view

gabiayako:
JIRA ticket: https://asaphub.atlassian.net/browse/CRN-1070

### DIFF
--- a/apps/crn-frontend/src/network/__tests__/Network.test.tsx
+++ b/apps/crn-frontend/src/network/__tests__/Network.test.tsx
@@ -11,6 +11,7 @@ import { MemoryRouter, Route } from 'react-router-dom';
 import { RecoilRoot } from 'recoil';
 import Network from '../Network';
 import { getTeams } from '../teams/api';
+import { getGroups } from '../groups/api';
 import { useUsers } from '../users/state';
 
 jest.mock('../users/state', () => ({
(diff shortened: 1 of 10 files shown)


apps/crn-server/src/data-providers/groups.data-provider.ts
```diff
@@ -69,6 +69,10 @@ export class GroupSquidexDataProvider implements GroupDataProvider {
       if (options.filter.userId) {
         filterList.push({ 'data/leaders/iv/user': options.filter.userId });
       }
+
+      if ('active' in options.filter) {
```
gabiayako, line 73: I could not use

```suggestion
      if (options.filter.active) {
```

because active can come as `false` then it won't enter this condition

apps/crn-frontend/src/network/__tests__/Network.test.tsx
```diff
@@ -162,6 +164,30 @@ it('allows selection of filters', async () => {
...
+  userEvent.click(checkbox);
+  expect(checkbox).toBeChecked();
+  await waitFor(() =>
+    expect(mockGetGroups).toHaveBeenLastCalledWith(
+      {
+        currentPage: 0,
+        filters: new Set(['Active']),
+        pageSize: 10,
+        searchQuery: '',
+      },
```
kmaid, line 185: consider using expect.ObjectMatching to make your tests more robust if we add things to requests
```suggestion
      expect.objectMatching({
        filters: new Set(['Active']),
      },
```

apps/crn-frontend/src/network/__tests__/Network.test.tsx
```diff
@@ -162,6 +164,30 @@ it('allows selection of filters', async () => {
...
+      'Bearer access_token',
```
kmaid, line 186:
```suggestion
      expect.anything()
```

kmaid (approved):

PR #2173, opened 2022-10-14
feat/Working Group resources tab

markcsims:
https://asaphub.atlassian.net/browse/GP2-60

- [x] Each working group has resources tab next to Overview tab
- [x] Resources tab only visible to users that are a part of said Working group
- [x] Resources tab has a list of all added resources with show more functionality on the page if more than X resources have been added
- [x] If resource added is a link, resource has generic link icon
- [x] If resource is not a link, no icon is added
- [x] If resource is a link, it has a View button
- [x] View button takes user to an external site


apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -107,68 +111,186 @@ describe('Working Group Data Provider', () => {
...
+        const title = 'working group title';
+        const description = 'working group description';
+        const resource = getGraphQLWorkingGroupResource();
+        resource.type = WorkingGroupsDataResourcesTypeEnum.Note;
+        resource.title = title;
+        resource.description = description;
+        resource.externalLink = 'some external link';
+        workingGroup.flatData.resources = [resource];
+        const { resources } = parseWorkingGroupToDataObject(workingGroup);
+        expect(resources).toEqual([
```
peterstarling, line 210: I think you don't want an `externalLink: undefined` here either in which case I would suggest 
```suggestion
        expect(resources).toStrictEqual([
```

packages/model/src/gp2/working-group.ts
```diff
@@ -15,6 +15,16 @@ export type WorkingGroupMember = {
   avatarUrl?: string;
 };
 
+const workingGroupResourceType = ['Link', 'Note'] as const;
+export type WorkingGroupResourceType = typeof workingGroupResourceType[number];
+
+export type WorkingGroupResource = {
```
peterstarling, line 21: as per https://github.com/yldio/asap-hub/blob/87ad25cbbd8dccb272e13c00126eb84eaa5dcf79/apps/gp2-server/test/data-providers/working-group.data-provider.test.ts#L199
the WorkingGroup data provider should not return externalLinks for the WorkingGroup of type 'Note' and I believe your types should reflect it. 

I wouldn't mind having the data provider return whatever shit comes out squidex but since you are already making sure this doesn't happen why not have it typed correctly.

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -107,68 +111,186 @@ describe('Working Group Data Provider', () => {
...
+    describe('resources', () => {
+      test('should map a resource note', () => {
+        const workingGroup = getGraphQLWorkingGroup();
+        const title = 'working group title';
+        const description = 'working group description';
+        const resource = getGraphQLWorkingGroupResource();
+        resource.type = WorkingGroupsDataResourcesTypeEnum.Note;
+        resource.title = title;
+        resource.description = description;
+        workingGroup.flatData.resources = [resource];
```
peterstarling, line 189: have you considered having a fixture of a working group with a resource available by default?

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
peterstarling: or if you don't mind having undefined - even better, I would then just go for 
```suggestion
        expect(resources.externalLink).toBeUndefined()
```

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
```diff
@@ -107,68 +111,186 @@ describe('Working Group Data Provider', () => {
...
+          },
+        ]);
+      });
+      test('should ignore an external link for a resource note', () => {
+        const workingGroup = getGraphQLWorkingGroup();
+        const title = 'working group title';
+        const description = 'working group description';
+        const resource = getGraphQLWorkingGroupResource();
+        resource.type = WorkingGroupsDataResourcesTypeEnum.Note;
+        resource.title = title;
```
peterstarling, line 205: do you really need all these defined in every test? If you follow my suggestion from above than her all you have to do is just add the link

peterstarling: the fact that each of these tests has all this setup repeated makes it difficult to understand what we are testing

apps/gp2-server/test/controllers/working-group.controller.test.ts
```diff
@@ -32,10 +32,37 @@ describe('Working Group controller', () => {
         total: 0,
         items: [],
       });
-      const result = await workingGroupController.fetch();
+      const result = await workingGroupController.fetch('11');
 
       expect(result).toEqual({ items: [], total: 0 });
     });
+
+    test('Should removes the resource if the user is not a member of the working group', async () => {
```
peterstarling, line 40: I think I would like to see another test (and it could be the only test really) which should show that a user needs to be the same user of that in the WG members list. In here the list is empty so its not exactly clear

apps/gp2-server/test/controllers/working-group.controller.test.ts
```diff
@@ -46,17 +73,37 @@ describe('Working Group controller', () => {
...
-      const result = await workingGroupController.fetchById('working-group-id');
+      const result = await workingGroupController.fetchById(
+        'working-group-id',
+        '11',
+      );
 
       expect(result).toEqual(getWorkingGroupResponse());
     });
+
+    test('Should not return the resources when the user is not part of the team', async () => {
```
peterstarling, line 92: same as above - add a member, just one that is not the same as the user who is fetching the data

peterstarling (changes requested):

apps/gp2-server/test/data-providers/working-group.data-provider.test.ts
markcsims: done

apps/gp2-server/test/controllers/working-group.controller.test.ts
markcsims: done

peterstarling (approved):

PR #2174, opened 2022-10-14
[CRN-1115] fix: User frontend regressions

ricardocarvalh0:
- Link is missing for user names on Users List

- Alumni badge should be below user’s name at mobile version of User Profile


packages/react-components/src/molecules/UsersList.tsx
```diff
@@ -74,27 +74,27 @@ const UsersList: FC<UsersListProps> = ({
...
           ) : (
             <div css={userStyles}>
-              {link ? (
-                <ImageLink link={link}>
-                  <Avatar {...user} imageUrl={getPlaceholderAvatarUrl()} />
-                </ImageLink>
-              ) : (
-                <Avatar {...user} imageUrl={getPlaceholderAvatarUrl()} />
-              )}
+              <Avatar {...user} imageUrl={imageUrl} />
```
gabiayako, line 97: This change removes the possibility of not internal users clicking in the avatar that it's currently possible, right? Is this intentional?

In dev
clickable-avatar

kmaid: I think its a bug. External users don't have a user profile

ricardocarvalh0: Exactly. Confirmed with Manoel.

gabiayako (approved):

PR #2178, opened 2022-10-17
feat: CRN-944 Split News Controller

peterstarling:
implement the data provider

change calendar default import to a named one


apps/crn-server/src/data-providers/news.data-provider.ts
```diff
@@ -0,0 +1,127 @@
...
+    : undefined,
+  title: item.data.title.iv,
+  type: item.data.type.iv,
+});
+
+const notTutorialFilter: Filter = {
+  path: 'data.type.iv',
+  op: 'ne',
+  value: 'Tutorial',
+};
```
gabiayako, line 77: I think we can remove this now, right? Also remove from the query 😄  but I believe we could leave it to another PR if you prefer

gabiayako (approved):

PR #2180, opened 2022-10-17
Teams Tabbed Card

kmaid:


packages/react-components/src/organisms/TeamsTabbedCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+    activeTabIndex={1}
+    tabs={[
+      {
+        tabTitle: 'Active Teams (0)',
+        items: [],
+        truncateFrom: 8,
+        disabled: true,
+      },
+      {
+        tabTitle: `Past Teams (${teams.length})`,
```
gabiayako, line 52: I feel like it's a bit weird to filter inactive teams at the top and in here show them all 😛 so this case could happen where the counts don't match. 



Link to the page https://2180.hub.asap.science/network/groups/27b21989-8d51-44c5-8a5e-432efa740140/about#group-teams-83224df9-b0cd-4dd5-a0c5-652b8358d8a5 

Just leaving a comment here to ask if this was indeed expected 😉

gabiayako (approved):

packages/react-components/src/organisms/TeamsTabbedCard.tsx
kmaid: Good spot! @oratiu ticket removes the count in the header from inactive groups. This count reflects active teams.

PR #2186, opened 2022-10-18
CRN-1040 Alumni user profile message

lctrt:


packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -196,127 +202,144 @@ const UserProfileHeader: React.FC<UserProfileHeaderProps> = ({
...
-              {degree ? (
-                <Display styleAsHeading={2}>, {degree}</Display>
-              ) : isOwnProfile ? (
-                <div css={{ color: tin.rgb }}>
-                  <Display styleAsHeading={2}>, Degree</Display>
-                </div>
-              ) : null}
+    <>
+      {alumniSinceDate && (
+        <Toast accent="info">
```
gabiayako, line 207: It seems there's a problem when the text is too large inside the `Toast` for mobile view because it has a fixed height (https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/organisms/Toast.tsx#L55)

kmaid (approved):

PR #2189, opened 2022-10-19
feat/resources tab for projects

markcsims:
https://asaphub.atlassian.net/browse/GP2-118

- [x] Create resources component
- [x] Add resources to schema
- [x] add resources to api
- [x] add project resources


packages/model/src/gp2/working-group.ts
```diff
@@ -1,4 +1,5 @@
 import { ListResponse } from '../common';
+import { Resource } from './common';
```
csbarbosa, line 2: To simplify, you could put:
import { ListResponse, Resource } from '../common';

csbarbosa (approved):

packages/model/src/gp2/project.ts
```diff
@@ -1,4 +1,5 @@
 import { ListResponse } from '../common';
+import { Resource } from './common';
```
csbarbosa, line 2: import { Resource, ListResponse } from '../common';

PR #2192, opened 2022-10-19
CRN-1119 - Teardown Contentful Environment

m-doughty:
# Teardown Contentful Environment

## Ticket

[Here](https://asaphub.atlassian.net/browse/CRN-1079) - subtask CRN-1119

## Functionality

- On closing a PR, checks whether we have a Contentful environment specifically for this branch
- If we do, destroys it via Contentful API
- Will explicitly NOT destroy `Production` or `Development` envs
- If destruction doesn't return 204, errors

## Testing

I manually ran this workflow [here](https://github.com/yldio/asap-hub/actions/runs/3283163547/jobs/5407530992).

You can check that `crn-2190` Environment was deleted [here](https://app.contentful.com/spaces/5v6w5j61tndm/settings/environments).

You can repeat this step by:

- manually creating `crn-2192` Contentful environment
- adding a migration to this branch to set `on-branch env` flag
- running the `on-remove` action manually on this branch targetting `2192`.

Note: Other steps are failing because they were already destroyed on previous unsuccessful test runs. We are only interested in `crn-contentful-remove` step.


.github/workflows/on-remove.yml
```diff
@@ -125,6 +125,41 @@ jobs:
...
+    environment: Branch
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: Branch
+      - name: Contentful checks
```
peterstarling, line 145: I think I would rename this to something like 'Get Contentful branch environment name' or something, that way I would know immediately what we are calling the contentful-checks action in this scenario

.github/actions/setup-environment/action.yml
```diff
@@ -65,6 +65,9 @@ outputs:
   contentful-environment:
     description: 'Contentful Environment'
     value: ${{ steps.vars.outputs.contentful-environment }}
+  max-contentful-envs:
+    description: 'Maximum Contentful environments'
+    value: ${{ steps.vars.outputs.max-contentful-envs }}
```
peterstarling, line 70: I don't mind this to be static like this but isn't there a way to do it via the contentful-checks action somehow?

.github/actions/contentful-teardown/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Remove Contentful PR Environment
+      if: ${{ inputs.contentful-env != 'Development' }} && ${{ inputs.contentful-env != 'Production' }}
+      shell: bash
+      run: |
+        STATUS_CODE=$(curl --silent \
```
peterstarling, line 22: i take it there is no way to use the cli for that?

peterstarling (approved):
approved but with minor comments

.github/actions/contentful-teardown/action.yml
```diff
@@ -0,0 +1,40 @@
+name: Contentful Teardown
```
markcsims, line 1: Do we need this action? It's only being called in 1 place. We could inline it into the workflow.

.github/actions/contentful-teardown/action.yml
```diff
@@ -0,0 +1,40 @@
...
+      shell: bash
+      run: |
+        STATUS_CODE=$(curl --silent \
+                           --request DELETE \
+                           --output /dev/null \
+                           --write-out "%{http_code}" \
+                           --header "Authorization: Bearer ${CONTENTFUL_MANAGEMENT_ACCESS_TOKEN}" \
+                           https://api.contentful.com/spaces/${CONTENTFUL_SPACE_ID}/environments/${CONTENTFUL_ENV_ID})
+
+        if [ "$STATUS_CODE" == "204" ]; then
```
markcsims, line 29: Should this be:

```suggestion
        if [ "$STATUS_CODE" = "204" ]; then
```
or
```suggestion
        if [[ "$STATUS_CODE" == "204" ]]; then
```

.github/actions/contentful-teardown/action.yml
m-doughty: There is, but you have to do multiple steps (setup, authenticate, delete and parse all the responses) to do what the API does in one.

.github/workflows/on-remove.yml
m-doughty: We're getting the branch environment as well as whether we're on a branch environment. Will rename.

.github/actions/contentful-teardown/action.yml
m-doughty: Potentially not, I put it here for readability as well as potential future reuse (potential use case where there WAS a migration on a PR but it gets removed -- seems edge-casey). Happy to move it.

.github/workflows/on-remove.yml
m-doughty: Done.

PR #2175, opened 2022-10-14
CRN-1079 | PR 1 | Validate whether we need to create new Contentful environment

m-doughty:
This action provides the 'brains' for all Contentful-dependent actions.

**Should we create?**
```
on-branch-env == True && 
contentful-env-exists == False
```

**Should we migrate?**
```
on-branch-env == True && 
contentful-env-exists == True && 
contentful-env-migrations-up-to-date == False
```

**Should we tear down?**
```
on-branch-env == True &&
contentful-env-exists == True
```

**Do we need a warning because we reached env limit?**
```
on-branch-env == True &&
contentful-limit-reached == True
```

**What env should we point deployment at?**
```
contentful-env
```

PR #2197, opened 2022-10-20
CRN-1120 | Create Contentful Environment on Branch

m-doughty:
# Create Contentful Environment on Branch

## Ticket

Parent ticket is [CRN-1079](https://asaphub.atlassian.net/jira/software/c/projects/CRN/boards/6?modal=detail&selectedIssue=CRN-1079).

## Changes

- If a migration exists on branch which has not been run on `Development`, creates a branch environment.
- If we're at Contentful limit, put out a useful message.

### Test Without Migration

[Example](https://github.com/yldio/asap-hub/actions/runs/3291249732/jobs/5425161807)

### Test With Migration

[Example](https://github.com/yldio/asap-hub/actions/runs/3291239290/jobs/5425137426)

### Test With Migration at Limit

[Example](https://github.com/yldio/asap-hub/actions/runs/3291221624/jobs/5425097338)


.github/scripts/contentful/create.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Target is Production branch; nothing to do."
+  exit 0
+fi
+
+if [[ "$CONTENTFUL_ENV_ID" == "Development" ]]; then
+  echo "Target is Development branch; nothing to do."
+  exit 0
+fi
+
+STATUS_CODE=$(curl --silent \
```
peterstarling, line 18: why are we not using the [contentful cli](https://www.contentful.com/developers/docs/tutorials/cli/environment-management/)?

.github/workflows/reusable-create-environment.yml
```diff
@@ -48,15 +48,32 @@ jobs:
...
+        if: ${{ steps.get-contentful-state.outputs.on-branch-env == 'True' && steps.get-contentful-state.outputs.contentful-env-exists == 'False' && steps.get-contentful-state.outputs.contentful-limit-reached == 'True' }}
+        shell: bash
+        run: |
+          echo "Contentful Environment required but all available slots are in use."
+          exit 1
+      - name: Create Contentful environment
+        id: create-contentful-environment
+        if: ${{ steps.get-contentful-state.outputs.on-branch-env == 'True' && steps.get-contentful-state.outputs.contentful-env-exists == 'False' && steps.get-contentful-state.outputs.contentful-limit-reached == 'False' }}
+        shell: bash
+        run: bash .github/scripts/contentful/create.sh
```
peterstarling, line 71: do we need a script for it? why not just do it inline?

peterstarling (changes requested):

.github/workflows/reusable-create-environment.yml
markcsims: We were creating these as actions, but we felt a script was better. It takes the functionality away from the framework of the workflows.

Also scripts give you syntax highlighting/support.

.github/scripts/contentful/create.sh
m-doughty: ## Agreed & Consistent Approach

I thought that using the API was an agreed approach as it was specified in [the ticket](https://asaphub.atlassian.net/browse/CRN-1079) which we all refined and estimated, and approved on three previous PRs:

- [Contentful checks](https://github.com/yldio/asap-hub/pull/2175)
- [Contentful teardown](https://github.com/yldio/asap-hub/pull/2192)
- [Contentful teardown hotfix](https://github.com/yldio/asap-hub/pull/2196)

I believe that if we were to switch to CLI we should use it consistently in all of these places. This would also mean checks would be parsing plain-text responses to set mission critical outputs.

## CLI does not provide an advantage

I don't believe that there is any advantage to be gained from using the CLI, it takes a single HTTP call and replaces it with three actions: login, configure, and then whatever the action to be performed is. 

In a long-running interactive session this would make sense because the stored login and configuration would reduce verbosity of subsequent calls, but in our situation we would have to call all three steps every time.

The output is also plain text which is much harder to parse than status codes + JSON.

## CLI just uses API under the hood anyway

The CLI is [just using the HTTP API under the hood anyway](https://github.com/contentful/contentful-cli/blob/master/lib/cmds/login.js), but we'd be making 3x as many HTTP requests in exchange for losing the easily-parsed JSON + status code outputs.

.github/workflows/reusable-create-environment.yml
m-doughty: > We were creating these as actions, but we felt a script was better. It takes the functionality away from the framework of the workflows.
> 
> Also scripts give you syntax highlighting/support.

Easier to test locally as well.

peterstarling (approved):

PR #2205, opened 2022-10-21
CRN 1121/run contentful migrations on branch

markcsims:
- Add a migration & add emacs tempfiles to gitignore.
- migration for pr environment


m-doughty:
Workflow with no migrations to run: [here](https://github.com/yldio/asap-hub/actions/runs/3346202284)

PR #2207, opened 2022-10-21
feat: CRN-1122 Set Contentful Env for Serverless

peterstarling:
Notable changes:
* add a `setup` job before the `build`, `create-environment` and `deployment`
* user lowercase for pseudo-boolean values
* move the contentful env limit check to where the env is actually created

Outstanding tasks for another PR
* use contentful cli to create the env

### DIFF
--- a/.github/scripts/contentful/create.sh
+++ b/.github/scripts/contentful/create.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [[ "$ON_BRANCH_ENV" == "false" ]]; then
-  echo "No branch Environment required; not deleting environment."
+  echo "No branch Environment required; not creating environment."
   exit 0
 fi
 
--- a/.github/scripts/contentful/create.sh
+++ b/.github/scripts/contentful/create.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ "ON_BRANCH_ENV" == "False" ]]; then
+if [[ "ON_BRANCH_ENV" == "false" ]]; then
   echo "No branch Environment required; not deleting environment."
   exit 0
 fi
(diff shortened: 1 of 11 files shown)


m-doughty (changes requested):
Due to the case change in the `checks` outputs (I don't have strong opinions on it either way), `on-remove.yml` will need to be changed where it checks.

I think there will also be changes to the guard clauses in the shell scripts (`create.sh` and `teardown.sh`).

.github/scripts/contentful/create.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ "ON_BRANCH_ENV" == "False" ]]; then
+if [[ "ON_BRANCH_ENV" == "false" ]]; then
```
markcsims, line 3:
```suggestion
if [[ "$ON_BRANCH_ENV" == "false" ]]; then
```

.github/scripts/contentful/create.sh
```diff
@@ -1,6 +1,6 @@
...
   echo "No branch Environment required; not deleting environment."
```
markcsims, line 4: this message is confusing to me.

The variable is 'ON_BRANCH_ENV' i.e. whether we are on a dev branch. 

And this message is printed when we are not on dev.

m-doughty: Yeah agree, this should probably be something like "No branch environment to delete."

m-doughty: I have made this change @markcsims.

peterstarling, line 4:
```suggestion
  echo "No branch Environment required; not creating environment."
```

markcsims (approved):

PR #2212, opened 2022-10-24
Added leaders tabbed card component

oratiu:


packages/react-components/src/organisms/LeadersTabbedCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+        tabTitle: 'Active Leaders (0)',
+        items: [],
+        disabled: true,
+      },
+      {
+        tabTitle: `Past Leaders (${leaders.length})`,
+        items: leaders,
+      },
+    ]}
+    getShowMoreText={(showMore) => `View ${showMore ? 'Less' : 'More'} Leaders`}
```
kmaid, line 45: This should be removed as you can never trigger it because it is never truncated.

packages/react-components/src/organisms/LeadersTabbedCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+  >
+    {({ data }) => (
+      <div css={containerStyles}>
+        <MembersList
+          members={data.map(({ user, role }) => ({
+            ...user,
+            firstLine: user.displayName,
+            secondLine: role,
+            thirdLine: user.teams.length <= 1 ? user.teams : 'Multiple Teams',
```
kmaid, line 54: You need two tests. One for a leader with a single team and another with multiple teams

packages/react-components/src/molecules/TabbedCard.tsx
```diff
@@ -80,7 +80,7 @@ const TabbedCard = <T extends object>({
       {showShowMoreButton && (
         <div css={showMoreStyles}>
           <Button linkStyle onClick={() => setShowMore(!showMore)}>
-            {getShowMoreText(showMore)}
+            {!!getShowMoreText && getShowMoreText(showMore)}
```
kmaid, line 83: I think this needs some more thought

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -58,17 +58,17 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
         tools={tools}
         active={active}
       />
-      <div id={membersSectionId} css={[!active && membersSectionStyles]}>
+      <div id={membersSectionId}>
         {active ? (
           <GroupMembersSection teams={teams} leaders={leaders} />
         ) : (
-          <>
+          <div css={membersSectionStyles}>
```
kmaid, line 65: 👍

kmaid (approved):

PR #2215, opened 2022-10-24
feat/onboarded for gp2 users

markcsims:
https://asaphub.atlassian.net/browse/GP2-103

- [x] refactor migration handler to server-common
- [x] add migration handler to GP2
- [x] add onboarded to gp2 user schema
- [x] add gp2 to api
- [x] add onboarded migration
- [ ] run migration on dev
- [ ] run migration on prod

Outstanding tasks for another PR

- Reduce tests on handlers in packages


peterstarling (approved):
Approved, although I would like to see a follow up work to remove the test duplication for the migration package.

Also the webpack config is incorrect as it copies the migrations for every lambda (which are packaged individually as per sls config) but since we already have this issue in CRN i approved it anyway

PR #2220, opened 2022-10-25
CRN-1084 Team Members in Team

gabiayako:


packages/react-components/src/molecules/TabbedCard.tsx
```diff
@@ -51,7 +55,7 @@ const TabbedCard = <T extends object>({
     getShowMoreText && truncateFrom && items.length > truncateFrom;
 
   return (
-    <Card padding={false}>
+    <Card padding={false} overrideStyles={cardStyles}>
```
gabiayako, line 58: Without this change, it was appearing like this in mobile views

lctrt (approved):

PR #2223, opened 2022-10-26
[CRN-1106] remove Group Members from details page and storybook

ricardocarvalh0:


packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -59,9 +57,7 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
         active={active}
       />
       <div id={membersSectionId}>
-        {active ? (
-          <GroupMembersSection teams={teams} leaders={leaders} />
-        ) : (
+        {!active && (
```
gabiayako, line 60: This means that when the team is active is not going to appear neither the leaders or teams while it's not replaced by the new designs. Is this the idea? I don't remember if we addressed this in refinement and it's not that clear in the ticket 😛

ricardocarvalh0: I checked with Tiffany and leaders and teams should now appear for both active and inactive groups. I'll apply it now. Thanks for spotting that @gabiayako !😃

kmaid (approved):

apps/storybook/src/LeadersTabbedCard.stories.tsx
```diff
@@ -14,11 +14,34 @@ export const Normal = () => (
...
           role: 'Lead PI - Chair' as GroupRole,
         }),
       ),
     ]}
     title={text('Title', 'Interest Group Teams')}
+    disableActiveTab={false}
+  />
+);
+
+export const ActiveTabDisabled = () => (
```
kmaid, line 30: Could this not be a simple boolean knob in the normal case

kmaid: The other less desirable solution is not to copy paste but take the data out

packages/react-components/src/organisms/TeamsTabbedCard.tsx
```diff
@@ -29,32 +30,31 @@ type TeamsTabbedCardProps = Pick<
...
-        disabled: true,
-      },
-      {
-        tabTitle: `Past Teams (${teams.length})`,
-        items: teams,
-        truncateFrom: 8,
-        disabled: false,
-      },
-    ]}
+    activeTabIndex={disableActiveTab ? 1 : 0}
```
kmaid, line 48: How does this work for more than two tabs?

packages/react-components/src/organisms/TeamsTabbedCard.tsx
```diff
@@ -29,32 +30,31 @@ type TeamsTabbedCardProps = Pick<
   ComponentProps<typeof TabbedCard>,
   'title' | 'description'
 > & {
-  teams: ReadonlyArray<Pick<TeamResponse, 'id' | 'displayName'>>;
+  teams: ReadonlyArray<
+    Pick<TeamResponse, 'id' | 'displayName' | 'inactiveSince'>
+  >;
+  disableActiveTab?: boolean;
```
kmaid, line 36: I think this raises issues about the disabled requirement. What happens if we have an active group/team/whatever with only alumni members? You would default to a disabled tab with no content which is a very odd behaviour

packages/react-components/src/organisms/__tests__/TeamsTabbedCard.test.tsx
```diff
@@ -6,18 +6,41 @@ import TeamsTabbedCard from '../TeamsTabbedCard';
...
+        teams={[
+          { displayName: 'team 1', id: '123' },
+          { displayName: 'team 2', id: '123', inactiveSince: '2021-01-01' },
+        ]}
+        title="Example"
+        disableActiveTab={disableActiveTab}
+      />,
+    );
+
+  it('renders for an Active Group', () => {
```
kmaid, line 26: Should this cover disabled rules for tabs?

kmaid, line 26: Should this cover disabled rules for tabs? Check first with the above regarding how we should handle the edge cases to do with disabled tabs.

kmaid (changes requested):

apps/storybook/src/LeadersTabbedCard.stories.tsx
ricardocarvalh0: Didn't know about that knob stuff! A new world unfolds, thank you.

packages/react-components/src/organisms/TeamsTabbedCard.tsx
ricardocarvalh0: I'll add a 'There are no past leaders.' paragraph when the tab is empty

packages/react-components/src/utils/common.ts
```diff
@@ -75,3 +76,66 @@ export function equals(a: Array<string>, b: Array<string>): boolean {
     a.length === b.length && a.every((element, index) => element === b[index])
   );
 }
+
+export function splitListBy<T>(
```
kmaid, line 80: I think this function is legit. We do want to split lists based on the the presence of a key being there or not. Could be optimized into a single reduce that returns a tupple. I think there is a better way to type this generically using something like whats in packages/dom-test-utils/src/css.ts except instead of css object it uses the generic passed in if that will work

apps/storybook/src/TeamsTabbedCard.stories.tsx
```diff
@@ -9,7 +9,13 @@ export default {
...
   <TeamsTabbedCard
-    teams={createListTeamResponse(number('Teams', 10)).items}
+    teams={createListTeamResponse(number('Teams', 10)).items.map(
+      (team, index) => ({
+        ...team,
+        inactiveSince: index === 2 ? '2021-01-01' : undefined,
+      }),
+    )}
     title={text('Title', 'Interest Group Teams')}
+    isGroupInactive={boolean('Is Group Inactive?', false)}
```
kmaid, line 19: I don't like this prop. This is a Teams Tabbed Card, either it needs to be a GroupTeamsTabbedCard or isGroupInactive should be something more generic

packages/react-components/src/utils/__tests__/common.test.tsx
```diff
@@ -112,3 +114,24 @@ describe('equals', () => {
     expect(equals(testArray, testArray.slice(0, 1))).toBeFalsy();
   });
 });
+
+describe('tabbed card utils', () => {
```
kmaid, line 118: This is just a generic function or it belongs in a new file
```suggestion
```

packages/react-components/src/templates/GroupProfileAbout.tsx
```diff
@@ -59,17 +57,18 @@ const GroupProfileAbout: React.FC<GroupProfileAboutProps> = ({
...
+        <div css={membersSectionStyles}>
+          <GroupLeadersTabbedCard
+            title="Interest Group Leaders"
+            leaders={leaders}
+            isGroupInactive={!active}
+          />
+          <GroupTeamsTabbedCard
+            title="Interest Group Teams"
+            teams={teams}
+            isGroupInactive={!active}
```
kmaid, line 69:
```suggestion
            isGroupActive={active}
```
Suretly this is better?

packages/react-components/src/organisms/__tests__/GroupTeamsTabbedCard.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { fireEvent, render, screen } from '@testing-library/react';
+import GroupTeamsTabbedCard from '../GroupTeamsTabbedCard';
+
+const props: ComponentProps<typeof GroupTeamsTabbedCard> = {
+  teams: [],
+  title: '',
+  isGroupInactive: false,
+};
+
+describe('renders the teams tabbed card', () => {
```
kmaid, line 12:
```suggestion
```

packages/react-components/src/organisms/__tests__/GroupTeamsTabbedCard.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { ComponentProps } from 'react';
+import { createListTeamResponse } from '@asap-hub/fixtures';
+import { fireEvent, render, screen } from '@testing-library/react';
+import GroupTeamsTabbedCard from '../GroupTeamsTabbedCard';
+
+const props: ComponentProps<typeof GroupTeamsTabbedCard> = {
```
kmaid, line 6: 👍

packages/react-components/src/organisms/__tests__/GroupTeamsTabbedCard.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        teams={[
+          { displayName: 'team 1', id: '123' },
+          { displayName: 'team 2', id: '123', inactiveSince: '2021-01-01' },
+        ]}
+        title="Example"
+        isGroupInactive={isGroupInactive}
+      />,
+    );
+
+  it('renders for an Active Group', () => {
```
kmaid, line 26:
```suggestion
  it('splits active and inactive teams', () => {
```

packages/react-components/src/organisms/__tests__/GroupTeamsTabbedCard.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    expect(screen.getByRole('heading').textContent).toEqual('Example');
+    expect(screen.getByText('Past Teams (1)')).toBeVisible();
+    expect(screen.getByText('Active Teams (1)')).toBeVisible();
+    expect(screen.getByText(/team 1/)).toBeVisible();
+
+    fireEvent.click(screen.getByText('Past Teams (1)'));
+    expect(screen.getByText(/team 2/)).toBeVisible();
+  });
+
+  it('renders for an Inactive Group', () => {
```
kmaid, line 37:
```suggestion
  it('shows all teams as past when group is inactive', () => {
```

packages/react-components/src/organisms/__tests__/GroupTeamsTabbedCard.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    expect(screen.getByText(/team 2/)).toBeVisible();
+  });
+});
+
+it('shows the correct more and less button text', () => {
+  render(
+    <GroupTeamsTabbedCard
+      {...props}
+      teams={createListTeamResponse(20).items}
+      title="Example"
```
kmaid, line 51:
```suggestion
]```

File: packages/react-components/src/organisms/__tests__/GroupLeadersTabbedCard.test.tsx [hunk 15]
```diff
@@ -0,0 +1,109 @@
...
+                displayName: 'Test team',
+                role: 'Project Manager',
+                id: 'team-id-1',
+              },
+            ],
+          },
+          role: 'Project Manager',
+        },
+      ]}
+      title="Leaders title"
```
kmaid, line 104:
```suggestion
```

apps/storybook/src/GroupLeadersTabbedCard.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import { GroupLeadersTabbedCard } from '@asap-hub/react-components';
+import { createUserResponse } from '@asap-hub/fixtures';
+import { number, boolean } from '@storybook/addon-knobs';
+import { GroupRole } from '@asap-hub/model';
+
+export default {
+  title: 'Organisms /Group Leaders Tabbed Card',
```
kmaid, line 7:
```suggestion
  title: 'Organisms / Group Leaders Tabbed Card',
```

packages/react-components/src/organisms/GroupLeadersTabbedCard.tsx
```diff
@@ -0,0 +1,66 @@
+import { GroupLeader } from '@asap-hub/model';
+import { css } from '@emotion/react';
+import React from 'react';
+import { MembersList, TabbedCard } from '../molecules';
+import { rem, tabletScreen } from '../pixels';
+import { splitListBy } from '../utils';
+
+const containerStyles = css({
+  listStyle: 'none',
```
kmaid, line 9: Does this do anything? The container is a div. Margin simmilarly

packages/react-components/src/organisms/GroupTeamsTabbedCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+          tabTitle: `Past Teams (${inactiveTeams.length})`,
+          items: inactiveTeams,
+          truncateFrom: 8,
+          disabled: inactiveTeams.length === 0,
+        },
+      ]}
+      getShowMoreText={(showMore) => `View ${showMore ? 'Less' : 'More'} Teams`}
+    >
+      {({ data }) => (
+        <ul css={containerStyles}>
```
kmaid, line 74: Looks like the above was copy pasted from here.

packages/react-components/src/organisms/__tests__/GroupLeadersTabbedCard.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { createUserResponse, createUserTeams } from '@asap-hub/fixtures';
+import { render, screen, fireEvent } from '@testing-library/react';
+import GroupLeadersTabbedCard from '../GroupLeadersTabbedCard';
+
+const props: ComponentProps<typeof GroupLeadersTabbedCard> = {
+  leaders: [],
+  isGroupActive: false,
+};
+
+describe('renders the leaders tabbed card', () => {
```
kmaid, line 11:
```suggestion

```
This is implied by it being the group leader tabbed card test

kmaid (approved):

PR #2230, opened 2022-10-28
GP2 120 allow adding resources to working group page server

markcsims:
https://asaphub.atlassian.net/browse/GP2-120

Adds a '/working-group/:workingGroupId/resources' endpoint which takes an array of resources.


apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -55,14 +69,30 @@ export class WorkingGroupSquidexDataProvider
...
+    return this.squidexGraphqlClient.request<
       FetchWorkingGroupQuery,
       FetchWorkingGroupQueryVariables
     >(FETCH_WORKING_GROUP, { id });
   }
 }
 
+const convertToSquidexWorkingGroup = (
+  workingGroup: gp2Model.WorkingGroupUpdateDataObject,
+) =>
```
peterstarling, line 90: I guess it would be useful to declare the output as `gp2Squidex.InputWorkingGroup["data"]`?

apps/gp2-server/test/controllers/working-group.controller.test.ts
```diff
@@ -137,4 +133,23 @@ describe('Working Group controller', () => {
       });
     });
   });
+
+  describe('update', () => {
+    beforeEach(jest.resetAllMocks);
+
+    test('Should return the newly updated user', async () => {
```
peterstarling, line 140:
```suggestion
    test('Should return the newly updated working group', async () => {
```

apps/gp2-server/test/controllers/working-group.controller.test.ts
```diff
@@ -137,4 +133,23 @@ describe('Working Group controller', () => {
...
+      workingGroupDataProviderMock.fetchById.mockResolvedValue(mockResponse);
+      const result = await workingGroupController.update(
+        '7',
+        { resources: [] },
+        '11',
+      );
+
+      expect(result).toEqual(getWorkingGroupResponse());
+      expect(workingGroupDataProviderMock.update).toHaveBeenCalledWith('7', {
+        resources: [],
```
peterstarling, line 151: probably best to test with actual data instead?

apps/gp2-server/test/routes/working-group.route.test.ts
```diff
@@ -120,4 +118,138 @@ describe('/working-groups/ route', () => {
...
+    test.each(gp2.userRoles.filter((role) => role !== 'Administrator'))(
+      'Should return 403 if the user is not in the Administrator role - %s',
+      async (role) => {
+        const loggedInUserId = '11';
+        const loggedUser: gp2.UserResponse = {
+          ...getUserResponse(),
+          id: loggedInUserId,
+          role,
+        };
+        const getLoggedUser = jest.fn().mockReturnValue(loggedUser);
```
peterstarling, line 198: looks like a lot of repetition, can you move out the common stuff into describe or beforeEach/beforeAll?

apps/gp2-server/test/routes/working-group.route.test.ts
```diff
@@ -120,4 +118,138 @@ describe('/working-groups/ route', () => {
...
+        const response = await supertest(appWithUser)
+          .put('/working-group/23/resources')
+          .send([]);
+
+        expect(response.status).toBe(403);
+        expect(workingGroupControllerMock.update).not.toBeCalledWith();
+      },
+    );
+    test('Should return 403 if the user is not in part of the working group', async () => {
+      const loggedInUserId = 'not-in-working-group';
```
peterstarling, line 224: i found it hard to see how this test was different from others, maybe it would be better to still use the same user ID and change the members list instead to make it more obvious that this user is no longer part of that?

apps/gp2-server/test/validation/working-group.validation.test.ts
```diff
@@ -0,0 +1,107 @@
+import { gp2 } from '@asap-hub/model';
+import { validateWorkingGroupPatchRequest } from '../../src/validation/working-group.validation';
+
+describe('Working Group validation', () => {
```
peterstarling, line 4: this is a little against the current pattern by which we test validation on a route level. That way we are not only testing that the validation rules are fine but also that it is attached correctly to the router. We often extract and manipulate data that comes out of the validation so I think that's much better to do it at that level.

I see little value in those tests on their own like this

apps/gp2-server/test/controllers/working-group.controller.test.ts
markcsims: done

apps/gp2-server/src/data-providers/working-group.data-provider.ts
markcsims: done

apps/gp2-server/test/controllers/working-group.controller.test.ts
markcsims: an empty array is valid too. I've updated it.

apps/gp2-server/test/routes/working-group.route.test.ts
markcsims: done

apps/gp2-server/test/routes/working-group.route.test.ts
markcsims: done

apps/gp2-server/test/validation/working-group.validation.test.ts
markcsims: the tests were really helpful when trying to understand if we were covering all the logic. 

I'll remove them

peterstarling (approved):

PR #2231, opened 2022-10-28
CRN-1052 Contentful - News content model

gabiayako:
This PR adds a migration adding the rest of fields in `news` content type in Contentful **except** the `type` field as we are currently in the process of separating the working groups from news so this field won't be necessary.

Below there are screenshots comparing the fields and input types in both CMS's.

Squidex | Contentful
:---------:|:---------:
news-squidex | news-contentful


packages/contentful/package.json
```diff
@@ -14,6 +14,7 @@
     "space:migrate:crn": "yarn ctf-migrate up --all",
     "space:migrate:crn:dryrun": "yarn ctf-migrate up --all --dry-run",
     "space:migrate:crn:count": "yarn ctf-migrate list --all | grep \\\\[pending\\\\] | wc -l",
+    "generate:migration:crn": "contentful space generate migration -s $CONTENTFUL_SPACE_ID -e $CONTENTFUL_ENV_ID",
```
peterstarling, line 17: I thought we were using this command `ctf-migrate create ...` as per https://github.com/yldio/asap-hub/blob/602236c1624ecbc2d8b640526251d98bc9c1ba2b/README.md#L206

README.md
```diff
@@ -214,6 +214,29 @@ yarn workspace @asap-hub/contentful ctf-migrate create add-foo-field -c bar
...
+### Generating a migration script from content model
+
+It's possible to make changes in Contentful UI content type, adding or changing fields and then generate the migration.
+
+To do that you need to follow these steps below:
+
+1. Create an empty migration with the command from the section above
+
+   ```sh
+   yarn workspace @asap-hub/contentful ctf-migrate create <name> -c <content_type>
```
peterstarling, line 226: isn't this the same as https://github.com/yldio/asap-hub/blob/602236c1624ecbc2d8b640526251d98bc9c1ba2b/README.md#L206 ?

README.md
```diff
@@ -214,6 +214,29 @@ yarn workspace @asap-hub/contentful ctf-migrate create add-foo-field -c bar
...
+   ```
+
+   :warning: Copy the path of the generated file, you will use that in step 5.
+
+2. Add the generated file, create a commit and push to remote and create a draft PR. In the pipeline there's a step called `create-environment/crn-contentful` that will create a PR env in Contentful if there's room for it (max of 6 envs).
+3. Access the PR env and make the changes you want to.
+4. In `.env` fill `CONTENTFUL_SPACE_ID`, `CONTENTFUL_ACCESS_TOKEN` and `CONTENTFUL_ENV_ID`. This field `CONTENTFUL_ENV_ID` should be the PR env that was created and you made changes to it in step 3, something like `crn-2227`.
```
peterstarling, line 233: I'm not sure what this part of README is about, where you trying to describe what a local development should look like?

packages/contentful/migrations/news/20221028130957-add-fields-to-news-content-type.js
```diff
@@ -0,0 +1,154 @@
+module.exports.description = 'Adds fields in news content type.';
+
+module.exports.up = function (migration) {
+  const news = migration.editContentType('news');
+  news.deleteField('id');
```
peterstarling, line 5: don't we need an ID?

peterstarling (changes requested):

README.md
gabiayako: (I'm replying the comments above in this one)

Not only local, it would be like when we change schema in squidex and sync it with local. The idea is to use ` yarn workspace @asap-hub/contentful ctf-migrate create (...)` to generate only the empty file as the `create` command generates an empty file

> create
Creates an empty time stamped file in the content-type's migrations folder.
https://github.com/deluan/contentful-migrate#create
 
Then from the empty file we can manually type all the fields we want to add and then run something like `yarn workspace @asap-hub/contentful ctf-migrate up -e crn-2227 -d` **or** we can use the generate to sync with the changes we made in the UI in the PR env. So this section is to describe how to do that :)

packages/contentful/migrations/news/20221028130957-add-fields-to-news-content-type.js
gabiayako: I wasn't sure about this. I thought we could use this id generated by Contentful (in "sys") or we want to copy the id from Squidex when we migrate the data?

Screen Shot 2022-10-28 at 08 41 38

peterstarling: well for News we probably don't need to but I think eventually we are going to have to use a unique ID which is the same as in squidex. Otherwise after the migration all urls will change - question is do we care?

README.md
peterstarling: So the issue is that if you generate an empty migration and commit it and push - this migration will run on your PR branch and it will only run once. So you cant really add anything to it later unless you remember to delete it (ie modify migration state) on the dedicated environment. 

I think you should wait until this https://asaphub.atlassian.net/browse/CRN-1121 is complete to suggest any changes to the development and PR dev flow

gabiayako: ok! 👍 I removed the instructions to run the generate and the script in package json

peterstarling (approved):

PR #2239, opened 2022-11-01
Active team members and groups components

oratiu:


packages/react-components/src/templates/TeamProfileAbout.tsx
```diff
@@ -49,36 +47,13 @@ const TeamProfileAbout: React.FC<TeamProfileAboutProps> = ({
...
-              firstName,
-              lastName,
-              id,
-            }),
-          )}
-        />
-      </section>
-    ) : null}
+    <section id={teamListElementId}>
+      <TeamMembersTabbedCard
```
gabiayako, line 51: I noticed that this problem where the tabbed card occupies more space than it should it's happening again. I think it's maybe due to this  `section` that is wrapping `TeamMembersTabbedCard`. Is it needed? :)
2239 hub asap science_network_teams_8c88a42b-9fc7-4ec1-8606-204be75d252c_about(iPhone SE)

oratiu: Kept the section but passed the overflow style, it needed the id because there is a functionality which moves the screen to the members section upon clicking the +X members in the header.

gabiayako: nice! It makes sense :)

gabiayako (approved):
Nice! 🎉

apps/crn-frontend/src/network/teams/groups/GroupsCard.tsx
```diff
@@ -17,14 +14,11 @@ const GroupsCard: React.FC<{ id: string; isInactive?: string }> = ({
...
-    groups.total > 0 ? (
-      <TeamGroupsCard groups={groups.items} />
-    ) : null
-  ) : (
-    <TeamTabbedGroupsCard
+  return (
+    <TeamGroupsTabbedCard
       title={'Team Interest Groups'}
       groups={groups.items}
+      inactive={isInactive}
```
gabiayako, line 21: very minor, but I wonder if it would be cleaner to make this a boolean like

```
isTeamInactive={!!isInactive}
```

packages/react-components/src/templates/TeamProfileAbout.tsx
```diff
@@ -49,36 +51,13 @@ const TeamProfileAbout: React.FC<TeamProfileAboutProps> = ({
...
-            }),
-          )}
-        />
-      </section>
-    ) : null}
+    <section id={teamListElementId} css={membersCardStyles}>
+      <TeamMembersTabbedCard
+        title="Team Members"
+        members={members}
+        inactive={inactiveSince}
```
gabiayako, line 58: the same thing here as the comment above, maybe it would be cleaner to make it

```
isTeamInactive={!!inactiveSince}
```

gabiayako (approved):
Looks good! I just made a minor suggestion 😄

gabiayako (approved):
Awesome, it looks good to me 🎉

PR #2242, opened 2022-11-01
CRN-1052 News Data Provider

gabiayako:
This PR adds the data provider for news using the contentful client library (https://github.com/contentful/contentful.js/tree/beta-v10). It also selects the news data provider (squidex or contentful) based on an env var called `isContentfulEnabled`.


packages/contentful/package.json
```diff
@@ -17,6 +17,12 @@
     "schema:update:crn": "graphql-codegen --config codegen.yml"
   },
   "dependencies": {
+    "@asap-hub/errors": "workspace:*",
+    "@asap-hub/model": "workspace:*",
+    "@babel/runtime-corejs3": "7.19.6",
+    "@contentful/rich-text-html-renderer": "^15.13.1",
+    "@contentful/rich-text-types": "^15.13.2",
+    "contentful": "beta-v10",
```
gabiayako, line 25: This was installed in this specific beta version to make it possible to use typescript https://www.contentful.com/developers/docs/javascript/tutorials/typescript-in-javascript-client-library/

apps/crn-server/src/app.ts
```diff
@@ -177,8 +194,14 @@ export const appFactory = (libs: Libs = {}): Express => {
   const groupDataProvider =
     libs.groupDataProvider ||
     new GroupSquidexDataProvider(squidexGraphqlClient);
-  const newsDataProvider =
+  const newsSquidexDataProvider =
     libs.newsDataProvider || new NewsSquidexDataProvider(newsRestClient);
+  /* istanbul ignore next */
```
gabiayako, line 199: I am ignoring this line for now, the only way I thought about testing this was with an integration test. What do you think. If we follow this path, then some more work needs to be done in CI I believe to create the env in contentful

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+export type FetchNewsProviderOptions = FetchPaginationOptions & {
+  filter?: FetchNewsFilter & {
+    title?: string;
+  };
+};
+
+const parsetNews = (item: ContentfulRestNews): NewsDataObject => ({
+  id: item.fields.id,
+  title: item.fields.title,
+  type: 'News',
```
gabiayako, line 77: This can't be removed for now, if we remove the type from the response, the news card gets a little crazy 😝 https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/organisms/NewsCard.tsx#L75

I think this can only be removed when we completely separate the working groups from news, then we won't need the `type`

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+        skip: options?.skip,
+        order: '-sys.createdAt',
+      });
+
+      return {
+        total,
+        items: items.map(parsetNews),
+      };
+    } catch (error) {
+      if (JSON.parse((error as Error).message)?.status === 404) {
```
peterstarling, line 39: this is strange. This is a fetch method so why would we look for a 404 at all here? I

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+        'fields.title[match]': options?.filter?.title,
+        'fields.frequency[in]': options?.filter?.frequency?.join(','),
+        limit: options?.take,
+        skip: options?.skip,
+        order: '-sys.createdAt',
+      });
+
+      return {
+        total,
+        items: items.map(parsetNews),
```
peterstarling, line 36:
```suggestion
        items: items.map(parseNews),
```

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
gabiayako: There is a test for this in the squidex data provider https://github.com/yldio/asap-hub/blob/master/apps/crn-server/test/data-providers/news.data-provider.test.ts#L67, I thought we would want the same behaviour

packages/contentful/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import * as contentful from 'contentful';
+import { CreateClientParams, ContentfulClientApi } from 'contentful';
+
+export const getClient = ({
```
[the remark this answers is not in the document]
gabiayako: I'm calling this `getClient` in `crn-server/app` :) 

So the `crn-server` doesn't call something from `contentful` directly. But I can move this to `crn-server` if you prefer

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+  fetchById(id: string): Promise<NewsDataObject | null>;
+  fetch: (options?: FetchNewsProviderOptions) => Promise<ListNewsDataObject>;
+}
+
+export class NewsContentfulDataProvider {
+  constructor(private contentfulClient: ReturnType<typeof getClient>) {}
+
+  async fetch(options?: FetchNewsProviderOptions) {
+    try {
+      const { total, items } = await this.contentfulClient.getEntries<News>({
```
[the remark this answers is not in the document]
gabiayako: It makes sense the graphql approach for me! I'm working on it since yesterday. I think I'll be able to open a PR maybe today :)

gabiayako:
closing the rest version as we're going to follow with graph ql

PR #2247, opened 2022-11-03
CRN-1052 News Data Provider - GraphQL

gabiayako:


peterstarling (changes requested):

apps/crn-server/src/controllers/news.ts
```diff
@@ -12,7 +13,9 @@ export interface NewsController {
 }
 
 export default class News implements NewsController {
-  constructor(private newsDataProvider: NewsDataProvider) {}
+  constructor(
+    private newsDataProvider: NewsDataProvider | NewsContentfulDataProvider,
```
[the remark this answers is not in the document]
gabiayako: nice! it makes sense, it was committed in here https://github.com/yldio/asap-hub/pull/2247/commits/2075b70fef5d6fb7540465cd212360f1b815bc54

apps/crn-server/src/config.ts
```diff
@@ -71,3 +76,9 @@ export const clientId = SQUIDEX_CLIENT_ID || 'squidex-client-id';
 export const clientSecret = SQUIDEX_CLIENT_SECRET || 'squidex-client-secret';
 export const appName = SQUIDEX_APP_NAME || 'asap-local';
 export const sentryTraceSampleRate = 1.0;
+export const contentfulSpaceId = CONTENTFUL_SPACE_ID || 'contentful-space-id';
+export const contentfulAccessToken =
+  CONTENTFUL_ACCESS_TOKEN || 'contentful-access-token';
+export const contentfulEnvId = CONTENTFUL_ENV_ID || 'contentful-environment-id';
+export const contentfulHost = CONTENTFUL_HOST || 'https://cdn.contentful.com';
+export const isContentfulEnabled = IS_CONTENTFUL_ENABLED || false;
```
peterstarling, line 84: maybe this should be evaluated into a boolean here rather than later?

gabiayako: Do you mean using something like this https://github.com/node-config/node-config/issues/272#issuecomment-162567709? :)

gabiayako:
rebase to master to solve code conflict

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -0,0 +1,87 @@
...
+      Pick<QueryNewsCollectionArgs, 'limit' | 'skip' | 'order'> & {
+        where?: Partial<NewsFilter>;
+      }
+    >(FETCH_NEWS, {
+      limit: options?.take,
+      skip: options?.skip,
+      order: [NewsOrder.SysFirstPublishedAtDesc],
+      where: {
+        frequency_in: options?.filter?.frequency as string[],
+        title_contains: options?.filter?.title,
```
gabiayako, line 37: now we have the possible filters appearing in TS as suggestions

types-gql

apps/crn-server/test/app.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+
+  test('news controller uses squidex data provider when IS_CONTENTFUL_ENABLED is false', async () => {
+    process.env.IS_CONTENTFUL_ENABLED = 'false';
+
+    const { appFactory } = require('../src/app');
```
peterstarling, line 31: why are you having to require?

apps/crn-server/test/fixtures/news.fixtures.ts
```diff
@@ -39,19 +80,26 @@ export const newsSquidexApiResponse: {
...
+  isContentfulResponse = false,
+): NewsDataObject => ({
   id: 'news-1',
   title: 'News 1',
   type: 'News',
+  frequency: 'News Articles',
   shortText: 'Short text of news 1',
   text: '<p>text</p>',
-  thumbnail: `${baseUrl}/api/assets/${appName}/thumbnail-uuid1`,
+  thumbnail: isContentfulResponse
```
peterstarling, line 92: im not a huge fan of this - the DataObject should be the same for both. Now we know that contentful url's will be different but perhaps there is a way to store the base path in a config or something?

apps/crn-server/test/fixtures/news.fixtures.ts
```diff
@@ -39,19 +80,26 @@ export const newsSquidexApiResponse: {
...
+    ? `http://images.ctfassets.net/thumbnail-id.png`
+    : `${baseUrl}/api/assets/${appName}/thumbnail-uuid1`,
   created: '2020-09-08T16:35:28.000Z',
 });
 
-export const getListNewsDataObject = (): ListNewsDataObject => ({
+export const getListNewsDataObject = (
+  isContentfulResponse = false,
```
peterstarling, line 99: again not a huge fan of this. Having to choose which data source we are getting this data object from when we said data objects should be the same for both sources is a little bit of an anti pattern to me

packages/contentful/src/crn/queries/news.queries.ts
```diff
@@ -0,0 +1,47 @@
...
+      text {
+        json
+      }
+    }
+  }
+`;
+
+export const FETCH_NEWS_BY_ID = gql`
+  query FetchNewsById($id: String!) {
+    newsCollection(where: { id: $id }) {
```
peterstarling, line 27: why exactly are you using the collection when we could just use the `news` query which has an ID param?
image

peterstarling (changes requested):

packages/contentful/src/crn/queries/news.queries.ts
gabiayako: This id is the contentful id. If we want to fetch by squidex id that we will migrate and populate this id field we need to query using the collection to filter properly

apps/crn-server/test/app.test.ts
```diff
@@ -0,0 +1,105 @@
+import supertest from 'supertest';
+import { authHandlerMock } from './mocks/auth-handler.mock';
+
+jest.mock('../src/data-providers/news.data-provider', () => ({
```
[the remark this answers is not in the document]
gabiayako: hmm I only understood the suggestion now, I think that's it [615fab3](https://github.com/yldio/asap-hub/pull/2247/commits/615fab317c8978abdd1985bf6408456fd0dec459), right?

apps/crn-server/test/fixtures/news.fixtures.ts
gabiayako: I'm making the url the same as squidex then [a5631a3](https://github.com/yldio/asap-hub/pull/2247/commits/a5631a3cfdf93bd8a7858ba4bd8f6bc0863d0dac), is it ok?

peterstarling (changes requested):

apps/crn-server/test/app.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { NewsDataProvider } from '../src/data-providers/types';
+
+describe('app is set with correct data provider for news', () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
+    jest.resetModules();
+    jest.resetAllMocks();
+    process.env = { ...OLD_ENV };
```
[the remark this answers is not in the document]
gabiayako: Without the resetAllMocks, the `newsSquidexDataProviderMock.fetch` seems that is called more than once in the tests that came after the first one that has a call to `newsSquidexDataProviderMock.fetch`.

And without `resetModules` even in the test where the calls to `newsSquidexDataProviderMock.fetch` should be 0, it appears as not zero because of the cache of previous tests.

peterstarling (approved):

gabiayako:
rebase to master to solve conflict

PR #2256, opened 2022-11-09
fix: CRN-1157 Contentful Environment Teardown

peterstarling:
First run:
image
Second run against the same PR:
image

### DIFF
--- a/.github/workflows/on-remove.yml
+++ b/.github/workflows/on-remove.yml
@@ -153,11 +153,10 @@ jobs:
           max-contentful-envs: ${{ steps.setup.outputs.max-contentful-envs }}
       - name: Contentful Teardown
         id: contentful-teardown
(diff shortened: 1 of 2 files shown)


markcsims (approved):

.github/workflows/on-remove.yml
```diff
@@ -153,11 +153,10 @@ jobs:
           max-contentful-envs: ${{ steps.setup.outputs.max-contentful-envs }}
       - name: Contentful Teardown
         id: contentful-teardown
-        if: ${{ steps.contentful-branch-state.outputs.on-branch-env == 'true' && steps.contentful-branch-state.outputs.contentful-env != 'Development' && steps.contentful-branch-state.outputs.contentful-env != 'Production' }}
```
gabiayako, line 156: is there another place where we check that this is not Development or Production?

peterstarling: yes, here:
https://github.com/yldio/asap-hub/blob/65383833db8a6b53932edf45b5d6849a7692ac9a/.github/scripts/contentful/teardown.sh#L3-L11

PR #2259, opened 2022-11-09
feat: CRN-1155 Add API Access to PR Environment in Contentful

peterstarling:
Replaces the current create environment bash script with a TS one and adds the access-key entry bit


.github/workflows/on-push-branch.yml
```diff
@@ -80,6 +80,7 @@ jobs:
       GP2_SQUIDEX_CI_CLIENT_SECRET: ${{ secrets.GP2_SQUIDEX_CI_CLIENT_SECRET }}
       GP2_SQUIDEX_SHARED_SECRET: ${{ secrets.GP2_SQUIDEX_SHARED_SECRET }}
       CONTENTFUL_MANAGEMENT_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
```
gabiayako, line 82: just out of curiosity, will this `CONTENTFUL_MANAGEMENT_TOKEN` be shared between CRN and GP2? Just now with your PR I realised that we've not been adding the suffix CRN to most of env vars

gabiayako (approved):

.github/workflows/on-push-branch.yml
peterstarling: actually I think it won't? unlike the access token I think this one works for all spaces. Or am I wrong?

markcsims: I don't think it will be shared?

markcsims (approved):

PR #2264, opened 2022-11-10
chore: Add jest onlyChanged flag to the CI

peterstarling:
- add onlyChanged flag
- add a test util WIP


.github/workflows/on-push-branch.yml
```diff
@@ -36,10 +36,42 @@ jobs:
...
+          path: |
+            .jest-cache/
+          key: jest-cache-${{ github.sha }}
+          restore-keys: |
+            jest-cache-
+      - name: Get number of CPU cores
+        id: cpu-cores
+        uses: SimenB/github-actions-cpu-cores@v1
+      - name: Run all tests
+        run: yarn test --coverage --maxWorkers ${{ steps.cpu-cores.outputs.count }} --ci
```
markcsims, line 66: could this be an action we reuse in the reusable test?

.github/actions/test-unit/action.yml
```diff
@@ -0,0 +1,49 @@
+name: 'Run unit tests'
+description: 'Check if a string is a number'
+inputs:
+  sinceMaster:
```
markcsims, line 4: should be `since-master`

markcsims (changes requested):

.github/actions/test-unit/action.yml
```diff
@@ -0,0 +1,49 @@
...
+      id: cpu-cores
+      uses: SimenB/github-actions-cpu-cores@v1
+    - name: Fetch master branch
+      shell: bash
+      if: ${{ inputs.sinceMaster == 'true' }}
+      run: |
+        chown -R root /__w/asap-hub/asap-hub
+        git fetch --no-tags --depth=1 origin master
+        git checkout -b master origin/master
+        git checkout -
```
markcsims, line 33: git switch

.github/actions/test-unit/action.yml
```diff
@@ -0,0 +1,49 @@
...
+          jest-cache-
+    - name: Get number of CPU cores
+      id: cpu-cores
+      uses: SimenB/github-actions-cpu-cores@v1
+    - name: Fetch master branch
+      shell: bash
+      if: ${{ inputs.sinceMaster == 'true' }}
+      run: |
+        chown -R root /__w/asap-hub/asap-hub
+        git fetch --no-tags --depth=1 origin master
```
markcsims, line 31: is there a way we can use the Checkout Action?

markcsims (changes requested):

.github/actions/test-unit/action.yml
peterstarling: doesnt seem to work as expected

.github/actions/test-unit/action.yml
peterstarling: i tried but it didn't work for me

markcsims (approved):

PR #2265, opened 2022-11-10
returns projects and working groups in user query

markcsims:
https://asaphub.atlassian.net/browse/GP2-129


apps/gp2-server/src/data-providers/user.data-provider.ts
```diff
@@ -213,6 +199,45 @@ export const parseGraphQLUserToDataObject = ({
...
+      if (!project.status) {
+        throw new TypeError('Status not defined');
+      }
+      return {
+        id: projectId,
+        status: project.status,
+        title: project.title || '',
+        members:
+          project.members?.map((member) => {
+            const user = member.user && member.user[0];
```
jhonny-m, line 214: this is weird, it that thing from squidex were we can add multiple members to the same role?

jhonny-m (approved):

PR #2269, opened 2022-11-14
CRN-1068 News data migration

gabiayako:
Run from the root

```
yarn workspace @asap-hub/cms-migration migrate:news
```


yarn.lock
```diff
@@ -481,6 +481,20 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@asap-hub/cms-migration@workspace:packages/cms-migration":
```
gabiayako, line 484: adding a comment in here because I can't add a comment in `.pnp.cjs`. I don't know what is going on with my local .env but everytime I run `yarn clean cache && yarn install` this file `.pnp.cjs` get's deleted. Without deleting it when I run the script I get this error

```
 yarn workspace @asap-hub/cms-migration migrate:news
/Users/gabiayako/asap/asap-hub/.pnp.cjs:78169
    throw firstError;
    ^

Error: Your application tried to access yn, but it isn't declared in your dependencies; this makes the require call ambiguous and unsound.
```

packages/cms-migration/src/news-migration.ts
```diff
@@ -0,0 +1,103 @@
...
+  const newsFields = [
+    'title',
+    'frequency',
+    'link',
+    'linkText',
+    'shortText',
+    // 'thumbnail', // TODO: deal with thumbnail
+  ];
+
+  const firstNews = squidexNews.items[0];
```
gabiayako, line 66: running only one news at once to see the errors, not planning to leave this as it is 😉

@types/contentful-html-rich-text-converter/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'contentful-html-rich-text-converter' {
+  function parseHtml(html: string): {};
```
gabiayako, line 2: this would need a better look, I only added a type because TS was complaning about `contentful-html-rich-text-converter` not having a type

peterstarling: this is fine, it doesn't come with it's own types and I take it you have looked in definitely-typed too?

I would just do
```suggestion
  function parseHtml(html: string): unknown;
```

I don't think `{}` is a good type, if you want to make it an object you could use some sort of Record type, the question is what do you need for it to work with the client lib.

@types/contentful-html-rich-text-converter/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
peterstarling, line 1: yeah you shouldn't need that? You should just add a type declaration file

@types/contentful-html-rich-text-converter/index.d.ts
peterstarling: btw maybe you should move it into the @types dir in the root?

packages/cms-migration/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
peterstarling, line 1: so you made this a separate package because in the end its all just throwaway code, yeah? I thought maybe it should be part of squidex package then. Just thinking loud..

packages/cms-migration/src/news-migration.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  [firstNews!].forEach(async ({ data, id }) => {
```
peterstarling, line 68: obvs this works for one there is just one and i guess we could still have dome processed one by one although since we are sending them along with their IDs than means they dont need to be inserted in a correct order and therefore we could insert like 10 at once or something. First I would see if the client supports batches and second I would maybe consider sending out more than 10 requests in parallel.

Now that I think about it - for News all this makes no sense cause we have like 10 news items in total, maybe something for us to think about for RO migration

packages/cms-migration/src/news-migration.ts
```diff
@@ -0,0 +1,103 @@
...
+    await contentfulEnvironment
+      .getEntry(id)
+      .then((entry) => entry.unpublish())
```
peterstarling, line 71: use await?

packages/cms-migration/src/news-migration.ts
```diff
@@ -0,0 +1,103 @@
...
+        if (newsFields.includes(key)) {
+          news[key] = { 'en-US': val?.iv };
+        } else if (key === 'text') {
+          news.text = { 'en-US': parseHtml(val?.iv as string) };
+        }
+
+        return news;
+      },
+      {},
+    );
```
peterstarling, line 87: so this is 11 lines of something we cant really read too well, wouldnt it make more sense to just declare the desired object that we want to use for inserting later? maybe have it typed already?

packages/cms-migration/src/news-migration.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (!news.frequency) {
+      news.frequency = { 'en-US': 'News Articles' };
+    }
+    news.text!['en-US']!.content = news.text?.['en-US']?.content.filter(
```
peterstarling, line 92: not sure what this does but... see above - you're having to assert not null here with `!` but i think you'd be better off just doing:
```typescript
const news = {
   title: ...,
   text: ....
}
```
right at the beginning

also consider moving the parser away from the loops etc, just have it as a function or something

yarn.lock
peterstarling: weird, is it just on your local?

packages/cms-migration/package.json
gabiayako: I thought about adding it to `package/squidex` or even `package/contentful`. But for me it seemed a little bit weird to install contentful libs in `package/squidex` (for example, contentful-management) or install `@asap/squidex` in `package/contentful`

gabiayako:
https://github.com/yldio/asap-hub/pull/2269#discussion_r1021546430
couldn't answer in the thread, I don't know why 😝 

I haven't tried to run the script outside local yet. But for the graphql data provider PR the same thing was happening: every time I tried to run `yarn cache clean && yarn install`, `.pnp.cjs` was getting deleted, I haven't committed the deletion in the graphql data provider PR though and CI was green

packages/cms-migration/package.json
gabiayako: so I created a separate package

peterstarling: yeah good point, this is fine then

@types/contentful-html-rich-text-converter/package.json
gabiayako: I thought I needed because looking at the other types folders, they all have package.json files
 
Screen Shot 2022-11-15 at 15 06 44

I can delete it, but it wouldn't follow the pattern of the other folders. Do you think it's better to remove this file?

m-doughty:
Closing old PR.

PR #2270, opened 2022-11-14
Initial commit of Recommended users

oratiu:


packages/react-components/src/atoms/Ellipsis.tsx
```diff
@@ -8,8 +8,15 @@ export const ellipsisStyles = css({
   textOverflow: 'ellipsis',
 });
 
-const Ellipsis: React.FC = ({ children }) => (
-  <span css={ellipsisStyles}>{children}</span>
+type EllipsisProps = {
+  readonly children?: React.ReactNode;
+  readonly title?: string;
```
kmaid, line 13: unit test

packages/react-components/src/organisms/DashboardRecommendedUsers.tsx
```diff
@@ -0,0 +1,31 @@
...
+    gridTemplateColumns: '1fr 1fr 1fr',
+    gridGap: `${15 / perRem}em`,
+  },
+});
+
+type DashboardRecommendedUsersProps = {
+  recommendedUsers: UserResponse[];
+};
+
+const DashboardRecommendedUsers: React.FC<DashboardRecommendedUsersProps> = ({
```
kmaid, line 21: This component doesn't seem to have allot in it. Why was it separated from recommended user? Is that design element reused?

packages/react-components/src/organisms/RecommendedUser.tsx
```diff
@@ -0,0 +1,91 @@
...
+          overrideStyles={avatarStyles}
+          imageUrl={user.avatarUrl}
+          firstName={user.firstName}
+          lastName={user.lastName}
+        />
+      </ImageLink>
+      <LinkHeadline
+        href={network({}).users({}).user({ userId: user.id }).$}
+        level={3}
+        noMargin
```
kmaid, line 58: I wonder if LinkHeadline should have had title set? I think it might already take title

packages/react-components/src/organisms/RecommendedUser.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <div css={roleStyles}>
+          <Paragraph hasMargin={false} accent="lead">
+            {user.teams[0].role} on
+          </Paragraph>
+          <Paragraph hasMargin={false} accent="lead">
+            {user.teams[0].displayName}
+          </Paragraph>
+        </div>
+      )}
+      <div css={tagListStyles}>
```
kmaid, line 75: There is already a component called taglist which this seems to be duplicating. What is the functionality difference?

kmaid (changes requested):
See comments

lctrt:
needs more tests but looks good to me otherwise :)

packages/react-components/src/molecules/TagList.tsx
```diff
@@ -62,22 +73,36 @@ interface TagListProps {
...
   enabled = true,
+  isVertical = false,
 }) =>
   tags.length ? (
-    <ul css={[listStyles, { counterReset: `tags ${tags.length}` }]}>
+    <ul
+      css={[
+        listStyles,
+        { counterReset: `tags ${tags.length}` },
+        isVertical && verticalList,
```
kmaid, line 91: Could this match MemberList?

lctrt (approved):

PR #2276, opened 2022-11-15
CRN-982 update sidebar

lctrt:


apps/storybook/src/Images.stories.tsx
```diff
@@ -13,8 +13,8 @@ import {
 export default { title: 'Atoms / Images' };
 
 export const Asap = () => <img alt="ASAP" src={asapImage} />;
-export const AsapPadded = () => <img alt="ASAP" src={asapPaddedImage} />;
-export const AsapWhite = () => <img alt="ASAP" src={asapPaddedWhiteImage} />;
+export const AsapPadded = () => asapPaddedImage;
+export const AsapWhite = () => asapPaddedWhiteImage;
```
kmaid, line 17: Are these still images or are they SVGs?

packages/react-components/src/organisms/MainNavigation.tsx
```diff
@@ -51,6 +51,9 @@ const listStyles = css({
     12 / perRem,
     'em',
   )})`,
+  li: {
```
kmaid, line 54: We try to avoid selecting other elements from parent elements to encapsulate their logic a notable exception being SVGs. Going to approve this as its only targeting things in the same file so its still encapsulated

apps/storybook/src/Images.stories.tsx
lctrt: They are SVG now

oratiu (approved):
Looks good.

PR #2279, opened 2022-11-15
Gp2-123 export user list

jhonny-m:


apps/gp2-frontend/src/users/UserList.tsx
```diff
@@ -36,6 +60,11 @@ const UserList: React.FC<UserListProps> = ({ displayFilters = false }) => {
         currentPageIndex={currentPage}
         renderPageHref={renderPageHref}
         onFiltersClick={onFiltersClick}
+        onExportClick={exportUsers}
+        isAdministrator={isAdministrator}
+        searchQuery=""
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        onSearchQueryChange={() => {}}
```
markcsims, line 67: Are these for when we implement the search?

apps/gp2-frontend/src/users/__tests__/UserList.test.tsx
```diff
@@ -48,7 +66,9 @@ const renderUserList = async (
   );
   await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
 };
-
+afterEach(() => {
+  jest.clearAllMocks();
```
markcsims, line 70: I think this should be resetAllMocks. If you change it a test will fail, I think this is because it's using a mock set up in another test.

apps/gp2-frontend/src/users/__tests__/export.test.ts
```diff
@@ -0,0 +1,233 @@
...
+import { Stringifier } from 'csv-stringify';
+
+import {
+  MAX_SQUIDEX_RESULTS,
+  squidexUsersResponseToStream,
+  userToCSV,
+} from '../export';
+
+afterEach(() => {
+  jest.clearAllMocks();
```
markcsims, line 13: resetAllMocks? But this spec is a copy/updated from CRN?

apps/gp2-frontend/src/users/__tests__/export.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    expect(userToCSV(output).secondaryPosition).toBeUndefined();
+    expect(userToCSV(output).terciaryPosition).toBeUndefined();
+  });
+  it('flattens and orders projects', () => {
+    const output: gp2Model.UserResponse = {
+      ...gp2Fixtures.createUserResponse(),
+      projects: [
+        { id: '1', title: 'project 1', status: 'Active' as const, members: [] },
+        { id: '2', title: 'project 2', status: 'Active' as const, members: [] },
+        { id: '3', title: 'project 3', status: 'Active' as const, members: [] },
```
markcsims, line 133: I don't think we need the as const

```suggestion
        { id: '1', title: 'project 1', status: 'Active', members: [] },
        { id: '2', title: 'project 2', status: 'Active', members: [] },
        { id: '3', title: 'project 3', status: 'Active', members: [] },
```

apps/gp2-frontend/src/users/export.ts
```diff
@@ -0,0 +1,90 @@
...
+  onboarded: output.onboarded ? 'Yes' : 'No',
+  primaryPosition: output.positions[0]
+    ? `${output.positions[0].role} in ${output.positions[0].department} at ${output.positions[0].institution}`
+    : undefined,
+  secondaryPosition: output.positions[1]
+    ? `${output.positions[1].role} in ${output.positions[1].department} at ${output.positions[1].institution}`
+    : undefined,
+  terciaryPosition: output.positions[2]
+    ? `${output.positions[2].role} in ${output.positions[2].department} at ${output.positions[2].institution}`
+    : undefined,
```
markcsims, line 47: Could be a function that takes a positions and an index, then returns the formatted string or not.
terciary should be tertiary

apps/gp2-frontend/src/users/export.ts
```diff
@@ -0,0 +1,90 @@
...
+    skip,
+  }: Pick<gp2.FetchUsersOptions, 'take' | 'skip'>) => Readonly<
+    Promise<gp2.ListUserResponse>
+  >,
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  transform: (result: gp2.UserResponse) => Record<string, any>,
+) => {
+  let morePages = true;
+  let currentPage = 0;
+  // csvStream.write(userFields);
```
markcsims, line 77: Should this be uncommented or removed?

jest.config.js
```diff
@@ -49,6 +62,7 @@ module.exports = {
     ...packageTestConfigs,
     ...appTestConfigs,
     ...appPaths,
+    ...packagePaths,
```
markcsims, line 65: Will this change impact CRN?

packages/frontend-utils/src/csv-export/__test__/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  close: jest.fn(),
+};
+jest.mock('streamsaver', () => ({
+  createWriteStream: jest.fn(() => ({
+    getWriter: jest.fn(() => mockWriteStream),
+  })),
+}));
+
+afterEach(() => {
+  jest.clearAllMocks();
```
markcsims, line 22: resetAllMocks?

packages/gp2-components/src/organisms/FilterSearchExport.tsx
```diff
@@ -0,0 +1,73 @@
...
+  grid: '"filter search export" /min-content auto min-content',
+  gridColumnGap: rem(32),
+  alignItems: 'center',
+  [mobileQuery]: {
+    gap: rem(24),
+    grid: `
+    "search search" max-content
+    "filter export" max-content
+    / 1fr 1fr
+    `,
```
markcsims, line 25: grid - nice :)

packages/react-components/src/molecules/SearchField.tsx
```diff
@@ -32,8 +32,10 @@ type SearchProps = Pick<
   ComponentProps<typeof TextField>,
   'value' | 'onChange'
 > &
-  Required<Pick<ComponentProps<typeof TextField>, 'placeholder'>>;
-const SearchField: React.FC<SearchProps> = (props) => {
+  Required<Pick<ComponentProps<typeof TextField>, 'placeholder'>> & {
+    padding?: boolean;
+  };
+const SearchField: React.FC<SearchProps> = ({ padding = 'true', ...props }) => {
```
markcsims, line 38: padding is a boolean, but we've defaulted it to a string. Shouldn't the typing have complained?

markcsims (changes requested):

jest.config.js
jhonny-m: it will just make it so if a package has a custom jest config it will use that one instead

apps/gp2-frontend/src/users/__tests__/export.test.ts
jhonny-m: yeah CRN is using the same

apps/gp2-frontend/src/users/UserList.tsx
jhonny-m: yeah, I've added it so it looks right, but it's still not used

packages/react-components/src/molecules/SearchField.tsx
jhonny-m: yeah idk how it doesnt complain about this one

markcsims (approved):

PR #2283, opened 2022-11-16
Attempt to get rollup working for auth0

kmaid:


apps/auth0-rules/package.json
```diff
@@ -11,6 +11,7 @@
     "typecheck": "tsc"
   },
   "dependencies": {
+    "@vedicium/auth0-actions-sdk": "^0.1.3",
```
peterstarling, line 14: this package doesn't look like it's been maintained, are we sure we don't want to just copy the types we need?

apps/auth0-rules/src/__tests__/on-execute-post-login.test.ts
```diff
@@ -0,0 +1,309 @@
...
+
+const apiBase = {
+  access: {
+    deny: jest.fn(),
+  },
+  idToken: {
+    setCustomClaim: jest.fn(),
+  },
+} as Partial<Auth0PostLoginApi> as Auth0PostLoginApi;
+beforeEach(() => {
```
peterstarling, line 63: I would add a `describe` before
```suggestion
describe('Auth0 Action - Add User Metadata', () => {
beforeEach(() => {
```
the way it was done previously

apps/auth0-rules/src/__tests__/on-execute-post-login.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        ...eventBase.secrets,
+        AUTH0_SHARED_SECRET: token,
+      },
+    },
+    apiBase,
+  );
+  expect(nock.isDone()).toBe(true);
+});
+
+describe('CRN', () => {
```
peterstarling, line 186:
```suggestion
describe('For a CRN user', () => {
```

apps/auth0-rules/src/on-execute-post-login.ts
```diff
@@ -70,53 +64,56 @@ const extractUser = (response: Auth0UserResponse): User | gp2Auth.User => {
...
+      );
+    }
   } catch (err) {
-    const error = handleError(err);
-    return callback(error);
+    let errorMessage;
+    if (err instanceof Error) {
+      errorMessage = err.message;
+    }
+    return api.access.deny(errorMessage ?? 'Unexpected Error');
```
peterstarling, line 116:
```suggestion
    let errorMessage;
    if (err instanceof Error) {
      errorMessage = err.message;
    }
    return api.access.deny(err instanceof Error ? errorMessage.err : 'Unexpected Error');
```
to me this still reads ok and avoids let

peterstarling (approved):
just tiny bits really

apps/auth0-rules/src/__tests__/on-execute-post-login.test.ts
kmaid: This doesn't make sense when the file already provides this information in the test output

apps/auth0-rules/package.json
kmaid: I rather not copy paste and maintain our own types if we don't have to.

I doubt Auth0 will release any breaking changes based on our experience thus far.

Why would we copy paste the types?

apps/auth0-rules/src/on-execute-post-login.ts
kmaid: Its test coverage I was trying to avoid. Writing it this way would require me to write mocks for got instead of using Nock for a test case demanded by typescript  I doubt will actually happen.

PR #2289, opened 2022-11-18
Crn 1068 news data migration with assets

gabiayako:
To run locally, run from the root

yarn workspace @asap-hub/cms-data-sync migrate:news

---

❗️ **Important note**: Rich text conversion is not working properly, so the script tries to convert it and publish the entry, if it gets an error, it publishes **without** the rich text.

---

This PR adds the script to migrate the news from Squidex to Contentful. In order to do that, the script

1. Fetches news from squidex using graphql: it uses graphql instead of rest, because in rest the thumbnail didn't come with its filename, mimeType, etc 
2. Clears the past entries of `news` contentful content-type
3. Converts the flat data to the respective fields
4. Converts the thumbnail: checks if it exists, if not, it uploads the thumbnail and link to the news
5. Converts rich text: parses the html to Document type expected by contentful API
7. Publishes the entries


gabiayako:
closing this one as commits were cherry picked here https://github.com/yldio/asap-hub/pull/2302

PR #2290, opened 2022-11-21
CRN-1068 News data migration (with assets, without action)

gabiayako:
To run locally, run from the root

yarn workspace @asap-hub/cms-data-sync migrate:news

---

❗️ Important note: Rich text conversion is not working properly, so the script tries to convert it and publish the entry, if it gets an error, it publishes without the rich text.

---

This PR adds the script to migrate the news from Squidex to Contentful. In order to do that, the script

Fetches news from squidex using graphql: it uses graphql instead of rest, because in rest the thumbnail didn't come with its filename, mimeType, etc
Clears the past entries of news contentful content-type
Converts the flat data to the respective fields
Converts the thumbnail: checks if it exists, if not, it uploads the thumbnail and link to the news
Converts rich text: parses the html to Document type expected by contentful API
Publishes the entries


packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,117 @@
...
+      linkText: { 'en-US': linkText! },
+      text: { 'en-US': null },
+      thumbnail: { 'en-US': null },
+    };
+
+    await convertThumbnail(newsPayload, thumbnail);
+    await convertText(newsPayload.text, text, id);
+
+    try {
+      const entry = await contentfulEnvironment.createEntryWithId('news', id, {
```
peterstarling, line 93: it would be great to use the data provider for this too (see comment above) but unfortunately we don't even expose a method for creating news so maybe let's keep the data creating bit as is

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,117 @@
...
+  linkText: { 'en-US': string | undefined };
+  text: { 'en-US': Document | null };
+  thumbnail: { 'en-US': SysLink | null };
+};
+
+const migrateNews = async () => {
+  const { contentfulEnvironment, squidexGraphqlClient } =
+    await getSquidexAndContentfulClients();
+
+  const convertThumbnail = async (
```
peterstarling, line 39: i found it a little confusing that this function was declared inside `migrateNews`, though i understand it's because you want to be able to use stuff from it's scope (like `contentfulEnvironment`) - maybe it would be better to have it declared as factory and have it accept the `contentfulEnvironment`?

or just have the function accept it as the first argument, same as `clearContentfulEntries`

packages/cms-data-sync/src/assets/assets.data-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+  contentfulEnvironment: Environment,
+  thumbnail: Pick<
+    Asset,
+    'id' | 'fileName' | 'thumbnailUrl' | 'mimeType' | 'fileType'
+  >[],
+) => {
+  if (thumbnail.length) {
+    const assetPromises: Promise<ContentfulAsset>[] = [];
+
+    for await (const thumb of thumbnail) {
```
peterstarling, line 36:
```suggestion
    for (const thumb of thumbnail) {
```

packages/cms-data-sync/src/assets/assets.data-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+              },
+            },
+          }),
+        );
+      }
+    }
+
+    const assets = await Promise.all(assetPromises);
+
+    const createdAssetsPromises: Promise<ContentfulAsset>[] = [];
```
peterstarling, line 65: looks like what you need is `assets.map`

packages/cms-data-sync/src/assets/assets.data-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+    assets.forEach((asset) =>
+      createdAssetsPromises.push(asset.processForAllLocales()),
```
peterstarling, line 67: so you create first that process for all locales, whatever that is. Wouldn't it be better to have this done for every promise in sequence? As in - you could just create an array of promises, each of which would be creating and then doing `processForAllLocales`, and then you would await it all.

packages/cms-data-sync/src/assets/assets.data-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+    const processedAssets = await Promise.all(createdAssetsPromises);
+
+    const processedAssetsPromises: Promise<ContentfulAsset>[] = [];
```
peterstarling, line 71: once again it looks like what you should use is a simple map :D

packages/cms-data-sync/src/utils/entries.ts
```diff
@@ -0,0 +1,73 @@
...
+) => {
+  const entries = await contentfulEnvironment.getEntries({
+    content_type: contentType,
+  });
+  console.log(
+    `Cleaning Contentful Entries from content-type ${contentType}...`,
+  );
+
+  const unpublishPromises: Promise<Entry>[] = [];
+  entries.items.forEach(async (entry) => {
```
peterstarling, line 16: again this could be a simple map

packages/cms-data-sync/src/utils/entries.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (entry.isPublished()) {
+      unpublishPromises.push(entry.unpublish());
+    }
+  });
+
+  await Promise.all(unpublishPromises);
+
+  const deletePromises: Promise<Entry>[] = [];
+  entries.items.forEach(async (entry) => {
```
peterstarling, line 25: same

packages/cms-data-sync/src/utils/rich-text.ts
```diff
@@ -0,0 +1,43 @@
...
+/* eslint-disable no-param-reassign */
+
+import { Document } from '@contentful/rich-text-types';
+import { parseHtml } from 'contentful-html-rich-text-converter';
+
+export const clearParsedHtmlOutput = (htmlDocument: Document) => ({
+  ...htmlDocument,
+  content: htmlDocument?.content
+    .map((node) =>
+      node.nodeType === 'unordered-list' || node.nodeType === 'ordered-list'
```
peterstarling, line 11: you may want to add a comment about what we are doing here

packages/cms-data-sync/src/utils/rich-text.ts
```diff
@@ -0,0 +1,43 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .filter((node: any) => node?.nodeType !== 'text'),
+});
+
+export const convertText = async <T extends { 'en-US': Document | null }>(
+  payload: T,
+  text: string | null,
+  id: string,
+) => {
+  if (text) {
```
peterstarling, line 33: im of an opinion that such conditionals should be done before sending it to the function, which means your text should not be null ever

packages/cms-data-sync/src/utils/rich-text.ts
```diff
@@ -0,0 +1,43 @@
...
+    const textWithoutDivTag = text.replace(/<[\\/]{0,1}(div)[^><]*>/g, '');
+
+    try {
+      const parsedHtml = parseHtml(textWithoutDivTag) as Document;
+      payload['en-US'] = clearParsedHtmlOutput(parsedHtml);
+    } catch {
+      console.log(`There is a problem converting rich text from entry ${id}`);
```
peterstarling, line 40: maybe we should be throwing a specific error here and then catching that in the main function and decide what to do with such entries there?

peterstarling (changes requested):

packages/cms-data-sync/src/news/news.data-migration.ts
gabiayako: It makes sense, I've committed in here https://github.com/yldio/asap-hub/pull/2290/commits/8abdb5e1a862cb415e625a97e28a70bc94851973

packages/cms-data-sync/src/assets/assets.data-migration.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/09981580b9770a597ef1f3f167ebc23b9d7d10e8

packages/cms-data-sync/src/assets/assets.data-migration.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/09981580b9770a597ef1f3f167ebc23b9d7d10e8

packages/cms-data-sync/src/assets/assets.data-migration.ts
gabiayako: it makes sense! I don't remember why I separated them 😛 
committed here https://github.com/yldio/asap-hub/pull/2290/commits/09981580b9770a597ef1f3f167ebc23b9d7d10e8

packages/cms-data-sync/src/utils/entries.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/fddabef2153bb8586f2cabcaee3b8cccf89e44bf

packages/cms-data-sync/src/utils/entries.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/fddabef2153bb8586f2cabcaee3b8cccf89e44bf

packages/cms-data-sync/src/utils/rich-text.ts
gabiayako: added in here https://github.com/yldio/asap-hub/pull/2290/commits/7c70e7be9e43e274abd5a374ec74d9194c9ba040

packages/cms-data-sync/src/utils/rich-text.ts
gabiayako: nice, added in here https://github.com/yldio/asap-hub/pull/2290/commits/796388b7b47a7b5ded531880f0b5404b4f3eec03 + tests fixes here https://github.com/yldio/asap-hub/pull/2290/commits/9daf1b73701bc0f0c9ccaf7412eacb81cbdfc4ce

packages/cms-data-sync/src/utils/rich-text.ts
gabiayako: added in here https://github.com/yldio/asap-hub/commit/796388b7b47a7b5ded531880f0b5404b4f3eec03 + tests fixes here https://github.com/yldio/asap-hub/commit/9daf1b73701bc0f0c9ccaf7412eacb81cbdfc4ce

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,117 @@
...
+            linkType: 'Asset',
+            // in thumbnail it's possible to load only one asset
+            id: thumbnail[0]?.id,
+          },
+        },
+      };
+    }
+  };
+
+  const gqlNews = await squidexGraphqlClient.request<
```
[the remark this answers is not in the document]
gabiayako: Replying in here so we have the answer in github as well :)

I haven't use the data provider because in rest the thumbnail didn't come with its filename, mimeType, etc and we use this to upload the asset in contentful

.yarnrc.yml
```diff
@@ -32,6 +32,7 @@ plugins:
     spec: 'https://unpkg.com/yarn-plugin-dotenv@0.1.1/bundles/@yarnpkg/plugin-dotenv.js'
 
 pnpEnableEsmLoader: false
+nodeLinker: 'pnp'
```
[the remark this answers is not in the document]
gabiayako: Maybe this is not needed 😅 I'll remove it tomorrow morning
I was having issues with module not found so I started to run `yarn cache clean` and this `nodeLinker: 'pnp'` apparently solved the issue of my `.pnp.cjs` file getting deleted after I ran `yarn install`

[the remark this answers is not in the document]
gabiayako: actually, I've noticed that if I remove this line, this error appears to me
Screen Shot 2022-11-23 at 09 34 31

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,110 @@
...
+  frequency: { 'en-US': string | null };
+  link: { 'en-US': string | undefined };
+  linkText: { 'en-US': string | undefined };
+  text: { 'en-US': Document | null };
+  thumbnail: { 'en-US': SysLink | null };
+};
+
+export const migrateNews = async () => {
+  const clients = await getSquidexAndContentfulClients();
+  if (clients) {
```
peterstarling, line 34: you shouldn't have to do this, make sure that `getSquidexAndContentfulClients` has the right return type cause i dont think we want it to be optional

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,110 @@
...
+      FetchNewsQuery,
+      FetchNewsQueryVariables
+    >(newsQuery);
+
+    await clearContentfulEntries(contentfulEnvironment, 'news');
+
+    const entries: Entry[] = [];
+
+    const newsItems = gqlNews.queryNewsAndEventsContents || [];
+    for await (const news of newsItems) {
```
peterstarling, line 46:
```suggestion
    for (const news of newsItems) {
```

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import {
+  clearContentfulEntries,
+  convertHtmlToContentfulFormat,
+  createAssetLink,
+  getSquidexAndContentfulClients,
+  publishContentfulEntries,
+} from '../utils';
+
+type NewsPayload = {
```
peterstarling, line 22: although it may sound like an overkill i think we should be able to have these types generated in the contentful package. We use graphql ones there at the moment cause we decided it was easier but i think they provide tools for working with their client libraries? If we use the generated types it will be less error prone, especially since we don't do any integration tests for this migration so types is the only real test available

packages/cms-data-sync/test/fixtures/contentful-environment.fixtures.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const contentfulEnvironmentLink: SysLink = {
+  sys: { id: 'env-id', type: 'Link', linkType: 'Environment' },
+};
+
+export const contenfulUserLink: Link<'User'> = {
+  sys: { type: 'Link', linkType: 'User', id: 'user-id' },
+};
+
+const baseEnvironment: Environment = {
```
peterstarling, line 15: this is not really a fixture, more like a mock no? maybe move to `mocks/`?

packages/cms-data-sync/test/news/news.data-migration.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    it('for a news that contains thumbnail', async () => {
+      squidexGraphqlClientMock.request.mockResolvedValueOnce(
+        squidexResponseWithThumb,
+      );
+
+      await migrateNews();
+
+      const createAssetLinkMock = createAssetLink as jest.Mock;
+
+      expect(createAssetLinkMock).toHaveBeenCalled();
```
peterstarling, line 136: you still want to see if that asset link was added right?

packages/cms-data-sync/test/news/news.data-migration.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    it('for a news that contains text', async () => {
+      squidexGraphqlClientMock.request.mockResolvedValueOnce(
+        squidexResponseWithText,
+      );
+      const convertHtmlToContentfulFormatMock =
+        convertHtmlToContentfulFormat as jest.Mock;
+
+      await migrateNews();
+
+      expect(convertHtmlToContentfulFormatMock).toHaveBeenCalled();
```
peterstarling, line 148: maybe you want to see if the output of that conversion was correctly added to the payload (same as the test above)

packages/cms-data-sync/test/news/news.data-migration.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    });
+  });
+
+  describe('error handling', () => {
+    it('outputs a message when converting from html to contenful format gives an error', async () => {
+      console.log = jest.fn();
```
peterstarling, line 154: maybe do this in a beforeAll/Each and then make sure it's assigned back in an afterAll?

packages/cms-data-sync/test/news/news.data-migration.test.ts
```diff
@@ -0,0 +1,232 @@
...
+        convertHtmlToContentfulFormat as jest.Mock;
+
+      await migrateNews();
+
+      expect(convertHtmlToContentfulFormatMock).toHaveBeenCalled();
+    });
+  });
+
+  describe('error handling', () => {
+    it('outputs a message when converting from html to contenful format gives an error', async () => {
```
peterstarling, line 153: maybe also test that when such error occurs the entry is not added?

packages/cms-data-sync/test/utils/assets.test.ts
```diff
@@ -0,0 +1,121 @@
...
+import {
+  squidexAsset,
+  contenfulAsset,
+  getContentfulEnvironmentMock,
+  contenfulUploadAssetFields,
+} from '../fixtures';
+
+describe('checkIfAssetAlreadyExistsInContentful', () => {
+  let envMock: Environment;
+  console.log = jest.fn();
```
peterstarling, line 16: you may want to assign globals back to what they were in an afterAll

packages/cms-data-sync/src/utils/entries.ts
```diff
@@ -0,0 +1,71 @@
...
+  const entries = await contentfulEnvironment.getEntries({
+    content_type: contentType,
+  });
+  console.log(
+    `Cleaning Contentful Entries from content-type ${contentType}...`,
+  );
+
+  await Promise.all(
+    entries.items.map(async (entry) => {
+      if (entry.isPublished()) {
```
peterstarling, line 17: can we not always unpublish? was just wondering if we need that extra check

peterstarling (changes requested):

packages/cms-data-sync/src/utils/entries.ts
gabiayako: although it's not flagged in the documentation (https://www.contentful.com/developers/docs/references/content-management-api/#/reference/entries/entry-publishing/unpublish-an-entry/console/js)
Screen Shot 2022-11-23 at 14 46 18

it gives an error when we try to unpublish a draft entry
Screen Shot 2022-11-23 at 14 46 29

packages/cms-data-sync/src/news/news.data-migration.ts
gabiayako: I've just tried this lib https://github.com/intercom/contentful-typescript-codegen

and it generated this

```
// THIS FILE IS AUTOMATICALLY GENERATED. DO NOT MODIFY IT.

import { Asset, Entry } from 'contentful';
import { Document } from '@contentful/rich-text-types';

export interface IMigrationFields {
  /** Migration State */
  state: Record<string, any>;

  /** Content Type ID */
  contentTypeId: string;
}

/** Meta data to store the state of content model through migrations */

export interface IMigration extends Entry<IMigrationFields> {
  sys: {
    id: string;
    type: string;
    createdAt: string;
    updatedAt: string;
    locale: string;
    contentType: {
      sys: {
        id: 'migration';
        linkType: 'ContentType';
        type: 'Link';
      };
    };
  };
}

export interface INewsFields {
  /** Title */
  title: string;

  /** Short Text */
  shortText?: string | undefined;

  /** Thumbnail */
  thumbnail?: Asset | undefined;

  /** Frequency */
  frequency: 'Biweekly Newsletter' | 'CRN Quarterly' | 'News Articles';

  /** External Link */
  link?: string | undefined;

  /** External Link Text */
  linkText?: string | undefined;

  /** Text */
  text?: Document | undefined;
}

/** ASAP Hub News */

export interface INews extends Entry<INewsFields> {
  sys: {
    id: string;
    type: string;
    createdAt: string;
    updatedAt: string;
    locale: string;
    contentType: {
      sys: {
        id: 'news';
        linkType: 'ContentType';
        type: 'Link';
      };
    };
  };
}

export type CONTENT_TYPE = 'migration' | 'news';

export type IEntry = IMigration | INews;

export type LOCALE_CODE = 'en-US';

export type CONTENTFUL_DEFAULT_LOCALE_CODE = 'en-US';
```

It's not like the payload for creating an entry but what comes back from the API. I google it for some time but I could not find another lib that does that. Since it doesn't have the types that we need, I won't commit that, is that ok?

packages/cms-data-sync/src/news/news.data-migration.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/5a9e5992424256a2347b07e0c6c02f5246b31781

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -0,0 +1,110 @@
...
+        shortText: { 'en-US': shortText },
+        frequency: { 'en-US': frequency || 'News Articles' },
+        link: { 'en-US': link! },
+        linkText: { 'en-US': linkText! },
+        text: { 'en-US': null },
+        thumbnail: { 'en-US': null },
+      };
+
+      if (thumbnail?.length) {
+        newsPayload.thumbnail['en-US'] = await createAssetLink(
```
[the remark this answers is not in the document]
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/1d51ff82f4c0fa1915e1f61a18495dd2b307394a

packages/cms-data-sync/test/news/news.data-migration.test.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/a35023f7732ada35ece5037ebd8d96545d481d0c

packages/cms-data-sync/test/news/news.data-migration.test.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/e62fc93f1ebde09cb70a5ac0a15352ef60fdf5c4

packages/cms-data-sync/src/news/news.data-migration.ts
gabiayako: Actually this wasn't giving any Typescript error, I just added to make the test pass without giving me an error that `{ contentfulEnvironment, squidexGraphqlClient }` could not be destructured this way because we can't destructure `undefined`.

I spent some hours yesterday and today trying to make the tests pass without this and I couldn't find a way. I've tried that https://github.com/yldio/asap-hub/pull/2290/commits/e999c8b052e4634c0d86a5e2ec6439be43a5eb48 that was passing locally but this way, but it fail on CI because it requires the env vars

packages/cms-data-sync/test/fixtures/contentful-environment.fixtures.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/ea48e78be5f28babe4bad30b406bb7466506473e

packages/cms-data-sync/test/news/news.data-migration.test.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/4604b7aefa73f806801f038969bc5782e7d67a22

packages/cms-data-sync/test/utils/assets.test.ts
gabiayako: committed here https://github.com/yldio/asap-hub/pull/2290/commits/5b0a38472172a5267a15b2335793a7516e8c6ca5

packages/cms-data-sync/src/news/news.data-migration.ts
peterstarling: yeah its fine i thought it would generate the input. Weird

packages/cms-data-sync/src/news/news.data-migration.ts
peterstarling: ok i definitely don't like adding something just for tests, surely if you add a jest mock and implement it to always return the clients it would never return undefined?

peterstarling (approved):

PR #2291, opened 2022-11-21
adds extra contact information to user

markcsims:
Adding the following the GP2 users:

- [x] telephone number
- [x] telephone country code
- [x] secondary email


apps/gp2-server/src/controllers/user.controller.ts
```diff
@@ -20,26 +27,31 @@ export default class Users implements UserController {
...
-    return this.fetchById(id);
+    return this.fetchById(id, currentUserId);
   }
 
   async fetch(options: gp2.FetchUsersOptions): Promise<gp2.ListUserResponse> {
     const { total, items: users } = await this.userDataProvider.fetch(options);
 
-    const items = total > 0 ? users.map(parseUserToResponse) : [];
+    const items =
+      total > 0 ? users.map((user) => parseUserToResponse(user)) : [];
```
jhonny-m, line 40: that's the same as before no?

apps/gp2-server/src/routes/user.route.ts
```diff
@@ -56,7 +56,9 @@ export const userRouteFactory = (userController: UserController): Router => {
   userRoutes.get<{ userId: string }>('/users/:userId', async (req, res) => {
     const { userId } = validateUserParameters(req.params);
 
-    const user = await userController.fetchById(userId);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const loggedInUserId = req.loggedInUser!.id;
+    const user = await userController.fetchById(userId, loggedInUserId);
```
jhonny-m, line 61: dont we know here if it it the same id? and we could just pass a boolean?

apps/gp2-server/src/controllers/user.controller.ts
markcsims: No. parseUserToResponse now takes 2 parameters the user, and the current user id.

Like this users.map(parseUserToResponse), parseUserToResponse will be called with the user as the first param, and the index as the second.

i.e. the parameters of the callback. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

apps/gp2-server/src/routes/user.route.ts
markcsims: Yes, I was looking to put all the logic in the controller and leave the routing to do routes. Doing it like this is consistent with the project and working groups controllers - where they too take the loggedInUserId

jhonny-m (approved):

PR #2292, opened 2022-11-21
feat: CRN-1118 Add Working Group profile page

peterstarling:
- Adds API endpoint for working-groups 
- Adds WorkingGroupProfile page and exposes that through `network/working-groups/<id>` url


lctrt (approved):

apps/crn-server/src/app.ts
```diff
@@ -259,6 +270,8 @@ export const appFactory = (libs: Libs = {}): Express => {
   const userController =
     libs.userController || new Users(userDataProvider, assetDataProvider);
   const labsController = libs.labsController || new Labs(squidexGraphqlClient);
+  const workingGroupsController =
+    libs.workingGroupsController || new WorkingGroups(workingGroupDataProvider);
```
gabiayako, line 274: really small one, but maybe all variables should be in singular or plural, like

`workingGroupsController` / `workingGroupsDataProvider`

or 

`workingGroupController` / `workingGroupDataProvider`


I always have a hard time to find using ctrl + F when the variables are mixed in singular and plural 😅

apps/crn-frontend/src/network/working-groups/WorkingGroupProfile.tsx
```diff
@@ -0,0 +1,41 @@
...
+      lastUpdated: new Date(workingGroup.lastModifiedDate).toISOString(),
+    };
+
+    return (
+      <WorkingGroupPage {...props}>
+        <WorkingGroupAbout
+          members={props.members}
+          description={props.description}
+          pointOfContact={props.pointOfContact}
+        />
```
gabiayako, line 33: I wonder if we don't need something like this

https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/network/groups/GroupProfile.tsx#L140


If we access https://2292.hub.asap.science/network/working-groups/37381e21-1e96-44fb-b55d-cd729a307e55

Screen Shot 2022-11-21 at 13 50 06

The tab "About" appears as unselected, but if we access https://2292.hub.asap.science/network/groups/f4823066-418e-4bc4-9cda-0042e59cf47e and we wait, the url changes to `/about` and "About" tab appears as selected

gabiayako (approved):

apps/crn-frontend/src/network/working-groups/WorkingGroupProfile.tsx
peterstarling: agree, although that would be part of the `WorkingGroupHeader.tsx` file and I think that bit is still WIP so it can be tackled later, for instance when more tabs are added, what do you think?

gabiayako: I wonder if it isn't part of [apps/crn-frontend/src/network/working-groups/WorkingGroupProfile.tsx](https://github.com/yldio/asap-hub/pull/2292/files/76c6a487cf880f63d79cb8285fce1e3bc90a5b57#diff-e6e86287bdeab9dcfb3ac0c6241bb7b581e788f50bfdf8b3b2784c8209e985d9) because in the example I've sent it's part of `GroupProfile` page https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/network/groups/GroupProfile.tsx#L140

PR #2304, opened 2022-11-25
CRN-1172 | Use yld fork of contentful-html-rich-text-converter.

m-doughty:
# CRN-1172 | Use yld fork of `contentful-html-rich-text-converter`

## Description of Problem

We rely on `contentful-html-rich-text-converter` for our Squidex to Contentful data sync operations.

The upstream `contentful-html-rich-text-converter` has not received an update for four years, so we've forked it to make fixes.

## Description of Fix

This PR changes our dependency from upstream `contentful-html-rich-text-converter` to the YLD fork.


gabiayako (approved):
🎉

PR #2310, opened 2022-11-29
User Profile - Contact information, Keywords, Biography and Funding Streams

markcsims:
- [x] add to squidex schema
- [x] return values in api
- [ ] add biography to page
- [ ] add contact information to page
- [ ] add keywords to page
- [ ] add funding streams to page
- [ ] check export


apps/gp2-server/src/validation/user.validation.ts
```diff
@@ -69,6 +69,16 @@ const userPatchRequestValidationSchema: JSONSchemaType<gp2.UserPatchRequest> = {
         number: { type: 'string', nullable: true },
       },
     },
+    keywords: {
+      type: 'array',
+      items: {
+        type: 'string',
+        enum: [...keywords, '', null],
```
jhonny-m, line 76: can it be an empty string?

jhonny-m (approved):
Can you check if the funding streams affects the export?

apps/gp2-server/src/validation/user.validation.ts
markcsims: no, and it can't be null either. This is copied over from CRN.

PR #2315, opened 2022-11-29
User Profile - Contact information, Keywords, Biography and Funding Streams - Client

markcsims:
- [x] add to squidex schema
- [x] return values in api
- [x] add biography to page
- [x] add contact information to page
- [x] add keywords to page
- [x] add funding streams to page
- [x] check export


packages/gp2-components/src/templates/UserOverview.tsx
```diff
@@ -0,0 +1,85 @@
...
+  padding: `${rem(32)} 0 ${rem(48)}`,
+});
+
+const contentStyles = css({
+  paddingTop: rem(32),
+});
+
+const columnStyles = css({
+  display: 'grid',
+  columnGap: rem(32),
```
jhonny-m, line 33: won't just `gap` work for both?

packages/gp2-components/src/templates/UserOverview.tsx
```diff
@@ -0,0 +1,85 @@
...
+const UserOverview: React.FC<UserOverviewProps> = ({
+  biography,
+  email,
+  secondaryEmail,
+  keywords,
+  fundingStreams,
+}) => (
+  <div css={containerStyles}>
+    <div css={columnStyles}>
+      <Card overrideStyles={cardStyles}>
```
jhonny-m, line 50: to avoid the overrideStyles, should we just pass this to a container div in the card?

jhonny-m (approved):

packages/gp2-components/src/templates/UserOverview.tsx
markcsims: Done.

packages/gp2-components/src/templates/UserOverview.tsx
markcsims: done

PR #2316, opened 2022-11-30
Gp2 111 modal

jhonny-m:


apps/gp2-frontend/src/users/__tests__/api.test.tsx
```diff
@@ -95,6 +95,48 @@ describe('getUsers', () => {
   });
 });
 
+describe('patchUser', () => {
+  it('makes an authorized PATCH request for the user id', async () => {
+    const patch = { firstName: 'New Name' };
+    nock(API_BASE_URL, { reqheaders: { authorization: 'Bearer x' } })
+      .patch('/users/42')
+      .reply(200, {});
```
markcsims, line 103: How does this work with the nock?

In the other describe blocks we have a nock.cleanAll? Shouldn't we need this here?

packages/gp2-components/src/organisms/EditUserModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+    width: '100%',
+  },
+});
+
+type EditUserModalProps = Partial<gp2.UserResponse> & {
+  title: string;
+  description: string;
+  dirty: boolean;
+  backHref: string;
+  onSave?: () => void | Promise<void>;
```
markcsims, line 41:
```suggestion
  onSave?: () => Promise<void>;
```

packages/gp2-components/src/organisms/EditUserModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  children: (
+    state: { isSaving: boolean },
+    asyncOnSave: () => void,
```
markcsims, line 44: Should this return a Promise?

apps/gp2-frontend/src/users/__tests__/api.test.tsx
jhonny-m: I just did the same test as crn, which I mostly for coverage, we are just testing that a call to the right endpoint with the right arguments is made

packages/gp2-components/src/organisms/__tests__/EditUserModal.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    });
+    expect(screen.getByRole('dialog')).toContainElement(
+      screen.getByText('Content'),
+    );
+  });
+  it('calls the onSave function when the save button is pressed', () => {
+    const handleSave = jest.fn();
+    render(
+      <EditUserModal {...defaultProps} onSave={handleSave}>
+        {() => 'content'}
```
csbarbosa, line 44: Why is this here?

jhonny-m: the component expects the children to be a function

packages/gp2-components/src/organisms/EditUserModal.tsx
markcsims: I think we discussed this and decided to leave it as it.

PR #2327, opened 2022-12-01
CRN-1075 Filtering non alumni users from dashboard latest users

gabiayako:


packages/model/src/user.ts
```diff
@@ -5,7 +5,9 @@ import { TeamRole } from './team';
 export const userRole = ['Staff', 'Grantee', 'Guest', 'Hidden'] as const;
 export type Role = typeof userRole[number];
 
-export const userTags = ['CRN Member', 'Alumni Member'] as const;
+export const activeUserTag = 'CRN Member';
+export const inactiveUserTag = 'Alumni Member';
```
gabiayako, line 9: I've decided to put this in a variable because in case we change how we call an active/inactive user it's easier to propagate the change

ricardocarvalh0 (approved):

PR #2328, opened 2022-12-01
CRN-1192 | Adds optional verbose debug logs to cms-data-sync

m-doughty:
# CRN-1192 | Adds optional verbose debug logs to cms-data-sync

## Ticket

[CRN-1192](https://asaphub.atlassian.net/browse/CRN-1192)

## Description

- Adds an environment variable, `VERBOSE_DATA_SYNC` to enable verbose logging.
- Colors & tags log outputs for readability.
- The logs show the input & output for each entry.

## Screenshot

image

## Testing

Enable `VERBOSE_DATA_SYNC=true` in .env & run `yarn workspace @asap-hub/cms-data-sync migrate:news`


packages/cms-data-sync/src/utils/rich-text.ts
```diff
@@ -34,7 +35,26 @@ export const convertHtmlToContentfulFormat = (html: string) => {
   // an important part of rich text input anyway, so we
   // can just remove them here
   const htmlWithoutDivTag = html.replace(/<[\\/]{0,1}(div)[^><]*>/g, '');
+  verboseLog(`HTML pre-parsed:\n${html}`);
+  verboseLog(`HTML post-parsed:\n${html}`);
```
gabiayako, line 39: should the post-parsed be `htmlWithoutDivTag`?

m-doughty: Good spot.

m-doughty:
> I think we can remove this line now https://github.com/yldio/asap-hub/blob/master/packages/cms-data-sync/src/news/news.data-migration.ts#L90 :)

I think the error log should happen whether in verbose mode or not.

packages/cms-data-sync/src/utils/setup.ts
```diff
@@ -6,6 +6,9 @@ import { createClient } from 'contentful-management';
 
 import { getAccessTokenFactory, SquidexGraphql } from '@asap-hub/squidex';
 
+export const isVerbose = () =>
+  process.env.VERBOSE_DATA_SYNC && process.env.VERBOSE_DATA_SYNC == 'true';
```
gabiayako, line 10: it seems the test is failing due to that in CI 
```suggestion
  process.env.VERBOSE_DATA_SYNC && process.env.VERBOSE_DATA_SYNC === 'true';
```

packages/cms-data-sync/src/utils/verbose.ts
```diff
@@ -0,0 +1,10 @@
+import { isVerbose } from './setup';
+
+export const verboseLog = (log: string): null => {
+  if (!isVerbose()) {
+    return null;
+  }
+
+  console.log('\x1b[32m', `[DEBUG] ${log}`);
```
gabiayako, line 8: the other error is due to this console log :) 

we could add `/* eslint-disable no-console */`

gabiayako (approved):

packages/cms-data-sync/test/news/news.data-migration.test.ts
```diff
@@ -246,7 +246,8 @@ describe('migrateNews', () => {
       await migrateNews();
 
       expect(console.log).toHaveBeenCalledWith(
-        'There is a problem converting rich text from entry news-1',
+        '\x1b[31m',
```
gabiayako, line 249: we can call this `RED_COLOR` now and replace the code by the variable (RED_COLOR/GREEN_COLOR/etc) in the other tests, right? :)

gabiayako (approved):

packages/cms-data-sync/test/news/news.data-migration.test.ts
m-doughty: imo correct way to do this would be to mock `logger` instead of `console.log` as we shouldn't be testing the underlying implementation of `logger` at this level.

PR #2343, opened 2022-12-07
feat: CRN-1051 Add Contentful GraphQl validation check

peterstarling:


.github/workflows/reusable-verify.yml
```diff
@@ -93,6 +96,32 @@ jobs:
...
+          environment-name: ${{ inputs.environment-name }}
+      - name: Verify CRN Contentful Graphql Schema
+        if: ${{ inputs.environment-name=='Branch' }}
+        run: |
+          yarn workspace @asap-hub/contentful schema:update:crn
+          git --git-dir=.git diff --exit-code ./packages/contentful/src/crn/autogenerated-gql/
+        env:
+          CONTENTFUL_ACCESS_TOKEN: ${{ secrets.CRN_CONTENTFUL_ACCESS_TOKEN }}
+          CONTENTFUL_SPACE_ID: ${{ steps.setup.outputs.contentful-space-id }}
+          CONTENTFUL_ENV_ID: Development
```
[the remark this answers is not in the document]
m-doughty: We could use the `contentful-setup` Action to get the right Env?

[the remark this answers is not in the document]
peterstarling: you guys are right, this should validate against the PR environment not dev, i made a mistake

.github/workflows/reusable-verify.yml
```diff
@@ -93,6 +100,32 @@ jobs:
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: ${{ inputs.environment-name }}
```
gabiayako, line 118: is this needed even if we add `setup` in here?

```
  verify:
    needs: [setup, deployment]
```

.github/workflows/on-push-branch.yml
```diff
@@ -122,13 +122,15 @@ jobs:
       GP2_AUTH0_SHARED_SECRET: ${{ secrets.GP2_AUTH0_SHARED_SECRET }}
 
   verify:
-    needs: [deployment]
+    needs: [setup, deployment]
```
gabiayako, line 125: this would need to be added in `on-push-master` as well, right? :)

gabiayako (approved):
Nice! 🎉

.github/workflows/reusable-verify.yml
peterstarling: it is, at the on-push-branch level we do a different setup, one which runs all the contentful checks (which the one above does not), see:
https://github.com/yldio/asap-hub/blob/4adc41c58b228c739bf69999e6923d8102155904/.github/workflows/on-push-branch.yml#L27

PR #2354, opened 2022-12-12
feat: CRN-1159 Add Page data migration

peterstarling:


packages/cms-data-sync/src/news/news.queries.ts
```diff
@@ -2,7 +2,7 @@ import { gql } from 'graphql-tag';
 
 export const newsQuery = gql`
   query FetchNews {
-    queryNewsAndEventsContents {
+    queryNewsAndEventsContents(top: 100) {
```
peterstarling, line 5: I think it would have only done 8 by default previously

packages/cms-data-sync/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
-    "migrate:news": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' ts-node ./src/news/news.data-migration.ts",
+    "migrate:news": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' ts-node ./src/index.ts",
```
gabiayako, line 14: now we're not migrating only news in this script 😛 but if you want, we can rename it in another PR

packages/cms-data-sync/src/news/news.data-migration.ts
```diff
@@ -1,113 +1,89 @@
...
-
 import {
-  clearContentfulEntries,
   convertHtmlToContentfulFormat,
   createAsset,
   getSquidexAndContentfulClients,
   logger,
-  publishContentfulEntries,
 } from '../utils';
+import { migrateFromSquidexToContentfulFactory } from '../utils/migration';
```
gabiayako, line 15: maybe it would be nice to add 
```
export * from './migration';
```
to `packages/cms-data-sync/src/utils/index.ts` so we can import this from `'../utils'` only

packages/cms-data-sync/src/pages/pages.data-migration.ts
```diff
@@ -0,0 +1,78 @@
+import { Document } from '@contentful/rich-text-types';
+import {
+  FetchPagesQuery,
+  FetchPagesQueryVariables,
+} from '../autogenerated-gql/graphql';
+import { logger } from '../utils/logs';
+import { migrateFromSquidexToContentfulFactory } from '../utils/migration';
+import { convertHtmlToContentfulFormat } from '../utils/rich-text';
+import { getSquidexAndContentfulClients } from '../utils/setup';
```
gabiayako, line 9: we can import both from `'../utils'` :)

gabiayako (approved):
Really nice the refactor to `utils/migration` 👏👏👏

packages/cms-data-sync/src/news/news.data-migration.ts
peterstarling: so i'm not a fan having those index files anywhere, we only had them in packages because they were necessary - we had to point package.json to a single file where all the exports were, but for internal use i think its a little misleading. TypeScript actually helps you with auto completing these paths so it's not like you have to type them every time and I see no other benefit of having those re-exporting index files

gabiayako (approved):

PR #2368, opened 2022-12-16
Backend developer logging

kmaid:


apps/crn-server/test/middleware/user-logger-handler.test.ts
```diff
@@ -32,7 +32,7 @@ describe('User info logging handler', () => {
     throw new Error('error message');
   });
   logRoutes.get('/events/custom-log', async (req, res) => {
-    res.send('foo');
+    res.log('test');
```
kmaid, line 35: This fixed the test but I am not sure if we should be replacing the logger on response another way that makes send work aswell

peterstarling (approved):

PR #2369, opened 2022-12-19
CRN-1202 Filter speakers without teams from event response

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-1202

This PR removes speakers without team from the event response. Before, we were throwing an error when this happened, so the user would see:

not-filtering

Now, if a speaker is added without a team in squidex

squidex-without-team

the speaker will not be display in the event and the backend **will not** throw an error
filtering

Link Event with Speaker without Team: https://2369.hub.asap.science/events/1a5a010f-a789-4261-a3ba-70d5dcb947d1


peterstarling (approved):

apps/crn-server/test/entities/event.test.ts
```diff
@@ -34,21 +34,21 @@ describe('events entity', () => {
...
-          ],
+          speakers: [squidexSpeaker],
         },
       };
-      expect(() => parseGraphQLEvent(event)).toThrowError(
-        'Either team or external author is required',
-      );
+      expect(parseGraphQLEvent(event)).toEqual({
+        ...getEventResponse(),
+        speakers: [],
```
peterstarling, line 50: maybe add another speaker with a team and then expect only that user in the response? that way you are testing filtering them out, whereas this technically still leaves a lot of room for error (ie - if any user is missing a team it could always return an empty array)

peterstarling (changes requested):

peterstarling (approved):

PR #2371, opened 2022-12-19
feat: GP2-142 Add User Projects and Working Groups

peterstarling:


packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+} from '@asap-hub/react-components';
+import { css } from '@emotion/react';
+import { useState } from 'react';
+import { StatusPill } from '../molecules';
+import colors from '../templates/colors';
+
+const { rem, tabletScreen } = pixels;
+
+const contentStyles = css({
+  padding: `${rem(16)} 0`,
```
markcsims, line 19: try rowGap, we've been trying to use gap/rowGap and columnGap.

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+});
+const textStyles = css({
+  color: lead.rgb,
+});
+const hideStyles = css({
+  [`:nth-of-type(n+6)`]: { display: 'none' },
+});
+const headingTopStyles = css({
+  display: 'none',
+  [`@media (min-width: ${tabletScreen.min}px)`]: {
```
markcsims, line 37: looks like this is not used much in gp2. Try using the media queries from 

https://github.com/yldio/asap-hub/blob/dace4dcfa6933c8a6b1063d6fbde1efb130b59ed/packages/gp2-components/src/layout.ts#L15-L16

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+const headingTopMainStyles = css({
+  flex: '40% 0 0',
+});
+const headinTopSecondaryStyles = css({
+  flex: '30% 0 0',
+});
+const listElementStyles = css({
+  display: 'flex',
+  alignItems: 'flex-start',
+  gap: '32px',
```
markcsims, line 50:
```suggestion
  gap: rem(32),
```

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+    <>
+      <div css={[contentStyles]}>
+        <Paragraph hasMargin={false} accent="lead">
+          {firstName} has been involved in the following GP2 projects:
+        </Paragraph>
+      </div>
+      <div css={headingTopStyles}>
+        <h4 css={headingTopMainStyles}>Name</h4>
+        <h4 css={headinTopSecondaryStyles}>Role</h4>
+        <h4 css={headinTopSecondaryStyles}>Status</h4>
```
markcsims, line 114: for the h4 tags here and below, we could try to use the https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/atoms/Headline4.tsx component.

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+      </div>
+      {projects.map((project, index) => (
+        <div key={`user-project-${index}`} css={getProjectListStyles()}>
```
markcsims, line 117: instead of index, would be good to see if we could use project.id

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <div css={[listElementStyles, listElementMainStyles]}>
+            <h4 css={headingListStyles}>Name:</h4>
+            <Link underlined={true} href="https://google.com">
```
markcsims, line 120:
```suggestion
            <Link underlined href="https://google.com">
```

markcsims: and update it to link through to the project details page.

markcsims: not sure about it needing to be underlined?

packages/gp2-components/src/organisms/UserProjects.tsx
```diff
@@ -0,0 +1,154 @@
...
+      ))}
+      {projects.length > minimumProjectsToDisplay && (
+        <div css={buttonWrapperStyles}>
+          <Button linkStyle onClick={() => setExpanded(!expanded)}>
+            <span
+              css={{
+                display: 'inline-grid',
+                verticalAlign: 'middle',
+                paddingRight: rem(12),
+              }}
```
markcsims, line 142: all the other stylings are extracted to a variable (even if they are used once)

packages/gp2-components/src/organisms/__tests__/UserProjects.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    }));
+  const firstName: gp2.UserResponse['firstName'] = 'John';
+  const id = 'user-id';
+  const renderUserProjects = (projects: Project[]) =>
+    render(<UserProjects projects={projects} firstName={firstName} id={id} />);
+  it('renders the short text when there are no projects', () => {
+    renderUserProjects([]);
+    expect(
+      screen.getByText(`${firstName} has been involved`, { exact: false }),
+    ).toBeInTheDocument();
```
markcsims, line 23: recently we've moved to toBeVisible when checking for existence on the page. 

When we check the element is not there we use not.toBeInTheDocument.

packages/gp2-components/src/organisms/UserProjects.tsx
peterstarling: its underlined as per designs but im not sure about the property. @kmaid seems to think we should have a theme for this
which makes sense to me except we're gonna need some sort of secondary theme anyway cause until now all links in GP2 didn't have an underline either

packages/gp2-components/src/organisms/UserProjects.tsx
peterstarling: this is copied from https://github.com/yldio/asap-hub/blob/dace4dcfa6933c8a6b1063d6fbde1efb130b59ed/packages/gp2-components/src/organisms/ProjectMilestones.tsx#L73-L78

markcsims (approved):

PR #2375, opened 2022-12-21
addind dashboard page and welcome back banner

jhonny-m:


apps/gp2-frontend/src/AuthenticatedApp.tsx
```diff
@@ -79,7 +84,14 @@ const AuthenticatedApp: FC<Record<string, never>> = () => {
       <Switch>
         <Route exact path={path}>
           <Frame title="Dashboard">
-            <Dashboard />
+            <UserContext.Provider
```
markcsims, line 87: Why do we need to use context, we could pass these to the Dashboard using props.

markcsims (approved):

csbarbosa (approved):

PR #2388, opened 2022-12-28
Crn 1183 add article output form for working groups

oratiu:


packages/react-components/src/utils/researchOutputForm.ts
```diff
@@ -0,0 +1,186 @@
...
+      {
+        label: team?.displayName || '',
+        value: team?.id || '',
+        isFixed: true,
+      },
+    ]
+  );
+};
+
+export const isDirtyEditMode = (
```
kmaid, line 63: I think that this is very repetitive. Could we make a function that takes the initial state of the form (eg empty or the api version of the research output data) it return a function that you can call with the current state and it will compare. I don't think this will need to vary between the forms as

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  if (canCreateUpdate && workingGroup) {
+    return (
+      <Frame title="Share Working Group Research Output">
+        <ResearchOutputPage
+          tagSuggestions={researchSuggestions}
+          documentType="Article"
+          getLabSuggestions={getLabSuggestions}
+          getAuthorSuggestions={(input) =>
+            getAuthorSuggestions(input).then((users) =>
```
kmaid, line 71: This should be a hook like the others

apps/crn-frontend/src/network/working-groups/api.ts
```diff
@@ -42,3 +48,27 @@ export const getWorkingGroup = async (
   }
   return resp.json();
 };
+
+export const createWorkingGroupResearchOutput = async (
```
kmaid, line 52: I don't think this is required as its a duplicate of the existing. I think because the endpoint is the same at worst we should just have the function take both types of research output

apps/crn-frontend/src/network/working-groups/state.ts
```diff
@@ -122,3 +133,70 @@ export const useWorkingGroups = (options: GetListOptions) => {
...
+  workingGroup: WorkingGroupDataObject | undefined,
+): boolean => {
+  const user = useCurrentUserCRN();
+  if (workingGroup === undefined || user === null) return false;
+
+  const { leaders } = workingGroup;
+
+  const hasPermissions = leaders.some((leader) => {
+    if (leader.user.id === user.id && leader.role === 'Project Manager') {
+      return true;
```
kmaid, line 184: This definitely needs a unit test

apps/crn-frontend/src/network/working-groups/state.ts
```diff
@@ -122,3 +133,70 @@ export const useWorkingGroups = (options: GetListOptions) => {
...
+    }
+    return false;
+  });
+  return hasPermissions;
+};
+
+export const usePostWorkingGroupResearchOutput = () => {
```
kmaid, line 191: Duplicate

packages/model/src/research-output.ts
```diff
@@ -128,6 +136,19 @@ export const researchOutputMapType = (
...
+  publishingEntity?: string | null,
+): ResearchOutputPublishingEntities | null => {
+  if (
+    publishingEntity &&
+    (publishingEntity === 'Team' || publishingEntity === 'Working Group')
+  ) {
+    return publishingEntity;
+  }
+
+  return null;
```
kmaid, line 149: This a case that should never happen. Either we throw because it shouldn't happen as we have migrated. Or we should default to Team and just have new entities correctly tagged.

packages/model/tests/research-output.test.ts
```diff
@@ -34,6 +35,17 @@ describe('Research Output Model', () => {
       expect(researchOutputMapType('NotACloningType')).toBeNull();
     });
   });
+
+  describe('Publishing Entities', () => {
+    it('should map to the correct entity', () => {
+      expect(researchOutputMapPublishingEntity('Team')).toEqual('Team');
+    });
+    it('should map to null', () => {
```
kmaid, line 43: Should be removed

packages/react-components/src/organisms/ResearchOutputContributorsCard.tsx
```diff
@@ -78,7 +81,7 @@ const ResearchOutputContributorsCard: React.FC<
     <AuthorSelect
```
kmaid, line 81: This field is marked as required however there is no validation to ensure that is been included. A plan needs to be made to address that

packages/react-components/src/templates/ResearchOutputPage.tsx
```diff
@@ -21,16 +25,21 @@ const ResearchOutputPage: React.FC<ResearchOutputPageProps> = ({
   documentType,
   researchOutputData,
   isEditMode,
+  publishingEntity = 'Team',
   ...formProps
```
kmaid, line 29: I think this file just makes things more complex and will need to be duplicated due to the new Working Group Report type. I think we should just move all the logic from Research Output Form into this file and rename it to ResearchOutputTeamForm.tsx . We can then have another called ResearchOutputWorkingGroupForm.tsx which will have the special handling around that

packages/react-components/src/organisms/ResearchOutputContributorsCard.tsx
oratiu: Created ticket, Tiff told me we will refine it next week.

packages/react-components/src/organisms/ResearchOutputFormSharingCard.tsx
```diff
@@ -135,7 +138,11 @@ const ResearchOutputFormSharingCard: React.FC<
       <LabeledDropdown<ResearchOutputType | ''>
         title="Type"
         subtitle="(required)"
-        description={`Select the option that applies to this ${documentType.toLowerCase()}.`}
+        description={`${
```
oratiu, line 141: Pass in the description directly from the parent component based on the publishing entity.

kmaid (approved):

lctrt (approved):

PR #2393, opened 2023-01-02
GP2-137 - Get the right svg's for icons

csbarbosa:
(Ticket:https://asaphub.atlassian.net/browse/GP2-137)
Get the right svg's for icons


packages/react-components/src/atoms/NavigationLink.tsx
```diff
@@ -67,18 +67,26 @@ const squareBorderStyles = css({
...
+  svg: { stroke: 'initial' },
+  '&.active-link > p > span > svg': { stroke: 'initial' },
+});
+
 interface NavigationLinkProps {
   readonly href: string;
   readonly enabled?: boolean;
   readonly icon?: JSX.Element;
   readonly squareBorder?: boolean;
+  readonly hasStrokeWidth?: boolean;
```
jhonny-m, line 81: I don't like the name of the prop, I would rather name it like `useStroke`

packages/react-components/src/atoms/NavigationLink.tsx
```diff
@@ -89,6 +97,7 @@ const NavigationLink: React.FC<NavigationLinkProps> = ({
         activeClassName={activeClassName}
         css={({ colors }) => [
           styles,
+          !hasStrokeWidth && noStrokeStyles,
```
jhonny-m, line 100: instead of re-writting styles, we can just separate the stroke styles and apply them if the variable is `true`

PR #2411, opened 2023-01-06
Added isRequired validaiton on multiSelect

oratiu:


apps/crn-frontend/src/network/working-groups/__tests__/WorkingGroupOutput.test.tsx
```diff
@@ -94,8 +94,6 @@ describe('WorkingGroupOutput', () => {
     userEvent.click(screen.getByText('Abu-Remaileh, M 1'));
     userEvent.click(screen.getByRole('textbox', { name: /Labs/i }));
     userEvent.click(screen.getByText('Example 1 Lab'));
-    userEvent.click(screen.getByRole('textbox', { name: /Authors/i }));
-    userEvent.click(screen.getByText('Person A 3'));
```
kmaid, line 98: ?

packages/react-components/src/organisms/AuthorSelect.tsx
```diff
@@ -57,6 +57,8 @@ const AuthorSelect: React.FC<AuthorSelectProps> = (props) => (
   <LabeledMultiSelect<AuthorOption>
     {...props}
     creatable={true}
+    isRequired={props.isRequired}
```
kmaid, line 60:
```suggestion
```
its already there from ...props.

packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -122,6 +124,8 @@ export type MultiSelectProps<T extends MultiSelectOptionsType> = {
   readonly values?: OptionsType<T>;
   readonly sortable?: boolean;
   readonly creatable?: boolean;
+  readonly isRequired?: boolean;
```
kmaid, line 127: Every other field has this as "required" please refactor this to be consistent.

apps/crn-frontend/src/network/working-groups/__tests__/WorkingGroupOutput.test.tsx
oratiu: I moved it to the mandatory fields as authors are mandatory

packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -159,6 +164,32 @@ const MultiSelect = <T extends MultiSelectOptionsType>({
...
+
+  const initialRender = useRef(true);
+  const checkValidation = useCallback(() => {
+    if (initialRender.current) {
+      initialRender.current = false;
+    } else {
+      validate();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
kmaid, line 181: Why isn't initialRender in this list? Shouldn't it be called inputRef?

packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -234,7 +264,17 @@ const MultiSelect = <T extends MultiSelectOptionsType>({
...
-      <div css={validationMessageStyles}>{validationMsg}</div>
+      <input
+        {...validationTargetProps}
+        tabIndex={-1}
+        autoComplete="off"
+        value={values.map((value) => value.label).join('')}
+        required={isRequired}
+        disabled={!enabled}
+        hidden={true}
+        onChange={noop}
```
kmaid, line 275: Why is this set to noop? Can this be omitted or set to undefined?

packages/react-components/src/templates/ResearchOutputTeamForm.tsx
```diff
@@ -321,6 +321,7 @@ const ResearchOutputPage: React.FC<ResearchOutputPageProps> = ({
                 onChangeTeams={setTeams}
                 getTeamSuggestions={getTeamSuggestions}
                 isEditMode={isEditMode}
+                authorsRequired={publishingEntity === 'Working Group'}
```
kmaid, line 324: Why is there a condition here? Won't the Team Output form only deal with Team Outputs?

packages/react-components/src/atoms/MultiSelect.tsx
oratiu: Following the approach we have in Dropdown.tsx, we don't want to validate on the first render so we don't display the error message directly.

packages/react-components/src/organisms/ResearchOutputContributorsCard.tsx
```diff
@@ -49,6 +50,7 @@ const ResearchOutputContributorsCard: React.FC<
   isSaving,
   isEditMode,
   authorsSubtitle = '(optional)',
+  authorsRequired = false,
```
kmaid, line 53:
```suggestion
  authorsRequired
```

packages/react-components/src/templates/ResearchOutputTeamForm.tsx
oratiu: We can either temporarily send a prop from the WorkingGroupOutputs component and have that prop added in the TeamForm, which will have to be removed once we add the WorkingGroupForm, or have this in place, which will also have to be removed once we add the WorkingGroupForm.

PR #2412, opened 2023-01-06
CRN-1178 add working groups card to user profile

lctrt:


packages/react-components/src/organisms/WorkingGroupsTabbedCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+  groups: ReadonlyArray<WorkingGroupMembership>;
+  isUserAlumni: boolean;
+};
+
+const WorkingGroupsTabbedCard: React.FC<WorkingGroupsTabbedCardProps> = ({
+  userName,
+  groups,
+  isUserAlumni,
+}) => {
+  const [inactiveMemberships, activeMemberships] = splitListBy(
```
gabiayako, line 29: Hey, I was testing manually and then I saw this error
reduce

Looking at console it seems the problem is with the reduce in `splitListBy` function

console1

console2


This function assumes that `groups` is never undefined, so while it's loading it gives this error.

I took a look at how is done for groups while it's loading and it seems that the whole section does not appear while the data hasn't arrived yet https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/network/users/groups/GroupsCard.tsx#L17

I would do the same here. Something like

```
if (!groups) return null
```
in line 29. But I would just check with Tiff if that's ok :)

apps/crn-server/src/data-providers/users.data-provider.ts
```diff
@@ -347,6 +355,24 @@ export const parseGraphQLUserToDataObject = (
...
+        }, 'Member');
+
+        return {
+          id: wg.id,
+          name: wg.flatData.title || '',
+          role: wgRole as 'Chair' | 'Project Manager' | 'Member',
+          active: !wg.flatData.complete,
+        };
+      })
+    : [];
```
gabiayako, line 374: I was wondering if this could be rewritten as 

```
  const workingGroups = (item.referencingWorkingGroupsContents || []).map(
    (wg) => {
      const leader = wg.flatData.leaders?.find(
        (leader) => leader.user?.[0]?.id === item.id,
      );
      const wgRole = leader ? leader.role : 'Member';

      return {
        id: wg.id,
        name: wg.flatData.title || '',
        role: wgRole as 'Chair' | 'Project Manager' | 'Member',
        active: !wg.flatData.complete,
      };
    },
  );
```

so maybe it's more readable the way we find the working groups role.

But it's just a suggestion, you don't need to implement if you don't like it :)

gabiayako (changes requested):
Looks very nice, just requesting changes to adjust the issue with the `reduce` error message

packages/react-components/src/organisms/WorkingGroupsTabbedCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+type WorkingGroupsTabbedCardProps = {
+  userName: string;
+  groups?: ReadonlyArray<WorkingGroupMembership>;
+  isUserAlumni: boolean;
+};
+
+const WorkingGroupsTabbedCard: React.FC<WorkingGroupsTabbedCardProps> = ({
+  userName,
+  groups = [],
```
gabiayako, line 26: I think there's an issue here. If `groups` is initially `undefined` because it hasn't loaded yet it will come as `[]` here, so in the tabbed card appears as if the person is not associated to any working groups. But if we refresh the page we can see that actually the person is associated with one

https://user-images.githubusercontent.com/16595804/211801595-49e2a581-d56a-449b-8efe-e0206fadb2c9.mov

with `return null` I believe we don't need to refresh the page :)

gabiayako (changes requested):

gabiayako (commented):
Sorry, I think I misunderstood the issue with working groups not appearing 😅 But I might need some help from @kmaid here.

I’ll recap it, so it makes more sense to everyone. The issue is that the workingGroups is undefined when we see this `reduce` error below

reduce

In `packages/react-components/src/organisms/WorkingGroupsTabbedCard.tsx` we have this call
```
  const [inactiveMemberships, activeMemberships] = splitListBy(
    groups,
    (group) => isUserAlumni || !group?.active,
  );
```

and `splitListBy` calls this reduce

```
export const splitListBy = <T>(
  items: ReadonlyArray<T>,
  splitBy: (item: T) => boolean,
) =>
  items.reduce<[T[], T[]]>(
    (split, item) => {
      splitBy(item) ? split[0].push(item) : split[1].push(item);
      return split;
    },
    [[], []],
  );
```

But the problem is not that the data is still loading, it’s that the data is “dirty” somehow.

We have this

```
const userState = selectorFamily<UserResponse | undefined, string>({
  key: 'user',
  get:
    (id) =>
    ({ get }) =>
      get(patchedUserState(id)) ?? get(initialUserState(id)),
});
```

in https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/network/users/state.ts#L91

And console logging `get(patchedUserState(id))` I've got this for my user

Screenshot 2023-01-11 at 14 16 45

Note that this does not have `workingGroups`

while the `get(initialUserState(id))` has `workingGroups`

Screenshot 2023-01-11 at 14 16 58

But when we call `userState` it comes with this `get(patchedUserState(id))` data that does not have workingGroups, that's why it gives an error. I just don't know how to fix it though. Do you have a suggestion @kmaid ?

gabiayako (approved):

PR #2418, opened 2023-01-09
CRN-1245 Remove Working Groups from discover endpoint

gabiayako:
This PR removes the data `working groups` from the response from the endpoint `/discover` since it's not needed anymore. The tab `Working Groups` was removed here https://github.com/yldio/asap-hub/pull/2413


m-doughty (approved):
This all looks sensible to me. Good work.

PR #2421, opened 2023-01-10
CRN-1246 Migration to delete working groups from news

gabiayako:
This PR introduces a migration that deletes the News that have type `Working Groups`.

I've ran the script in the PR env:

us-east-1 console aws amazon com_lambda_home_region=us-east-1 (1)

And I could verify that there isn't any News with type `Working Groups` now:

squidex-after

### DIFF
--- a/apps/crn-server/src/migrations/1673349371-remove-working-groups-from-news.ts
+++ b/apps/crn-server/src/migrations/1673349371-remove-working-groups-from-news.ts
@@ -1,0 +1,18 @@
+/* istanbul ignore file */
+import { Migration } from '@asap-hub/server-common';
+import { RestNews } from '@asap-hub/squidex';
+import { applyToAllItemsInCollection } from '../utils/migrations';
+
+export default class RemoveWorkingGroupsFromNews extends Migration {
+  up = async (): Promise<void> => {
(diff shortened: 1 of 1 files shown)


apps/crn-server/src/migrations/1673349371-remove-working-groups-from-news.ts
```diff
@@ -0,0 +1,20 @@
...
+      'news-and-events',
+      async (news, squidexClient) => {
+        if (news.data.type.iv === 'Working Groups') {
+          await squidexClient.delete(news.id);
+        }
+      },
+    );
+  };
+  down = async (): Promise<void> => {
+    // this is not needed
```
gabiayako, line 18: I just added this comment because lint was giving an error that an empty function is not allowed

Anyway, I don't think this is needed right? If so, the idea is to move these records to a backup schema and move them back in case we want to rollback?

peterstarling: nah too much effort, this is fine

peterstarling (approved):
i mean this is fine but do we really need a data migration for deleting like 10 records :D?

PR #2424, opened 2023-01-10
Added zero outputs component on working groups

oratiu:


apps/crn-frontend/src/network/teams/Outputs.tsx
```diff
@@ -134,7 +135,7 @@ const Outputs: React.FC<OutputsProps> = ({ teamId }) => {
           searchQuery={debouncedSearchQuery}
           filters={filters}
           hasOutputs={hasOutputs}
-          ownTeam={ownTeam}
+          ownEntity={ownTeam}
```
kmaid, line 138: ?

apps/crn-frontend/src/network/working-groups/Outputs.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const currentUserId = useCurrentUserCRN()?.id;
+  const userAssociationMember = !!useWorkingGroupById(
+    workingGroupId,
+  )?.members.filter((member) => member.user.id === currentUserId).length;
+  return (
+    <article>
+      <SearchFrame title="">
+        <OutputsList
+          workingGroupId={workingGroupId}
+          searchQuery={''}
```
kmaid, line 71: Why is this stubbed? Won't this mean that the search bar wont allow you to type?

apps/crn-frontend/src/network/working-groups/Outputs.tsx
```diff
@@ -0,0 +1,81 @@
...
+          filters={new Set()}
+          hasOutputs={hasOutputs}
+          displayName={workingGroup?.title ?? ''}
```
kmaid, line 74: Doesn't there have to be a working group? If there isn't shouldn't this page not be found

apps/crn-frontend/src/network/working-groups/__tests__/Outputs.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+it('renders the no outputs component correctly', async () => {
+  const { getByText } = await renderOutputs('');
+  expect(
+    getByText('This working group hasn’t shared any research.'),
+  ).toBeVisible();
+});
+
+it('renders the no outputs component correctly for own team', async () => {
+  const { getByText } = await renderOutputs('working-group-id', 'user-id-0');
```
kmaid, line 61: How does this work? Are we relying on the fixture's id to match the user id here?

packages/react-components/src/templates/ProfileOutputs.tsx
```diff
@@ -0,0 +1,102 @@
...
+  cardViewHref,
+  exportResults,
+  listViewHref,
+  hasOutputs,
+  userAssociationMember,
+  contactEmail,
+  publishingEntity,
+}) => (
+  <div css={containerStyles}>
+    {hasOutputs ? (
```
kmaid, line 40: So we have the number of items below how does this differ from hasOutputs?

packages/react-components/src/templates/ProfileOutputs.tsx
```diff
@@ -0,0 +1,102 @@
...
+        numberOfItems={numberOfItems}
+        numberOfPages={numberOfPages}
+        currentPageIndex={currentPageIndex}
+        renderPageHref={renderPageHref}
+        isListView={isListView}
+        cardViewHref={cardViewHref}
+        listViewHref={listViewHref}
+      />
+    ) : userAssociationMember ? (
+      <NoOutputsPage
```
kmaid, line 53: This feels like allot of duplication. I think it would be better to use two ternary's in the props over duplicating the entire description

kmaid (commented):
Please leave me a direct link to your PR to test this functionality out

oratiu:
https://2424.hub.asap.science/network/working-groups/0182aef4-0d46-4ec7-aae0-584d0b0e3bad/about

packages/react-components/src/templates/ProfileOutputs.tsx
oratiu: This is the same as above, I've just change the name of ownTeam to userAssociationMember in order to be more general.

packages/react-components/src/templates/ProfileOutputs.tsx
oratiu: I've mostly refactored TeamProfileOutputs to fit the needs of WorkingGroupProfileOutputs, this wasn't added by me and I was unsure how far down the refactoring hole I should go.

apps/crn-frontend/src/network/working-groups/Outputs.tsx
oratiu: Followed the same approach from team/Outputs.tsx, but I guess you are right, I will render the NotFoundPage in case we cannot get the workingGroup by Id

apps/crn-frontend/src/network/working-groups/Outputs.tsx
oratiu: Search bar is removed completely until we deal with the ticket in which we actually want to display the research output list. As we are only displaying the NoResearchOutputs page right now, I thought it would be better to remove the search bar component and to pass in a empty value.

packages/react-components/src/templates/ProfileOutputs.tsx
kmaid: My issue is that your duplicating the entire description twice for a couple word changes. It would be less duplication by just including two ternary operators within the description and title. Less duplication = better

apps/crn-frontend/src/network/teams/Outputs.tsx
```diff
@@ -115,7 +114,7 @@ const Outputs: React.FC<OutputsProps> = ({ teamId }) => {
   }).total;
   const team = useTeamById(teamId);
 
-  const ownTeam = !!(useCurrentUserCRN()?.teams ?? []).filter(
+  const userAssociationMember = !!(useCurrentUserCRN()?.teams ?? []).filter(
```
kmaid, line 117: You want to use some not filter.

apps/crn-frontend/src/network/teams/Outputs.tsx
```diff
@@ -133,8 +132,7 @@ const Outputs: React.FC<OutputsProps> = ({ teamId }) => {
           teamId={teamId}
           searchQuery={debouncedSearchQuery}
           filters={filters}
-          hasOutputs={hasOutputs}
-          ownTeam={ownTeam}
+          userAssociationMember={userAssociationMember}
           contactEmail={team?.pointOfContact?.email}
```
kmaid, line 136: This is also funky, if there isn't a team we should display 404.

kmaid: Not for you to fix just an observation

apps/storybook/src/ProfileOutputs.stories.tsx
```diff
@@ -19,8 +19,12 @@ export const Normal = () => {
       isListView={false}
       cardViewHref={''}
       listViewHref={''}
-      ownTeam={boolean('Own Team', true)}
-      hasOutputs={!!numberOfOutputs}
+      userAssociationMember={boolean('Own Entity', true)}
```
kmaid, line 22:
```suggestion
      userAssociationMember={boolean('User is association member, true)}
```

kmaid (approved):

apps/crn-frontend/src/network/teams/Outputs.tsx
oratiu: Will change in both places

apps/crn-frontend/src/network/teams/Outputs.tsx
oratiu: I am going to send the team not the team id, as the parent component also retrieves a team by id and renders or not all the child component based on that team (which can be null and for that case it does render the NotFoundPage )

apps/storybook/src/ProfileOutputs.stories.tsx
oratiu: ... damn, ofc I forgot that

PR #2426, opened 2023-01-10
[CRN-153] TEAMS: Past Team Members

ricardocarvalh0:
https://asaphub.atlassian.net/browse/CRN-1153

- [x] Create functionality in Squidex for identifying whether a user is an inactive member of an active team
- [x] Where a user is set as inactive then that user should be displayed in the past members section in the team detail page on the hub


packages/squidex/schema/crn/schemas/users.json
```diff
@@ -243,6 +243,28 @@
...
+              "defaultValue": "Active",
+              "allowedValues": ["Active", "Inactive"],
+              "isUnique": false,
+              "isEmbeddable": false,
+              "inlineEditable": true,
+              "createEnum": false,
+              "contentType": "Unspecified",
+              "editor": "Dropdown",
+              "label": "Status",
+              "hints": "Attention: Check if this user needs to be added to Smart Simple",
```
gabiayako, line 263: is this hint supposed to be here? what is Smart Simple? 😛

packages/squidex/src/entities/crn/user.ts
```diff
@@ -16,10 +16,12 @@ import { GraphqlTeam } from './team';
...
   ? {
+      status: string;
       role: TeamRole;
       id: T[];
     }
   : {
+      status: string;
       role: TeamRole;
       id: T[];
     };
```
gabiayako, line 27: I think it's weird that the ternary returns the same thing if the condition is true or false, I took a look at the code that introduced this ternary (https://github.com/yldio/asap-hub/pull/632/files)

Screenshot 2023-01-11 at 09 36 37

and the returning value was different when the condition was true or false, but now it does not make sense the ternary, it could be just

```
export type UserTeamConnection<T = string> = {
  status: 'Active' | 'Inactive';
  role: TeamRole;
  id: T[];
};
```

right?

packages/react-components/src/organisms/TeamMembersTabbedCard.tsx
```diff
@@ -40,7 +40,10 @@ const TeamMembersTabbedCard: React.FC<TeamMembersTabbedCardProps> = ({
 }) => {
   const [alumniMembers, activeMembers] = splitListBy(
```
gabiayako, line 41:
```suggestion
  const [pastMembers, activeMembers] = splitListBy(
```

packages/react-components/src/organisms/TeamMembersTabbedCard.tsx
```diff
@@ -40,7 +40,10 @@ const TeamMembersTabbedCard: React.FC<TeamMembersTabbedCardProps> = ({
...
     members,
-    (member) => isTeamInactive || !!member?.alumniSinceDate,
+    (member: TeamMember) =>
+      isTeamInactive ||
+      !!member?.alumniSinceDate ||
+      member?.status === 'Inactive',
```
gabiayako, line 46: Since this `status` is a required field we don't need `?` here 😉 
```suggestion
      member.status === 'Inactive',
```

gabiayako (changes requested):
Looks good! I've just made some suggestions :)

gabiayako (approved):

ricardocarvalh0:
Closing this as the squidex environment doesn't have all the users

PR #2450, opened 2023-01-13
Minor refactor and naming fixes

kmaid:


packages/model/src/research-output.ts
```diff
@@ -225,6 +226,7 @@ export type ResearchOutputDataObject = ResearchOutputCoreObject & {
   organisms: string[];
   subtype?: string;
   teams: Pick<TeamResponse, 'id' | 'displayName'>[];
+  workingGroups: Pick<WorkingGroupResponse, 'id' | 'title'>[];
```
peterstarling, line 229: your intention was to only add workingGroups property to the response type and add a dummy implementation - I would do just that and add the property here instead:
https://github.com/yldio/asap-hub/blob/48e1e34f3aeaaf56233207c0c0192965f4dd0ec5/packages/model/src/research-output.ts#L261

That way you are not going to have to implement it anywhere other than at the controller/route level. Same goes for the input

peterstarling (changes requested):

peterstarling (approved):

oratiu (approved):
Nothing that I can comment on, looks really good

PR #2466, opened 2023-01-17
[CRN-1243] fix feature flag override from cookie

ricardocarvalh0:


peterstarling (changes requested):
looks like there should be a test attached and there isnt

ricardocarvalh0:
> looks like there should be a test attached and there isnt

done 👍

PR #2468, opened 2023-01-17
add missing schema updates

ricardocarvalh0:


kmaid (changes requested):
@gabiayako removed these fields. We should not add them back in

PR #2471, opened 2023-01-17
upload avatar

jhonny-m:


packages/gp2-components/src/organisms/UserDetailHeaderCard.tsx
```diff
@@ -1,14 +1,21 @@
...
+  Avatar,
+  Headline3,
+  Link,
+  pixels,
+  uploadIcon,
+} from '@asap-hub/react-components';
 import { css } from '@emotion/react';
-import { addIcon } from '../icons';
+import { addIcon, editIcon } from '../icons';
 import locationIcon from '../icons/location-icon';
```
csbarbosa, line 11: Merge these imports into one, if possible
```suggestion
import {addIcon, editIcon, locationIcon, roleIcon} from '../icons';
```

packages/gp2-components/src/organisms/UserDetailHeaderCard.tsx
```diff
@@ -1,14 +1,21 @@
...
 import roleIcon from '../icons/role-icon';
 import { usersHeaderImage } from '../images';
 import { mobileQuery, nonMobileQuery } from '../layout';
 import CardWithBackground from '../molecules/CardWithBackground';
```
csbarbosa, line 15: Merge these imports into one, if possible
```suggestion
import { CardWithBackground, IconWithLabel, UserRegion} from '../molecules';
```

apps/gp2-server/src/routes/user.route.ts
```diff
@@ -104,6 +106,37 @@ export const userRouteFactory = (userController: UserController): Router =>
...
+      // convert bytes to MB and check size
+      // 3MB = 2.8MB (2MB Base64 image) + some slack
+      if (avatar.length / 1e6 > 3) {
+        throw Boom.entityTooLarge('Avatar must be smaller than 2MB');
+      }
+      const result = await userController.updateAvatar(
+        userId,
+        avatar,
+        parsed.contentType,
+        userId,
```
csbarbosa, line 137: Shouldn't be loggedInUser.id instead of userId? Seems a lil bit odd to use it twice. Maybe change it here or refactor the function in the userController?

apps/gp2-server/test/controllers/user.controller.test.ts
```diff
@@ -173,6 +174,63 @@ describe('Users controller', () => {
...
+    beforeEach(() => {
+      jest.resetAllMocks();
+    });
+
+    test('should return 200 when syncs asset and updates users profile', async () => {
+      const user = getUserDataObject();
+      const userId = user.id;
+      assetDataProviderMock.create.mockResolvedValueOnce('42');
+      userDataProviderMock.fetchById.mockResolvedValueOnce(user);
+      const result = await userController.updateAvatar(
```
csbarbosa, line 187: Using userId seems a lil bit odd, maybe refactor the function?

csbarbosa (commented):
Some import changes and maybe a refactor of a function

apps/gp2-server/src/controllers/user.controller.ts
```diff
@@ -39,24 +34,22 @@ export default class Users implements UserController {
   async update(
     id: string,
     update: gp2.UserUpdateRequest,
-    loggedInUserId: string,
   ): Promise<gp2.UserResponse> {
     await this.userDataProvider.update(id, update);
-    return this.fetchById(id, loggedInUserId);
+    return this.fetchById(id, id);
```
csbarbosa, line 39: Does fetchById function need also to receive both id's? It seems that it's not used anywhere else except in user.route.ts, but could be mistaken

jhonny-m: when we parse the squidex data to our UserResponse, we check if the user making the request is the one being fetched and if it is true we will add telephone data, else we don't (for security/privacy concerns)
we we make a update we always returned the user, but for an update we already validate you are the user we are changing

apps/gp2-frontend/src/onboarding/CoreDetails.tsx
```diff
@@ -4,25 +4,55 @@ import {
...
+    setAvatarSaving(true);
+    imageCompression(file, { maxSizeMB: 2 })
+      .then((compressedFile) =>
+        imageCompression.getDataUrlFromFile(compressedFile),
+      )
+      .then((encodedFile) => postUserAvatar(encodedFile))
+      .catch(() =>
+        toast('There was an error and we were unable to save your picture'),
+      )
+      .finally(() => setAvatarSaving(false));
```
markcsims, line 45: Looks strange to see a promise chain, and the result not being returned. 

I guess this is how it's handled in CRN?

markcsims (approved):

csbarbosa (approved):

dianaguimaraes (approved):
Just a couple of small typos

apps/gp2-frontend/src/onboarding/CoreDetails.tsx
jhonny-m: yeah, it's what crn is doing

PR #2477, opened 2023-01-18
Crn 153 past team members

ricardocarvalh0:
It covers both
https://asaphub.atlassian.net/browse/CRN-1153
and
https://asaphub.atlassian.net/browse/CRN-1281


packages/react-context/src/__tests__/auth.test.tsx
```diff
@@ -6,6 +6,7 @@ import {
   useCurrentUserCRN,
   useCurrentUserGP2,
   useCurrentUserTeamRolesCRN,
+  useCurrentUserTeamStatusesCRN,
```
kmaid, line 9: 😢 statuses

kmaid (approved):
I never want to read "Active" again

packages/squidex/schema/crn/schemas/users.json
```diff
@@ -243,6 +243,41 @@
...
+              "contentType": "Unspecified",
+              "editor": "Dropdown",
+              "label": "Status",
+              "isRequired": false,
+              "isRequiredOnPublish": false,
+              "isHalfWidth": true
+            }
+          },
+          {
+            "name": "inactiveSinceDate",
```
peterstarling, line 269: i wonder if we need both of these fields at the same time?

Let me ask you this - can a user be inactive but not have an `inactiveSinceDate` set? Cause with your current proposal that would be possible, and your types seem to reflect it:
https://github.com/yldio/asap-hub/blob/0eb3e0b6dbf0bcc8d4a3c7e3d174a3d272cd3b19/packages/model/src/user.ts#L109-L110

What is even more strange is that the ticket does not seem to require such field, nor do any of the FE components seem to use it. 

If we know we are going to need such date then I suggest we make sure the requirements are clear. To me - if a user is set to inactive a date will be required, in which case I would be tempted to remove the `status` field and just use the `inactiveSinceDate`. You could leave both of them in the Response type so it's easier for FE to check the status, but the data provider would only return this one.

peterstarling (changes requested):

PR #2479, opened 2023-01-18
GP2 135 user activated account yes or no

markcsims:
- adds activated date field to user
- sets activatedDate when user clicks on invitation
- activated date is updated in controller
- returns activated date from squidex
- fixes activedDate in schema
- updates schema


apps/gp2-frontend/src/users/export.ts
```diff
@@ -20,7 +20,8 @@ export const userFields = {
   workingGroups: 'Working groups',
   fundingStreams: 'Funding streams',
   contributingCohorts: 'Contributing cohorts',
-  createdDate: 'Activated account',
+  createdDate: 'Account Created',
+  activatedDate: 'Activated account',
```
jhonny-m, line 24: shouldn't it be Account Activated?

apps/gp2-server/src/controllers/user.controller.ts
```diff
@@ -96,6 +96,7 @@ export default class Users implements UserController {
       {
         email: user.email,
         connections: [...(user.connections || []), { code: authUserId }],
+        activatedDate: user.activatedDate ?? new Date().toISOString(),
```
jhonny-m, line 99: why do we default it to a new Date?

jhonny-m, line 99: whats the scenario where we would have a date already?

jhonny-m (approved):

apps/gp2-frontend/src/users/export.ts
markcsims: It's what we had, but I think it should change.

apps/gp2-server/src/controllers/user.controller.ts
markcsims: There is. All goes back to the conversations we're had in RTM about connections. 

I think a user will be allow multiple connection, so could go past this code more that once. 

If that happens, I think we should keep the first activated date.

PR #2480, opened 2023-01-19
Gp2 138 frontend

csbarbosa:
(Ticket:https://asaphub.atlassian.net/browse/GP2-138)
Enable filter pills for user filters


packages/gp2-components/src/molecules/FilterPill.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+interface PillProps {
+  readonly value: ValueProps;
+  onRemove: (value: ValueProps) => void;
+}
+
+const FilterPill: React.FC<PillProps> = ({ value, onRemove }) => (
+  <div css={[pillStyles]}>
+    <div css={css({ display: 'flex', padding: `8px 12px` })}>
```
dianaguimaraes, line 44: is this extra div really needed? can't these styles be applied on the parent div?

packages/gp2-components/src/templates/UsersPage.tsx
```diff
@@ -25,9 +27,56 @@ type UsersPageProps = ComponentProps<typeof FilterSearchExport> & {
...
+  const mapping: FilterMappingType = {
+    keywords: (x: string) => x,
+    regions: (x: string) => x,
+    projects: (x: string) => getLabel(projects, x),
+    workingGroups: (x: string) => getLabel(workingGroups, x),
+  };
+
+  return Object.entries(filters).reduce(
+    // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+    (acc: any, [key, value]) => [
```
dianaguimaraes, line 69: can we avoid this any?

apps/storybook/src/gp2-components/FilterPill.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+  title: 'GP2 / Molecules / Filter Pill',
+  component: FilterPill,
+};
+
+const props: ComponentProps<typeof FilterPill> = {
+  value: {
+    id: 'region-Africa',
+    label: 'Africa',
+  },
+  onRemove: (value) => alert(`${value.label} erased!`),
```
markcsims, line 14: Not sure we should have this alert here.

The other ways I've seen this are with either a noop or a () => Promise.resolve()

apps/storybook/src/gp2-components/FilterPills.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+    {
+      id: 'region-Africa',
+      label: 'Africa',
+    },
+    {
+      id: 'region-Asia',
+      label: 'Asia',
+    },
+  ],
+  onRemove: (value) => alert(`${value.label} erased!`),
```
markcsims, line 20: same as above

packages/gp2-components/src/templates/UsersPage.tsx
```diff
@@ -45,6 +93,23 @@ const UsersPage: React.FC<UsersPageProps> = ({
...
+        (i) => i !== val,
+      );
+      const updatedFilters = {
+        ...filters,
+        [key]: newFilter,
+      };
+      updateFilters(backHref, updatedFilters);
+    },
+    [filters, backHref, updateFilters],
+  );
```
markcsims, line 111: I'm not sure I understand this method. 

It's called onRemove but it's setting a filter? Maybe we could have a chat to help me understand it?

packages/gp2-components/src/templates/UsersPage.tsx
```diff
@@ -25,9 +27,55 @@ type UsersPageProps = ComponentProps<typeof FilterSearchExport> & {
...
+const mapFilters = (
+  filters: FiltersType,
+  projects: ProjectsType,
+  workingGroups: WorkingGroupsType,
+) => {
+  const mapping: FilterMappingType = {
+    keywords: (x: string) => x,
+    regions: (x: string) => x,
+    projects: (x: string) => getLabel(projects, x),
+    workingGroups: (x: string) => getLabel(workingGroups, x),
```
markcsims, line 64: What is x?

markcsims (changes requested):

packages/gp2-components/src/molecules/FilterPill.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { ReactNode } from 'react';
+import colors from '../templates/colors';
+
+const { rem } = pixels;
+
+const pillStyles = css({
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  width: 'fit-content',
```
jhonny-m, line 16: I don't think it needs a lot of this css code, the height and width are gonna be set by the content + padding, we don't need to justify the content and I'm not sure we need the align either

packages/gp2-components/src/molecules/FilterPill.tsx
```diff
@@ -0,0 +1,54 @@
...
+  gap: rem(8),
+  borderStyle: 'solid',
+  borderWidth: `${borderWidth}px`,
+  borderRadius: rem(24),
+  borderColor: colors.neutral500.rgb,
+  color: colors.neutral1000.rgb,
+});
+
+const iconStyles = css({
+  display: 'flex',
```
jhonny-m, line 28: this is just a button with icon in, I dont think we need any flex styling in, It would also be nice to add the `cursor: 'pointer'` we do for Button

packages/gp2-components/src/organisms/FilterPills.tsx
```diff
@@ -0,0 +1,39 @@
...
+  gap: rem(8),
+  paddingTop: rem(24),
+});
+
+type FiltersType = keyof Pick<
+  gp2Model.FetchUsersFilter,
+  'keywords' | 'regions' | 'workingGroups' | 'projects'
+>;
+type FilterPillsProps = {
+  pills: { typeOfFilter: FiltersType; label: string; id: string }[];
```
jhonny-m, line 22: maybe `type` or `filterType`

jhonny-m: I don't like the name pills here, since its not the pills yet, it's more of a filterList or appliedFiltersList

packages/gp2-components/src/organisms/FilterPills.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onRemove: (id: string, typeOfFilter: FiltersType) => void;
+};
+
+const FilterPills: React.FC<FilterPillsProps> = ({ pills, onRemove }) => (
```
jhonny-m, line 26: do we need this component? we are just mapping through an array

packages/gp2-components/src/templates/UsersPage.tsx
```diff
@@ -45,6 +88,22 @@ const UsersPage: React.FC<UsersPageProps> = ({
   const { users } = gp2;
   const backHref = users({}).$;
   const onBackClick = () => changeLocation(backHref);
+
+  const onRemove = useCallback(
```
jhonny-m, line 92: this is more of a removeFilter than a onRemove

packages/gp2-components/src/templates/UsersPage.tsx
```diff
@@ -45,6 +88,22 @@ const UsersPage: React.FC<UsersPageProps> = ({
...
+    (id: FilterPillId, typeOfFilter: FiltersType) => {
+      const newFilter = (filters[typeOfFilter] || []).filter(
```
jhonny-m, line 94: we can do it directly, I think the name newFilter here is a bit confusing since we are removing

packages/gp2-components/src/molecules/FilterPill.tsx
csbarbosa: The width : 'fit-content' is necessary, because it will cause the pill to expand to the other side of the page, if not applied. The align is also necessary because it will move away from the center. After testing a little bit more, the one I can remove is the justifyContent.

packages/gp2-components/src/molecules/FilterPill.tsx
csbarbosa: Without the flex it will not center in it's div, but I've tested a lil bit more and it was possible to remove the other 2 properties and added the cursor : 'pointer'

packages/gp2-components/src/organisms/__tests__/FilterPills.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { ComponentProps } from 'react';
+import FilterPills from '../FilterPills';
+
+describe('FilterPills', () => {
+  beforeEach(jest.resetAllMocks);
```
jhonny-m, line 7: what are we mocking?

packages/gp2-components/src/organisms/__tests__/FilterPills.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    onRemove: jest.fn(),
+  };
+  it('should render all the pills', () => {
+    render(<FilterPills {...props} />);
+    expect(screen.getByText('Epidemiology')).toBeVisible();
+    expect(screen.getByText('Asia')).toBeVisible();
+    expect(screen.getByText('Project 1')).toBeVisible();
+    expect(screen.getByText('Working Group 1')).toBeVisible();
+  });
+  it("shows the remove button for the second pill and it's clickable", () => {
```
jhonny-m, line 26: I think the description is a bit wrong

packages/react-components/src/atoms/Tag.tsx
```diff
@@ -33,24 +34,43 @@ const disabledStyles = css({
   color: tin.rgb,
 });
 
+const iconStyles = css({
```
jhonny-m, line 37: it should be buttonStyles rather than icon

jhonny-m (approved):

markcsims (approved):

PR #2482, opened 2023-01-19
CRN 1241 - Inline images appear in the hub

gabiayako:
The function `documentToHtmlString` from the external lib `@contentful/rich-text-html-renderer` (https://www.npmjs.com/package/@contentful/rich-text-html-renderer) does not converts inline images automatically as we can see in this answer https://stackoverflow.com/questions/54083103/contentful-documenttohtmlstring-doesnt-include-embedded-image-in-rich-text, so in this PR it's being added the snippet from stack overflow to convert inline images.

Also, to bring the url and alt description from images, in the query I had to add `text` -> `links` -> `assets` -> `block` -> `url`/`description`.
Screenshot 2023-01-20 at 10 02 02

But this made the query too become too complex
complex-query

So one thing that it's possible to decrease the complexity is to define a maximum limit of linked assets/entries/links (read more here https://www.contentful.com/developers/docs/references/graphql/#/introduction/query-complexity-limits)
limit-ui

I am limiting the number of links by 10 (it was super random, we can change it later), so I added two migrations to make this change in news and pages.

---

### Validation

I've made 2482 point to `rtf-test` contentful env in AWS  https://2482.hub.asap.science/news/17973b96-4fa3-4829-9d36-089eedc3eec3

Then I could see that the inline images are now appearing in the hub https://2482.hub.asap.science/news/17973b96-4fa3-4829-9d36-089eedc3eec3
Screenshot 2023-01-20 at 10 21 43 (2)

### DIFF
--- a/packages/contentful/src/utils.ts
+++ b/packages/contentful/src/utils.ts
@@ -48,7 +48,7 @@ export const parseRichText = (rtf: RichTextFromQuery) => {
           const { url, description, contentType, width, height } =
             assetById[assetId];
           const dimensions =
-            width && height ? ` width="${width}" height="${height}` : '';
+            width && height ? ` width="${width}" height="${height}"` : '';
 
           switch (contentType) {
             case 'application/pdf':
--- a/packages/contentful/test/utils.test.ts
+++ b/packages/contentful/test/utils.test.ts
@@ -122,7 +122,7 @@ describe('parseRichText', () => {
       },
     };
     expect(parseRichText(rtf)).toEqual(
-      `<p>Here it will appear an image</p><iframe src=\"https://images.ctfassets.net/envId/image1/file-1.pdf\" width=\"200\" height=\"300><p>below there&#39;s a cat</p><iframe src=\"https://images.ctfassets.net/envId/image2/file-2.pdf\"><p> </p>`,
+      `<p>Here it will appear an image</p><iframe src=\"https://images.ctfassets.net/envId/image1/file-1.pdf\" width=\"200\" height=\"300\"><p>below there&#39;s a cat</p><iframe src=\"https://images.ctfassets.net/envId/image2/file-2.pdf\"><p> </p>`,
     );
   });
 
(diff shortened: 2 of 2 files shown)


apps/crn-server/src/data-providers/contentful/news.data-provider.ts
```diff
@@ -76,6 +77,6 @@ export const parseContentfulGraphQlNews = (item: NewsItem): NewsDataObject => ({
   thumbnail: item.thumbnail?.url ?? undefined,
   link: item.link ?? undefined,
   linkText: item.linkText ?? undefined,
-  text: item.text ? parseRichText(item?.text.json) : undefined,
+  text: item.text ? parseRichText(item?.text as RichTextFromQuery) : undefined,
```
peterstarling, line 80: this should work
```suggestion
  text: item.text ? parseRichText(item.text as RichTextFromQuery) : undefined,
```
also why do we have to assert?

packages/contentful/src/utils.ts
```diff
@@ -1,5 +1,53 @@
...
+      },
+    }),
+    {},
+  );
+
+  const options = {
+    renderNode: {
+      'embedded-asset-block': (node: Node) => {
+        const assetId = node.data.target.sys.id as string;
+        if (assetById && assetId && assetById[assetId]) {
```
peterstarling, line 42: what if its a different asset like pdf?

packages/contentful/test/utils.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    const rtf = {
+      json: document,
+      links,
+    };
+    expect(parseRichText(rtf)).toEqual(
+      `<p>Here it will appear an image</p><img class="img-fluid" src="${baseAssetUrl}/${assetId_1}/penguin.jpeg" alt=A very cute baby penguin walking/><p>below there&#39;s a cat</p><img class="img-fluid" src="${baseAssetUrl}/${assetId_2}/cat.png" alt=cat/><p> </p>`,
+    );
+  });
+
+  test('throws an error if rtf is linked to an image that does not exist', () => {
```
peterstarling, line 100: we will have to support pdfs and videos alter so i would add tests for these, especially since they are not supported yet

peterstarling (changes requested):

packages/contentful/src/utils.ts
gabiayako: in contentful it seems it's the same :) so I just needed to update this `embedded-asset-block` render node

apps/crn-server/src/data-providers/contentful/news.data-provider.ts
gabiayako: this `parseRichText` will receive the text field from pages and news. I didn't want to make the type to be the one autogenerated from news or pages because I think it would be weird to choose one of them. I didn't discover a smart way to do that, so I wrote the type manually

peterstarling (approved):

PR #2501, opened 2023-01-23
GP2 Biography Modal

markcsims:


packages/gp2-components/src/organisms/BiographyModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+    >
+      {({ isSaving }) => (
+        <LabeledTextArea
+          value={newBiography}
+          onChange={setNewBiography}
+          enabled={!isSaving}
+          required
+          maxLength={2500}
+          title="Background"
+          subtitle="(Required)"
```
jhonny-m, line 40: I think we have them as `(required)` in other places

packages/gp2-components/src/organisms/__tests__/BiographyModal.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    expect(screen.getByRole('dialog')).toContainElement(
+      screen.getByRole('heading', { name: 'Biography' }),
+    );
+  });
+
+  it('renders biography', () => {
+    renderModal();
+    expect(
+      screen.getByRole('textbox', {
+        name: 'Background (Required)',
```
jhonny-m, line 37: will a regex for background only work here?

packages/gp2-components/src/organisms/BiographyModal.tsx
markcsims: Done

packages/gp2-components/src/organisms/__tests__/BiographyModal.test.tsx
markcsims: Done

jhonny-m (approved):

PR #2510, opened 2023-01-24
Crn 1236 introduce working groups on events calendar page

oratiu:


packages/react-components/src/templates/__tests__/EventsCalendar.tsx
```diff
@@ -0,0 +1,37 @@
+import { render } from '@testing-library/react';
+
+import EventsCalendar from '../EventsCalendar';
+
+it('Renders calender list', () => {
```
lctrt, line 5:
```suggestion
it('Renders calendar list', () => {
```

packages/fixtures/src/calendars.ts
```diff
@@ -1,101 +1,31 @@
...
-  if (itemIndex > calendarResponses.length) {
-    throw new Error(`Exceeds fixture data limit ${calendarResponses.length}`);
-  }
-  return calendarResponses[itemIndex];
-};
+export const createCalendarResponse = (
+  { group = false, workingGroup = false }: FixtureOptions = {},
+  itemIndex = 0,
+): CalendarResponse => ({
+  id: `calendar-${itemIndex}`,
```
lctrt, line 12: 🎉

packages/react-components/src/templates/__tests__/EventsCalendar.tsx
oratiu: Woops

lctrt (approved):

apps/crn-server/test/data-providers/calendars.data-provider.test.ts
```diff
@@ -172,7 +172,12 @@ describe('Calendars data provider', () => {
           order: 'data/name/iv asc',
         },
       );
-      expect(result).toEqual({ total: 1, items: [getCalendarDataObject()] });
+      expect(result).toEqual({
+        total: 1,
+        items: [
+          { ...getCalendarDataObject(), workingGroup: true, group: true },
```
peterstarling, line 178: why not update the fixture? It's strange you are having to do this here and not in here https://github.com/yldio/asap-hub/blob/2471f1141b14d49b3ebd2b1aeec95cdf59e994d3/apps/crn-server/test/data-providers/calendars.data-provider.test.ts#L99

apps/crn-server/test/entities/working-group.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    expect(
+      parseGraphQlWorkingGroup(workingGroupWithCalendars).calendars,
+    ).toEqual([
+      {
+        id: 'hub@asap.science',
+        color: '#B1365F',
+        name: 'ASAP Hub',
+      },
+    ]);
+  });
```
peterstarling, line 36: if you are going to test the parser separately you might as well add another test for the entire input and output

packages/model/src/calendar.ts
```diff
@@ -57,6 +57,8 @@ export interface CalendarDataObject {
   resourceId?: string | null;
   expirationDate?: number | null;
   version: number;
+  group?: boolean;
```
peterstarling, line 60: this doesn't look right - first of all an optional boolean makes no sense to me. Second - are we sure we don't just want to return the list of groups and workingGroups associated with it? Weren't you supposed to remove the calendars from the list of working-groups which are inactive?

peterstarling (changes requested):
I think you have missed an important edge case that we spoke about. By leaving the functionality that removes a calendar if it belongs to an inactive group you are effectively removing any calendar that belongs to both an inactive group and a working-group. Similarly you should make sure that for a calendar that is linked with more than one group and one of them is inactive the calendar is still present (something that should have been captured previously but worth having another look).

Although if you think such edge case does not need to be handled and we are ok to just remove the calendar if it belongs to two - you should still write a test for it so we know it was intended.

apps/crn-server/test/entities/working-group.test.ts
oratiu: Wrote a test for the entire input and output instead as it also contains the calendar part

apps/storybook/src/CalendarList.stories.tsx
```diff
@@ -14,7 +14,7 @@ export const Normal = () => (
   <CalendarList
     page={select<ComponentProps<typeof CalendarList>['page']>(
       'Page',
-      ['event', 'calendar', 'group'],
+      ['event', 'calendar', 'group', 'calendar-working-group'],
```
kmaid, line 17: Why is this calendar-working-group? Shouldn't it just be working-group?

apps/storybook/src/EventsCalendar.stories.tsx
```diff
@@ -1,16 +1,21 @@
...
-  calendars: calendars.slice(0, number('Calendar item', 1, { min: 0, max: 6 })),
-});
-
-export const Normal = () => <EventsCalendar {...props()} />;
+export const Normal = () => (
+  <EventsCalendar
+    calendars={
+      createListCalendarResponse(number('Calendar item', 1, { min: 0 }), {
+        activeGroups: boolean('Calendar connected to Group', true),
+        incompleteWorkingGroups: boolean(
```
kmaid, line 14: Should this be activeWorkingGroups to be consistent?

packages/fixtures/src/calendars.ts
```diff
@@ -1,101 +1,34 @@
...
-};
+export const createCalendarResponse = (
+  {
+    activeGroups = false,
+    incompleteWorkingGroups = false,
+  }: FixtureOptions = {},
+  itemIndex = 0,
+): CalendarResponse => ({
+  id: `calendar-${itemIndex}`,
+  color: '#B1365F',
```
kmaid, line 16: Can we vary the colour?

packages/fixtures/src/working-groups.ts
```diff
@@ -58,10 +60,13 @@ export const createWorkingGroupResponse = (
...
   lastModifiedDate: '2020-11-09T20:36:54Z',
   externalLink: `https://www.example.com/working-group-${itemIndex}`,
-  deliverables: createDeliverables(options?.deliverables ?? 1),
+  deliverables: createDeliverables(deliverables ?? 1),
   complete: false,
-  members: createWorkingGroupMembers(options?.members ?? 1),
-  leaders: createWorkingGroupLeaders(options?.members ?? 1),
+  members: createWorkingGroupMembers(members ?? 1),
+  leaders: createWorkingGroupLeaders(members ?? 1),
+  calendars: Array.from({ length: calendarsCount }, (_, index) =>
```
kmaid, line 67: I don't think you will get your default if you set deliverables to 1 because calendars count will be undefined. If thats not the case can you change teh others to be consistent with this approach

apps/storybook/src/EventsCalendar.stories.tsx
oratiu: Well it depends, the actual property is "complete" which means the working group is no longer active, I tried keeping consistency with how it's seen on Squidex

apps/storybook/src/CalendarList.stories.tsx
oratiu: That will happen once we add the calendar list on the Working Group page, basically "calendar" is meant for the Calendar page, where we showed the Groups Calendar list. I think I should rather rename "calendar" to "calendar-group"

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -54,67 +70,113 @@ const orderList = css({
...
+  const [showMore, setShowMore] = useState(false);
+  const displayShowMoreButton = calendars.length > truncateFrom;
+
+  return (
+    <div css={containerStyles}>
+      <Card padding={false}>
+        <div css={headerStyles}>
+          <Headline3 noMargin>
+            Subscribe to{' '}
+            {page === 'group'
```
kmaid, line 102: I think this would be better done with a Record object. There are too many ternary operators to follow.

`
{
'group' : {
	headline: 'aaa',
	description: 'bbb',
	}
}

`

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -54,67 +70,113 @@ const orderList = css({
...
+        )}
+        {displayShowMoreButton && (
+          <div css={showMoreStyles}>
+            <Button linkStyle onClick={() => setShowMore(!showMore)}>
+              View {showMore ? 'Less' : 'More'}
+            </Button>
+          </div>
+        )}
+      </Card>
+      {showBottomMessage && (
```
kmaid, line 164:
```suggestion
      {showSupportText && (
```
or something

packages/react-components/src/templates/EventsCalendar.tsx
```diff
@@ -13,10 +13,27 @@ type EventsCalendarPageProps = {
...
+    <GoogleCalendar calendars={calendars} />
+    <CalendarList
+      page="calendar"
+      calendars={calendars.filter(
+        (calendar) =>
+          calendar.activeGroups === true ||
+          (calendar.activeGroups === false &&
+            calendar.incompleteWorkingGroups === false),
+      )}
+      showBottomMessage={false}
```
kmaid, line 29:
```suggestion
	hideSupportText
```

apps/crn-server/src/data-providers/calendars.data-provider.ts
```diff
@@ -156,4 +173,12 @@ export const parseGraphQlCalendarToDataObject = (
   version: item.version,
   expirationDate: item.flatData.expirationDate,
   syncToken: item.flatData.syncToken,
+  activeGroups:
+    item.referencingGroupsContents?.findIndex(
+      (group) => group.flatData.active === true,
+    ) !== -1,
+  incompleteWorkingGroups:
+    item.referencingWorkingGroupsContents?.findIndex(
```
kmaid, line 181: You should really use .some

apps/storybook/src/CalendarList.stories.tsx
kmaid: Starting to wonder if you should just inject the title and description to simplify the whole thing. How many times is the component used? 4 times?

packages/react-components/src/organisms/__tests__/CalendarList.test.tsx
```diff
@@ -1,12 +1,14 @@
...
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
+import { createListCalendarResponse } from '@asap-hub/fixtures';
 
 import CalendarList from '../CalendarList';
 
 const props: ComponentProps<typeof CalendarList> = {
   calendars: [],
-  page: 'calendar',
+  title: 'Subscribe to this Interest Groups Calendar',
```
kmaid, line 9:
```suggestion
  title: '',
```
You don't want to include anything here that can be asserted on to encourage people to set the data within the tests for clarity,

kmaid (approved):

packages/model/src/calendar.ts
```diff
@@ -72,6 +74,8 @@ export interface CalendarResponse {
   id: string;
   name: string;
   color: GoogleLegacyCalendarColor;
+  activeGroups?: boolean;
```
peterstarling, line 77:
```suggestion
  hasActiveGroups?: boolean;
```

packages/model/src/calendar.ts
```diff
@@ -72,6 +74,8 @@ export interface CalendarResponse {
...
+  incompleteWorkingGroups?: boolean;
```
peterstarling, line 78:
```suggestion
  hasIncompleteWorkingGroups?: boolean;
```

peterstarling (changes requested):

packages/model/src/calendar.ts
kmaid: I really don't like prefixing has on variables. I think its very easily inferred from the type and overly verbose. I think it also encourages good practices because if we ever needed to return an actual list of groups it wouldn't allow you to also maintain this flag with the same variable name

peterstarling: i agree we should not prefix properties such as `active` with prefixes like `has` or `is`but i would actually use your example as a reason why we should do it here - if we ever want to return an actual list of `groups` or `activeGroups` we will not be able to as the name will already be taken. 

It is very confusing to have a `groups` property which is not an array of groups (like everywhere else) but a boolean. 

We want to be consistent and to leave the interfaces open for extension, not for modification (see what i did here)

peterstarling (changes requested):

PR #2512, opened 2023-01-25
navigation header

jhonny-m:


packages/gp2-components/src/molecules/OnboardedTabLink.tsx
```diff
@@ -41,22 +42,50 @@ const mobileStyle = css({
   },
 });
 
+interface OnboardedTabLinkProps {
```
markcsims, line 45: interface or type?

packages/gp2-components/src/molecules/OnboardedTabLink.tsx
```diff
@@ -2,15 +2,12 @@ import { pixels } from '@asap-hub/react-components';
 import { css } from '@emotion/react';
 import { ReactNode } from 'react';
 import { NavLink } from 'react-router-dom';
+import onboardingCompletedIcon from '../icons/onboarding-completed-icon';
```
csbarbosa, line 5: Maybe add the icons to the index file and then import them from there

markcsims (approved):

PR #2519, opened 2023-01-26
add Keywords modal to Onboarding

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/GP2-149

### DIFF
--- a/packages/gp2-components/src/organisms/__tests__/KeywordsModal.test.tsx
+++ b/packages/gp2-components/src/organisms/__tests__/KeywordsModal.test.tsx
@@ -1,0 +1,95 @@
+import { gp2 as gp2Fixtures } from '@asap-hub/fixtures';
+import { Keyword } from '@asap-hub/model/src/gp2';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { ComponentProps } from 'react';
+import { StaticRouter } from 'react-router-dom';
+import KeywordsModal from '../KeywordsModal';
+
+describe('KeywordsModal', () => {
+  const getSaveButton = () => screen.getByRole('button', { name: 'Save' });
+
+  beforeEach(jest.resetAllMocks);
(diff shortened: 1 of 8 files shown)


packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -148,6 +149,7 @@ const MultiSelect = <T extends MultiSelectOptionsType>({
   components,
   enabled = true,
   placeholder = '',
+  maxMenuHeight,
```
csbarbosa, line 152: Shouldn't be added in here a default value, since this prop is optional?

packages/gp2-components/src/organisms/__tests__/KeywordsModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const textbox = screen.getByRole('textbox', {
+      name: /Keywords/i,
+    });
+    expect(textbox).toBeVisible();
+    expect(screen.getByText(expected)).toBeVisible();
+  });
+
+  it('calls onSave with the right arguments', async () => {
+    const onSave = jest.fn();
+    const keywords = ['Genetics'] as Keyword[];
```
csbarbosa, line 49: Instead of importing Keyword directly, use the gp2 import and put "gp2.Keyword[]"

packages/react-components/src/atoms/MultiSelect.tsx
dianaguimaraes: No, as you said it is optional. If it's not set I don't want to use it.

packages/gp2-components/src/organisms/__tests__/KeywordsModal.test.tsx
dianaguimaraes: I would agree if there were other imports from `gp2`, which is not the case here. So I don't see a point in importing the whole `gp2` if I just need `Keyword`

csbarbosa (approved):

jhonny-m (approved):

PR #2524, opened 2023-01-27
Crn 1236 introduce working groups on events calendar page

oratiu:


apps/crn-server/test/data-providers/calendars.data-provider.test.ts
```diff
@@ -204,7 +207,7 @@ describe('Calendars data provider', () => {
     });
 
     describe('Active filter', () => {
-      test('Should skip the calendars which belong to an inactive group when active is set to true', async () => {
+      test('Should skip the calendars which belong to an inactive group and complete working group when active is set to true', async () => {
```
peterstarling, line 210: to be fair this test was easy to read when it was just that one condition (ie skip calendars linked to an inactive group) but now its a bit of a nightmare. 

Plus - do you honestly thing with the extra requirement you have tested all the scenarios? I would add tests for when
1) Active flag is on - calendar belongs to an inactive group and a working group that is not complete - it shows
2) Active flag is on - calendar belong to an active group and a working group that is complete - it shows
3) Active flag is on - calendar belongs to an inactive group and a working group that is complete - it does not show - this is technically the test i am commenting in now - however maybe its better to run this scenario in isolation, meaning only one calendar in the input fixture
4) Active flag is on - multiple calendars, one belongs to an active group and a complete working group - it shows, second belongs to an inactive group and an incomplete working group - it shows, third belongs to an inactive group and a working group that is completed - it does not show, fourth does not belong to any groups or working groups - it shows
5) Active flag is on - calendar belongs to multiple groups of which one is inactive - it shows (this test already exists https://github.com/yldio/asap-hub/blob/14e72948d47f9b70ef1c1342cd9b732a232ea991/apps/crn-server/test/data-providers/calendars.data-provider.test.ts#L344)
6) Active flag is on - calendar belogns to multiple working groups of which one is completed - it shows
7) Active flag is off - same as 4 except this time all calendars are shown

packages/model/src/calendar.ts
```diff
@@ -72,6 +76,8 @@ export interface CalendarResponse {
   id: string;
   name: string;
   color: GoogleLegacyCalendarColor;
+  groups?: Pick<GroupDataObject, 'id' | 'active'>[];
```
peterstarling, line 79: I don't think this makes much sesnse, we don't need groups or working-groups on the create method, especially not with both ID and Active properties. 

There is currently no requirement for creating calendars which associations and if you had looked at where this type is actually being used you would have found that the create method is not actually used anywhere while the update method of the data provider is exposed in a controller but only uses this input https://github.com/yldio/asap-hub/blob/14e72948d47f9b70ef1c1342cd9b732a232ea991/packages/model/src/calendar.ts#L85-L88

packages/model/src/working-group.ts
```diff
@@ -57,6 +58,7 @@ export type WorkingGroupDataObject = {
   pointOfContact?: WorkingGroupLeader;
   complete: boolean;
   shortText: string;
+  calendars: ReadonlyArray<CalendarResponse>;
```
peterstarling, line 61: This should be:
```suggestion
  calendars: CalendarDataObject[];
```
You are mixing types from different abstractions. There is a clear line between data providers and controllers. In terms of calendars the data provider returns all data from the database (CMS) and then the controller only picks the stuff that we want to return to the API (as that is it's responsibility as a Controller), see here:
https://github.com/yldio/asap-hub/blob/789ba1b2a27f01d360f5d12a96b936dc93308c38/apps/crn-server/src/controllers/calendars.ts#L58-L67

Similarly - in here you might need to return the entire CalendarDataObject and then parse this in the controller to reduce it but that it the whole point.

peterstarling (changes requested):

packages/react-components/src/templates/EventsCalendar.tsx
```diff
@@ -13,10 +13,31 @@ type EventsCalendarPageProps = {
...
+  calendars,
+}) => (
   <div css={containerStyles}>
-    <GoogleCalendar {...props} />
-    <CalendarList page="calendar" {...props} />
+    <GoogleCalendar calendars={calendars} />
+    <CalendarList
+      calendars={calendars.filter(
+        (calendar) =>
+          calendar.groups?.some(({ active }) => active) ||
```
kmaid, line 24: Why are groups potentially undefined? Should always be an array even if its an empty one

packages/react-components/src/templates/__tests__/EventsCalendar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  expect(getByText('Test Event')).toBeVisible();
+});
+
+it('Renders the calendar both for working and interest group', () => {
+  const { getAllByText } = render(
+    <EventsCalendar
+      calendars={[
+        {
+          color: '#0D7813',
+          name: 'Test Event',
```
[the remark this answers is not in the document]
kmaid: Please check the rest of your code for calling a calendar an event. I saw it in at least one other place

peterstarling (approved):

packages/react-components/src/organisms/CalendarList.tsx
```diff
@@ -54,67 +70,102 @@ const orderList = css({
...
+            <Paragraph accent="lead" noMargin>
+              {description}
+            </Paragraph>
+          </div>
+        )}
+        {!!calendars.length && (
+          <ul
+            css={[
+              orderList,
+              !displayShowMoreButton && {
```
kmaid, line 115: This is going to make warnings in the test output because a boolean is not serialized strings

kmaid (approved):

PR #2539, opened 2023-02-01
fix: CRN-1270 Support Multiple Auth Providers

peterstarling:
Fixes the following issues:
1. When the user added a new auth0 connection any previous connections were removed - it should now add them to the list and only remove the welcome code - this was fixed in the Controllers
2. When a user was published again it wiped all their connections - it should now create a new welcome code but keep the previous connections - this was fixed in the Invite Handler
3. The `alumniSinceDate` was an optional property that should have been provided for the Team, Group, Working Group and Event Responses however it was only implemented for the UserResponse
4. More optional properties were added to fixtures to avoid unimplemented optional properties in User Data Provider


apps/crn-server/src/queries/events.queries.ts
```diff
@@ -58,6 +58,7 @@ export const eventContentFragment = gql`
             id
             flatData {
               alumniSinceDate
+              alumniLocation
```
gabiayako, line 61: I believe we don't need this `alumniLocation` everywhere. We do need this in `user` because it appears in the user profile page
user-location


but in the event, for example, we need the `alumniSinceDate` to flag with this "A" badge beside the name of the user, but we don't show the `alumniLocation`

event

I haven't took a look at groups and research output in the hub, but I believe it's the same thing

apps/crn-server/src/controllers/users.ts
```diff
@@ -109,9 +109,11 @@ export default class Users implements UserController {
     if (user.connections?.find(({ code }) => code === userId)) {
       return parseUserToResponse(user);
     }
+    const existingConnections =
+      user.connections?.filter(({ code }) => code !== welcomeCode) || [];
```
markcsims, line 113: Don't we want to keep the welcome code? 

I thought that was needed for some reason?

peterstarling: i dont think so? why would we want to, do you think?

markcsims (approved):

gabiayako (approved):

PR #2567, opened 2023-02-08
Do sharding

kmaid:


.github/workflows/reusable-test.yml
```diff
@@ -52,8 +56,26 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [unit]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/download-artifact@v3
+        with:
+          name: coverage-artifacts
+          path: coverage
+      - name: Merge Code Coverage
+        run: npx nyc merge coverage/ merged-output/merged-coverage.json
```
peterstarling, line 71: yarn?

markcsims (approved):

.github/workflows/reusable-test.yml
kmaid: There is no NPX equivalent in yarn. I don't think its worth adding nyc into our project just for this. Would be slower as you would need to check out the project which i might be doing unnecessarily anyway

peterstarling (approved):

PR #2581, opened 2023-02-09
Crn 1300 further cleanup of research output removal of publishing entity

oratiu:


apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -161,17 +160,28 @@ export const getResearchOutputDataObject = (): ResearchOutputDataObject => ({
...
-  publishingEntity: 'Team',
 });
 
 export const getListResearchOutputDataObject =
-  (): ListResearchOutputResponse => ({
-    total: 1,
-    items: [getResearchOutputDataObject()],
-  });
+  (): ListResponse<ResearchOutputDataObject> => {
+    const researchOutput = getResearchOutputDataObject();
```
peterstarling, line 167: this seems unnecessary?

packages/validation/src/research-output-guards.ts
```diff
@@ -0,0 +1,16 @@
...
+  ResearchOutputWorkingGroupPostRequest,
+  ResearchOutputWorkingGroupResponse,
+} from '@asap-hub/model/src/research-output';
+
+export const isResearchOutputWorkingGroup = (
+  researchOutputData: ResearchOutputResponse,
+): researchOutputData is ResearchOutputWorkingGroupResponse =>
+  !!researchOutputData.workingGroups?.length;
+
+export const isResearchOutputWorkingGroupRequest = (
```
peterstarling, line 13: what do you need this for?

packages/model/src/research-output.ts
```diff
@@ -295,11 +297,17 @@ export type ResearchOutputPostRequest = {
...
   usageNotes?: string;
   usedInPublication?: boolean;
-  publishingEntity: ResearchOutputPublishingEntities;
 };
 
 export type ResearchOutputPutRequest = ResearchOutputPostRequest;
 
+type NonEmptyArray<T> = [T, ...T[]];
+
+export type ResearchOutputWorkingGroupPostRequest =
```
peterstarling, line 306: do we need a separate type for the Post request?

peterstarling (changes requested):

packages/model/src/research-output.ts
oratiu: I'd say we need one since we have a error message that can be either addressed to teams or working groups. The researchOutputData used there is of type ResearchOutputPostRequest and I'm using this type in the type guard.

peterstarling (approved):

kmaid (approved):

PR #2599, opened 2023-02-13
CRN-1206 update schema

lctrt:


gabiayako (changes requested):
I've noticed that the name of the working group is not clickable in the card. Not sure if it's part of this ticket though

Screenshot 2023-02-14 at 09 05 21

Another thing is that even when there is a research output it is not appearing in the working group profile page
Screenshot 2023-02-14 at 09 09 13

gabiayako (approved):
Approving it because the things I pointed it out were not part of this ticket :)

PR #2603, opened 2023-02-13
CRN-1282 Add publishDate in News content type

gabiayako:
This PR:

- Adds a migration to add a new field `publishDate` to `News` content type in contentful: [20230213120700-add-publish-date-field.js](https://github.com/yldio/asap-hub/pull/2603/files#diff-2e1c067d6bece83a30a467403216f76f57ba95daf85c1310299b94eda2cea3b2)
- Adds squidex `created` field to cms-data sync query that will be the value of `publishDate` in Contentful: [cms-data-sync/src/news/news.queries.ts](https://github.com/yldio/asap-hub/pull/2603/files#diff-6cb365b0a9e925c705dfd08becc0678f0501dd9ca6935ca38b48682a7de26838)
- Changes the ordering of the news from contentful's `firstPublishedAt` to the new `publishDate`

--- 

### Considerations:

I've made `publishDate` as required for now, which is a bad user experience in my view. But we can try to make it fill automatically after this spike https://asaphub.atlassian.net/browse/CRN-1295

Some notes:
- I thought about trying to make squidex `created` date to be contentful's `firstPublishedAt`, but it's not possible, read more here https://github.com/contentful/contentful-import/issues/16
- I thought about leaving `publishDate` as an optional field and in the graph ql parser I would do a logic to make the `created` date to be squidex `created` date if it exists and it is less than contentful's `firstPublishedAt` otherwise it would be contentful's `firstPublishedAt`. And I removed the `order` from the graphql request and in the return I would do something like this
```
    return {
      total: newsCollection?.total,
      items: newsCollection?.items
        .filter((x): x is NewsItem => x !== null)
        .map(parseContentfulGraphQlNews)
        .sort((a, b) => {
          dateA = new Date(a.created);
          dateB = new Date(b.created);
          return dateB - dateA
        }
    };
```
I implemented it, but then I saw what appeared in the hub and it wasn't working. I think this was because we fetch using pagination, so this code above is sorting only the 10 news that came via graphql not the whole set of news. So I ended up thinking that we do need to use `order`:

```
  async fetch(options?: FetchNewsProviderOptions) {
    const { newsCollection } = await this.contentfulClient.request<
      FetchNewsQuery,
      FetchNewsQueryVariables
    >(FETCH_NEWS, {
      limit: options?.take || null,
      skip: options?.skip || null,
      order: [NewsOrder.PublishDateDesc],                        <---
      where: {
        frequency_in: options?.filter?.frequency as string[],
        title_contains: options?.filter?.title || null,
      },
    });
```

and for that we do need to make `publishDate` required


packages/contentful/migrations/news/20230213120700-add-publish-date-field.js
```diff
@@ -0,0 +1,24 @@
...
+  const news = migration.editContentType('news');
+  news
+    .createField('publishDate')
+    .name('Publish Date')
+    .type('Date')
+    .localized(false)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
```
peterstarling, line 13: can we add a default to today?

peterstarling (approved):

packages/contentful/migrations/news/20230213120700-add-publish-date-field.js
gabiayako: hmm good idea, let me test that

gabiayako: Trying to add default value here https://github.com/yldio/asap-hub/pull/2603/commits/733a373e0ad67aadbb701e2d8d2574206ab05149

It becomes a snapshot of the date when the migration was created 

schema

content-create

I will revert this last commit

PR #2604, opened 2023-02-13
[CRN-1214] refactor(working group): move member logic to backend

Riccardo-Brown:


packages/model/src/working-group.ts
```diff
@@ -39,6 +40,7 @@ export type WorkingGroupMember = {
     | 'email'
     | 'avatarUrl'
   >;
+  readonly isActive: boolean;
```
peterstarling, line 43: by adding `isActive` property to the `WorkingGroupMember` you are effectively adding it to `WorkingGroupDataObject` parent object which I think is not a valid approach. The reason is that our data-providers should only be an abstraction layer on top of the data store (CMS) which send the data back and forth and they should be fairly 'dumb' in a sense that they should not really perform any complicated logic and should not return properties calculated 'on the go'.

Although `WorkingGroupResponse` is currently the same as `WorkingGroupDataObject` I think you should make sure that the new property is added only to members on the `WorkingGroupResponse` type. This means you are going to have to perform the check in the controller but that's ok.

peterstarling (changes requested):

apps/crn-server/test/fixtures/working-groups.fixtures.ts
```diff
@@ -50,7 +50,7 @@ export const getWorkingGroupDataObject = (): WorkingGroupDataObject => ({
 });
 
 export const getWorkingGroupResponse = (): WorkingGroupResponse =>
-  getWorkingGroupDataObject();
+  getWorkingGroupDataObject() as WorkingGroupResponse;
```
peterstarling, line 53: i dont think this is true though? there is that inActive property marked as required that you added to leaders and members and this fixture doesn't have it

packages/model/src/event.ts
```diff
@@ -77,7 +77,7 @@ export interface EventResponse {
 
   calendar: CalendarResponse;
   group?: GroupResponse;
-  workingGroup?: WorkingGroupResponse;
+  workingGroup?: WorkingGroupDataObject;
```
peterstarling, line 80: incorrect, you shouldn't mix these. This Response type should only reference another Response type

peterstarling (changes requested):

apps/crn-server/test/fixtures/working-groups.fixtures.ts
Riccardo-Brown: So for all the fixtures we don't populate the leaders or member properties, so the type of `WorkingGroupResponse` is the same as `WorkingGroupObject`. Should i just copy getWorkingGroupDataObject function instead of calling it?

peterstarling (approved):

PR #2630, opened 2023-02-16
outputs in gp2

markcsims:


apps/gp2-server/src/app.ts
```diff
@@ -189,7 +215,13 @@ export const appFactory = (libs: Libs = {}): Express => {
   const eventDataProvider =
     libs.eventDataProvider ||
     new EventSquidexDataProvider(eventRestClient, squidexGraphqlClient);
+  const outputDataProvider =
+    libs.outputDataProvider ||
+    new OutputSquidexDataProvider(squidexGraphqlClient, outputRestClient);
 
+  const externalAuthorDataProvider =
+    libs.externalAuthorDataProvider ||
+    new ExternalAuthorSquidexDataProvider(externalAuthorRestClient);
```
jhonny-m, line 224: we don't use the squidexGraphqlClient for this?

apps/gp2-server/src/controllers/output.controller.ts
```diff
@@ -0,0 +1,233 @@
...
+    return output;
+  }
+
+  async fetch(options: FetchOptions): Promise<gp2Model.ListOutputResponse> {
+    const { filter: fetchFilter, ...fetchOptions } = options;
+
+    const filter: gp2Model.FetchOutputOptions['filter'] = Array.isArray(
+      fetchFilter,
+    )
+      ? { documentType: fetchFilter }
```
jhonny-m, line 37: I'm a bit confused :D

apps/gp2-server/src/controllers/output.controller.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    const errors = (
+      await Promise.all([
+        this.validateTitleUniqueness(outputData, outputId),
+        this.validateLinkUniqueness(outputData, outputId),
+      ])
+    ).filter(isError);
+
+    if (errors.length > 0) {
+      // TODO: Remove Boom from the controller layer
```
jhonny-m, line 127: When are we doing this?

apps/gp2-server/src/utils/odata.ts
```diff
@@ -0,0 +1,24 @@
+import buildQuery, { Filter } from 'odata-query';
+
+const FILTER_URL_QUERY_PARAM = '?$filter=';
+
+export const buildODataFilter = (filterObj: Filter): string =>
+  buildQuery({ filter: filterObj }).replace(FILTER_URL_QUERY_PARAM, '');
+
+export const buildEqFilterForWords = (
```
jhonny-m, line 8: where is this used?

jhonny-m: if it's the same as crn we can move it to a shared package

apps/gp2-server/src/app.ts
markcsims: Because there is only a create method. For creating things we use Rest.

In CRN, External Authors are fetched from Algolia. We need to discuss this, when we need to get them.

apps/gp2-server/src/controllers/output.controller.ts
markcsims: Copied from the research output controller: https://github.com/yldio/asap-hub/blob/a275282553687f40e935bf158c07c805d9384b35/apps/crn-server/src/controllers/research-outputs.ts#L61-L65

It confused me too, the filter can take a string[]. If it does, then it's an array of documentTypes.

apps/gp2-server/src/controllers/output.controller.ts
markcsims: When are we removing the TODO?

apps/gp2-server/src/utils/odata.ts
markcsims: it's a util package copied over from CRN.

Once we move to Contentful, it should be deleted.

jhonny-m: but are we using the second function?

jhonny-m (approved):

PR #2639, opened 2023-02-17
Added permission for editing working group research outputs

oratiu:


apps/crn-frontend/src/shared-research/ResearchOutput.tsx
```diff
@@ -17,13 +20,25 @@ const ResearchOutput: React.FC = () => {
   const { path } = useRouteMatch();
   const researchOutputData = useResearchOutputById(researchOutputId);
   const backHref = useBackHref() ?? sharedResearch({}).$;
-  const canCreateUpdate = useCanCreateUpdateResearchOutput(
+  const canCreateUpdateTeam = useCanCreateUpdateResearchOutput(
```
kmaid, line 23: We are trying to keep Research Output agnostic of being Team or WG. Can you move these two permission functions to either Team or WG output? It might also make sense to move the context up there too

apps/crn-frontend/src/shared-research/state.ts
```diff
@@ -160,3 +161,19 @@ export const useSetResearchOutputItem = () => {
     setRefresh(refresh + 1);
   };
 };
+
+export const useCanUpdateWorkingGroupResearchOutput = (
```
kmaid, line 165: This isn't really state as it doesn't touch recoil. Why do we have this function twice with the one thats currently in RO. (( Self: Come back and figure this out better ))

PR #2654, opened 2023-02-22
Crn 1181 team associations at user profile

ricardocarvalh0:


packages/model/src/user.ts
```diff
@@ -104,6 +104,7 @@ export interface OrcidWork {
 export interface UserTeam {
   id: string;
   displayName?: string;
+  teamInactiveSince?: string;
```
Riccardo-Brown, line 107: Should this field be optional as a value is always set https://github.com/yldio/asap-hub/pull/2654/files#diff-9c6a0821d4e3970c62d1271d10fa7de6fe33823b99083824f6b340875adc5613R256

ricardocarvalh0: I can be set as undefined

oratiu: If it can be set as undefined, shouldn't it be teamInactiveSince: string | undefined ?

packages/react-components/src/organisms/UserTeamsTabbedCard.tsx
```diff
@@ -0,0 +1,173 @@
...
+  teams: UserTeam[];
+};
+
+const UserTeamsTabbedCard: React.FC<UserTeamsTabbedCardProps> = ({
+  isUserAlumni,
+  userName,
+  teams,
+}) => {
+  const sortedTeams = [...teams].sort(
+    (a, b) => priorities[a.role] - priorities[b.role],
```
oratiu, line 86: This has to pass the coverage, there is no test file for this component

packages/react-components/src/organisms/UserTeamsTabbedCard.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+const titleStyle = css({
+  fontWeight: 'bold',
+});
+
+const detailsContentStyle = css({
+  marginBottom: `${24 / perRem}em`,
+});
+
+const priorities: Record<TeamRole, number> = {
```
oratiu, line 61: Nice

packages/react-components/src/organisms/UserTeamsTabbedCard.tsx
```diff
@@ -0,0 +1,173 @@
...
+  'ASAP Staff': 6,
+  'Scientific Advisory Board': 7,
+};
+
+type UserTeamsTabbedCardProps = Pick<
+  ComponentProps<typeof TabbedCard>,
+  'description'
+> & {
+  userName: string;
+  isUserAlumni: boolean;
```
oratiu, line 76: Tom suggested we drop "is", "has" etc from variable naming due to the fact that we use TypeScript and thus it's easy to see what type that specific property has. So userAlumni would suffice.

oratiu (changes requested):

packages/model/src/user.ts
ricardocarvalh0: Afaik `teamInactiveSince?: string;` covers both scenarios. When the prop is not set and when it's set as undefined.

oratiu (approved):

PR #2667, opened 2023-02-24
CRN-792 Shared Output: Save Draft

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-792

---

This PR:

#### Refactors and unifies the logic of permissions in [permissions/research-output.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-a2d2c48ee4a62b3c20507554b499f65dc16e6f96181a49c97a193640ebef232a)

The role permission now is handled by `getUserRole` and it was created 3 small functions to get the permissions: `hasShareResearchOutputPermission`, `hasEditResearchOutputPermission` and `hasPublishResearchOutputPermission`.

#### Makes it possible to submit a draft research output

In squidex, if we call the `create` method from the rest client passing `published` as `false` it creates the entry as draft. So some files were changed in order to be able to pass this `published` value as false.

In the backend:
- Data Provider: [research-outputs.data-provider.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-335d39a6d3239d2e50e6d49644deed3e455234cd469fe6a10c150aaf6fb6d55b)
- Controller: [controllers/research-outputs.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-0844f18f1195226c257bc5b66c57d291c8724b24d84daa78559542734e2be16d)
- Routes: [research-outputs.route.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-3e0d42e033394d10aea4f686a57853792c297de218f3d6953a843a4bb92ea63a)

In the frontend:
- Create and edit requests to backend in api.ts: [api.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-0392bae59e98156dbdeb1a6c8282988298487d7a608550327976e9dfc2d43190)
- Functions that pass `published` as props to be linked to `Save`/`Publish` or `Save Draft` buttons: [TeamOutput.tsx](https://github.com/yldio/asap-hub/pull/2667/files#diff-d5a475e3660bbdce7b7935b0ec9ef76091eceb2fe44e3ffb47a20654f4ec5620) and [WorkingGroupOutput.tsx](https://github.com/yldio/asap-hub/pull/2667/files#diff-f7fbc1c6db856f0b7fc204ce0a5d6ed5fb32a6ed896ec53c82fc6ae79e510481)

#### Makes Share an Output and Edit buttons visible by team/wg members behind a feature flag

- Share an Output button: [WorkingGroupPageHeader.tsx](https://github.com/yldio/asap-hub/pull/2667/files#diff-5eb46f54f93b38fd5ec9982f0a550b480fb6e58361d5bf8defbf6d0935fce8b3) and [TeamProfileHeader.tsx](https://github.com/yldio/asap-hub/pull/2667/files#diff-46d379951376a2e3edc84f65ba24113afee04c2dd13f0e956d48ce359d93b963)
- Edit button: [SharedResearchOutput.tsx](https://github.com/yldio/asap-hub/pull/2667/files#diff-941f382146f83a4f7dfda62eff635275030346a7f17846770572dd7b76aba2b4)

#### Displays form buttons accordingly to permissions and research output status

To do that, it was introduced a `published` field in the research output model [packages/model/src/research-output.ts](https://github.com/yldio/asap-hub/pull/2667/files#diff-20e35ac4b9b198b232ec54b753925d8b9c4a3825d1ef24e44bd553586b68b241). That currently **is not** being retrieved properly from the backend, it is hardcoded now, the requirement to bring draft from the backend is not part of this ticket, so this was not added.
 
Screenshot 2023-02-28 at 12 34 20

---

#### Note

If you click Save Draft it will give an error for now because right after saving the RO, it fetches the same RO but since it is a draft, it can’t be found yet. The part of fetching draft is part of another ticket. But you can still verify that the draft was created in squidex.

### DIFF
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -56,8 +56,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
 
   const createResearchOutput = usePostResearchOutput({ publish: true });
   const createDraftResearchOutput = usePostResearchOutput({ publish: false });
-  const updateResearchOutput = usePutResearchOutput({ publish: true });
-  const updateDraftResearchOutput = usePutResearchOutput({ publish: false });
+  const updateResearchOutput = usePutResearchOutput();
 
   const getLabSuggestions = useLabSuggestions();
   const getAuthorSuggestions = useAuthorSuggestions();
@@ -123,7 +122,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
           }
           onSaveDraft={(output) =>
             researchOutputData
-              ? updateDraftResearchOutput(researchOutputData.id, output).catch(
+              ? updateResearchOutput(researchOutputData.id, output).catch(
                   handleError(['/link', '/title'], setErrors),
                 )
               : createDraftResearchOutput(output).catch(
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -9,13 +9,12 @@ import {
   ResearchOutputForm,
   ResearchOutputHeader,
 } from '@asap-hub/react-components';
-import { ResearchOutputPermissionsContext } from '@asap-hub/react-context';
 import {
   network,
   TeamOutputDocumentTypeParameter,
   useRouteParams,
 } from '@asap-hub/routing';
-import React, { useContext, useState } from 'react';
+import React, { useState } from 'react';
 import researchSuggestions from './research-suggestions';
 import { useTeamById } from './state';
 import {
@@ -28,6 +27,7 @@ import {
   usePutResearchOutput,
   usePostResearchOutput,
 } from '../../shared-research';
+import { useResearchOutputPermissions } from '../../shared-research/state';
 
 const useParamOutputDocumentType = (
   teamId: string,
@@ -54,10 +54,6 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
 
-  const { canEditResearchOutput } = useContext(
-    ResearchOutputPermissionsContext,
-  );
-
   const createResearchOutput = usePostResearchOutput({ published: true });
   const createDraftResearchOutput = usePostResearchOutput({ published: false });
   const updateResearchOutput = usePutResearchOutput({ published: true });
@@ -70,7 +66,13 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
 
   const published = researchOutputData ? !!researchOutputData.published : false;
 
-  if (canEditResearchOutput && team) {
+  const permissions = useResearchOutputPermissions(
+    'teams',
+    [teamId],
+    published,
+  );
+
+  if (permissions.canEditResearchOutput && team) {
     return (
       <Frame title="Share Research Output">
         <ResearchOutputHeader
@@ -109,6 +111,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
             }),
           )}
           published={published}
+          permissions={permissions}
           onSave={(output) =>
             researchOutputData
               ? updateResearchOutput(researchOutputData.id, output).catch(
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -54,7 +54,9 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
 
-  const { permissions } = useContext(ResearchOutputPermissionsContext);
+  const { canEditResearchOutput } = useContext(
+    ResearchOutputPermissionsContext,
+  );
 
   const createResearchOutput = usePostResearchOutput({ published: true });
   const createDraftResearchOutput = usePostResearchOutput({ published: false });
@@ -68,11 +70,7 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
 
   const published = researchOutputData ? !!researchOutputData.published : false;
 
-  const isForbidden =
-    (!permissions.saveDraft && !permissions.publish) ||
-    (published && permissions.saveDraft && !permissions.publish);
-
-  if (!isForbidden && team) {
+  if (canEditResearchOutput && team) {
     return (
       <Frame title="Share Research Output">
         <ResearchOutputHeader
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -54,10 +54,10 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
 
-  const createResearchOutput = usePostResearchOutput({ published: true });
-  const createDraftResearchOutput = usePostResearchOutput({ published: false });
-  const updateResearchOutput = usePutResearchOutput({ published: true });
-  const updateDraftResearchOutput = usePutResearchOutput({ published: false });
+  const createResearchOutput = usePostResearchOutput({ publish: true });
+  const createDraftResearchOutput = usePostResearchOutput({ publish: false });
+  const updateResearchOutput = usePutResearchOutput({ publish: true });
+  const updateDraftResearchOutput = usePutResearchOutput({ publish: false });
 
   const getLabSuggestions = useLabSuggestions();
   const getAuthorSuggestions = useAuthorSuggestions();
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -54,17 +54,25 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
 
-  const { canCreateUpdate } = useContext(ResearchOutputPermissionsContext);
+  const { permissions } = useContext(ResearchOutputPermissionsContext);
 
-  const createResearchOutput = usePostResearchOutput();
-  const updateResearchOutput = usePutResearchOutput();
+  const createResearchOutput = usePostResearchOutput({ published: true });
+  const createDraftResearchOutput = usePostResearchOutput({ published: false });
+  const updateResearchOutput = usePutResearchOutput({ published: true });
+  const updateDraftResearchOutput = usePutResearchOutput({ published: false });
 
   const getLabSuggestions = useLabSuggestions();
   const getAuthorSuggestions = useAuthorSuggestions();
   const getTeamSuggestions = useTeamSuggestions();
   const researchTags = useResearchTags();
 
-  if (canCreateUpdate && team) {
+  const published = researchOutputData ? !!researchOutputData.published : false;
+
+  const isForbidden =
+    (!permissions.saveDraft && !permissions.publish) ||
+    (published && permissions.saveDraft && !permissions.publish);
+
+  if (!isForbidden && team) {
     return (
       <Frame title="Share Research Output">
         <ResearchOutputHeader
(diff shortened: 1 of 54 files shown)


apps/crn-server/src/entities/research-output.ts
```diff
@@ -115,6 +115,7 @@ export const parseGraphQLResearchOutput = (
         id: group.id,
         title: group.flatData.title || '',
       })) || [],
+    published: true,
```
gabiayako, line 118: This is the part that is hard coded for now that I've mentioned in the description of the PR. This is part of another ticket [Shared Output: Display Draft](https://asaphub.atlassian.net/browse/CRN-1304)

apps/crn-frontend/src/shared-research/state.ts
```diff
@@ -153,10 +153,14 @@ export const useSetResearchOutputItem = () => {
...
-  researchOutputData: ResearchOutputResponse | undefined,
-): userPermissions => {
+export const useUserPermissions = ({
+  entity,
+  entityIds,
+}: {
+  entity: 'teams' | 'workingGroups';
+  entityIds: string[];
+}): UserPermissions => {
   const user = useCurrentUserCRN();
```
gabiayako, line 163: We've discussed in the last RTM that maybe here we shouldn't get the user from Auth0, but I believe if we do this we should split into another ticket/PR since this one is already huge.

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -376,6 +377,7 @@ export interface ResearchOutputController {
   fetchById: (id: string) => Promise<ResearchOutputResponse>;
   create: (
     researchOutputRequest: ResearchOutputCreateData,
+    published?: boolean,
```
peterstarling, line 380: I think it should be 
```suggestion
    publish?: boolean,
```
the past tense just doesn't make much sense to me, what do you think?

peterstarling: In fact maybe better if it was an object with a property `publish` ? That way if we ever need to add more options to the this method we don't need to add another argument, just expand this one. I would call this object CreateOptions or something

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -45,44 +47,61 @@ export const researchOutputRouteFactory = (
...
+    const permissions = getUserPermissions(
+      loggedInUser as UserResponse,
+      'teams',
+      createRequest.teams,
+    );
+
+    const options = validateResearchOutputRequestQueryParameters(query);
+    const published = options.published ?? true;
+
+    const isRequestAllowed =
```
peterstarling, line 63: I would say this is part of checking if the user has the permissions. I find this much more difficult to read now and I think there is no such separation of concerns as before. 

Maybe you should consider sending the `publish` query parameter across to the `hasCreateUpdateResearchOutputPermissions` function and keep the code more similar to what i was before? 

Before it was: `if i dont have permissions then throw` 
Now it is: `do i have permissions. But also is this a draft save or normal. And then do i have permissions if its one or the other, then throw`

Do you see what i mean?

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -45,44 +47,61 @@ export const researchOutputRouteFactory = (
...
-      )
-    ) {
+    const permissions = getUserPermissions(
+      loggedInUser as UserResponse,
+      'teams',
+      createRequest.teams,
+    );
+
+    const options = validateResearchOutputRequestQueryParameters(query);
+    const published = options.published ?? true;
```
peterstarling, line 61: once again I don't think we should use a past tense here, on the API request. If I want create a resource I can decide to either `publish` it or not

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -45,44 +47,61 @@ export const researchOutputRouteFactory = (
...
       validateResearchOutputPostRequestParametersIdentifiers(body);
 
-      if (
-        !loggedInUser ||
-        !hasCreateUpdateResearchOutputPermissions(
-          loggedInUser,
-          updateRequest.teams,
-        )
-      ) {
+      const permissions = getUserPermissions(
```
peterstarling, line 91: same as above, rather than split into stages i would just pass on the `published` option (should be `publish` but thats a different comment) to the `hasCreateUpdateResearchOutputPermissions` function which should determine permissions based on the input

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -153,13 +153,58 @@ describe('ResearchOutputs controller', () => {
...
+          .spyOn(global, 'Date')
+          .mockImplementation(() => mockDate);
+
+        const researchOutputCreateData = getResearchOutputCreateData();
+        const researchOutputId = 'research-output-id-1';
+        researchOutputDataProviderMock.create.mockResolvedValueOnce(
+          researchOutputId,
+        );
+
+        const result =
```
peterstarling, line 185: im not a huge fan of such logic inside tests, i think would prefer a `describe` about publishing within which we would have one test to prove that when its set to false - false is passed on to the controller, and when its missing - it defaults to true

packages/model/src/research-output.ts
```diff
@@ -261,7 +262,7 @@ export type ResearchOutputTeamResponse = ResearchOutputBaseResponse & {
 };
 
 export type ResearchOutputWorkingGroupResponse = ResearchOutputBaseResponse & {
-  workingGroups: [Pick<WorkingGroupResponse, 'id' | 'title'>];
```
peterstarling, line 264: what's the reason behind this change? I believe we wanted it to be a 'tuple' so essentially an array of one, cause it should only ever be one WG, no?

packages/model/src/research-output.ts
```diff
@@ -218,6 +218,7 @@ export type ResearchOutputDataObject = ResearchOutputCoreObject & {
   subtype?: string;
   teams: Pick<TeamResponse, 'id' | 'displayName'>[];
   workingGroups: Pick<WorkingGroupResponse, 'id' | 'title'>[];
+  published?: boolean;
```
peterstarling, line 221: why optional? The research output is either published or not, don't see how it could come back as undefined?

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -83,7 +83,7 @@ describe('ResearchOutputs data provider', () => {
       );
       const expectedResult = getResearchOutputDataObject();
 
-      expect(result).toEqual(expectedResult);
+      expect(result).toEqual({ ...expectedResult, published: true });
```
peterstarling, line 86: why is it not part of the fixture? You've added it to the DataObject so it's best to have it within the fixture? I guess it's because it was marked as undefined - well in my opinion it should not be undefined

packages/validation/src/permissions/research-output.ts
```diff
@@ -1,92 +1,50 @@
...
-  !!user.teams.find(
-    (team) =>
-      teams.map((t) => t.id).includes(team.id) &&
-      team.role === 'Project Manager',
-  );
+export const getUserPermissions = (
+  user: Omit<User, 'algoliaApiKey'> | UserResponse | null,
+  entity: 'teams' | 'workingGroups',
+  entityIds: string[],
+): UserPermissions => {
```
peterstarling, line 23: I see what you did here - you're trying to make this permission check more generic, but you're shifting the responsibility of saying what is allowed to the client of this function now. Imho we should keep those functions as they were, returning booleans, and if you need to check whether the user has permissions to save a draft - add another function. This should be composable anyway so it's not like it's going to be a huge repetition?

peterstarling (changes requested):

apps/crn-server/src/routes/research-outputs.route.ts
gabiayako: > Before it was: if i dont have permissions then throw

But before we didn't have to consider if the research output was published or not. So the permissions were only referring to user asap role/user team role

Not every condition in this Draft Feature depends on research output status (published/draft). For example, this Share Output Button does not depend on the research output status
Screenshot 2023-03-01 at 08 38 06

that's why I believe it's better to separate the permission referring to user asap role/user team role from the one of published or not

apps/crn-server/src/controllers/research-outputs.ts
gabiayako: fixed here [review: change published to create options](https://github.com/yldio/asap-hub/pull/2667/commits/b072177f672bd5f5e7ce1a2eeea1d168638ed9ce)

apps/crn-server/src/routes/research-outputs.route.ts
gabiayako: fixed here [review: change published to create options](https://github.com/yldio/asap-hub/pull/2667/commits/b072177f672bd5f5e7ce1a2eeea1d168638ed9ce)

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
gabiayako: fixed here [review: make published mandatory](https://github.com/yldio/asap-hub/pull/2667/commits/7ab15402dfc9a32c8bb9335ea9e6e2bf4110fd82)

apps/crn-server/test/controllers/research-outputs.test.ts
gabiayako: fixed here [review: change published to create options](https://github.com/yldio/asap-hub/pull/2667/commits/b072177f672bd5f5e7ce1a2eeea1d168638ed9ce)

apps/crn-frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -54,17 +54,24 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
   const team = useTeamById(teamId);
   const [errors, setErrors] = useState<ValidationErrorResponse['data']>([]);
 
-  const { canCreateUpdate } = useContext(ResearchOutputPermissionsContext);
```
gabiayako, line 57: Why I removed context here:

`TeamOutput` is called by `TeamProfile` and `ResearchOutput`

`TeamProfile` -> `TeamOutput` -> `ResearchOutputForm` (Buttons Save Draft / Publish)  -> this requires `published` to be calculated
`ResearchOutput` -> `TeamOutput` -> `ResearchOutputForm` (Buttons Save Draft / Publish)  -> this requires `published` to be calculated

But `TeamProfile` does not have `published` since it’s related to a team not a specific research output. So the value never got to the ResearchOutputForm when we went by the flow `TeamProfile` -> `TeamOutput` -> `ResearchOutputForm`. It worked with context before because we didn't have to consider `published`.

The same thing applies to WorkingGroupOutput

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -376,6 +377,7 @@ export interface ResearchOutputController {
   fetchById: (id: string) => Promise<ResearchOutputResponse>;
   create: (
     researchOutputRequest: ResearchOutputCreateData,
+    createOptions?: { publish: boolean },
```
peterstarling, line 380:
```suggestion
    createOptions?: { publish?: boolean },
```

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -45,25 +50,31 @@ export const researchOutputRouteFactory = (
...
     validateResearchOutputPostRequestParametersIdentifiers(createRequest);
 
-    if (
-      !loggedInUser ||
-      !hasCreateUpdateResearchOutputPermissions(
-        loggedInUser,
-        createRequest.teams,
-      )
-    ) {
+    const userRole = getUserRole(
```
peterstarling, line 57: any reason why we need to get the role first only to pass it on the `hasEditResearchOutputPermission`? Could it not accept the logged in user and teams? is it used elsewhere where these params are not available?

peterstarling (approved):

apps/crn-server/src/routes/research-outputs.route.ts
gabiayako: no, but in my opinion having `getUserRole` separated is more organized. Also here
 
Screenshot 2023-03-03 at 09 02 11

if we had `getUserRole` as part of each function, it would be called 3 times instead of 1. This wouldn't make a difference, but I don't think we gain a lot moving `getUserRole` inside each function as well ¯\_(ツ)_/¯

apps/crn-server/src/controllers/research-outputs.ts
gabiayako: Hey, I think we don't need this `?` because are always sending `publish` to the controller here

https://github.com/yldio/asap-hub/blob/CRN-792/apps/crn-server/src/routes/research-outputs.route.ts#L76
 right?

apps/crn-server/src/routes/research-outputs.route.ts
peterstarling: well i think its separation of concerns really. Once again having to determine the role by the client and then having to send it across to the permission service means shifting that responsibility onto the clients and they become aware of our role-based system. I realise some architectural patterns may seem less performant but this could haunt us in the future so I was trying to avoid it. But it's fine, if you feel it's better this way lets leave it for now

PR #2679, opened 2023-02-27
CRN-1316 update no events messages

lctrt:


lennym (commented):
It looks like this is hardcoding the content to "working group" where the ticket specifies different content for interest groups and team groups.

I'd guess there's some parameterisation required in place of "working".

Edit:

From looking at it, it seems that the `type` prop could be used directly with some small modification upstream to pass the expected values for that prop (which should be safe as it doesn't seem to be used anywhere else (or indeed anywhere _at all_ right now).

It might need a special case defining for users though as the ticket doesn't seem to specify a behaviour for those, so I'm _assuming_ that should continue to render the name?

lennym (approved):

PR #2702, opened 2023-03-01
adds gp2 event handler

markcsims:


packages/server-common/src/utils/sync-google-event.ts
```diff
@@ -16,21 +12,29 @@ export type SyncEvent = (
   googleCalendarId: string,
   squidexCalendarId: string,
   defaultTimezone: string,
-) => Promise<EventResponse>;
+) => Promise<EventResponse | gp2.EventResponse>;
```
jhonny-m, line 15: can we use the BasicEvent for this?

jhonny-m (approved):

PR #2704, opened 2023-03-01
gp2-270-frontend

jhonny-m:


packages/validation/src/user-guards.ts
```diff
@@ -3,3 +3,8 @@ import { UserResponse, ExternalAuthorResponse } from '@asap-hub/model';
 export const isInternalUser = (
   author: ExternalAuthorResponse | UserResponse,
 ): author is UserResponse => (author as UserResponse).email !== undefined;
+
+export const isExternalUser = (
+  author: ExternalAuthorResponse | UserResponse,
+): author is ExternalAuthorResponse =>
+  isInternalUser(author as ExternalAuthorResponse) === false;
```
markcsims, line 10: Why do we need this? Isn't it `!isInternalUser`

markcsims (approved):

packages/validation/src/user-guards.ts
jhonny-m: not quite, it's a typescript thing

jhonny-m: this validation is casting as ExternalUser not UserResponse what I need is the UserResponse + href

PR #2717, opened 2023-03-02
CRN-1334 add report type to research outputs

lctrt:


gabiayako (changes requested):
Hey, I've noticed that after creating a CRN Report, if you try to edit, the field `Type` appears as required and if you choose a type for it and save it, the research output which had type `Research` becomes an `Article`.


https://user-images.githubusercontent.com/16595804/223479224-073be222-40f0-4a8d-903c-c71e2d36a2be.mov

gabiayako (approved):
nice fix 🎉

PR #2718, opened 2023-03-02
adds calendars to the projects and wg schema

markcsims:


dianaguimaraes (approved):

apps/crn-server/src/controllers/calendars.ts
```diff
@@ -1,21 +1,16 @@
...
   ListCalendarResponse,
 } from '@asap-hub/model';
-import {
-  CalendarController,
-  CalendarDataProvider,
-} from '@asap-hub/server-common';
 
 export default class Calendars implements CalendarController {
-  dataProvider: CalendarDataProvider;
+  constructor(private dataProvider: CalendarDataProvider) {}
```
jhonny-m, line 12: does this do the same thing? never seen it before 😄

markcsims: It does the same thing, @peterstarling showed it to me.

jhonny-m (approved):

packages/model/src/data-providers.ts
```diff
@@ -0,0 +1,41 @@
...
+} from '.';
+
+export type FetchCalendarProviderOptions = {
+  maxExpiration?: number;
+  active?: boolean;
+  resourceId?: string;
+};
+
+export interface CalendarDataProvider<
+  CalendarDataObject = DefaultCalendarDataObject,
```
peterstarling, line 20: so the only thing I would maybe suggest is that here we should not only assign default but also extend the some BaseCalendarDataObject type. This will mean that our interface will then return or ask for an object that at minimum has the properties from the extended type - otherwise we could provide any value for the generic and implement this interface anywhere. That is btw also an option since all of our data-providers only have the 4 methods with similar arguments and returns but I don't think thats what we are aiming for here

peterstarling (changes requested):

packages/model/src/data-providers.ts
markcsims: I was wondering about this too. The data providers and the controllers have some standard methods on them.

I left it for a refactoring task later on.

markcsims: Shall we try pairing on it together to see how it looks?

PR #2719, opened 2023-03-02
Rewrite Tennant.yaml

kmaid:


markcsims (approved):

apps/auth0/tsconfig.json
```diff
@@ -5,6 +5,11 @@
     "rootDir": "src",
     "outDir": "build"
   },
+  "ts-node": {
```
kmaid, line 8: @peterstarling  should put this in the BE config to avoid doing the export before ts node calls

peterstarling: do we need it though? in same places we had it in-line i was able to drop it (or should drop) because we changed the type to module

apps/auth0/scripts/dependencies-sync.ts
```diff
@@ -0,0 +1,40 @@
...
+    const actions = data.actions?.map((action) => {
+      const dependencies = action.dependencies?.map(
+        (dependency: { name: keyof typeof packageJson.dependencies }) => {
+          const version = packageJson.dependencies[dependency.name];
+          return version
+            ? {
+                ...dependency,
+                version,
+              }
+            : dependency;
```
peterstarling, line 23: I believe this means that if the dependency isn't listed in package json we just leave the dependency without the version - does it work that way? Maybe we should throw if the dependency isn't listed?

apps/auth0/package.json
```diff
@@ -9,11 +9,13 @@
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
     "typecheck": "tsc",
-    "deploy": "a0deploy import --input_file ./tenant.yaml"
+    "deploy": "yarn sync:dependencies && yarn deploy:tennant",
```
peterstarling, line 12: i guess this makes sense for dev and prod deployment since these are the only places we run this anyway, but I thought maybe we want to build in PR stage? I've added a comment to the build script to throw if the deps aren't found, if you did that it would be worth running the build pre-master merge, otherwise this is good

peterstarling (approved):
just minor suggestions, if you want to ignore them feel free to merge as is

apps/auth0/scripts/dependencies-sync.ts
kmaid: Yeah, it gets latest

apps/auth0/tsconfig.json
kmaid: Better one place than repeating it in multiple places

apps/auth0/package.json
kmaid: If the dependencies aren't found then then they aren't in package.json and the unit tests will fail. I think the risk is much higher the other way round but you said not to overthink it

PR #2733, opened 2023-03-06
working groups create output page

jhonny-m:


packages/routing/src/gp2/projects.ts
```diff
@@ -1,15 +1,26 @@
...
 import resource from './resource';
+import { outputDocumentTypeParser } from './working-groups';
 
 const project = (() => {
   const edit = route('/edit', {}, { resource });
   const add = route('/add', {}, {});
   const overview = route('/overview', {}, {});
   const resources = route('/resources', {}, { add, edit });
+
+  const upcoming = route('/past', {}, {});
```
csbarbosa, line 11: Why do we have the upcoming and the past with the same route? Shouldn't be /upcoming?

packages/routing/src/gp2/working-groups.ts
```diff
@@ -1,15 +1,39 @@
...
+  serialize: (data: OutputDocumentTypeParameter): string => data,
+};
+
 const workingGroup = (() => {
   const edit = route('/edit', {}, { resource });
   const add = route('/add', {}, {});
   const overview = route('/overview', {}, {});
   const resources = route('/resources', {}, { add, edit });
+
+  const upcoming = route('/past', {}, {});
```
csbarbosa, line 24: Shouldn't be /upcoming in here?

packages/routing/src/gp2/projects.ts
jhonny-m: yeah, my mistake

packages/routing/src/gp2/working-groups.ts
jhonny-m: copy/paste :D

packages/gp2-components/src/molecules/__tests__/ShareOutputButton.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  it('renders the right links', () => {
+    render(<ShareOutputButton entityType={'workingGroup'} id={'123'} />);
+    userEvent.click(screen.getByRole('button', { name: /share an output/i }));
+    expect(
+      screen
+        .getAllByRole('link')
+        .map((element) => (element as HTMLAnchorElement).href),
+    ).toEqual([
+      expect.stringContaining('article'),
+      expect.stringContaining('code%2Fsoftware'),
```
csbarbosa, line 21: What does %2F means in this case? From what I found out, it means a forward slash, in case it's in a query string, but want to make sure if it's the case

csbarbosa (approved):

PR #2738, opened 2023-03-07
Add support for running serverless locally

lennym:
Adds offline support for serverless with eventbridge and ssm to enable local development of lambda handlers.

```
yarn start:backend:crn:sls
```

Note: also run `yarn:watch` to enable live rebuilding of code.

To enable inbound webhooks from Squidex you will need an ngrok tunnel (or similar) to port 3333, and configure Squidex to send webhook to the tunnel endpoint.

Some environment variables normally set in CI deployments are needed to be set in `.env` to run. See `.env.example` for details.


apps/crn-server/src/handlers/webhooks/webhook-squidex.ts
```diff
@@ -21,7 +21,16 @@ export const squidexWebhookFactory = (
   return lambda.http(squidexHandler);
 };
 
-const eventBridge = new EventBridge();
+const eventBridge = new EventBridge(
+  /* istanbul ignore next */
+  process.env.IS_OFFLINE
```
peterstarling, line 26: put in the config maybe? just so we don't have process.env all over the place 

not a huge fan of such conditionals just for local dev tbh but meh

peterstarling (approved):

apps/crn-server/src/handlers/webhooks/webhook-squidex.ts
lennym: I agree, I just wasn't sure how possible it was to emulate the pre-existing behaviour of sending literally nothing to `new EventBridge()` if defined in config, and I didn't want to risk breaking prod environments.

I guess I can try undefined parameters and see how it works though.

lennym: Now updated, and seems to be working correctly in the review environment as far as I can tell.

PR #2740, opened 2023-03-07
Inject secrets from CI

kmaid:
This PR is to use the Pipelines variable injection system. 

I have also fixed not being able to go through the invitation flow. I have verified that redirect_uri is present


apps/auth0/src/types.d.ts
```diff
@@ -24,8 +24,8 @@ export type DeepPartial<T> = {
 
 type Auth0Secrets = {
   secrets: {
-    PR_APP_DOMAIN: string; // hub.asap.science / gp2.asap.science
-    ASAP_API_URL: string; // https://api-dev.hub.asap.science / https://api-dev.gp2.asap.science
+    BASE_PR_APP_DOMAIN: string; // hub.asap.science / gp2.asap.science
+    API_URL: string; // https://api-dev.hub.asap.science / https://api-dev.gp2.asap.science
```
markcsims, line 28: why are these secrets?

kmaid: Because Auth0 doesn't have a concept of Environment Variables. I guess we could inline these variables when we compile but seems outside the scope of this ticket

apps/auth0/src/post-login-add-metadata.ts
```diff
@@ -69,29 +69,35 @@ const extractUser = (response: Auth0UserResponse): User | gp2Auth.User => {
   };
 };
 
-export const onExecutePostLogin = async (
-  event: Auth0PostLoginEventWithSecrets,
-  api: Auth0PostLoginApi,
-) => {
+const getApiUrl = (event: Auth0PostLoginEventWithSecrets) => {
```
peterstarling, line 72:
```suggestion
const getApiUrls = (event: Auth0PostLoginEventWithSecrets): [string, string] => {
```

apps/auth0/src/post-login-connect-user.ts
```diff
@@ -2,6 +2,27 @@ import { Auth0PostLoginApi } from '@vedicium/auth0-actions-sdk';
 import got from 'got';
 import type { Auth0PostLoginEventWithSecrets } from './types';
 
+const getApiUrl = (event: Auth0PostLoginEventWithSecrets) => {
```
peterstarling, line 5: any chance this could be made reusable across both actions? the input event is a little different but we only use the `event.request` which I think should be the same anyway?

peterstarling (changes requested):

apps/auth0/src/post-login-connect-user.ts
kmaid: No. Need rollup for that

peterstarling (approved):

PR #2742, opened 2023-03-07
Added new relatedResearch component

oratiu:
- [x] Add RelatedResearch in SharedResearchOutput

- [x] Add test in SharedResearchOutput

### DIFF
--- a/packages/react-components/src/atoms/Pill.tsx
+++ b/packages/react-components/src/atoms/Pill.tsx
@@ -30,11 +30,7 @@ export type AccentVariant =
   | 'neutral'
   | 'success'
   | 'warning'
-<<<<<<< HEAD
   | 'gray';
-=======
-  | 'relatedResearch';
->>>>>>> eee54ad71 (Added new relatedResearch component)
 
 export const accents: Record<AccentVariant, CSSObject> = {
   default: {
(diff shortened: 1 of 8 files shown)


packages/react-components/src/atoms/Pill.tsx
```diff
@@ -61,6 +62,14 @@ export const accents: Record<AccentVariant, CSSObject> = {
     color: colors.success500.rgb,
     borderColor: colors.success500.rgb,
   },
+  relatedResearch: {
```
jhonny-m, line 65: this doesn't look like an accent

packages/react-components/src/organisms/RelatedResearch.tsx
```diff
@@ -0,0 +1,202 @@
...
+import {
+  protocol,
+  article,
+  dataset,
+  bioinformatics,
+  labResource,
+  grantDocument,
+} from '../icons';
+
+export const getIconForDocumentType = (
```
jhonny-m, line 19: this should be a `Record` instead of a switch

packages/react-components/src/organisms/RelatedResearch.tsx
```diff
@@ -0,0 +1,202 @@
...
+    case 'Grant Document':
+      return grantDocument;
+    default:
+      return protocol;
+  }
+};
+
+const container = css({
+  display: 'grid',
+  paddingTop: `${32 / perRem}em`,
```
jhonny-m, line 42: you can set them all up with `padding`

lctrt (approved):

PR #2745, opened 2023-03-07
fix: Contentful Pages text field validation

gabiayako:
In the last squidex-contentful migration run (https://github.com/yldio/asap-hub/actions/runs/4354930475/jobs/7612848718#step:6:777)

ga-actions

The page with id `77ac91b8-6919-4157-b461-aa404e44a3c3` could not be published. Taking a look at Contentful

error

I could see that the text was there but there was a validation problem. Looking at the Page schema in `Production` I saw that embeds were not allowed, that was causing the validation problem.

embeds

So this PR adds a migration to add 

```
          'embedded-entry-block',
          'embedded-entry-inline',
          'embedded-asset-block',
```

to `enabledNodeTypes`.

### DIFF
--- a/packages/contentful/migrations/pages/20230307163242-text-enabled-node-types.js
+++ b/packages/contentful/migrations/pages/20230307163242-text-enabled-node-types.js
@@ -1,0 +1,189 @@
+module.exports.description = 'Change Pages text field validation';
+
+module.exports.up = (migration) => {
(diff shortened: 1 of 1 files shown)


peterstarling (approved):
approved. I've made the change manually in prod already but this needs to go in anyway

PR #2746, opened 2023-03-07
adding create outputs page to projects (#2739)

jhonny-m:


apps/gp2-frontend/src/outputs/state.ts
```diff
@@ -24,3 +24,11 @@ export const refreshOutputsState = atom<number>({
 
 export const useOutputs = (options: GetListOptions) =>
   useRecoilValue(outputsState(options));
+
+export const useCreateOutput = () => {
+  const authorization = useRecoilValue(authorizationState);
+  return async (payload: gp2.OutputPostRequest) => {
+    const output = await createOutput(payload, authorization);
+    return output;
+  };
```
markcsims, line 33:
```suggestion
  return (payload: gp2.OutputPostRequest) => {
    return createOutput(payload, authorization);
  };
```

packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [title, setTitle] = useState('');
+  const currentPayload = {
+    title,
+    documentType: documentTypeMapper[documentType],
+  };
+
+  return (
+    <Form onSave={() => createOutput(currentPayload)} dirty={title !== ''}>
+      {({ isSaving, onSave }) => (
+        <div>
```
markcsims, line 32: Do we need this div? <>?

markcsims (approved):

packages/gp2-components/src/templates/OutputForm.tsx
jhonny-m: currently I'm not doing any styling, but I will be needed in some time
but I can remove for now

PR #2747, opened 2023-03-07
fix: Empty links in Contentful News

gabiayako:
In the last squidex-contentful migration, the news with id `77d9e32d-ce79-492b-8a9e-7bf4e6b3fe0b` (https://github.com/yldio/asap-hub/actions/runs/4354930475/jobs/7612848718#step:6:660) was not published with an error of invalid url

Screenshot 2023-03-07 at 13 46 43 (2)

I saw in Squidex that if I type something in `link` and then erase it, `link` value becomes `""` and this empty string is not valid.

Screenshot 2023-03-07 at 13 46 43

So this PR makes link to be `null` when it is an empty string in squidex

### DIFF
--- a/packages/cms-data-sync/src/news/news.data-migration.ts
+++ b/packages/cms-data-sync/src/news/news.data-migration.ts
@@ -48,7 +48,7 @@ export const migrateNews = async () => {
       title: title!,
       shortText,
       frequency: frequency || 'News Articles',
-      link: link!,
+      link: link?.length ? link : null,
       linkText: linkText!,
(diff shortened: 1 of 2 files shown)


lennym:
It seems like the better fix here would be to modify [the validation regex](https://github.com/yldio/asap-hub/blob/master/packages/contentful/migrations/news/20221028130957-add-fields-to-news-content-type.js#L65) to accept an empty string.

The pattern you describe of someone removing a value from the text input field in the UI seems like something that should be able to happen without an error.

PR #2755, opened 2023-03-08
Crn 1320/integration test spike

lennym:


apps/crn-server/test/helpers/retryable.ts
```diff
@@ -0,0 +1,26 @@
+type Options = {
+  retries?: number;
+  delay?: number;
+};
+
+export const retryable = async (
```
peterstarling, line 6: maybe you want to use something from npm like async-retry? it's got exponential backoff, time limiting etc...and its from vercel

apps/crn-server/test/integration/reminders.integration-test.ts
```diff
@@ -201,38 +224,42 @@ describe('Reminders', () => {
...
-      const teamId = await teamDataProvider.create(teamCreateDataObject);
+      teamId = await teamDataProvider.create(teamCreateDataObject);
 
       const userCreateDataObject = getUserInput(teamId);
       userId = await userDataProvider.create(userCreateDataObject);
     });
 
-    afterAll(() => {
+    afterEach(async () => {
       jest.useRealTimers();
```
peterstarling, line 263: yeah i would probably extract the timers into a beforeAll and beforeEach

Also by running the setup each time for every test we are slowing it down quite significantly (I think you have already extended jest's timeout here to something really high). As far as I remember the team and user fixtures were reused between the tests - maybe we should consider tearing down just the data we know we need to purge after every tests?

Although I admit this would be dangerous, as we could accidentally run into the same problems we had before, plus there are some user and team fixtures down below that need to be torn down just as well, so yeah it's really just a hint in case it's got too slow that maybe we want to be more selective about which data we want to keep between the tests

peterstarling (changes requested):
this is really good stuff, just minor comments really

apps/crn-server/test/integration/reminders.integration-test.ts
lennym: I bumped the timer from 30s to 2 minutes. The actual test suite takes a little over 60 seconds to run now, but was occasionally timing out before when set at 30 seconds, so would need increasing anyway. 2 minutes seemed like a safe margin but without being excessively slow.

I agree it is sub-optimal, but given it's running in parallel with the jest unit test runners - all of which are currently taking a good time longer than this task still - and given the need for manual teardown otherwise, then I'd suggest leaving it as it is.

If you feel strongly about it I'm happy to revisit.

lennym: I will pull out the timers though.

lennym: Updated with the timers moved into `beforeAll`

I've run a few test runs locally and the typical execution time is 40-45 seconds, but the timeout should be a good deal longer to handle the possibility that tests may need to retry, hence two minutes.

apps/crn-server/test/helpers/retryable.ts
lennym: I don't have any particularly strong feeling either way on this one. There's arguments in favour of doing it either way.

I've swapped out for the package you suggested, with a small wrapper to override the default retries because 10 is way more than necessary for our needs I think.

peterstarling (approved):

PR #2756, opened 2023-03-08
fix: Save Draft type errors

gabiayako:
Making `onSaveDraft` optional as `packages/gp2-components/src/templates/OutputForm.tsx` does not require this props.

github error: https://github.com/yldio/asap-hub/actions/runs/4363587990/jobs/7629920136#step:6:109

### DIFF
--- a/packages/react-components/src/organisms/Form.tsx
+++ b/packages/react-components/src/organisms/Form.tsx
@@ -14,14 +14,14 @@ const styles = css({
 
 type FormProps<T> = {
   onSave: () => Promise<T | void>;
-  onSaveDraft: () => Promise<T | void>;
+  onSaveDraft?: () => Promise<T | void>;
   validate?: () => boolean;
   dirty: boolean; // mandatory so that it cannot be forgotten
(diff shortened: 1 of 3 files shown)


packages/react-components/src/organisms/Form.tsx
```diff
@@ -96,7 +96,9 @@ const Form = <T extends void | Record<string, unknown>>({
           onCancel,
           isSaving: status === 'isSaving',
           onSave: getWrappedOnSave(onSave),
-          onSaveDraft: getWrappedOnSave(onSaveDraft),
+          ...(onSaveDraft
```
jhonny-m, line 99: something we did for the modal, was instead of passing all the function we want wrapped, we just return the wrapper and we can wrap any function we want

jhonny-m (approved):
I'll approve to get master working, but we can have a chat about just returning the wrapper

packages/react-components/src/organisms/Form.tsx
gabiayako: nice, I'll open a new PR with that soon

PR #2759, opened 2023-03-08
CRN-1250 Create Teams content model and Contentful data provider

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-1250

Screenshot 2023-03-15 at 07 07 55
Screenshot 2023-03-15 at 07 08 23
Screenshot 2023-03-15 at 07 08 33
Screenshot 2023-03-15 at 07 08 46
Screenshot 2023-03-15 at 07 12 17


apps/crn-server/src/app.ts
```diff
@@ -254,9 +271,33 @@ export const appFactory = (libs: Libs = {}): Express => {
...
+    'IS_CONTENTFUL_ENABLED_V2',
+    false,
+  );
+
+  featureFlagDependencySwitch.setDependency(
+    'teams',
+    libs.teamContentfulDataProvider ||
+      new TeamContentfulDataProvider(
+        contentfulGraphQLClient,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
gabiayako, line 288: above I defined `contentfulRestClient` like this:

```
  let contentfulRestClient: Environment | undefined;

  (async () => {
    contentfulRestClient = await getContentfulRestClient({
      space: contentfulSpaceId,
      accessToken: contentfulManagementAccessToken,
      environment: contentfulEnvId,
    });
  })();
```

so here I was seeing an error that `contentfulRestClient` was being used before assignment. This non-null assertion avoids this error, I know it's not ideal but I could not find a better way to do that.

I haven't done this

```
const contentfulRestClient = await getContentfulRestClient({
    space: contentfulSpaceId,
    accessToken: contentfulManagementAccessToken,
    environment: contentfulEnvId,
  });
```

because then I would need to add `async` in `appFactory` and this seems a little weird to me. Also, this would make it necessary to call appFactory like this

```
  (async () => {
    app = await appFactory({
      ...
    });
  })();
```

where currently is called at the top level

peterstarling: I understand what you have done but it's a little risky.

I'm surprised this even works, the DI might actually be resolved before the var is assigned, I guess it's because when you pass this variable in the constructor it's still a reference, not a copy. 

I think we should avoid such risky patterns. Why don't you define this dependency as Promise<Environment> and then just make sure to await it before using inside `TeamContentfulDataProvider`? Yes I know this means having to await each time but I think that's still better.

Alternatively you could look into top level await - it should be supported in our node version, this should mean that we should be able to make the appFactory async and then await it rather than use the pattern you suggested.

@lennym @Riccardo-Brown thoughts?

peterstarling (changes requested):

apps/crn-server/setup-jest.ts
```diff
@@ -0,0 +1,13 @@
+import * as contentful from '@asap-hub/contentful';
+import { getContentfulEnvironmentMock } from './test/mocks/contentful-rest-client.mock';
+
+jest.mock('@asap-hub/contentful', () => {
```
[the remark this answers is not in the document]
gabiayako: For me it seems better to mock globally because the alternative is to mock in a **LOT** of different files

Most of the tests that failed here https://github.com/yldio/asap-hub/actions/runs/4405132410/jobs/7715680051 are due to this function not being mock 

This seems the way to override the mock :)
https://stackoverflow.com/questions/55900737/how-to-override-jest-mock-defined-in-a-setupfiles-jest-config-js

apps/crn-server/src/app.ts
lennym: I agree that passing the promise is probably the best approach here. All of the methods on the client are async already, so it's fairly straightforward to replace `const environment = this.restClient;` in the data provider with `const environment = await this.restClient;`.

packages/contentful/test/client.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      accessToken,
+    });
+
+    expect(
+      contentfulManagement.createClient({ accessToken }).getSpace,
+    ).toHaveBeenCalledWith(spaceId);
+
+    expect(
+      (
+        await contentfulManagement
```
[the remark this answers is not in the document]
peterstarling: that said - don't spend too much time on it. This is one of those tests you really just build for coverage, it doesn't really test much that hasn't already been tested by the proper use of typings, maybe just that you're passing the arguments correctly

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      const teamId = 'team-id-1';
+      const tool = {
+        name: 'Youtube Channel',
+        description: 'Youtube channel with team videos',
+        url: 'http://www.youtube.com/abcde',
+      };
+      const toolMock = getEntry({});
+
+      jest
+        .spyOn(await contentfulRestClientMock, 'createEntry')
```
peterstarling, line 425: why are you having to spy on a mock that is already injected into the data provider? I see you enjoy jest's mocking functions a lot :D but you could just assign the result of `getContentfulEnvironmentMock()` into a const and make assertions on that

peterstarling (changes requested):

apps/crn-server/setup-jest.ts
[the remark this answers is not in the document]
peterstarling: sorry but i have to disagree. I think mocking globally will introduce headaches. The reason we opt for very simple DI in BE is that we are able to track what is being called and where, at least in principle. You seem to follow that pattern and inject dependencies via constructor functions but then for some reason insist to just jest.mock and jest.spy which is not necessary with such simple DI. I still stand by what I said - i think this should go, even if you were going to just jest.mock and jest.spy in some places it should be in the file it is needed. 

Best if we ask the rest of team members @lennym @Riccardo-Brown

apps/crn-server/src/app.ts
```diff
@@ -155,6 +160,14 @@ export const appFactory = (libs: Libs = {}): Express => {
     accessToken: contentfulAccessToken,
     environment: contentfulEnvId,
   });
+
+  const getContentfulRestClientFunction = () =>
```
peterstarling, line 164:
```suggestion
  const getContentfulRestClientFactory = () =>
```

apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
```diff
@@ -0,0 +1,240 @@
...
+    id: item.sys.id ?? '',
+    displayName: item.displayName ?? '',
+    inactiveSince: item.inactiveSince ?? undefined,
+    projectTitle: item.projectTitle ?? '',
+    lastModifiedDate: new Date(item.sys.publishedAt).toISOString(),
+    expertiseAndResourceTags,
+    tools,
+    projectSummary: item.projectSummary ?? undefined,
+
+    // TODO implement below when users, labs and RO are ready
```
peterstarling, line 237: you might want to add the ticket numbers for reference

apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
```diff
@@ -0,0 +1,240 @@
...
+  NonNullable<FetchTeamsQuery['teamsCollection']>['items'][number]
+>;
+
+type TeamFilter = {
+  active?: boolean;
+};
+
+export type FetchTeamsOptions = FetchOptions<TeamFilter>;
+
+const getCleanTools = (tools: TeamTool[]) =>
```
peterstarling, line 35: this might sound like a silly requirement but maybe you can move all the helper functions below the class implementation? I think only the types need to be at the top

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,583 @@
...
+import { getContentfulEnvironmentMock } from '../../mocks/contentful-rest-client.mock';
+import {
+  TeamContentfulDataProvider,
+  teamUnreadyResponse,
+} from '../../../src/data-providers/contentful/teams.data-provider';
+import { getEntry } from '../../fixtures/contentful.fixtures';
+
+describe('Teams data provider', () => {
+  const contentfulGraphqlClientMock = getContentfulGraphqlClientMock();
+  const environment = getContentfulEnvironmentMock();
```
peterstarling, line 23: I'd do
```suggestion
  const environmentMock = getContentfulEnvironmentMock();
```
only because later you will be adding expectations on its methods. It's therefore important to make it obvious that this is a mock so that you don't need to use spies (which you actually did use - see my comment down below)

peterstarling (changes requested):
I'd approve if it wasn't for jest.spy which I think you should get rid of.

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,589 @@
...
+  });
+
+  const getContentfulTeamDataObject = () => ({
+    ...getTeamDataObject(),
+    ...teamUnreadyResponse,
+  });
+
+  describe('Fetch method', () => {
+    test('Should fetch the list of teams from Contentful GraphQl', async () => {
+      const result = await teamDataProviderMock.fetch({});
```
Riccardo-Brown, line 53: Why are we testing a mock
```suggestion
      const result = await teamDataProvider.fetch({});
```

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,589 @@
...
+
+      const result = await teamDataProvider.fetch({});
+
+      expect(result).toEqual({
+        items: [],
+        total: 0,
+      });
+    });
+
+    test('Should return teams', async () => {
```
Riccardo-Brown, line 101: How is this test different to the first test `Should fetch the list of teams from Contentful GraphQl`?

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,589 @@
...
+      );
+      const result = await teamDataProvider.fetch({ take: 8, skip: 5 });
+
+      expect(result).toEqual({
+        total: 1,
+        items: [getContentfulTeamDataObject()],
+      });
+    });
+
+    test('Should query data properly when request does not have options', async () => {
```
Riccardo-Brown, line 113:
```suggestion
    test('Should use default query params when request does not have any', async () => {
```

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,589 @@
...
+          limit: 8,
+          order: ['displayName_ASC'],
+          skip: 0,
+          where: {},
+        }),
+      );
+    });
+
+    describe('Active Filter', () => {
+      test('Should query data properly when active filter is true', async () => {
```
Riccardo-Brown, line 139:
```suggestion
      test('Should return active teams only when active filter is true', async () => {
```

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -0,0 +1,589 @@
...
+          expect.objectContaining({
+            limit: 8,
+            order: ['displayName_ASC'],
+            skip: 5,
+            where: { inactiveSince_exists: false },
+          }),
+        );
+      });
+
+      test('Should query data properly when active filter is false', async () => {
```
Riccardo-Brown, line 168:
```suggestion
      test('Should return inactive teams only when active filter is false', async () => {
```

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
gabiayako: The `teamDataProvider` and `teamDataProviderMock` are using different mocks as graphql client one of them uses https://github.com/yldio/asap-hub/blob/CRN-1250-create-teams-content-model-in-contentful/packages/contentful/src/mocks/graphql-client.mock.ts#L7 and the other https://github.com/yldio/asap-hub/blob/CRN-1250-create-teams-content-model-in-contentful/apps/crn-server/test/mocks/contentful-graphql-client.mock.ts#L3

gabiayako: to be more clear, one of them is testing the graphql query and the other is not 😉

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
gabiayako: related to this thread https://github.com/yldio/asap-hub/pull/2759#discussion_r1138422905

apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
```diff
@@ -0,0 +1,241 @@
...
+    await teamUpdated.publish();
+  }
+
+  async create(input: TeamCreateDataObject): Promise<string> {
+    const environment = await this.getRestClient();
+
+    const { tools, ...plainInput } = input;
+    let toolsInput = {};
+
+    if (tools) {
```
Riccardo-Brown, line 138: Is it possible to create the team first then add tools to them. As currently if we create some tools and any fail except the first one, we would have tools in the CMS that don't belong to a team. Creating the team first would mean at least successfully tool creations would be linked to the team.

apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
```diff
@@ -0,0 +1,241 @@
...
+      (acc, [key, value]) =>
+        value?.trim && value?.trim() === ''
+          ? acc // deleted field
+          : { ...acc, [key]: value },
+      {} as TeamTool,
+    ),
+  );
+
+const createAndPublishTools = (environment: Environment, tools: TeamTool[]) =>
+  Promise.all(
```
Riccardo-Brown, line 234: if my suggestion https://github.com/yldio/asap-hub/pull/2759/files#r1138461381 is implemented we would want to use  Promise.allSettled here.

apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
gabiayako: > Creating the team first would mean at least successfully tool creations would be linked to the team.
I didn't understand. Could you explain it again?

Riccardo-Brown: Let me try explaining with an example

Say we are creating a team with two tools (A,B) and creating tool B fails the whole createTeam request will fail. So if we retry creating the team and it succeeds the CMS would have the tools (A,A,B) where one A doesn't belong to a team.

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
gabiayako: It's testing that given a valid Contentful response provided by `contentfulGraphqlClientMock`, the data provider returns the expected response. But ok, I admit it's redundant. I wouldn't mind having them both. But I can remove this one if you prefer.

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
peterstarling: actually Gabi - Riccardo's suggestion is valid, it's just that this is NOT a mock and I think you got the naming wrong. As you said in another comment in here we have two  TeamContentDataProvider instances:
1. First one is using a graphql simple jest mock
2. Second one is using a mocked graphql server instead which enables us to pass through actual queries and get mocked data back, with graphql query validation

In Squidex tests we used different naming conventions to differentiate them, see:
https://github.com/yldio/asap-hub/blob/260775c669f63fc7b6d0cffbc15c171042e944f6/apps/crn-server/test/data-providers/teams.data-provider.test.ts#L26-L32

Perhaps you might want to rename `teamDataProviderMock` to `teamDataProviderMockGraphql` which indicates that the mocked part is the graphql server?

peterstarling (approved):
Although I've approved I actually think you should make the change to avoid the confusion

Riccardo-Brown (approved):

PR #2763, opened 2023-03-08
adding project/working group association to create output

jhonny-m:


apps/gp2-frontend/src/working-groups/__tests__/CreateWorkingGroupOutput.test.tsx
```diff
@@ -74,6 +74,8 @@ it('publishes the output', async () => {
     {
       title,
       documentType: 'Form',
+      workingGroups: expect.arrayContaining(['1']),
```
markcsims, line 77: the number 1 was a bit confusing. Maybe is should be working-group-id-1 or something. Same with projects.

PR #2770, opened 2023-03-09
feat: CRN-1240 Add feature flag cookie switch

peterstarling:
FE changes:
* add a new feature flag - CONTENTFUL - to determine which data source we want to use for selected API calls
* pass on the x-contentful-enabled header to the BE in the user endpoint (will need to be added to all API calls - would suggest an FE refactor to add an API abstraction)
* switch between indexes based on the presence of the feature flag

BE changes:
* add a feature flag proxy and middleware to support switching between two instantiated data-providers at runtime
* add a new feature flag IS_CONTENTFUL_ENABLED_V2 for all new data providers migrated to contentful (will need to be renamed once IS_CONTENTFUL_ENABLED is no longer in use for news,pages and dashboard)

### DIFF
--- a/packages/flags/src/index.ts
+++ b/packages/flags/src/index.ts
@@ -9,6 +9,7 @@ let overrides: Flags = {
   // flags already live in prod:
   // can also be used to manually disable a flag in development:
   ASAP_DRAFT_RESEARCH_OUTPUT: false,
+  CONTENTFUL: false,
 };
 
 const envDefaults: Record<string, boolean> = {


apps/crn-server/test/utils/feature-flag.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    featureFlagDependencySwitch.setFeatureFlag(
+      'IS_CONTENTFUL_ENABLED_V2',
+      true,
+    );
+    expect(() => dependency.fetch({})).toThrowError();
+  });
+
+  test('Should bind a depedency, call its method, do a feature flag switch and call the same method again to confirm it was called on both bound dependencies', async () => {
```
gabiayako, line 53:
```suggestion
  test('Should bind a dependency, call its method, do a feature flag switch and call the same method again to confirm it was called on both bound dependencies', async () => {
```

apps/crn-server/test/utils/feature-flag.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    featureFlagDependencySwitch.setFeatureFlag(
+      'IS_CONTENTFUL_ENABLED_V2',
+      true,
+    );
+
+    // Assert that the method was called on the other dependency after the switch
+    expect(userDataProviderMockTrue.fetch).not.toHaveBeenCalled();
+    await userDataProvider.fetch({});
+    expect(userDataProviderMockTrue.fetch).toHaveBeenCalled();
+  });
```
gabiayako, line 105: nice 🎉

apps/crn-server/test/app.test.ts
```diff
@@ -200,4 +202,105 @@ describe('Contentful feature flag', () => {
...
+        const response = await supertest(app)
+          .get('/users')
+          .set('X-Contentful-Enabled', 'false');
+
+        expect(response.status).toBe(200);
+        expect(userSquidexDataProviderMock.fetch).toHaveBeenCalledTimes(1);
+        expect(userContentfulDataProviderMock.fetch).not.toHaveBeenCalled();
+      });
+
+      test('user controller uses squidex data provider when the feature flag is set to false', async () => {
```
gabiayako, line 242:
```suggestion
      test('user controller uses contentful data provider when the feature flag is set to true', async () => {
```

gabiayako (approved):

PR #2772, opened 2023-03-09
Gp2 295 type subtype

jhonny-m:


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -4,33 +4,105 @@ import {
...
   const [title, setTitle] = useState('');
+  const [link, setLink] = useState('');
+  const [type, setType] = useState('');
+  const [subtype, setSubtype] = useState('');
   const currentPayload = {
     title,
     documentType: documentTypeMapper[documentType],
+    link,
+    type: type ? (type as gp2Model.OutputType) : undefined,
+    subtype: subtype ? (subtype as gp2Model.OutputSubtype) : undefined,
```
markcsims, line 56: You should be able to remove this casting by using generics on the useState

jhonny-m: using generic it doesn't allow for the empty string

markcsims (approved):

PR #2778, opened 2023-03-09
feat: add returning drafts for research outputs

Riccardo-Brown:


apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -173,6 +176,56 @@ export class ResearchOutputSquidexDataProvider
     };
   }
 
+  async fetchDrafts(options: {
```
peterstarling, line 179: I think I mentioned that in my previous review but we aim to only use `create` `update`, `fetch` and `fetchById` in our providers. To be fair I don't think you need a separate method either here or in your controller but since it's quite a common practice in this repo to add such methods to controllers, at least make sure to reduce it in the data-provider. 

It's come up recently elsewhere, maybe we should create a generic `DataProvider` interface to enforce this?

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -80,6 +80,30 @@ export default class ResearchOutputs implements ResearchOutputController {
     };
   }
 
+  async fetchDrafts(
```
peterstarling, line 83: looks like a copy of the above method, do you really need a whole new method for this?

apps/crn-server/src/routes/working-groups.route.ts
```diff
@@ -1,4 +1,5 @@
 import { validateFetchOptions } from '@asap-hub/server-common';
+import Boom from '@hapi/boom';
```
peterstarling, line 2: what do you need this for?

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -11,27 +12,59 @@ import {
...
     '/research-outputs',
     async (req, res: Response<ListResearchOutputResponse>) => {
       const { query } = req;
-
-      const options = validateFetchOptions(query);
-
-      const result = await researchOutputController.fetch(options);
+      let result;
+
+      if (query.status) {
```
peterstarling, line 39: so if no status is sent we will just return zero results? not sure if i am reading it right since there are no tests

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -11,27 +12,59 @@ import {
...
 
   researchOutputRoutes.get(
     '/research-outputs',
     async (req, res: Response<ListResearchOutputResponse>) => {
       const { query } = req;
-
-      const options = validateFetchOptions(query);
-
-      const result = await researchOutputController.fetch(options);
+      let result;
```
peterstarling, line 37: I'd probably extract the below into a separate function/method and use a const, unless you were going to do this refactor once the tests are in?

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -18,6 +19,32 @@ type ResearchOutputParameters = {
   researchOutputId: string;
 };
 
+type ResearchOutputDraftParameters = FetchPaginationOptions & {
+  status: string;
```
peterstarling, line 23: what is entityId and also why does it only apply to drafts? I take it you want to say that only should they be looking for drafts do we want to allow searching by some entityId? 

I would just use the basic research output query string validation and add the status property (and maybe the entityId - whatever that is - if it's a working-group ID then I guess you should name it that)

apps/crn-server/src/routes/working-groups.route.ts
Riccardo-Brown: just leftover from when i had the endpoint here, will remove

apps/crn-server/src/validation/research-output.validation.ts
Riccardo-Brown: entityId is either working group or team id. It only applies to drafts as we want only members of a team or working group to view the drafts.

apps/crn-server/src/controllers/research-outputs.ts
Riccardo-Brown: removed function

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
Riccardo-Brown: Will address this issue in a separate PR

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -27,11 +31,31 @@ export const researchOutputRouteFactory = (
   researchOutputRoutes.get(
     '/research-outputs',
     async (req, res: Response<ListResearchOutputResponse>) => {
-      const { query } = req;
+      const { query, loggedInUser } = req;
+
+      const { associationId, status, ...options } = query.status
+        ? validateResearchOutputDraftFetchOptions(query)
```
peterstarling, line 37: ok so i think you want to have a single query string validation per endpoint, doing this conditionally seems very strange. If you know that some properties only refer to a situation when the drafts are being fetched - you can still apply that in ajv

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -27,11 +31,31 @@ export const researchOutputRouteFactory = (
...
+      const hasTeamRole =
+        getUserRole(loggedInUser as UserResponse, 'teams', [associationId]) !==
+        'None';
+      const hasWorkingGroupRole =
+        getUserRole(loggedInUser as UserResponse, 'workingGroups', [
+          associationId,
+        ]) !== 'None';
 
-      const options = validateFetchOptions(query);
+      if (status && !hasTeamRole && !hasWorkingGroupRole) {
```
peterstarling, line 48: what is status?

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -27,11 +31,31 @@ export const researchOutputRouteFactory = (
...
+        throw Boom.forbidden();
+      }
 
-      const result = await researchOutputController.fetch(options);
+      const result = await researchOutputController.fetch({
+        ...options,
+        ...(status && {
+          includeDrafts: true,
+          filter: `status eq 'Draft' and (data/workingGroups/iv in ['${associationId}'] or data/teams/iv in ['${associationId}'])`,
```
peterstarling, line 56: leaking CMS filtering syntax outside of data-provider is a no-go

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -95,7 +95,7 @@ export class ResearchOutputSquidexDataProvider
     take?: number;
     skip?: number;
     search?: string;
-    filter?: ResearchOutputFilter;
+    filter?: ResearchOutputFilter | string;
```
peterstarling, line 98: as states previously you were trying to make your life easier by exposing filtering outside of data-provider. How would it work with Contentful? you definitely don't want to do that

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -453,3 +445,11 @@ type ResearchOutputParsedTags = {
   environments: string[];
   subtype?: string;
 };
+
+export type ResearchOutputFetchOptions = {
+  take?: number;
+  skip?: number;
+  search?: string;
+  filter?: ResearchOutputFilter;
```
peterstarling, line 453: since we already have `includeDrafts` isn't your task simply to make sure ONLY drafts are coming through? I guess you could expand this filter:
https://github.com/yldio/asap-hub/blob/23c7ae9481a5a043f564226be09b1e11da4591ef/apps/crn-server/src/controllers/research-outputs.ts#L427-L434
and add a status option there (after documentType, title, link)

apps/crn-server/src/validation/research-output.validation.ts
peterstarling: pretty sure this validation logic is used for just that endpoint and i actually think you should just expand the existing validation to add support for filtering by status and by a workingGroupId and by a teamId. I think we want to be explicit so I would add all three filters.

In fact I wouldnt bother tying it to just drafts, i would just expose such filter for the entire search. I think implementation-wise this will actually make it a lot easier.

peterstarling (changes requested):

apps/crn-server/src/routes/research-outputs.route.ts
Riccardo-Brown: status is a query param, as we only want to do association check for drafts, i make it clear with the following 
```suggestion
      if (status === 'draft' && !hasTeamRole && !hasWorkingGroupRole) {
```

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -27,11 +30,35 @@ export const researchOutputRouteFactory = (
...
+        ]) !== 'None';
+
+      if (isRequestingDrafts && !hasTeamRole && !hasWorkingGroupRole) {
+        throw Boom.forbidden();
+      }
 
-      const result = await researchOutputController.fetch(options);
+      const result = await researchOutputController.fetch({
+        ...options,
+        ...(isRequestingDrafts && {
```
peterstarling, line 54: i dont know why you keep insisting on handling the case individually when drafts are requested and then use a different set of filters? 
Previously we had the `includeDrafts` option right? That option allowed us to ALSO return drafts, yeah? So that should stay.

and now we are introducing a status filter - where we want to say 'give me either published or unpublished outputs', correct? To me you could just leave them separate and if they do `includeDrafts: false` and `status: Draft` (or whatever the value is for the status) then return an empty list. 

As for the `associationId` - sorry but i still don't fully understand - in FE do we know what is being provided, so `workingGroupId` or `teamId`? I'm pretty sure we do in which case you should have a filter for those two and not a generic `associationId`, especially if we introduce more filters like this.

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -18,6 +19,49 @@ type ResearchOutputParameters = {
...
+      search: { type: 'string', nullable: true },
+      filter: {
+        type: 'array',
+        items: { type: 'string' },
+        nullable: true,
+      },
+      status: { type: 'string', const: 'draft' },
+      associationId: { type: 'string' },
+    },
+    if: {
```
peterstarling, line 42: although I told you that you COULD do these conditionally I actually stand by what I said earlier which is that these should all be completely independent filters. I don't see why we can't just expose a teamId or workingGroupId filters. Similarily why can't we not just have a `status` filter?

peterstarling (changes requested):

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -86,6 +86,79 @@ describe('/research-outputs/ route', () => {
...
+
+        expect(response.status).toBe(403);
+      });
+
+      test("Should return 403 if user is not associated to output's working group", async () => {
+        const response = await supertest(app).get('/research-outputs').query({
+          take: 15,
+          skip: 5,
+          status: 'draft',
+          workingGroupId: 'workingGroup-id-403',
```
peterstarling, line 156: to be fair both this and the test above should also contain a fixture for the loggedInUser - in this file we don't actually have a mock function for the currently logged user but there is a way. We can keep it as is for now but I know @oratiu is working on a similar piece of work where he is adding such a fixture, might be a good time to then add them here too

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -150,7 +154,9 @@ export const getSquidexGraphqlResearchOutput = (): NonNullable<
   },
 });
 
-export const getResearchOutputDataObject = (): ResearchOutputDataObject => ({
+export const getResearchOutputDataObject = ({
+  published = true,
```
peterstarling, line 158: we don't generally use the pattern of overrides as parameters for fixtures in BE - we decided it's easier to mutate the fixture in the test. I would just leave the function as it was, have the `published` property set to `true` but in your test just do 
```
const researchOutputDataObject = getResearchOutputDataObject();
researchOutputDataObject.published = false
```
I think in your test you actually need a list so you would have to reference an element in the list but that should be fine too

peterstarling (commented):
I've left you comments but you still need to add data-provider tests, if you had them I would have approved

apps/crn-server/test/fixtures/research-output.fixtures.ts
Riccardo-Brown: ok will revert this change and mutate in the test

apps/crn-server/test/routes/research-outputs.route.test.ts
Riccardo-Brown: do we have a JIRA ticket about this so i can add this file to?

peterstarling (approved):

PR #2785, opened 2023-03-13
CRN-946 - Refactor data fetching for external authors into data provider

lennym:
Move the logic in the `fetch` and `fetchById` functions from the external authors controller and into the data provider.

Note: the "new" tests for the data provider methods are a direct copy of the old controller tests.


apps/crn-server/src/data-providers/external-authors.data-provider.ts
```diff
@@ -19,4 +43,48 @@ export class ExternalAuthorSquidexDataProvider
...
+  }
+
+  async fetchById(id: string): Promise<ExternalAuthorResponse> {
+    const { findExternalAuthorsContent } =
+      await this.squidexGraphqlClient.request<
+        FetchExternalAuthorQuery,
+        FetchExternalAuthorQueryVariables
+      >(FETCH_EXTERNAL_AUTHOR, { id });
+    if (!findExternalAuthorsContent) {
+      throw Boom.notFound();
```
peterstarling, line 85: I think we are trying to move away from calling Boom anywhere outside of routes and controllers - this is because Boom is quite heavily REST focused. There should be a generic NotFound error you should be able to use here instead

package.json
```diff
@@ -43,7 +43,7 @@
     "start:backend:gp2": "yarn workspace @asap-hub/gp2-server start",
     "start:frontend:gp2": "yarn workspace @asap-hub/gp2-frontend start",
     "start:storybook": "yarn workspace @asap-hub/storybook start",
-    "test": "LANG=en_US TZ=UTC jest",
+    "test": "LANG=en_US TZ=UTC LOG_LEVEL=silent jest",
```
peterstarling, line 46: if you do that it will override the LOG_LEVEL setting from the .env file which I think is often helpful to have. Unless the .env config takes precedence?

peterstarling (changes requested):

package.json
lennym: It will override `.env`, and deliberately so. When I'm _running_ the app I want super-verbose runtime logs so I have `LOG_LEVEL=debug` set in my `.env`, but I don't want any of that in a test run. Switching the value in `.env` in and out depending on context isn't really a good option.

apps/crn-server/src/data-providers/external-authors.data-provider.ts
lennym: I think it actually looks like there's a more prevailing pattern of returning null from `fetchById` in the case of the entity not found. Will converge towards that.

package.json
kmaid: I agree. This has been annoying me for awhile

peterstarling: except i don't know how easy it is to then override if i wanted to debug why my test is failing?

peterstarling (approved):

package.json
Riccardo-Brown: i dont think we should be overriding it npm scripts as i expect env files to be source of truth for env var.  Doesn't running the following solve your issue without modifying the npm scripts
```
LOG_LEVEL=silent yarn test
```

PR #2788, opened 2023-03-13
Gp2 216 - Projects Empty State

csbarbosa:
(Ticket: https://asaphub.atlassian.net/browse/GP2-216)
Add empty state when there aren't any projects available

### DIFF
--- a/packages/gp2-components/src/molecules/EmptyState.tsx
+++ b/packages/gp2-components/src/molecules/EmptyState.tsx
@@ -1,13 +1,8 @@
-import {
-  contentSidePaddingWithNavigation,
-  Display,
-  Paragraph,
-  pixels,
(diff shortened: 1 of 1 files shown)


packages/gp2-components/src/molecules/EmptyState.tsx
```diff
@@ -0,0 +1,49 @@
...
+    36,
+    largeDesktopScreen,
+    72,
+    'px',
+  )} ${contentSidePaddingWithNavigation()}`,
+
+  display: 'grid',
+  textAlign: 'center',
+
+  [`@media (min-width: ${mobileScreen.width + 1}px)`]: {
```
jhonny-m, line 30: this seems a bit of a weird query, what is it for?

PR #2802, opened 2023-03-15
feat: CRN-1275 Setup Contentful Webhook for Teams

peterstarling:
## Notable changes
* Added a check for the Contentful webhook setup file, when changes are detected a new dedicated environment is created (much like when outstanding migrations are present)
* Adding new source to event-bus in case we want to split the subscribing handlers
* Adding webhook for Contentful events
* Improvement on typings
* Adding support for base64 encoded payload body
* Redefine typings for webhooks
* Add Contentful subscriber for external-authors indexing
* Create dependency resolvers for sorting the DI in more complex cases to avoid repetition


apps/auth0/tsconfig.json
```diff
@@ -5,11 +5,6 @@
     "rootDir": "src",
     "outDir": "build"
   },
-  "ts-node": {
```
peterstarling, line 8: moving to root

apps/crn-server/serverless.ts
```diff
@@ -541,13 +547,13 @@ const serverlessConfig: AWS = {
...
             pattern: {
-              source: [eventBusSource],
+              source: [eventBusSourceSquidex],
               'detail-type': [
                 'TeamsPublished',
                 'TeamsUpdated',
                 'TeamsUnpublished',
                 'TeamsDeleted',
-              ],
+              ] satisfies WebhookDetailType[],
```
lennym, line 556: Is there a reason this line has been added only here and not to the equivalent place on each handler?

lennym:
I'm curious about the hard-coding of `source: [eventBusSourceSquidex],` in the handler configs in serverless.ts.

It seems that this effectively disables being able to handle contentful sourced events at all, which defeats the purpose of having the indexers be able to use either based on feature flag.

Should the event source in config instead be dynamic based on feature flag, or have some logic in the handler to determine whether it can handle a particular event or not?

peterstarling:
> I'm curious about the hard-coding of `source: [eventBusSourceSquidex],` in the handler configs in serverless.ts.
> 
> It seems that this effectively disables being able to handle contentful sourced events at all, which defeats the purpose of having the indexers be able to use either based on feature flag.
> 
> Should the event source in config instead be dynamic based on feature flag, or have some logic in the handler to determine whether it can handle a particular event or not?

I take it this was before we had our huddle yesterday? should I ignore this comment then?

lennym:
Yes it was. Feel free to disregard.

.github/actions/contentful-checks/action.yml
```diff
@@ -42,14 +42,23 @@ runs:
         CONTENTFUL_SPACE_ID: ${{ inputs.contentful-space-id }}
         CONTENTFUL_ENV_ID: ${{ inputs.contentful-master-env }}
         CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ inputs.contentful-token }}
+      # Determine whether the Contentful webhook setup has changed since master
+    - uses: technote-space/get-diff-action@v6
+      with:
+        PATTERNS: |
+          +src/**/*.ts
```
gabiayako, line 49: if we add this pattern isn't going to select all files that match this pattern?

looking at the repo (https://github.com/technote-space/get-diff-action#usage) it seems it would
Screenshot 2023-03-30 at 08 25 58

and we want to see the changes only in `setup-webhook.ts` right?

apps/crn-server/src/handlers/webhooks/webhook-contentful.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  if (request.headers.authorization !== webhookAuthenticationToken) {
+    return false;
+  }
+
+  return true;
+};
+
+const getDetailTypeFromRequest = (
+  request: lambda.Request<ContentfulWebhookPayload<'teams' | 'news'>>,
```
lennym, line 75: Does `teams | news` need to be expanded to include other things? Or possibly just removed completely? I'm not sure I see the value in defining the type as a generic here.

packages/contentful/src/crn/types/webhook.ts
```diff
@@ -0,0 +1,88 @@
...
+    revision: number;
+    createdAt: string;
+    updatedAt: string;
+    deletedAt: string;
+  };
+}
+
+export type ContentfulWebhookPayload<T extends 'teams' | 'news'> =
+  | ContentfulWebhookPublishPayload<T>
+  | ContentfulWebhookUnpublishPayload<T>;
```
lennym, line 88: It seems like there's possible value in pulling `'teams' | 'news'` out into a `type ContentfulWebhookContentType` since it's a list we know we're going to want to add to over time, and so being able to do so once in a single place might be worthwhile.

packages/model/src/webhook.ts
```diff
@@ -0,0 +1,24 @@
+export type WebhookDetailType =
+  | 'ExternalAuthorsPublished'
+  | 'LabsCreated'
+  | 'LabsDeleted'
+  | 'LabsPublished'
+  | 'LabsUnpublished'
+  | 'LabsUpdated'
+  | 'LabsUpdated'
+  | 'NewsPublished'
```
gabiayako, line 9: Don't we need the rest (`NewsUnublished`, `NewsDeleted`, etc)?

apps/crn-server/src/handlers/webhooks/webhook-contentful.ts
```diff
@@ -0,0 +1,86 @@
...
+): WebhookDetailType => {
+  const actions = request.headers['x-contentful-topic']?.split('.') ?? [];
+  const action = actions[actions.length - 1] as 'publish' | 'unpublish';
```
gabiayako, line 78: Do we want to handle only `publish` and `unpublish`?

I saw here https://www.contentful.com/developers/docs/references/content-management-api/#/reference/webhooks that this `x-contentful-topic` can have these other values
Screenshot 2023-03-30 at 14 39 46

lennym: I think because contentful allows creating/editing _without_ publishing then these are probably the only events we'll need. To archive or delete you need to have previously unpublished.

lennym: One possible concern that might be worth checking though is that if we unpublish some content, the management API will (potentially) still return it, so the code that deletes from algolia when a 404 is returned from the `fetchById` might need to be double checked that it works correctly to delete when content is unpublished.

peterstarling: i should have probably added it to this PR's description - for our purposes we are only interested in publish and `unpublish` events. Any delete/archive action is, like @lennym said, preceded by an `unpublish` event

apps/crn-server/src/dependencies/external-authors.dependencies.ts
```diff
@@ -0,0 +1,64 @@
...
+export const getExternalAuthorDataProvider = (): ExternalAuthorDataProvider => {
+  if (isContentfulEnabledV2) {
+    const contentfulGraphQLClient = getContentfulGraphQLClient({
+      space: contentfulSpaceId,
+      accessToken: contentfulAccessToken,
+      environment: contentfulEnvId,
+    });
+
+    const getContentfulRestClientFactory = () =>
+      getContentfulRestClient({
```
lennym, line 35: It feels like there's a case here for allowing data providers to be read-only and so not provide a rest client at all. The reason there's no coverage here is because this argument is never used in this context, so it's possibly better to not provide it at all.

.github/actions/contentful-checks/action.yml
lennym: Also are these paths relative to the project root, or some other subdirectory? It looks like they're intended to be relative to the `./packages/contentful/` directory, but I can't see what code there is that causes that to be the case.

apps/crn-server/test/dependencies/external-authors.dependencies.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    );
+    const externalAuthorDataProvider =
+      getExternalAuthorDataProviderModule.getExternalAuthorDataProvider();
+
+    expect(externalAuthorDataProvider).toBeInstanceOf(
+      ExternalAuthorSquidexDataProvider,
+    );
+  });
+
+  it('Should resolve External-Author Squidex Data Provider when the Contentful feature flag is off', async () => {
```
gabiayako, line 23:
```suggestion
  it('Should resolve External-Author Squidex Data Provider when the Contentful feature flag is on', async () => {
```

.github/actions/contentful-checks/action.yml
gabiayako, line 49: Reminder to remove the Patterns line

gabiayako (changes requested):

gabiayako (approved):
Nice, I've tested creating an external author in crn-2802 env and it appeared in contentful's algolia 2802 index 🎉

.github/actions/contentful-checks/action.yml
```diff
@@ -42,14 +42,23 @@ runs:
         CONTENTFUL_SPACE_ID: ${{ inputs.contentful-space-id }}
         CONTENTFUL_ENV_ID: ${{ inputs.contentful-master-env }}
         CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ inputs.contentful-token }}
+      # Determine whether the Contentful webhook setup has changed since master
+    - uses: technote-space/get-diff-action@v6
+      with:
+        PATTERNS: |
+          +packages/contentful/scripts/setup-webhook.ts
```
lennym, line 49: Is there a case for adding `packages/contentful/migrations` to this path and getting rid of the migration count logic? That would help the problem with false positives when CI fails on master.

Happy to consider that question OOS for this PR though.

.github/workflows/on-remove.yml
```diff
@@ -147,14 +147,22 @@ jobs:
...
           CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
           CONTENTFUL_ENV_ID: ${{ steps.setup.outputs.crn-contentful-env-id }}
+      - name: Contentful Webhook Teardown
+        id: contentful-webhook-teardown
+        shell: bash
+        run: yarn workspace @asap-hub/contentful space:remove-webhook
+        env:
+          CONTENTFUL_SPACE_ID: ${{ steps.setup.outputs.contentful-space-id }}
+          CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
+          CONTENTFUL_ENVIRONMENT: ${{ steps.setup.outputs.crn-contentful-env-id }}
```
lennym, line 165: It might not matter, but is there a chance that the environment teardown fails - if for example the environment was manually deleted to free space - that this leaves lingering webhooks behind? Should the two steps run in parallel?

.github/actions/contentful-checks/action.yml
peterstarling: interesting.. at first I thought that what we were only interested in the migrations that have not been run yet, and not a diff, especially when fails in the pipeline or if the migration itself is broken and fails. But then I thought - so what, it's not like we want a dedicated environment in that case anyway. 

The only other issue I can think of is that the diff does not say if something has been added, it just tells you if its different from master, and again I don't know if that matters. Obvs its OOS for this PR but 100% worth the discussion, ill create a bug-ticket to address the issue with unnecessary environment creation and we can tackle it there, does this work for you @lennym ?

.github/workflows/on-remove.yml
peterstarling: well actually if it fails to take down the contetnful environment then it shouldn't really proceed to deleting the webhook, should it? so imho that's the best setup, doing it sequentially, I mean.

lennym (approved):

PR #2804, opened 2023-03-15
GP2 224 upcoming and past events on projects tab

markcsims:
- empty commit to create pr
- adds events upcoming and past to working groups
- adds test coverage for events
- adds missing coverage
- tests correct count is displayed
- starts adding events to projects page

### DIFF
--- a/packages/frontend-utils/src/events/index.ts
+++ b/packages/frontend-utils/src/events/index.ts
@@ -12,12 +12,10 @@ export type GetEventListOptions<C extends Constraint = EventConstraint> =
       | {
           before: string;
           after?: undefined;
-          sort: { sortBy: 'endDate'; sortOrder: 'desc' };
         }
       | {
           after: string;
           before?: undefined;
-          sort?: undefined;
         }
     ) & {
       constraint?: C;
@@ -52,14 +50,6 @@ export const getEventListOptions = <C extends Constraint = EventConstraint>(
     constraint,
     pageSize,
     filters: new Set(),
-    ...(past
-      ? {
-          before: time,
-          sort: {
-            sortBy: 'endDate',
-            sortOrder: 'desc' as const,
-          },
-        }
-      : { after: time }),
+    ...(past ? { before: time } : { after: time }),
   };
 };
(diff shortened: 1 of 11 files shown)


packages/gp2-components/src/organisms/ProjectDetailHeader.tsx
```diff
@@ -75,69 +77,82 @@ const ProjectDetailHeader: React.FC<ProjectDetailHeaderProps> = ({
...
-            </Paragraph>
-            <Link href={opportunitiesLink}>
-              <span css={css({ display: 'inline-flex' })}>
-                Read more <span css={infoIconStyles}>{externalLinkIcon}</span>
-              </span>
-            </Link>
+  upcomingTotal = 0,
+  pastTotal = 0,
+}) => {
+  const route = gp2Routing.projects({}).project({ projectId: id });
```
jhonny-m, line 83: we are only using it in one of the tablink, either we use it more or remove it 😄

packages/frontend-utils/src/events/index.ts
```diff
@@ -52,14 +50,6 @@ export const getEventListOptions = <C extends Constraint = EventConstraint>(
...
-    ...(past
-      ? {
-          before: time,
-          sort: {
-            sortBy: 'endDate',
-            sortOrder: 'desc' as const,
-          },
-        }
-      : { after: time }),
+    ...(past ? { before: time } : { after: time }),
```
jhonny-m, line 53: how does the ordering got changed?

packages/gp2-components/src/organisms/ProjectDetailHeader.tsx
markcsims: We use it in 5 places.

packages/frontend-utils/src/events/index.ts
markcsims: 1. The sort is no longer used by crn, the events are sourced from Algolia and this property is no longer used.
2. The past events should be ordered by endData desc. GP2 wasn't doing this.

jhonny-m: and we are gonna do it now?

jhonny-m: I'm just trying to understand if there is work missing for events

jhonny-m (approved):

PR #2813, opened 2023-03-16
CRN-1321 - Add Contentful backup restore scripts to github workflows

lennym:
These can be run on demand to restore a previous dev or prod backup to Contentful.

They take two inputs:
* The backup file to restore, which is the S3 key of the backup file created by a previous backup task, or `latest` to restore the most recent backup.
* The Contentful environment to restore into.

At present it is possible to restore any backup into any Contentful environment. We may want to add additional handrails so that non-Production backups cannot be restored into the production environment.


.github/workflows/on-demand-contentful-restore-prod.yml
peterstarling, line 1: consolidate into one

peterstarling (changes requested):

.github/workflows/on-demand-contentful-restore.yml
```diff
@@ -0,0 +1,48 @@
...
+    inputs:
+      environment:
+        required: true
+        description: Backup environment to restore
+        type: choice
+        options:
+          - Production
+          - Development
+      backup-version:
+        required: true
```
peterstarling, line 13:
```suggestion
        required: false
```

.github/workflows/on-demand-contentful-restore.yml
```diff
@@ -0,0 +1,48 @@
...
+        description: Backup file to restore
```
peterstarling, line 14:
```suggestion
        description: Backup file to restore, leave blank for latest
```

.github/workflows/on-demand-contentful-restore.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        description: Backup environment to restore
+        type: choice
+        options:
+          - Production
+          - Development
```
peterstarling, line 11: would we ever want to restore into a dedicated PR environment?

peterstarling (approved):

.github/workflows/on-demand-contentful-restore.yml
lennym: Perhaps, and that is possible. The contentful environment is a separate input to be able to support exactly that.

It's possible that these options need to be made clearer, but this input is a selection of the _source_ of the backup - and perhaps more importantly selects which of the [environment config files](https://github.com/yldio/asap-hub/tree/master/.github/environment) to use.

lennym: What we _might_ want to do is make the contentful environment optional and default to the source environment. It would mean creating a dependency that the contentful environments are always named exactly as the github workflow environments, but that might be tolerable.

PR #2828, opened 2023-03-17
gp2-321 fix users links on project and working group details pages

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/GP2-321

### DIFF
--- a/packages/react-components/src/molecules/MembersList.tsx
+++ b/packages/react-components/src/molecules/MembersList.tsx
@@ -75,15 +75,15 @@ interface MembersListProps {
   >;
   singleColumn?: boolean;
   readonly overrideNameStyles?: SerializedStyles;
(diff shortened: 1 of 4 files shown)


packages/react-components/src/molecules/MembersList.tsx
```diff
@@ -75,11 +75,15 @@ interface MembersListProps {
   >;
   singleColumn?: boolean;
   readonly overrideNameStyles?: SerializedStyles;
+  readonly overrideUserRoute?: ({ userId }: { userId: string }) => {
```
jhonny-m, line 78: can we just call it user route and default it to the crn one?

dianaguimaraes: done

jhonny-m (approved):

PR #2834, opened 2023-03-20
fix: add missing env var to remove-environment action

gabiayako:


.github/actions/sls-remove/action.yml
```diff
@@ -17,6 +17,9 @@ inputs:
   lightstep-token:
     description: 'Lightstep Token'
     required: true
+  crn-contentful-management-access-token:
+    description: 'CRN Contentful Management Access Token'
+    required: true
```
lennym, line 22: You don't need this if you're just setting the env var to a dummy value.

PR #2836, opened 2023-03-20
Draft detail page

oratiu:


apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -41,6 +45,14 @@ export default class ResearchOutputs implements ResearchOutputController {
       );
     }
 
+    if (!researchOutput.published) {
+      if (!hasAccessToDraft(loggedInUser, researchOutput)) {
+        throw Boom.notFound(
+          'You do not have permission to view this research-output',
```
oratiu, line 51: Have to double check this error type and error message

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -30,7 +31,10 @@ export default class ResearchOutputs implements ResearchOutputController {
     private externalAuthorDataProvider: ExternalAuthorDataProvider,
   ) {}
 
-  async fetchById(researchOutputId: string): Promise<ResearchOutputResponse> {
+  async fetchById(
+    researchOutputId: string,
+    loggedInUser?: UserResponse,
```
peterstarling, line 36: although doing this inside the controller allows you to skip some of the parsing logic i think the pattern is that we check the permissions in the routes rather than controllers, im not sure if that's valid but this is how we have done it until now (though I think if we were to reuse same controller methods in different routes same rules would apply so maybe we should move it to controllers at some point)

apps/crn-server/src/entities/research-output.ts
```diff
@@ -39,6 +39,7 @@ export const parseGraphQLResearchOutput = (
 
   const uniqueContactEmails = [...new Set(filteredContactEmails)];
 
+  const published = output.status === 'Published';
```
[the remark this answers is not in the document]
peterstarling: this is coming from Squidex and it's auto generated so we can't really define it there, and then below it's been typed as a boolean so not much room for improvement im afraid

peterstarling (changes requested):

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -41,10 +43,21 @@ export const researchOutputRouteFactory = (
...
       const { params } = req;
+
+      if (!req.loggedInUser) throw Boom.forbidden();
+
       const { researchOutputId } = validateResearchOutputParameters(params);
 
       const result = await researchOutputController.fetchById(researchOutputId);
 
+      if (!result.published) {
+        if (!hasAccessToDraft(req.loggedInUser, result)) {
```
peterstarling, line 54: i think you can just move this nested if to the one above?

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -120,6 +138,124 @@ describe('/research-outputs/ route', () => {
         researchOutputId,
       );
     });
+    test('Should return 404 when you are trying to get a draft and are not part of the team', async () => {
+      userMockFactory.mockReturnValueOnce(createUserResponse());
```
peterstarling, line 142: maybe be more specific here too as we don't really know which teams this user fixture contains? so yeah create user response but then modify the team list for the user to show it does not contain the team that the research output is assicated with

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -120,6 +138,124 @@ describe('/research-outputs/ route', () => {
...
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce(
+        researchOutputResponse,
+      );
+
+      const response = await supertest(app).get('/research-outputs/123');
+
+      expect(response.status).toEqual(404);
+    });
+    test('Should return 404 when you are trying to get a draft and are not part of the working group', async () => {
+      userMockFactory.mockReturnValueOnce(createUserResponse());
```
peterstarling, line 159: same as above, in fact make sure you defined both teams and working groups for the user, that way we know we didn't accidentally pass it because something else matched

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -120,6 +138,124 @@ describe('/research-outputs/ route', () => {
...
+      );
+
+      const response = await supertest(app).get('/research-outputs/123');
+
+      expect(response.status).toEqual(404);
+    });
+    test('Should return 200 when you are ASAP staff member and are trying to get a draft', async () => {
+      userMockFactory.mockReturnValueOnce({
+        ...createUserResponse(),
+        role: 'Staff',
```
peterstarling, line 179: same, add the teams, for what i know this test could be passing not because of the role but because of some team that the user belongs to

peterstarling (changes requested):

peterstarling (approved):

PR #2850, opened 2023-03-21
 Display Shared Research Drafts for Team and Work Group

kmaid:
This is ready for review. Please ignore the backend stuff as its going in on a different PR

### DIFF
--- a/apps/crn-frontend/src/network/teams/Outputs.tsx
+++ b/apps/crn-frontend/src/network/teams/Outputs.tsx
@@ -5,21 +5,26 @@ import {
   ResearchOutputsSearch,
   utils,
 } from '@asap-hub/react-components';
-import { useCurrentUserCRN } from '@asap-hub/react-context';
 import { network } from '@asap-hub/routing';
 import format from 'date-fns/format';
 import { ComponentProps } from 'react';
-import { TeamResponse } from '@asap-hub/model';
+import { ResearchOutputResponse, TeamResponse } from '@asap-hub/model';
+import { useRecoilValue } from 'recoil';
 
 import { usePagination, usePaginationParams, useSearch } from '../../hooks';
 import { useResearchOutputs } from '../../shared-research/state';
 
 import { useAlgolia } from '../../hooks/algolia';
-import { getResearchOutputs } from '../../shared-research/api';
+import {
+  getDraftResearchOutputs,
+  getResearchOutputs,
+} from '../../shared-research/api';
 import {
   algoliaResultsToStream,
   researchOutputToCSV,
+  squidexResultsToStream,
 } from '../../shared-research/export';
+import { authorizationState } from '../../auth/state';
 
 type OutputsListProps = Pick<
   ComponentProps<typeof ProfileOutputs>,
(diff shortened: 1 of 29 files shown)


apps/crn-frontend/src/network/working-groups/WorkingGroupProfile.tsx
```diff
@@ -112,11 +133,23 @@ const WorkingGroupProfile: FC<WorkingGroupProfileProps> = ({ currentTime }) => {
                   groupType="working"
                 />
               )}
+              DraftOutputs={() => (
+                <Outputs
+                  workingGroup={workingGroup}
+                  draftOutputs
+                  userAssociationMember
```
gabiayako, line 140:
```suggestion
                  userAssociationMember: {canShareResearchOutput}
```

gabiayako (changes requested):
I've noticed the text for the empty view now has `published` and `draft` in the copy. The figma is in the left side and the hub in the right side in both screenshots below

Screenshot 2023-03-22 at 10 31 58 (2)


Screenshot 2023-03-22 at 10 31 27 (2)

Another change that is part of this ticket is to add "Draft" tag in the cards :)

Screenshot 2023-03-22 at 10 33 47

apps/crn-frontend/src/network/working-groups/Outputs.tsx
```diff
@@ -64,28 +133,27 @@ const OutputsList: React.FC<OutputsListProps> = ({
       }
       userAssociationMember={userAssociationMember}
       workingGroupAssociation
+      exportResults={exportResults}
```
gabiayako, line 136: I believe it's necessary to pass

```
draftOutputs={draftOutputs}
```

here the same way it was done for `apps/crn-frontend/src/network/teams/Outputs.tsx` so the right text appears :)

packages/react-components/src/templates/ProfileOutputs.tsx
gabiayako, line 1: For the `NoOutputsPage`  when `draftOutputs` is false, it seems (https://www.figma.com/file/qM73Om8hMDi8o15nj09ECw/02-CRN-Screens?node-id=19683-311302&t=YEt1DzjtJeBWmP9U-0) that now the text ends with `doesn’t have any published outputs.`

gabiayako (changes requested):

apps/crn-frontend/src/network/teams/Outputs.tsx
```diff
@@ -50,30 +60,62 @@ const OutputsList: React.FC<OutputsListProps> = ({
...
+            { header: true },
+          ),
+          (paginationParams) =>
+            getDraftResearchOutputs(
+              {
+                filters,
+                searchQuery,
+                userAssociationMember,
+                associationId: teamId,
+                draftsOnly: true,
```
gabiayako, line 96: Not a big deal, but I was wondering if we could remove this `draftsOnly` from here because it seems kind of weird having to pass `draftsOnly: true` to a function called `getDraftResearchOutputs`. 

Maybe we could do something like
```
export const useResearchOutputs = (
  options: ResearchOutputListOptions,
  published = true,
) => {
  (...)
throw (
      !published && 'userAssociationMember' in options
        ? getDraftResearchOutputs(options, authorization)
        : getResearchOutputs(client, options).then(
            (data): ListResearchOutputResponse => ({
              total: data.nbHits,
              items: data.hits,
            }),
          )
(...)
```

in `apps/crn-frontend/src/shared-research/state.ts` so we would not pass `draftsOnly` inside `option`

kmaid: I don't think it matters too much. The types make it explicit that draftsOnly can only be true for that method. Its useful for type discrimination

PR #2851, opened 2023-03-21
CRN-1234 Create external authors content model and data provider

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/CRN-1234

Example of successful run of algolia sync https://github.com/yldio/asap-hub/actions/runs/4492402209/jobs/7902235174 - contentful data

It's possible to check that algolia results appear in the hub for external authors when contentful feature flag is enabled.

Screenshot 2023-03-22 at 14 11 47

Screenshot 2023-03-22 at 13 57 52

Screenshot 2023-03-22 at 13 57 17


apps/crn-server/src/handlers/external-author/index-handler.ts
```diff
@@ -15,45 +19,58 @@ import {
...
+        logger.debug(`Fetched external author ${externalAuthor.displayName}`);
+        return externalAuthor;
+      } catch {
         await algoliaClient.remove(event.detail.payload.id);
-        return;
+        return null;
       }
+    };
 
-      logger.debug(`Fetched external author ${externalAuthor.displayName}`);
```
gabiayako, line 46: I've made this piece

```
const externalAuthor = await externalAuthorDataProvider.fetchById(
        event.detail.payload.id,
      );
```

to be wrapped by a try catch block because now that we are using the controller, if the External Author is not found is going to throw a not found error, it's not going to return null

lennym (approved):

apps/crn-server/src/handlers/external-author/index-handler.ts
peterstarling: thats correct, yeah

apps/crn-server/src/handlers/external-author/index-handler.ts
```diff
@@ -70,18 +87,48 @@ const squidexGraphqlClient = new SquidexGraphql(getAuthToken, {
...
       externalAuthorRestClient,
       squidexGraphqlClient,
-    ),
-    algoliaSearchClientFactory({ algoliaApiKey, algoliaAppId, algoliaIndex }),
+    );
+
+/* istanbul ignore next */
+const selectedAlgoliaIndex = isContentfulEnabledV2
+  ? `${algoliaIndex}-contentful`
+  : algoliaIndex;
```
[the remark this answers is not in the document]
peterstarling: actually i don't think this is valid, this is sort of why this ticket is blocked by mine IF we want to add the live resource syncing with algolia for contentful. 

As it is proposed in this PR regardless of where the event was emitted from it would be indexed based on the feature flag - thats actually incorrect, right? We want to index in the  `-contentful` index if it came from Contentful.

apps/crn-server/test/handlers/external-author/index-handler.test.ts
```diff
@@ -1,28 +1,108 @@
...
+    process.env = { ...OLD_ENV };
+  });
+
   afterEach(() => jest.clearAllMocks());
 
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+
+  test('Should instantiate contentful data provider when feature flag is truthy', async () => {
```
peterstarling, line 30: see my comment above, i think this logic isn't correct, I think where this is indexed depends on the source. And I think this part should be blocked by my work. I suggest that you leave the live sync of Algolia out of scope for this one and just make sure that when we run on-demand algolia sync it gets indexed in the `-contentful` index. I can probably rebase on top of that and add the sync to external authors

peterstarling (changes requested):
i dont think you need these changes to indexers

apps/crn-server/src/handlers/external-author/index-handler.ts
peterstarling, line 1: to be fair to be more thorough we should change the data type that the data-provider returns so it's a `DataObject` not the `Response` type but it's fine

peterstarling (approved):

PR #2854, opened 2023-03-22
CRN-1331 adding field

lctrt:


apps/crn-frontend/src/network/teams/__tests__/TeamOutput.test.tsx
```diff
@@ -39,13 +39,13 @@ const mandatoryFields = async (
   {
     link = 'http://example.com',
     title = 'example title',
-    description = 'example description',
```
kmaid, line 42: What is the plan for migration? It was my understanding that we would maintain the old field but try and convert it into MD in the FE upon an old RO edit. This seems only to give the new MD field in which case I am not sure why the name changed.

apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -75,7 +75,9 @@ export const researchOutputToCSV = (
   rrid: output.rrid,
   accession: output.accession,
   labCatalogNumber: output.labCatalogNumber,
-  description: htmlToCsvText(output.description),
+  description: [htmlToCsvText(output.description), output.descriptionMD]
```
kmaid, line 78: Some concern here. I think this function stripped out HTML from description. If this is now MD is this still required? I would have thought MD is just readable in exports. Maybe a question to tiff.

package.json
```diff
@@ -59,7 +59,8 @@
     "expect-playwright": "0.8.0",
     "vfile": "4.2.1",
     "@typescript-eslint/eslint-plugin": "5.54.1",
-    "@typescript-eslint/parser": "5.54.1"
+    "@typescript-eslint/parser": "5.54.1",
+    "@contentful/field-editor-markdown@1.1.16": "patch:@contentful/field-editor-markdown@npm%3A1.1.16#./.yarn/patches/@contentful-field-editor-markdown-npm-1.1.16-297013de9c.patch"
```
kmaid, line 63: This probally shouldn't be there

packages/react-components/package.json
```diff
@@ -31,9 +31,11 @@
     "core-js": "3.29.1",
     "date-fns": "2.29.3",
     "date-fns-tz": "1.3.8",
+    "dompurify": "^2.2.9",
```
kmaid, line 34: Does markdown still leave us open to attack? I kind of feel doing this client side might be a mistake as what you see in Contentful might not reflect the hub

packages/react-components/package.json
```diff
@@ -31,9 +31,11 @@
...
     "emotion-normalize": "11.0.1",
     "fast-deep-equal": "3.1.3",
     "hast-util-sanitize": "3.0.2",
+    "markdown-to-jsx": "^7.1.1",
```
kmaid, line 38: This probablly can be removed

packages/react-components/src/atoms/__tests__/Markdown.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+it('renders processed markdown text', () => {
+  const { getByText } = render(
+    <Markdown value={`# some text content\n **foo**`} />,
+  );
+  expect(getByText('some text content').tagName).toBe('H4');
+  expect(getByText('foo').tagName).toBe('STRONG');
+});
+
+it('renders processed links', () => {
```
kmaid, line 13: It be nice to see a test that shows we won't allow XSS or something

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -108,15 +111,18 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
...
         <div css={cardsStyles}>
           <SharedResearchOutputHeaderCard {...props} published={published} />
-          {((description && !isGrantDocument) || !!tags.length) && (
+          {((hasDescription && !isGrantDocument) || !!tags.length) && (
             <Card>
-              {description && !isGrantDocument && (
+              {hasDescription && !isGrantDocument && (
                 <div css={{ paddingBottom: `${12 / perRem}em` }}>
                   <Headline2 styleAsHeading={4}>Description</Headline2>
+                  <Markdown value={descriptionMD}></Markdown>
```
kmaid, line 119: It seems likely that older documents could have both MD and Text. I don't think you have considered this case.

packages/react-components/package.json
lctrt: we're using `rehype-sanitize` instead of dompurify (also for rich text). I agree that it's worth considering. But I think that discussion should be separated from this, and we should take care of that sanitization on both markdown and rich text in the BE.

packages/react-components/src/templates/SharedResearchOutput.tsx
lctrt: I have! Until we deal with migration, we display both, markdown first.

apps/crn-frontend/src/network/teams/__tests__/TeamOutput.test.tsx
lctrt: That's correct but to be addressed in a separate ticket. At the moment both field coexist, the old one is not mandatory.

apps/crn-frontend/src/shared-research/export.ts
lctrt: I'm still cleaning up the old description if it exists, and adding the raw markdown next to it.

apps/crn-frontend/src/network/teams/__tests__/TeamOutput.test.tsx
kmaid: Why is the new one mandatory? I think some of the document types don't require a description

packages/react-components/src/templates/SharedResearchOutput.tsx
kmaid: Shouldn't we just prefer markdown if it exists?

apps/crn-frontend/src/network/teams/__tests__/TeamOutput.test.tsx
lctrt: if you look in `apps/crn-server/src/validation/research-output.validation.ts` you'll see the old one was mandatory

Riccardo-Brown (approved):

apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -169,7 +170,7 @@ const researchOutputPostRequestValidationSchema: JSONSchemaType<ResearchOutputPo
     },
     required: [
       'documentType',
-      'description',
+      'descriptionMD',
```
peterstarling, line 173: I see you're making the markdown description compulsory instead of the old one - but the types still say both are required:
https://github.com/yldio/asap-hub/blob/313d4a49d59f59919a7f17a8d9244e591d3d5a3a/packages/model/src/research-output.ts#L292-L293

To be fair `JSONSchemaType` should enforce the correct behaviour but unfortunately it only does the opposite check (so if you mark a property as optional then you have to give a `nullable` parameter).

peterstarling (changes requested):
only small type issues, otherwise looks good, surprisingly easy btw (not ot belittle your achievement here but I thought there would be a lot more issues, FE and BE).

One thing I am not a huge fan of is the MD suffix - it's the same as the medical degree which we actually have in our codebase

peterstarling (approved):

PR #2862, opened 2023-03-22
Add User content model and data provider for Contentful

lennym:
* Add migration to Contentful to create user and lab models
* Add Contentful data providers for users and avatar assets
* Add relationship between users, teams and labs
* Add feature switch to use Contentful for users
* Update filter options for users to only allow single filters

## OOS/TODO:

* Syncing ORCID works
* Invitation and onboarding journey
* Resolve race condition in profile edits
* Displaying team names in Contentful user UI.
* Working group membership
* Data migration

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/users.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/users.data-provider.ts
@@ -43,7 +43,6 @@ export type TeamMembership = NonNullable<
 >;
 
 export class UserContentfulDataProvider implements UserDataProvider {
-
   constructor(
     private contentfulClient: GraphQLClient,
     private getRestClient: () => Promise<Environment>,
@@ -121,11 +120,9 @@ export class UserContentfulDataProvider implements UserDataProvider {
     return usersCollection || { total: 0, items: [] };
   }
 
-  /* eslint-disable class-methods-use-this */
   async create(): Promise<string> {
     throw new Error('Method not implemented.');
   }
-  /* eslint-enable class-methods-use-this */
 
   async update(id: string, data: UserUpdateDataObject): Promise<void> {
     const fields = cleanUser(data);
--- a/apps/crn-server/src/dependencies/users.dependencies.ts
+++ b/apps/crn-server/src/dependencies/users.dependencies.ts
@@ -16,14 +16,9 @@ import {
 } from '../config';
 import { UserContentfulDataProvider } from '../data-providers/contentful/users.data-provider';
 import { AssetContentfulDataProvider } from '../data-providers/contentful/assets.data-provider';
-import {
-  UserDataProvider,
-  UserSquidexDataProvider,
-} from '../data-providers/users.data-provider';
-import {
-  AssetDataProvider,
-  AssetSquidexDataProvider,
-} from '../data-providers/assets.data-provider';
+import { UserDataProvider, AssetDataProvider } from '../data-providers/types';
+import { UserSquidexDataProvider } from '../data-providers/users.data-provider';
+import { AssetSquidexDataProvider } from '../data-providers/assets.data-provider';
 import { getAuthToken } from '../utils/auth';
 import { getContentfulRestClientFactory } from './clients.dependencies';
 
--- a/apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
+++ b/apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
@@ -14,7 +14,7 @@ import {
   getUserCreateDataObject,
 } from '../../fixtures/users.fixtures';
 import { getEntry } from '../../fixtures/contentful.fixtures';
-import { UserDataProvider } from '../../../src/data-providers/users.data-provider';
+import { UserDataProvider } from '../../../src/data-providers/types';
 import { UserContentfulDataProvider } from '../../../src/data-providers/contentful/users.data-provider';
 import { getContentfulGraphqlClientMock } from '../../mocks/contentful-graphql-client.mock';
 import { getContentfulEnvironmentMock } from '../../mocks/contentful-rest-client.mock';
@@ -579,7 +579,7 @@ describe('User data provider', () => {
   });
 
   describe('Create', () => {
-    test('pending', async () => {
+    test('not implemented', async () => {
       expect(async () =>
         userDataProvider.create(getUserCreateDataObject()),
       ).rejects.toThrow();
--- a/apps/crn-server/src/app.ts
+++ b/apps/crn-server/src/app.ts
@@ -69,10 +69,7 @@ import Users, { UserController } from './controllers/users';
 import WorkingGroups, {
   WorkingGroupController,
 } from './controllers/working-groups';
-import {
-  AssetDataProvider,
-  AssetSquidexDataProvider,
-} from './data-providers/assets.data-provider';
+import { AssetSquidexDataProvider } from './data-providers/assets.data-provider';
 import { CalendarSquidexDataProvider } from './data-providers/calendars.data-provider';
 import { CalendarContentfulDataProvider } from './data-providers/contentful/calendars.data-provider';
 import { AssetContentfulDataProvider } from './data-providers/contentful/assets.data-provider';
@@ -115,6 +112,7 @@ import {
   TutorialsSquidexDataProvider,
 } from './data-providers/tutorials.data-provider';
 import {
+  AssetDataProvider,
   NewsDataProvider,
   PageDataProvider,
   UserDataProvider,
--- a/apps/crn-server/src/controllers/users.ts
+++ b/apps/crn-server/src/controllers/users.ts
@@ -8,8 +8,7 @@ import {
   UserUpdateRequest,
 } from '@asap-hub/model';
 import Intercept from 'apr-intercept';
-import { AssetDataProvider } from '../data-providers/assets.data-provider';
-import { UserDataProvider } from '../data-providers/types';
+import { AssetDataProvider, UserDataProvider } from '../data-providers/types';
 import { parseUserToResponse } from '../data-providers/users.data-provider';
 import { fetchOrcidProfile, transformOrcidWorks } from '../utils/fetch-orcid';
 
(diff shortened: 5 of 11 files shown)


apps/crn-server/src/data-providers/contentful/teams.data-provider.ts
```diff
@@ -203,6 +210,44 @@ export const parseContentfulGraphQlTeams = (item: TeamItem): TeamDataObject => {
...
+          lastName: lastName ?? '',
+          email: email ?? '',
+          role: role ?? '',
+          inactiveSinceDate,
+          avatarUrl: avatar?.url ?? '',
+          displayName: `${firstName} ${lastName}`,
+        },
+      ];
+    },
+    [] as TeamMember[],
```
peterstarling, line 248:
```suggestion
    [],
```

apps/crn-server/src/data-providers/contentful/users.data-provider.ts
```diff
@@ -1,52 +1,273 @@
...
-      total: 1,
+      total: usersCollection?.total,
+      items: usersCollection?.items
+        .filter((x): x is UserItem => x !== null)
+        .map(parseContentfulGraphQlUsers),
     };
   }
 
   async create(): Promise<string> {
+    await this.getRestClient();
```
peterstarling, line 88: what do you need this for then if its not implemented?

peterstarling: in fact, why is it not? is it OOS?

apps/crn-server/src/data-providers/contentful/users.data-provider.ts
```diff
@@ -1,52 +1,273 @@
...
+  );
+
+  return {
+    id: item.sys.id,
+    createdDate: item.sys.firstPublishedAt,
+    lastModifiedDate: item.sys.publishedAt,
+    workingGroups: [],
+    ...item,
+    email: item.email ?? '',
+    contactEmail: item.contactEmail ?? '',
```
peterstarling, line 204: I see you haven't written the tests for this bit yet which might explain why this was missed but in the Squidex data provider we seem to be defaulting to undefined when Squidex returns values as Maybe (similar to Contentful). 

To be fair I can see that the test coverage for defaulting to undefined isn't great in the old data provider either.

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -1,18 +1,30 @@
+import { Environment } from '@asap-hub/contentful';
+
 import { UserContentfulDataProvider } from '../../../src/data-providers/contentful/users.data-provider';
 
-describe('User data provider', () => {
-  const userDataProvider = new UserContentfulDataProvider();
+import { getContentfulGraphqlClientMock } from '../../mocks/contentful-graphql-client.mock';
+import { getContentfulEnvironmentMock } from '../../mocks/contentful-rest-client.mock';
+
+describe.skip('User data provider', () => {
```
peterstarling, line 8: ?

packages/contentful/src/mocks/graphql-client.mock.ts
```diff
@@ -17,12 +17,16 @@ export const getContentfulGraphqlClientMockServer = (
   });
   const mocks = {
     JSON: () => '2021-10-12T15:42:05Z',
+    DateTime: () => null,
```
lennym, line 20: For reasons I cannot fathom this line shows as not covered, but also the tests fail completely if it is removed.

packages/contentful/src/crn/queries/teams.queries.ts
```diff
@@ -23,6 +23,38 @@ export const teamsContentQueryFragment = gql`
         url
       }
     }
+    linkedFrom {
+      teamMembershipCollection(limit: 100) {
```
lennym, line 27: This is sufficient based on current data (the largest team is ~50-60 users) but we may want to add an additional check so that if the users on a team exceeds 100 then it's users are aggregated across multiple queries.

packages/contentful/src/utils.ts
```diff
@@ -120,3 +121,20 @@ export const updateEntryFields = (
   });
   return updatedEntry;
 };
+
+export const patchAndPublish = async (
+  entry: Entry,
+  fields: Record<string, unknown>,
+): Promise<void> => {
+  const patch: OpPatch[] = Object.entries(fields).map(([key, value]) => ({
```
peterstarling, line 129: you can avoid adding the json-patch library to bundle by doing 
```suggestion
  const patch: Parameters<Entry["patch"]>[0] = Object.entries(fields).map(([key, value]) => ({
```

packages/model/src/user.ts
```diff
@@ -218,15 +218,34 @@ export interface UserAvatarPostRequest {
...
+      labId?: never;
+      role?: string[];
+      teamId?: string;
+      code?: never;
+      hidden?: never;
+      onboarded?: never;
+      orcid?: never;
+    };
 
 export type FetchUsersOptions = FetchOptions<FetchUsersFilter>;
```
peterstarling, line 250: as mentioned previously I think you want to disable search entirely no? in which case I would advise to remove it from types too

packages/contentful/src/mocks/graphql-client.mock.ts
peterstarling: yeah to be fair i wouldn't mind skipping coverage checks for this file

packages/contentful/src/crn/queries/teams.queries.ts
peterstarling: meh :D

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/users.data-provider.ts
```diff
@@ -1,52 +1,299 @@
...
-      ],
-      total: 1,
+      total: result?.total,
+      items: result?.items
+        .filter((x): x is UserItem => x !== null)
+        .map(parseContentfulGraphQlUsers),
     };
   }
 
+  async fetchUsers(options: FetchUsersOptions) {
```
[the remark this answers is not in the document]
lennym: That's correct, it's just pulled out to clean up the code. Will make it private.

apps/crn-server/src/data-providers/contentful/users.data-provider.ts
lennym: I had it in my head that I'd preserved the default types from the previous implementation, but apparently not. Will change to `undefined`.

packages/contentful/test/utils.test.ts
```diff
@@ -202,3 +203,40 @@ describe('updateEntryFields', () => {
...
+  test('patches with a "replace" op if field exists on entry', async () => {
+    entry.fields = {
+      foo: null,
+    };
+    await patchAndPublish(entry, { foo: 'bar', baz: 1 });
+    expect(entry.patch).toHaveBeenCalledWith([
+      { op: 'replace', path: '/fields/foo', value: { 'en-US': 'bar' } },
+      { op: 'add', path: '/fields/baz', value: { 'en-US': 1 } },
+    ]);
+  });
```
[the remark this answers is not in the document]
lennym: It took a while to work out. There's no mention in the contentful documentation, and the error it returns is next to useless.

😬

apps/crn-server/test/data-providers/users.data-provider.test.ts
```diff
@@ -648,95 +631,40 @@ describe('User data provider', () => {
...
-        expect.anything(),
-        {
-          top: 12,
-          skip: 2,
-          filter: expectedFilter,
-        },
-      );
-    });
-    test('Should escape double quotation mark by escaping it', async () => {
+    test('Should query with team filters and return the users', async () => {
```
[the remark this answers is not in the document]
lennym: I've removed `search` from the allowed fetch options [here](https://github.com/yldio/asap-hub/pull/2862/commits/01c6029095578374558185167fd3dacc29e27074). Is there anything else required?

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -1,34 +1,519 @@
...
+        total: 0,
+        items: [],
       });
     });
-  });
 
-  describe('Create', () => {
-    test('should throw an error', async () => {
-      await expect(userDataProvider.create()).rejects.toThrow();
+    describe('query parameters', () => {
```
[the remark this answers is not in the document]
lennym: So the types as defined [here](https://github.com/yldio/asap-hub/blob/01c6029095578374558185167fd3dacc29e27074/packages/model/src/user.ts#L222-L248) should specify exactly one filter per query for linked filters.

My assumption is that if we added a unit test to cover this case also then it would fail type-checking?

apps/crn-server/src/dependencies/users.dependencies.ts
```diff
@@ -0,0 +1,63 @@
...
+  UserSquidexDataProvider,
+} from '../data-providers/users.data-provider';
+import {
+  AssetDataProvider,
+  AssetSquidexDataProvider,
+} from '../data-providers/assets.data-provider';
+import { getAuthToken } from '../utils/auth';
+import { getContentfulRestClientFactory } from './clients.dependencies';
+
+const getRestClient = () =>
```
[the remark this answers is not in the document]
lennym: Not for mocking, just pulling some repetitive code out into single function.

apps/crn-server/src/data-providers/contentful/users.data-provider.ts
lennym: I've made the default values align across both data providers to the squidex values and added unit tests to cover both.

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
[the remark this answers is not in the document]
peterstarling: yeah no - i hear you, you would need to cast as unknown first and i think TS checks are good enough, I guess what I meant though was that i think you can still do a text search with these filters, whereas that functionality has actually been removed no?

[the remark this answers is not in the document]
peterstarling: nvm i see you've address it already

peterstarling (approved):

gabiayako (approved):

PR #2872, opened 2023-03-24
refactor: add generic data provider interface

Riccardo-Brown:
https://github.com/yldio/asap-hub/pull/2778#discussion_r1133948279


apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -69,17 +70,22 @@ export const makeDraftFilter = ({
...
-  ): Promise<string>;
-  update(id: string, input: ResearchOutputUpdateDataObject): Promise<string>;
-}
+type ResearchOutputFetchOptions = {
+  take?: number;
+  skip?: number;
+  search?: string;
+  filter?: FetchResearchOutputFilter;
+  includeDrafts?: boolean;
+};
```
lennym, line 79: It feels like there's going to be a lot of repetition of these properties. It might be worth insisting that all `fetch` implementations use the same options (modulo `includeDrafts`)

packages/model/src/data-providers.ts
```diff
@@ -16,26 +16,34 @@ export type FetchCalendarProviderOptions = {
   resourceId?: string;
 };
 
-export interface CalendarDataProvider<
-  CalendarDataObject = DefaultCalendarDataObject,
-  ListCalendarDataObject = DefaultListCalendarDataObject,
+export interface DataProvider<
+  TItemResponse = null,
+  TFetchOptions = null,
```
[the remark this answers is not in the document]
lennym: Turns out this suggestion doesn't work at all because lots of the existing data providers don't fit that pattern.

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
lennym: `FetchOptions< FetchResearchOutputFilter> & { includeDrafts?: boolean }`

packages/model/src/data-providers.ts
```diff
@@ -16,26 +16,34 @@ export type FetchCalendarProviderOptions = {
...
+  TListItemResponse = null,
+  TCreateData = null,
```
peterstarling, line 23: not sure if that works but maybe we should give it a `never` default which means that if its not given you will not be able to type anything into the create method? 
Same for update data maybe

apps/crn-server/test/mocks/user-data-provider.mock.ts
```diff
@@ -1,8 +1,8 @@
-import { UserDataProvider } from '../../src/data-providers/users.data-provider';
+import { UserDataProvider } from '../../src/data-providers/types';
 
-export const userDataProviderMock: jest.Mocked<UserDataProvider> = {
+export const userDataProviderMock = {
   fetchById: jest.fn(),
```
peterstarling, line 4: since its always these for methods now maybe we want a helper function to generate these? 
Although since create and update may be no-ops it's worth addressing that in that helper method too

Riccardo-Brown: So if i understand correctly we would had  one `getDataProviderMock` function instead of these specific mock for each content type.

apps/crn-server/test/mocks/asset-data-provider.mock.ts
```diff
@@ -1,5 +1,5 @@
 import { AssetDataProvider } from '../../src/data-providers/assets.data-provider';
 
-export const assetDataProviderMock: jest.Mocked<AssetDataProvider> = {
+export const assetDataProviderMock = {
   create: jest.fn(),
-};
+} as unknown as jest.Mocked<AssetDataProvider>;
```
lennym, line 5: I see this change a lot here. What does it do?

packages/model/src/data-providers.ts
Riccardo-Brown: i will change it to never and update the partial of create

apps/crn-server/src/data-providers/users.data-provider.ts
```diff
@@ -38,13 +39,17 @@ import { buildEqFilterForWords, buildODataFilter } from '../utils/odata';
...
 
-export interface UserDataProvider {
-  fetchById(id: string): Promise<UserDataObject | null>;
-  fetch(options: FetchUsersOptions): Promise<ListUserDataObject>;
-  create(input: UserCreateDataObject): Promise<string>;
-  update(id: string, update: UserUpdateDataObject): Promise<void>;
-}
-export class UserSquidexDataProvider implements UserDataProvider {
+export class UserSquidexDataProvider
+  implements
```
peterstarling, line 43: you've already defined this type in here https://github.com/yldio/asap-hub/blob/18e2b18c373d25434b47d74493420d15c7e51391/apps/crn-server/src/data-providers/types/users.data-provider.types.ts#L10
so use that instead, same goes for all implementations

packages/model/src/data-providers.ts
```diff
@@ -16,26 +16,37 @@ export type FetchCalendarProviderOptions = {
...
-  CalendarDataObject = DefaultCalendarDataObject,
-  ListCalendarDataObject = DefaultListCalendarDataObject,
+export interface DataProvider<
+  TItemResponse = null,
+  TFetchOptions = null,
+  TListItemResponse = {
+    items: TItemResponse[];
+    total: number;
+  },
+  TCreateData = never,
```
peterstarling, line 26: a nice alternative is to do this:
```
type DataProvider<T = {}, K = null>  = {
  fetchById: (id: string) => Promise<T | null>;
} & (K extends null ? {} : {
  create: K extends null ? undefined : (data: K) => Promise<string>;
});
```
you can still implement regular Types and with the above you are able to conditionally add create or update based on the input generic. Ill leave it up to you to decide which one is better

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/pages.data-provider.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable class-methods-use-this */
```
peterstarling, line 1: I think it's best to disable this rule for the entire crn-server

apps/crn-server/test/app.test.ts
```diff
@@ -142,15 +145,13 @@ describe('Contentful feature flag', () => {
...
-        fetch: jest.fn(),
-      };
-
-    const dashboardContentfulDataProviderMock: jest.Mocked<DashboardDataProvider> =
-      {
-        fetch: jest.fn(),
-      };
+    const dashboardSquidexDataProviderMock = {
+      fetch: jest.fn(),
+    } as unknown as jest.Mocked<DashboardSquidexDataProvider>;
```
peterstarling, line 150: why are you having to cast?

apps/crn-server/test/data-providers/assets.data-provider.test.ts
```diff
@@ -26,25 +26,37 @@ describe('Asset data provider', () => {
...
+          contentType: 'image/jpeg',
+        }),
       ).rejects.toThrow();
       expect(nock.isDone()).toBe(true);
     });
   });
+
+  describe('Fetch', () => {
+    test('should throw an error', async () => {
+      await expect(assetDataProvider.fetch()).rejects.toThrow();
```
peterstarling, line 53:
```suggestion
      await expect(assetDataProvider.fetch()).rejects.toThrow('Not implemented');
```

apps/crn-server/test/data-providers/assets.data-provider.test.ts
```diff
@@ -26,25 +26,37 @@ describe('Asset data provider', () => {
...
+    });
+  });
+
+  describe('Fetch-by-id', () => {
+    test('should throw an error', async () => {
+      await expect(assetDataProvider.fetchById()).rejects.toThrow();
```
peterstarling, line 59: same as above

apps/crn-server/test/data-providers/contentful/pages.data-provider.test.ts
```diff
@@ -94,4 +94,10 @@ describe('Pages Contentful Data Provider', () => {
       );
     });
   });
+
+  describe('Fetch-by-id', () => {
+    test('should throw an error', async () => {
+      await expect(pageDataProvider.fetchById()).rejects.toThrow();
```
peterstarling, line 100: same here - basically i think it's best to be more explicit as to why it's throwing

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -970,11 +970,12 @@ describe('ResearchOutputs data provider', () => {
...
-          researchOutputId,
-          researchOutputUpdateData,
-        );
-        expect(result).toEqual(researchOutputId);
+        await expect(
+          researchOutputDataProvider.update(
+            researchOutputId,
+            researchOutputUpdateData,
+          ),
+        ).resolves.not.toThrow();
```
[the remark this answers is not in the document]
peterstarling: that is a strange change i admit. Of course the important bit is the nock assertion and to be fair we could just call the update method and never expect anything, but maybe just for the sake of having something here can you not expect a result of the update like it did before?

peterstarling (changes requested):

apps/crn-server/test/app.test.ts
Riccardo-Brown: I can't really remember, but as the dashboard provider does not implement the generic data provider i will revert the change.

peterstarling (approved):

PR #2889, opened 2023-03-28
Crn 1359 team outputs search bar is not accepting words

oratiu:
Thank you @gabiayako for the help you provided.

This bug would have also applied to working group once we enabled the outputs search bar there.

I initially tried keeping the functionality of ProfileSwitch as unchanged as possible, tried wrapping the OutputsComponent in a useCallback hook but this resulted in code coverage issues since team and working group could be undefined yet it's not allowed to have them undefined in the Outputs component. @kmaid I tried passing the team to the hook but it would still re-render that way.
 Then I tried pulling them out of the Profile switch and just having the routs outside that component but it would mean having to control the Redirect.


gabiayako:
I've tested locally, I think this would work as well https://github.com/yldio/asap-hub/compare/suggestion?expand=1

kmaid (approved):

PR #2896, opened 2023-03-28
Handled case in which the draft output belongs to a team

oratiu:


packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -81,8 +81,13 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
...
       {!published && (
         <Toast accent="warning">
-          This draft is available to members in the working group listed below.
-          Only PMs can publish this output.
+          This draft is available to members in the{' '}
+          {props.workingGroups
+            ? 'working group'
+            : props.teams.length > 1
+            ? 'teams'
+            : 'team'}{' '}
```
kmaid, line 89: Can you please inject this from the parent. Also `team${props.teams.length > 1 ? 's':''}`

kmaid (approved):

PR #2898, opened 2023-03-29
CT-7 Calendars Contentful content model and data provider

gabiayako:
This PR adds Contentful Calendar content model and data provider.

Hub showing calendars from contentful:
calendar-hub

Contentful data:
calendar-contentful


packages/contentful/migrations/calendars/20230328165340-create-calendars-content-model.js
```diff
@@ -0,0 +1,150 @@
...
+    .omitted(false);
+
+  calendars
+    .createField('syncToken')
+    .name('Google Last Sync Token')
+    .type('Symbol')
+    .localized(false)
+    .required(false)
+    .validations([])
+    .disabled(true)
```
gabiayako, line 109: When disable is true, the field does not appear on Contentful

calendar-disabled-field

But they still can come in the response
calendar-graphql

In this case, I've left the field enabled, added a value to the entry and then disabled the field. That's why it has a value.

apps/crn-server/src/data-providers/contentful/calendars.data-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    return calendarEntry.sys.id;
+  }
+
+  async update(id: string, update: CalendarUpdateDataObject): Promise<void> {
+    const environment = await this.getRestClient();
+
+    const calendar = await environment.getEntry(id);
+
+    Object.entries(update).forEach(([fieldName, fieldValue]) => {
```
peterstarling, line 96: it's probably something we do all over the place, maybe there should be a helper function for that?

packages/contentful/migrations/calendars/20230328165340-create-calendars-content-model.js
peterstarling: isnt that what we want? that the field is not visible in the UI but still returned in the API? I mean it's quite handy to have these field in squidex for debugging but we can always just use the API for that. I don't know, what do you think?

peterstarling: oh sorry, you're just saying that you did in fact make it disabled but just explaining how you got the value into it yeah? sure that makes sense, it would only ever be updated via the API so this works for me

gabiayako: yeah, I was just explaining why the value does not appear in the ui but appears in the response :)

apps/crn-server/src/data-providers/contentful/calendars.data-provider.ts
gabiayako: committed here [moving update function to utils](https://github.com/yldio/asap-hub/pull/2898/commits/6d7e2795673f8de4eac34215a08dd37e6b6828fc)

PR #2910, opened 2023-03-30
[CRN-1317] chore: update aws-sdk from v2 to v3

Riccardo-Brown:


apps/crn-server/src/handlers/webhooks/webhook-squidex.ts
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
peterstarling, line 1: same as above

apps/gp2-server/src/handlers/webhooks/webhook-squidex.ts
```diff
@@ -19,7 +19,7 @@ export const squidexWebhookFactory = (
   return lambda.http(squidexHandler);
 };
 
-const eventBridge = new EventBridge();
+const eventBridge = new EventBridge({});
```
peterstarling, line 22: since they've changed the constructor function input does it mean it still resolves the credentials from lambda environment? basically - does it still work?

peterstarling (changes requested):
only small changes requested, the new aws sdk looks much better though!

also you may want to manually test all these affected handlers

apps/gp2-server/src/handlers/webhooks/webhook-squidex.ts
Riccardo-Brown: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html

V3 provides a default credential provider in Node.js. So you are not required to supply a credential provider explicitly. The default credential provider attempts to resolve the credentials from a variety of different sources in a given precedence, until a credential is returned from the one of the sources. If the resolved credential is from a dynamic source, which means the credential can expire, the SDK will only use the specific source to refresh the credential.

Here's the order of the sources where the default credential provider resolve credentials from:

1. Environment variables
2. The shared credentials file
3. Credentials loaded from the Amazon ECS credentials provider (if applicable)
4. Credentials loaded from AWS Identity and Access Management using the credentials provider of the Amazon EC2 instance (if configured in the instance metadata)

apps/crn-server/src/config.ts
```diff
@@ -78,8 +79,8 @@ export const userInviteReturn = EMAIL_RETURN || 'hub.invites.dev@asap.science';
 export const eventBus = EVENT_BUS || 'asap-events-dev';
 export const eventSource = EVENT_SOURCE || '';
 export const eventBridgeEndpoint = EVENT_BRIDGE_ENDPOINT;
-export const eventBridgeAccessKey = EVENT_BRIDGE_ACCESS_KEY;
-export const eventBridgeSecret = EVENT_BRIDGE_SECRET;
+export const eventBridgeAccessKey = EVENT_BRIDGE_ACCESS_KEY!;
```
peterstarling, line 82: actually you might want to consider using the `|| ''` pattern we used for all other env vars, sorry

peterstarling (approved):

PR #2911, opened 2023-03-30
CT-12 Migrate calendar data

gabiayako:


apps/crn-server/test/handlers/calendar/utils.test.ts
peterstarling, line 1: you may want to rebase once this is in https://github.com/yldio/asap-hub/blob/c3c6fb5c5f71fa0a569801006c115bdf835b8834/apps/crn-server/src/dependencies/external-authors.dependencies.ts
In my PR I am dealing with the same issue of having to switch between the providers based on the flag, I've created a new directory for resolving dependencies (i think we should put more dependency resolvers in there so they do not have to be repeated all the time) and then the test is quite simple compared to what you have in here.

packages/cms-data-sync/src/utils/setup.ts
```diff
@@ -52,5 +54,13 @@ export const getSquidexAndContentfulClients = async () => {
     CONTENTFUL_ENV_ID!,
   );
 
+  if (disableWebhook) {
```
peterstarling, line 57: have you thought about how to enable the webhook back? Once the migration is done it needs to be returned to it's normal state.

the issue with making this part of the client is that we only get the client once, so I think you may want to move this into the migration script, somewhere at the beginning
https://github.com/yldio/asap-hub/blob/40b0c652abaaff7975822d2196795ed4573886bb/packages/cms-data-sync/src/index.ts#L6
you might also want to make sure that should something go wrong during the migration, the webhooks are still enabled back

peterstarling (changes requested):

PR #2918, opened 2023-04-03
adding noUncheckedAccess to root

jhonny-m:


apps/crn-frontend/src/shared-research/ResearchOutput.tsx
```diff
@@ -52,10 +52,12 @@ const ResearchOutput: React.FC = () => {
...
-            <TeamOutput
-              teamId={researchOutputData.teams[0]?.id}
-              researchOutputData={researchOutputData}
-            />
+            researchOutputData.teams[0]?.id && (
+              <TeamOutput
+                teamId={researchOutputData.teams[0].id}
+                researchOutputData={researchOutputData}
+              />
+            )
```
markcsims, line 60: If `researchOutputData.teams[0]?.id` aren't we going to get undefined displayed in the document?

packages/gp2-components/src/organisms/DashboardNews.tsx
```diff
@@ -81,6 +81,7 @@ const DashboardNews: React.FC<DashboardNewsProps> = ({ items }) => {
   const [expanded, setExpanded] = useState(false);
   const news = expanded ? items : items.slice(0, 1);
   const firstNews = items[0];
+  if (firstNews === undefined) return null;
```
markcsims, line 84: Doesn't `items[0]!` work here?

jhonny-m: we can't do the noNull assertion here

apps/crn-frontend/src/shared-research/ResearchOutput.tsx
jhonny-m: undefineds dont get rendered

jhonny-m: a lot of code would be incorrect if that would be the case 😄

markcsims (approved):

PR #2922, opened 2023-04-04
CT-8 Create events content model

gabiayako:
**Only the content model, the data provider will be done in another PR**

About app extensions: https://www.notion.so/Contentful-App-Extensions-d216b4c9ec9b4652b2f779ce532007b5

### Added App Extensions

---

##### Why?
In events a lot of fields are not editable, including the title. If I selected the option to hide the field when editing, I believe the user experience would be bad as the person who is editing would not see a lot of fields. With this app extension the person would just see the text.
##### How does it look?
Screenshot 2023-04-13 at 08 32 56

---

##### Why?
When the user changes the value of materials we need to update the equivalent updatedAt field so we can show reminders in the dashboard.
##### How does it look?
After editing some material (notes, videos, presentation) and publishing the entry, it's possible to see the date in the equivalent updatedAt field
Screenshot 2023-04-13 at 08 38 21

---

##### Why?
A nice way to deal with JSON Objects without having to create another content model.
##### How does it look?
Screenshot 2023-04-13 at 08 45 39

---

##### Why?
Although it's also a json object, I couldn't find an easy way to link references. So, for speakers, I decided to create a content model for it. But with only team/user reference link, the entry title becomes "Untitled" and it would appear like this in the event speakers section.
Screenshot 2023-04-13 at 08 47 21

##### How does it look?
With this extension, I just make the title to be `[team name]-[user name]` or just `[team name]` or `[user name]` if only one is provided.
Screenshot 2023-04-13 at 08 47 53
Screenshot 2023-04-13 at 08 48 04


.prettierignore
```diff
@@ -7,7 +7,7 @@
 /coverage
 /apps/*/coverage
 /packages/*/coverage
-
+/packages/contentful-app-extensions
```
lennym, line 10: Is this deliberate? And if so, how come?

packages/contentful/migrations/events/20230404114001-create-events-content-model.js
```diff
@@ -0,0 +1,804 @@
...
+
+  eventSpeakers
+    .createField('user')
+    .name('User')
+    .type('Link')
+    .localized(false)
+    .required(false)
+    .validations([
+      {
+        linkContentType: ['externalAuthors'],
```
gabiayako, line 784: reminder to myself: rebase to https://github.com/yldio/asap-hub/pull/2862 and add `users` here

packages/contentful-app-extensions/disabled-fields/src/locations/Field.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+      startAutoResizer: jest.fn(),
+    },
+  }),
+}));
+
+describe('Field component', () => {
+  it('displays field text', () => {
+    const { getByText } = render(<Field />);
+
+    expect(getByText('Hello World')).toBeInTheDocument();
```
peterstarling, line 20: wouldn't that be in the document even if it was a regular input? I don't think I care as much but in case you do maybe you should look for the <p>

peterstarling (changes requested):
I've not really focused much on the FE testing bit - you might want to ask for another review for that

.github/workflows/reusable-contentful-extensions-deploy.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: ${{ inputs.environment-name }}
+      - name: Install app packages
+        run: npm run contentful-apps:install
```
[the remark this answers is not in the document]
gabiayako: I'm not sure if it's possible to run a npm script using yarn 🤔 

I've tried here and it didn't work. Having the command install in `packages/contentful-app-extensions/package.json`, when I run 
```
yarn workspace @asap-hub/contentful-app-extensions install
```
it completely ignores the script. I tried to google it to see if people try to run npm scripts with yarn but I couldn't find any reference


So I tried to run this 
```
npm run install --workspace 'packages/contentful-app-extensions/'
```
instead but this breaks the tests https://github.com/yldio/asap-hub/actions/runs/4746054923/jobs/8429207483#step:4:4428

Do you think is too bad to leave as it was?

packages/contentful-app-extensions/disabled-fields/src/locations/Field.spec.tsx
gabiayako: Yeah, I was not sure how to test that, it's just a simple paragraph that renders what it gets from `field.getValue`. In here `maybe you should look for the` what would be the suggestion? :)

peterstarling (approved):
just a minor comment - I would recommend asking @markcsims or @kmaid for a review since this is affecting our build process

.github/workflows/reusable-test.yml
```diff
@@ -119,6 +119,25 @@ jobs:
...
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: ${{ inputs.environment-name }}
+      - name: Run disabled-fields tests
```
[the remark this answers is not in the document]
gabiayako: Yeah, I forgot to rename it 😅 I'll do that now

gabiayako:
I'm going to squash the commits, rebase to master and force push

.gitignore
```diff
@@ -23,6 +23,7 @@ node_modules
 # production
 /apps/*/build*
 /packages/*/build*
+/packages/**/build*
```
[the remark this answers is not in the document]
kmaid: What build directories do we want to keep?

packages/contentful-app-extensions/event-additional-materials/src/locations/Field.spec.tsx
```diff
@@ -0,0 +1,196 @@
+import React from 'react';
```
kmaid, line 1: Not sure im keen on introducing a third convention on test file naming. In FE we have __test__ in be we duplicate the folder structure in a parralel directory. Now we are adding same directory under a *spec* filename?

package.json
```diff
@@ -23,7 +23,7 @@
     "algolia:set-settings": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' yarn ts-node --transpile-only apps/asap-cli/src/cli.ts algolia:set-settings",
     "build": "yarn build:typecheck && yarn build:babel && yarn typecheck && yarn build:frontend",
     "build:babel": "yarn workspaces foreach -vp -j 2 run build:babel",
-    "build:frontend": "yarn workspaces foreach -j 2 --exclude asap-hub -vp run build",
```
kmaid, line 26: How are contentful apps built?

.github/workflows/reusable-contentful-extensions-deploy.yml
```diff
@@ -0,0 +1,61 @@
...
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: ${{ inputs.environment-name }}
+      - name: Build apps
```
kmaid, line 31: So we build in the deploy process?

.github/environment/Base
```diff
@@ -4,7 +4,12 @@ aws-region=us-east-1
 cache-prefix-definitions=definitions-ouput
 cache-prefix-frontend=frontend-output
 cache-prefix-transpile=transpile-output
+contentful-app-disabled-fields=2finDNk15g5UtOq4DaLNxv
+contentful-app-event-additional-materials=3TfrYX8zVOgFtKxro89UVA
+contentful-app-event-speakers-title=6ZkXISzhv1b7jjgyaK2piv
+contentful-app-field-as-updated-at=mqvX9KU5AthRTlnIRhNNh
 contentful-environment=Production
```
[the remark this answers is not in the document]
gabiayako: They are generic to all environments :) The app id is the same across all CI envs (Branch, Development and Production) and the same across all spaces in contentful.

.github/workflows/on-push-app-extensions.yml
```diff
@@ -0,0 +1,92 @@
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-environment
+        with:
+          environment-name: Branch
```
gabiayako, line 61: The values are in `Base` so they are the same for every environment, I added one hardcoded because I didn't wanted to make a different file for branch or dev/prod. Is there a cleaner way or is this good enough? :)

kmaid: You could just hardcode this to prod or even a different filename specific for this deployment?

.github/workflows/on-push-app-extensions.yml
```diff
@@ -0,0 +1,92 @@
+name: Pipeline Contentful App Extensions
+
+on:
+  push:
```
kmaid, line 4: Does this mean we will deploy to prod even on PR branches?

gabiayako: Oh, that's a good point. At first I thought about deploying even in PR branches. The app id is unique so it would deploy to prod in contentful. But yeah, I think it's better to limit only to master

.github/workflows/on-push-app-extensions.yml
gabiayako: I'm going to change to `Production` because I'll change this to run only on `master`

packages/contentful-app-extensions/disabled-fields/src/setupTests.ts
```diff
@@ -0,0 +1,10 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
+import { configure } from '@testing-library/react';
+
+configure({
+  testIdAttribute: 'data-test-id',
+});
```
markcsims, line 10: Is this used, they advise not using testid

packages/contentful-app-extensions/event-additional-materials/src/setupTests.ts
markcsims, line 1: Think this can be deleted too

packages/contentful-app-extensions/event-additional-materials/src/test/locations/Field.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+  });
+
+  it('renders only input fields and add button initially when there is not any previous additional material', () => {
+    (useSDK as jest.Mock).mockReturnValue({
+      ...mockBaseSdk,
+      field: {
+        getValue: jest.fn(() => []),
+      },
+    });
+    const { queryByRole, getByLabelText, container } = render(<Field />);
```
markcsims, line 49: Prefer screen rather than us the return values from render

packages/contentful-app-extensions/event-additional-materials/src/test/locations/Field.test.tsx
markcsims, line 1: Prefer screen rather than us the return values from render

packages/contentful-app-extensions/event-speakers-title/src/setupTests.ts
markcsims, line 1: we can delete this?

packages/contentful-app-extensions/event-speakers-title/src/test/locations/Field.test.tsx
markcsims, line 1: Prefer screen rather than us the return values from render

packages/contentful-app-extensions/field-as-updated-at/src/setupTests.ts
markcsims, line 1: delete this?

packages/contentful-app-extensions/field-as-updated-at/src/test/locations/Field.test.tsx
markcsims, line 1: Prefer screen rather than us the return values from render

packages/contentful-app-extensions/field-as-updated-at/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "isolatedModules": true,
+    "jsx": "react",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noFallthroughCasesInSwitch": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+
+    "target": "es5",
```
markcsims, line 18: Is this the right target for these extensions? That's from 2009?

gabiayako: I've changed to es2015 now https://github.com/yldio/asap-hub/pull/2922/commits/3c135c50ffd0cf3d0bde6ac4e29685fefa23c013

kmaid (approved):

PR #2926, opened 2023-04-04
adding success message on save and publish

jhonny-m:


markcsims (approved):
We should consider using Context or something else to create a messaging system.

PR #2927, opened 2023-04-04
GP2 339 - Fix Navigation Bar

csbarbosa:
(Ticket: https://asaphub.atlassian.net/browse/GP2-339)
Fix the spacing between the components inside the Navigation Bar, aswell as the removal of some elements.

### DIFF
--- a/packages/gp2-components/src/organisms/NavigationHeader.tsx
+++ b/packages/gp2-components/src/organisms/NavigationHeader.tsx
@@ -6,12 +6,11 @@ import {
   steel,
   pixels,
   noop,
-  Anchor,
 } from '@asap-hub/react-components';
 
 import { css } from '@emotion/react';
 import { ComponentProps } from 'react';
-import { gp2LogoSmall } from '../icons';
+import { HeaderLogo } from '../molecules';
 import MainNavigation from './MainNavigation';
 import UserNavigation from './UserNavigation';
 
@@ -76,15 +75,7 @@ const NavigationHeader: React.FC<NavigationHeaderProps> = ({
       <div css={[menuButtonStyles]}>
         <MenuButton open={menuShown} onClick={() => onToggleMenu(!menuShown)} />
       </div>
-      <Anchor
-        css={{
-          display: 'flex',
-          padding: `${rem(16)} 0`,
-        }}
-        href={'/'}
-      >
-        {gp2LogoSmall}
-      </Anchor>
+      <HeaderLogo />
       <div css={desktopNavigationStyles}>
         <MainNavigation />
       </div>
--- a/packages/gp2-components/src/organisms/NavigationHeader.tsx
+++ b/packages/gp2-components/src/organisms/NavigationHeader.tsx
@@ -6,12 +6,12 @@ import {
   steel,
   pixels,
   noop,
+  Anchor,
 } from '@asap-hub/react-components';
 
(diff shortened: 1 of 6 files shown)


packages/gp2-components/src/organisms/UserNavigation.tsx
```diff
@@ -64,18 +78,31 @@ const UserNavigation: React.FC<UserNavigationProps> = (userNavigationProps) => {
...
+        onClick={(event) => {
+          setMenuShown(!menuShown);
+          event.preventDefault();
+        }}
       >
-        <span css={buttonTextStyles}>{`Hi, ${
-          firstName ?? 'Unknown User'
-        }`}</span>
-      </UserMenuButton>
+        <span
```
dianaguimaraes, line 89: Avatar is a `div` and therefore shouldn't be inside a `span`

packages/gp2-components/src/organisms/NavigationHeader.tsx
```diff
@@ -79,7 +76,15 @@ const NavigationHeader: React.FC<NavigationHeaderProps> = ({
       <div css={[menuButtonStyles]}>
         <MenuButton open={menuShown} onClick={() => onToggleMenu(!menuShown)} />
       </div>
-      <HeaderLogo />
```
jhonny-m, line 82: why did we stop using the HeaderLogo? can't we apply the changes in the component?

csbarbosa: As far as the new designs for the NavBar, we don't need to use the gp2Logo anymore

packages/gp2-components/src/organisms/UserNavigation.tsx
```diff
@@ -64,18 +78,31 @@ const UserNavigation: React.FC<UserNavigationProps> = (userNavigationProps) => {
   }, [setMenuShown, reference]);
   return (
     <div css={css({ position: 'relative' })} ref={reference}>
-      <UserMenuButton
```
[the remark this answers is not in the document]
csbarbosa: As far as I can it's also used in CRN and the changes are still considerable.

packages/gp2-components/src/organisms/UserNavigation.tsx
```diff
@@ -86,9 +86,9 @@ const UserNavigation: React.FC<UserNavigationProps> = (userNavigationProps) => {
           event.preventDefault();
         }}
       >
-        <span
+        <div
           css={{
-            display: 'inline-flex',
```
jhonny-m: why won't the inline-flex work?

packages/gp2-components/src/organisms/UserNavigation.tsx
```diff
@@ -1,19 +1,39 @@
...
+  backgroundColor: 'unset',
+  border: 'none',
+  outline: 'none',
+});
+const styles = css({
+  padding: `${rem(12)} 0`,
+  cursor: 'pointer',
+
+  display: 'grid',
+  gridTemplateColumns: `auto ${rem(48)}em auto auto`,
```
dianaguimaraes, line 33: `${rem(48)}em`
did you mean to have that `em`?
`rem` function already adds `em` so I'm thinking it will maybe end up being `<some_value>emem`, no?

dianaguimaraes (approved):

PR #2930, opened 2023-04-06
gp2 sentry handlers

markcsims:
- adds sentry handlers to gp2
- adds permissionHandler
- removes opentracing
- removes lightstep
- sentryTransactionId


apps/gp2-server/src/routes/user.route.ts
```diff
@@ -68,6 +70,14 @@ export const userRouteFactory = (userController: UserController): Router =>
 
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const loggedInUserId = req.loggedInUser!.id;
+
+        if (
+          req.loggedInUser?.onboarded !== true &&
```
peterstarling, line 75: isn't this exactly the same what is already in permissionHandler, which is being injected here anyway?

peterstarling (changes requested):

PR #2935, opened 2023-04-06
CT-12 Migrate calendar data

gabiayako:


packages/cms-data-sync/src/index.ts
```diff
@@ -1,8 +1,33 @@
...
+
+  const contentfulSpace = await contentfulClient.getSpace(CONTENTFUL_SPACE_ID!);
+  const webhook = await contentfulSpace.getWebhook(
+    `${CONTENTFUL_ENV_ID!.toLowerCase()}-webhook`,
+  );
+
+  webhook.active = false;
+  await webhook.update();
+
   await migrateTeams();
```
peterstarling, line 27: i don't think this file is tested but it might be worth adding some tests now since the logic is becoming a little bit more complex... 

anyway the reason i bring it up is because I think in the previous review of this I asked what would happen if one of the migrations failed after you disabled a webhook - looking at the code I think it would never activate the webhook back, but maybe you decided not do do it because technically the individual migrations should not throw - I still think something could go wrong and we definitely don't want to leave those webhooks inactive so you may want to add something here for that.

lennym (approved):

peterstarling (changes requested):

packages/cms-data-sync/src/calendars/calendars.queries.ts
```diff
@@ -0,0 +1,14 @@
+import { gql } from 'graphql-tag';
+
+export const calendarsQuery = gql`
+  query FetchCalendars {
+    queryCalendarsContents(top: 100) {
```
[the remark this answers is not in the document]
gabiayako: Yeah, there's 25 in prod
Screenshot 2023-04-10 at 12 12 25

packages/cms-data-sync/src/index.ts
gabiayako: Sorry, I've missed the idea of the other comment. I've added a try catch finally block now

packages/cms-data-sync/test/index.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  afterAll(() => {
+    console.log = consoleLogRef;
+  });
+
+  it('deactivates webhook and activates it again after running the migrations', async () => {
+    await runMigrations();
+
+    expect(spaceMock.getWebhook).toHaveBeenCalledWith('crn-env-id-webhook');
+    expect(webhookMock.update).toHaveBeenCalled();
+    expect(console.log).toHaveBeenNthCalledWith(
```
peterstarling, line 77: I take it this is your way of testing that the webhook was deactivated as technically we are unable to assert that webhookMock.active was set to false since immediately after running it is changed to true? I think we should try and test that better, even if this is a throwaway code. 

One way would be to create setter function on `webhookMock` here https://github.com/yldio/asap-hub/blob/06a948055b7c06d252691911f8a9a430a1027fe7/packages/cms-data-sync/test/index.test.ts#L49

That way you could set some global variable once active is set to false and then assert that that global var was set correctly, what do you think?

gabiayako: added here [review: setter active test](https://github.com/yldio/asap-hub/pull/2935/commits/ea92fda801e831b8592bfaa76fdcf42335361f8c)

packages/cms-data-sync/test/index.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    migrateCalendars: mockMigrateCalendars,
+  };
+});
+
+const mockContentfulManagement = contentfulManagement as jest.Mocked<
+  typeof contentfulManagement
+>;
+
+const webhookMock = { update: jest.fn(), active: true };
+const spaceMock = {
```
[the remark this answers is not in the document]
gabiayako: added here [review: add jest.Mocked to mocks](https://github.com/yldio/asap-hub/pull/2935/commits/4b946bb42481e947519470f4d93b1c966148976c)

packages/cms-data-sync/src/index.ts
```diff
@@ -1,8 +1,46 @@
...
+    await webhook.update();
+
+    logger('Webhook deactivated');
+
+    try {
+      await migrateTeams();
+      await migrateExternalAuthors();
+      await migrateCalendars();
+    } catch {
+      logger('Error migrating data', 'ERROR');
```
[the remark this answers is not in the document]
gabiayako: added here [review: throw error in finally](https://github.com/yldio/asap-hub/pull/2935/commits/a107c2ecba3223d0e0d6246c52403907daead17e)

peterstarling (changes requested):

peterstarling (approved):

PR #2944, opened 2023-04-11
refactor dashboard to use notification context

jhonny-m:


apps/gp2-frontend/src/NotificationMessages.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const addNotification = useCallback(
+    (notification: Notification) => {
+      setNotifications((prev) => [...prev, notification]);
+    },
+    [setNotifications],
+  );
+  const removeNotification = useCallback(
+    (notification: Notification) => {
+      setNotifications((prev) =>
+        prev.filter((n) => n.message !== notification.message),
```
markcsims, line 15: Should we check the page here too? For example the same message is displayed on the dashboard and outputs.

packages/gp2-components/src/templates/PageNotifications.tsx
```diff
@@ -0,0 +1,46 @@
...
+  page,
+  children,
+}) => {
+  const { notifications, removeNotification } = useNotificationContext();
+  const pageNotifications = notifications.filter(
+    (notification) => notification.page === page,
+  );
+  const displayNotification = pageNotifications[0];
+  useEffect(
+    () => () => displayNotification && removeNotification(displayNotification),
```
markcsims, line 24: Should this be run as part of the clean up code?

apps/gp2-frontend/src/AuthenticatedApp.tsx
```diff
@@ -35,22 +40,33 @@ const AuthenticatedApp: FC<Record<string, never>> = () => {
...
     user?.onboarded || false,
   );
+
+  const welcomeBackMessage = `Welcome back to the GP2 Hub${
+    user?.firstName ? `, ${user.firstName}` : ''
+  }!`;
+
+  useEffect(() => {
+    if (showWelcomeBackBanner) {
+      addNotification({
```
markcsims, line 55: Is this a hook? If so, it shouldn't be called conditionally?

jhonny-m: the hook is `useNotificationContext`, the context has notifications plus functions to add/remove them

packages/gp2-components/src/templates/PageNotifications.tsx
jhonny-m: when the components gets dismounted, it cleans up the notification associated to the page,
I thought this was a simple way to dismiss it when we change the route without having to add a listener to the route

apps/gp2-frontend/src/NotificationMessages.tsx
jhonny-m: that's a valid point, I can create a function to validate the notifications

markcsims (approved):

PR #2945, opened 2023-04-11
Crn 1305 shared output publishing a draft output

oratiu:


apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -500,7 +500,7 @@ const getEventMaterialsRemindersFromQuery = (
 
 const getSortDate = (reminder: ReminderDataObject): DateTime => {
   if (reminder.entity === 'Research Output') {
-    return DateTime.fromISO(reminder.data.addedDate);
+    return DateTime.fromISO(reminder.data.addedDate || '');
```
oratiu, line 503: I'm unsure how reminders handle draft research outputs, but I think we need some new type to split the ResearchOutput into DraftResearchOutput (which has not added date) and PublishedResearchOutput (which has added date).

packages/react-components/src/organisms/RecentSharedOutputs.tsx
```diff
@@ -118,9 +118,11 @@ const RecentSharedOutputs: React.FC<RecentSharedOutputProp> = ({ outputs }) => (
               {getIconForDocumentType(documentType)} {documentType}
             </p>
             <span css={[titleStyles, rowTitleStyles]}>Date</span>
-            <span>
-              {formatDateToTimezone(addedDate, 'E, d MMM y').toUpperCase()}
-            </span>
+            {addedDate && (
```
oratiu, line 121: Here I know that all the "Recent Shared outputs" are Published ones that we are displaying on the dashboard, and with the new type I could avoid doing this check.

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -134,7 +136,11 @@ export const researchOutputRouteFactory = (
 
       // TODO: update the published value in hasEditResearchOutputPermission in
       // display draft task. Currently we are not sending the value published
-      if (!loggedInUser || !hasEditResearchOutputPermission(userRole, false)) {
+
+      if (
+        !loggedInUser ||
+        !hasEditResearchOutputPermission(userRole, publish)
```
gabiayako, line 142: This needs to be the value of `published` to check if the research output is published or not 😉 For example, if the RO is already published and the user is just a member, `hasEditResearchOutputPermission` should give false. 
The `publish` that it's being passed here now is to publish the RO. I think we're not sending this `published` value to the BE yet, but we need to pass it so we can call this `hasEditResearchOutputPermission` :)

```suggestion
        !hasEditResearchOutputPermission(userRole, published)
```

apps/crn-server/src/data-providers/reminders.data-provider.ts
oratiu: I think we should discuss this at the RTM @peterstarling, I feel that since we have a lot of places where we only get published ROs and other places where we only get drafts, and there is a way to diff between them (one having an added date and the other not having it) it might warrant some discussion.

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -280,6 +285,10 @@ export class ResearchOutputSquidexDataProvider
       updatedBy: [researchOutputData.updatedBy],
     });
 
+    if (updateOptions.publish) {
```
Riccardo-Brown, line 288: Should this be after patch call, so we update the research output then publish?

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -121,20 +122,30 @@ export const researchOutputRouteFactory = (
   researchOutputRoutes.put(
     '/research-outputs/:researchOutputId',
     async (req, res) => {
-      const { body, params, loggedInUser } = req;
+      const { body, params, loggedInUser, query } = req;
       const { researchOutputId } = validateResearchOutputParameters(params);
       const updateRequest = validateResearchOutputPutRequestParameters(body);
       validateResearchOutputPostRequestParametersIdentifiers(body);
+      const options = validateResearchOutputRequestQueryParameters(query);
+      const publish = options.publish ?? true;
```
Riccardo-Brown, line 130: Should the default be false so that we don't publish be mistake?

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -588,8 +614,81 @@ describe('ResearchOutputs controller', () => {
...
+
+        const result = await researchOutputs.update(
+          researchOutputId,
+          researchOutputUpdateData,
+          { publish: false },
+        );
+
+        expect(result).toEqual(getResearchOutputResponse());
+
+        const researchOutputUpdateDataObject =
```
Riccardo-Brown, line 649: same here

apps/crn-server/src/routes/research-outputs.route.ts
oratiu: I based this off the post request which is publish true by default, I get your point but then I'd say we should probably change it there as well and in the controllers and data providers as well. I think that would warrant another ticket.

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -571,7 +568,35 @@ describe('ResearchOutputs controller', () => {
...
+
+      const result = await researchOutputs.update(
+        researchOutputId,
+        researchOutputUpdateData,
+        { publish: false },
+      );
+
+      expect(result).toEqual(getResearchOutputResponse());
+
+      const researchOutputUpdateDataObject =
```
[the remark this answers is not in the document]
oratiu: Good point

packages/squidex/test/publish.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    await expect(() => client.publish('42')).rejects.toThrow(GenericError);
+  });
+
+  it('return GenericError on HTTP Error', async () => {
+    nock(baseUrl)
+      .put(`/api/content/${appName}/${collection}/42/status`)
+      .reply(500);
+
+    await expect(() => client.publish('42')).rejects.toThrow(GenericError);
+  });
```
Riccardo-Brown, line 58: why is this test repeated?

apps/crn-frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -57,7 +57,11 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
 
   const createResearchOutput = usePostResearchOutput({ publish: true });
   const createDraftResearchOutput = usePostResearchOutput({ publish: false });
-  const updateResearchOutput = usePutResearchOutput();
+
+  const updateResearchOutput = usePutResearchOutput({
+    publish: !researchOutputData?.published,
```
kmaid, line 62: This doesn't seem like it will work. Won't edits of unpublised RO make the RO unpublished?

apps/crn-frontend/src/network/teams/api.ts
```diff
@@ -117,9 +117,10 @@ export const updateTeamResearchOutput = async (
   researchOutputId: string,
   researchOutput: ResearchOutputPostRequest,
   authorization: string,
+  publish: boolean = false,
 ): Promise<ResearchOutputResponse> => {
   const resp = await fetch(
-    `${API_BASE_URL}/research-outputs/${researchOutputId}`,
+    `${API_BASE_URL}/research-outputs/${researchOutputId}?publish=${publish}`,
```
kmaid, line 123: It is weird that this is a get parameter. Why wouldn't it just be setting `published` to true in the body?

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
```diff
@@ -51,7 +51,11 @@ const WorkingGroupOutput: React.FC<WorkingGroupOutputProps> = ({
 
   const createResearchOutput = usePostResearchOutput({ publish: true });
   const createDraftResearchOutput = usePostResearchOutput({ publish: false });
-  const updateResearchOutput = usePutResearchOutput();
+
+  const updateResearchOutput = usePutResearchOutput({
+    publish: !researchOutputData?.published,
```
kmaid, line 56: As above

apps/crn-frontend/src/shared-research/ResearchOutput.tsx
```diff
@@ -34,9 +44,13 @@ const ResearchOutput: React.FC = () => {
   if (researchOutputData) {
     return (
       <ResearchOutputPermissionsContext.Provider value={permissions}>
-        <Route exact path={path}>
+        <Route exact path={publishedNow ? publishedNowPath : path}>
```
kmaid, line 47: Can't you just use an array on path?

apps/crn-frontend/src/network/teams/api.ts
peterstarling: to be fair i thought the same when i saw this - why not just use the `published` property of the ResearchOutput? Or whatever it is that you would use to distinguish the published research from a draft one.

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -571,7 +568,34 @@ describe('ResearchOutputs controller', () => {
...
+      expect(researchOutputDataProviderMock.update).toBeCalledWith(
+        researchOutputId,
+        researchOutputUpdateDataObject,
+        { publish: false },
+      );
+    });
+
+    test('Should update a draft research output and return it', async () => {
+      const mockDate = new Date('2010-01-01');
+      const spy = jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
```
peterstarling, line 598: can't you just use jest mock timers method fake time? I realise this was probably copied from before but I think it would be a good time to look into a better alternative

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -571,7 +568,34 @@ describe('ResearchOutputs controller', () => {
...
+      );
+
+      const researchOutputUpdateDataObject =
+        getResearchOutputUpdateDataObject();
+
+      expect(result).toEqual(getResearchOutputResponse());
+      expect(researchOutputDataProviderMock.update).toBeCalledWith(
+        researchOutputId,
+        researchOutputUpdateDataObject,
+        { publish: false },
```
peterstarling, line 592: to be fair i am in two minds about this being a separate parameter. Like I mentioned earlier in this review (and I think so did @kmaid) it feels like we could just infer the actual state of the research-output by simply looking at the `published` property rather than have this separate 'option'

peterstarling (changes requested):

packages/react-components/src/organisms/RecentSharedOutputs.tsx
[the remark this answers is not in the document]
oratiu: Spoke with design, using created as fallback, test added

apps/crn-frontend/src/shared-research/ResearchOutput.tsx
oratiu: Issue is that publishedNow is seen as undefined.

apps/crn-server/src/data-providers/research-outputs.data-provider.ts
```diff
@@ -288,6 +290,10 @@ export class ResearchOutputSquidexDataProvider
       labs: researchOutput.labs,
       teams: researchOutput.teams,
     });
+
+    if (updateOptions.publish) {
```
peterstarling, line 294: Check out my comment in the controller tests - to me this line can stay as is and the only difference I would make is that I would tell the controller to always pass the `publish` option as true when the status is set to `published` - that way we would be calling the rest client `publish` method on every update - so the question is whether squidex might like it? If it doesn't - then I guess you could catch the error response although I would expect a 201

apps/crn-server/src/controllers/research-outputs.ts
peterstarling, line 1: can you add a condition to only enable turning drafts into published but not the other way round?

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -126,15 +123,22 @@ export const researchOutputRouteFactory = (
...
-      // TODO: update the published value in hasEditResearchOutputPermission in
-      // display draft task. Currently we are not sending the value published
-      if (!loggedInUser || !hasEditResearchOutputPermission(userRole, false)) {
+      const result = await researchOutputController.fetchById(researchOutputId);
+      const publish = updateRequest.published && !result.published;
+
+      if (
+        !loggedInUser ||
+        !hasEditResearchOutputPermission(userRole, result.published) ||
+        (publish && !hasPublishResearchOutputPermission(userRole))
```
peterstarling, line 140: I don't think you need this at all - the `hasEditResearchOutputPermission` seems to cover it

apps/crn-server/src/routes/research-outputs.route.ts
```diff
@@ -126,15 +123,22 @@ export const researchOutputRouteFactory = (
...
-        updateRequest.teams,
+        workingGroupOutput ? 'workingGroups' : 'teams',
+        workingGroupOutput ? updateRequest.workingGroups : updateRequest.teams,
       );
 
-      // TODO: update the published value in hasEditResearchOutputPermission in
-      // display draft task. Currently we are not sending the value published
-      if (!loggedInUser || !hasEditResearchOutputPermission(userRole, false)) {
+      const result = await researchOutputController.fetchById(researchOutputId);
+      const publish = updateRequest.published && !result.published;
```
peterstarling, line 135: this is unnecessary - i think you don't need to throw only when the status changes - just check if they have edit permissions - see below as well

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -959,6 +959,29 @@ describe('ResearchOutputs data provider', () => {
       test('Should update the existing research-output', async () => {
```
peterstarling, line 959: can you say that we are updating a draft

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -959,6 +959,29 @@ describe('ResearchOutputs data provider', () => {
...
+        await expect(
+          researchOutputDataProvider.update(
+            researchOutputId,
+            researchOutputUpdateData,
+            { publish: false },
+          ),
+        ).resolves.not.toThrow();
+      });
+
+      test('Should default to not publish the existing research-output', async () => {
```
peterstarling, line 982: i dont think this is valid

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -978,6 +1001,39 @@ describe('ResearchOutputs data provider', () => {
         ).resolves.not.toThrow();
       });
 
+      test('Should publish a draft research-output', async () => {
```
peterstarling, line 1004: why dont you create a separaete describe block where you test:
1. Sending the status option as Published
2. Sending the status option as Draft
3. Not sending it at all

In 3) we should show that the request to change the status does not happen

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -767,8 +709,16 @@ describe('/research-outputs/ route', () => {
...
       ...researchOutputPutRequest
     } = getResearchOutputPutRequest();
+    beforeEach(() => {
+      researchOutputControllerMock.fetchById.mockResolvedValue(
+        researchOutputResponse,
+      );
+    });
 
     test('Should send the data to the controller and return status 200 along with all the research output data', async () => {
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce(
```
peterstarling, line 719: should not happen anymore

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -796,6 +746,84 @@ describe('/research-outputs/ route', () => {
       expect(response.status).toBe(403);
     });
 
+    test('Should return 403 when user is not permitted to publish a research output', async () => {
```
peterstarling, line 749: we should say it is because you don't have a sufficient role

apps/crn-server/test/routes/research-outputs.route.test.ts
```diff
@@ -796,6 +746,84 @@ describe('/research-outputs/ route', () => {
...
+        teams: [{ id: 'team-1', displayName: 'team-1', role: 'Key Personnel' }],
+      });
+      researchOutputControllerMock.fetchById.mockResolvedValueOnce({
+        ...researchOutputResponse,
+        teams: [{ id: 'team-1', displayName: 'team-1' }],
+        published: false,
+      });
+      const response = await supertest(app)
+        .put('/research-outputs/abc123')
+        .send({ ...researchOutputPutRequest, teams: ['team-1'] });
```
peterstarling, line 761: be explicit - make sure that the `published: true` is set

apps/crn-server/test/routes/research-outputs.route.test.ts
peterstarling: we also probably want a test that is exactly like this one except we would be sending a `published: false` property and then it should be a 200

packages/model/src/reminder.ts
```diff
@@ -46,7 +46,7 @@ export interface ResearchOutputPublishedReminder
     researchOutputId: ResearchOutputDataObject['id'];
     documentType: ResearchOutputDataObject['documentType'];
     title: ResearchOutputDataObject['title'];
-    addedDate: ResearchOutputDataObject['addedDate'];
+    addedDate: string;
```
peterstarling, line 49: ?

peterstarling: this is valid but better go with 
```suggestion
    addedDate: NonNullable<ResearchOutputDataObject['addedDate']>;
```

peterstarling: or just use the PublishedResearchOutputDataObject

packages/model/src/research-output.ts
```diff
@@ -229,6 +229,14 @@ export type ResearchOutputDataObject = ResearchOutputCoreObject & {
   >;
 };
 
+export type DraftResearchOutputDataObject = Omit<
```
peterstarling, line 232: maybe do `ResearchOutputDraftDataObject`

peterstarling (changes requested):

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -139,6 +141,13 @@ export default class ResearchOutputs implements ResearchOutputController {
       );
     }
 
+    if (
+      currentResearchOutput.published &&
+      !researchOutputUpdateData.published
+    ) {
+      throw Boom.badRequest();
```
peterstarling, line 148: some message would be nice, something along the lines of 'cannot unpublish a research output' or whatever

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -596,24 +561,185 @@ describe('ResearchOutputs controller', () => {
...
+      researchOutputDataProviderMock.update.mockResolvedValueOnce(
+        researchOutputId,
+      );
+
+      await expect(
+        researchOutputs.update(researchOutputId, {
+          ...getResearchOutputUpdateData(),
+          published: false,
+        }),
+      ).rejects.toThrow(Error);
```
peterstarling, line 578: and this is where i think you should assert that the right error was thrown.

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -596,24 +561,185 @@ describe('ResearchOutputs controller', () => {
...
+    });
+
+    test('Should update a published existing research output and return it', async () => {
```
peterstarling, line 581: and how to we know that the existing research output is published? Is it that the default fixture has it? well i would expect to be explicit - use the `researchOutputDataProviderMock.fetchById.mockResolvedValue` same way you do above

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -596,24 +561,185 @@ describe('ResearchOutputs controller', () => {
...
+          {
+            ...getResearchOutputUpdateDataObject(),
+            addedDate: undefined,
+          },
+          { publish: false },
+        );
+      });
+      test('Should be created when publishing a draft', async () => {
+        const mockDate = new Date('2010-01-01');
+        const spy = jest
```
peterstarling, line 689: isn't it better to use `useFakeTimers`?

peterstarling (changes requested):

apps/crn-frontend/src/shared-research/ResearchOutput.tsx
```diff
@@ -13,6 +13,16 @@ const ResearchOutput: React.FC = () => {
   const { researchOutputId } = useRouteParams(
     sharedResearch({}).researchOutput,
   );
+
+  const publishedNowPath = useRouteMatch({
+    path: `${
+      sharedResearch({}).researchOutput({ researchOutputId }).$
+    }/publishedNow`,
```
kmaid, line 20: Shouldn't this route be with the other ones? I also wonder if it should have been a query parameter rather than a different route

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -89,7 +88,9 @@ export default class ResearchOutputs implements ResearchOutputController {
         researchOutputCreateData.authors ?? [],
       ),
       accession: researchOutputCreateData.accession,
-      addedDate: new Date(Date.now()).toISOString(),
+      addedDate: researchOutputCreateData.published
```
kmaid, line 91: Won't this reset the added date every time the RO is saved?

kmaid (approved):

apps/crn-server/src/controllers/research-outputs.ts
oratiu: Given it's the post method, when we first create the RO it will either set the addedDate or undefined (if it's not published/ is a draft)

apps/crn-frontend/src/shared-research/ResearchOutput.tsx
oratiu: It is with the other, corrected now.

peterstarling (approved):

PR #2953, opened 2023-04-13
Improve Algolia search ranking & results

kmaid:
This PR is to implement the website side of https://www.algolia.com/doc/guides/sending-events/connectors/google-tag-manager/

Algolia Client and GTM are both tagged with UserId so queries can be related to a user.
Result lists rendered from algolia data should be tagged with the index they were run on

- Events
- Users
- Research Outputs

Hits should be tagged with QueryId, Position and Object ID which is the same as our ID's 
I have written very high level unit tests to ensure tags are present on the relevant result lists to prevent regressions.

I have also refactored the algolia scripts to be within the algolia package.

### DIFF
--- a/apps/asap-cli/package.json
+++ b/apps/asap-cli/package.json
@@ -7,6 +7,12 @@
     "node": "18.x"
   },
   "scripts": {
+    "algolia:delete-index": "ts-node src/cli.ts algolia:delete-index",
+    "algolia:clear-index": "ts-node src/cli.ts algolia:clear-index",
+    "algolia:get-settings": "ts-node src/cli.ts algolia:get-settings",
+    "algolia:move-index": "ts-node src/cli.ts algolia:move-index",
+    "algolia:remove-records": "ts-node src/cli.ts algolia:remove-records",
+    "algolia:set-settings": "ts-node src/cli.ts algolia:set-settings",
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
     "typecheck": "tsc"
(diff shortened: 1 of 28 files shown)


apps/asap-cli/package.json
```diff
@@ -7,6 +7,12 @@
     "node": "18.x"
   },
   "scripts": {
+    "algolia:delete-index": "ts-node --transpile-only src/cli.ts algolia:delete-index",
```
gabiayako, line 10: isn't `TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}` necessary in some scripts like it was in the root package.json?

packages/react-components/src/templates/SharedResearchList.tsx
```diff
@@ -18,18 +19,26 @@ type SharedResearchListProps = Omit<
...
+} & Pick<ComponentProps<typeof AlgoliaHit>, 'algoliaQueryId'>;
 
 const SharedResearchList: React.FC<SharedResearchListProps> = ({
   researchOutputs,
+  algoliaQueryId,
   ...cardListProps
 }) => (
   <ResultList {...cardListProps}>
     {cardListProps.isListView ? (
       <SharedResearchListCard researchOutputs={researchOutputs} />
```
gabiayako, line 31: isn't `AlgoliaHit` necessary here as well?

gabiayako (approved):

apps/asap-cli/package.json
kmaid: No, we moved it into tsconfig in the TS_Node section for the entire app

packages/react-components/src/templates/SharedResearchList.tsx
kmaid: Good catch.

PR #2960, opened 2023-04-14
CT-32 User Migration

lennym:
* Migrate users and labs, including users' team memberships and avatars
* Add a created date to user model to use in place of `firstPublishedAt` when showing the date a user joined the hub because the built-in metadata cannot be backdated when migrating
* Add an adapter for the contentful management API which rate-limits requests to avoid crashing when Contentful responds with a 429 response
* Recursively fetch paginated data to handle datasets which exceed the maximum page size


apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -576,6 +582,16 @@ describe('User data provider', () => {
         },
       });
     });
+
+    test('sets `createdDate` when a user is onboarded', async () => {
```
peterstarling, line 586: what if they were already onboarded? shouldn't that be tested and handled?

packages/cms-data-sync/src/labs/labs.data-migration.ts
```diff
@@ -0,0 +1,62 @@
...
+export const migrateLabs = async () => {
+  const { contentfulEnvironment, squidexGraphqlClient } =
+    await getSquidexAndContentfulClients();
+
+  const migrateFromSquidexToContentful = migrateFromSquidexToContentfulFactory(
+    contentfulEnvironment,
+    logger,
+  );
+
+  const fetchAllData = async (skip: number = 0): Promise<LabItem[]> => {
```
peterstarling, line 24: seems a little unnecessary considered we only have a few of those, especially since we don't paginate through teams for example, but fair enough. 

Why haven't you used to helper function you created though?

peterstarling (changes requested):

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
lennym: Not sure I follow. Anyone already onboarded would be the users migrated from Squidex, which have this data point copied over. This is for new people joining the hub in future.

There is the edge case for people who have accounts created manually in the CMS with onboarding skipped, but for those the fallback to `firstPublishedAt` would suffice. Do you mean I should add a test case for that fallback?

packages/cms-data-sync/src/labs/labs.data-migration.ts
lennym: I wasn't sure what the numbers in prod were like, but if it was up to me then all the content types would be resilient to pagination overflow.

Which helper function do you mean? There is one for fetching from Contentful, which I could re-purpose into a generic paginated fetch method.

packages/cms-data-sync/src/labs/labs.queries.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const labsQuery = gql`
+  query FetchLabs($take: Int, $skip: Int) {
+    queryLabsContentsWithTotal(top: $take, skip: $skip) {
+      total
+      items {
+        id
+        created
+        lastModified
+        version
```
gabiayako, line 11: we're not using these fields :)

gabiayako (approved):

packages/cms-data-sync/src/labs/labs.data-migration.ts
lennym: I had a look in production squidex and there are 258 labs listed, so these will need fetching in a paginated way.

There are also a little over 1,300 external authors, so these will also need to be refactored to be paginated.

lennym: I've extracted a paginated fetch helper that can be repurposed across contentful and squidex queries, and applied it to external authors, I just need to rewrite the unit tests.

PR #2964, opened 2023-04-17
CRN-610 - enforce URL regex patterns on the schema defs

andreicovaciu:


apps/crn-server/src/migrations/1681729305-enforce-url-format.ts
```diff
@@ -0,0 +1,238 @@
...
+  RestTeam,
+  RestUser,
+} from '@asap-hub/squidex';
+import { Migration } from '@asap-hub/server-common';
+import { applyToAllItemsInCollection } from '../utils/migrations';
+
+export default class EnforceUrlFormat extends Migration {
+  isPatchingEnabled = false;
+
+  cleanupUrl = (url?: string) => {
```
peterstarling, line 15: maybe extract this function to a helper function and test it?

I would also probably extend it with the http/https adding logic, so in the end the function would take any string, trim, conver to lowercase and add http or https.

peterstarling (changes requested):

apps/crn-server/src/migrations/1681729305-enforce-url-format.ts
```diff
@@ -0,0 +1,238 @@
...
+    if (!url) {
+      return null;
+    }
+    // trim whitespaces and lowercase the http(s) protocol.
+    // Do not lowercase the rest of the url as it might contain query parameters that are case-sensitive (e.g. password of a zoom meeting).
+    const cleanedUrl = url
+      .trim()
+      .replace('HTTP://', 'http://')
```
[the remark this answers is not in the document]
andreicovaciu: There are link that looks like this: 
```https://zoom.us/j/84474388259?pwd=RFFvV2NTOHEyMkM5b3h3b0QwWll5QT09```
and the password query param is case-sensitive

apps/crn-server/src/migrations/1681729305-enforce-url-format.ts
andreicovaciu: As discussed we'll merge as it is as we need the two step implementation

peterstarling (approved):

PR #2986, opened 2023-04-20
feat: CT-2 Split Lab Controller

peterstarling:
- wip
- add lab data provider
- use data provider in controller and export


lennym:
Should the data provider have a Contentful counterpart?

lennym (approved):

PR #2989, opened 2023-04-21
adding country codes dropdown to modal

jhonny-m:


apps/gp2-frontend/src/users/country-codes-suggestions.ts
```diff
@@ -0,0 +1,1213 @@
+export const countryCodesSuggestions = [
+  {
+    name: 'Afghanistan',
+    dialCode: '+93',
+    code: 'AF',
```
markcsims, line 5: Do we need the code?

markcsims (approved):

apps/gp2-frontend/src/users/country-codes-suggestions.ts
jhonny-m: we don't use the codes
I can clean them up

apps/gp2-frontend/src/users/country-codes-suggestions.ts
dianaguimaraes, line 1: should this be moved to the components instead and avoid passing it on all `ContactInformationModal` components which seems to be what we're doing?

packages/react-components/src/molecules/LabeledDropdown.tsx
```diff
@@ -52,7 +56,9 @@ export default function LabeledDropdown<V extends string>({
             </span>
           )}
           <br />
```
dianaguimaraes, line 58: do we want the `br` even if there is no description?

apps/gp2-frontend/src/users/country-codes-suggestions.ts
jhonny-m: I just kept the behaviour of what has been done previously

PR #3002, opened 2023-04-26
Add support for Contentful invitation/onboarding journey

lennym:
* Updates invite handler factory function to use a data provider instance in place of the Squidex client to allow data provider agnostic invitation pipelines.
* Updates webhook handlers to use `resourceId` for accessing entity IDs
* Updates CRN user data provider to support setting connection codes
* Updates invite handler to never send invitation if there is any connection code attached to the profile to prevent repeat invitations being sent to users when publishing profile updates

### DIFF
--- a/apps/crn-server/serverless.ts
+++ b/apps/crn-server/serverless.ts
@@ -414,7 +414,7 @@ const serverlessConfig: AWS = {
         },
       ],
       environment: {
-        SES_REGION: `\${ssm:ses-region-${envAlias}}`,
+        SES_REGION: `\${ssm:ses-region-dev}`,
         EMAIL_SENDER: `\${ssm:email-invite-sender-${envAlias}}`,
         EMAIL_BCC: `\${ssm:email-invite-bcc-${envAlias}}`,
         EMAIL_RETURN: `\${ssm:email-invite-return-${envAlias}}`,
--- a/apps/crn-server/serverless.ts
+++ b/apps/crn-server/serverless.ts
@@ -95,6 +95,9 @@ const offlinePlugins = [
 const offlineSSM =
   SLS_STAGE === 'local'
     ? {
+        'algolia-app-id-dev': '${env:ALGOLIA_APP_ID}',
+        'algolia-index-api-key-dev': '${env:ALGOLIA_API_KEY}',
+        'algolia-search-api-key-dev': '${env:ALGOLIA_API_KEY}',
         'crn-algolia-app-id-dev': '${env:ALGOLIA_APP_ID}',
         'crn-algolia-index-api-key-dev': '${env:ALGOLIA_API_KEY}',
         'crn-algolia-search-api-key-dev': '${env:ALGOLIA_API_KEY}',
@@ -369,7 +372,7 @@ const serverlessConfig: AWS = {
         SENTRY_DSN: sentryDsnHandlers,
       },
     },
-    inviteUser: {
+    inviteUserSquidex: {
       handler: './src/handlers/user/invite-handler.handler',
       events: [
         {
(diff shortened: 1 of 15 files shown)


packages/server-common/src/handlers/user/invite.handler.ts
```diff
@@ -15,73 +11,69 @@ import {
...
+    );
+
+    const code = uuidV4();
 
     try {
-      await userClient.patch(user.id, {
-        connections: {
-          iv: [...(user.data.connections.iv || []), { code: newCode }],
-        },
+      await dataProvider.update(user.id, {
```
peterstarling, line 52: alright this one was actually here for a reason. See - we had this issue whereby sometimes squidex would send out the 'user-create' event for all dev users. Although the actually root cause of it was the order in which the data was impoted into the PR environments we made sure that the previous connections would not be removed when such invite handler was fired. 

I understand we want to skip creating connections for users that already have ANY connection code (unlike previously when we would only skip if they had an invitation code) but this line here - i don't think it needs to be modified? In which case you should make sure that we get the current user connection codes and add the new one

packages/server-common/test/fixtures/users.fixtures.ts
```diff
@@ -74,3 +74,11 @@ export const getUserWebhookPayload = (
     },
   },
 });
+
+export const userMock = (): UserDataObject => ({
```
peterstarling, line 78:
```suggestion
export const getUserDataObject = (): UserDataObject => ({
```

packages/server-common/test/handlers/user/invite.handler.test.ts
```diff
@@ -1,174 +1,134 @@
...
 
     await inviteHandler(event);
 
-    expect(userClient.fetchById).toBeCalledWith(userWithConnection.id);
+    expect(dataProvider.fetchById).toBeCalledWith(userWithConnection.id);
     expect(sendEmailMock).not.toBeCalled();
   });
 
-  test('Should find the user with an existing connection code and add an invitation code without removing the existing connection', async () => {
+  test('Should not send the invitation email for a user that already has an authentication code', async () => {
```
peterstarling, line 89: you have repurposed a test that was referring to a whole different part of the app which I don't think you should have changed (see my comment below) - unless I am mistaken?

If I am right - I think this should be an new test

peterstarling (changes requested):

packages/server-common/src/handlers/user/invite.handler.ts
lennym: By this point we have already checked that there are no existing user connection codes and exited early if any exist. 

`...(user.data.connections.iv || [])` is equivalent to `...[]` here, and so was removed.

packages/server-common/test/handlers/user/invite.handler.test.ts
lennym: The removed test and the new test are mutually exclusive - the old test asserts that the invitation code is added to a user with a pre-existing auth code, and the new test asserts that it is _not_.

apps/crn-server/serverless.ts
```diff
@@ -391,6 +394,32 @@ const serverlessConfig: AWS = {
...
+              'detail-type': ['UsersPublished'],
+            },
+            retryPolicy: {
+              maximumRetryAttempts: 2,
+            },
+          },
+        },
+      ],
+      environment: {
+        SES_REGION: `\${ssm:ses-region-${envAlias}}`,
```
peterstarling, line 417:
```suggestion
        SES_REGION: `\${ssm:ses-region-dev}`,
```

peterstarling (changes requested):

packages/server-common/src/handlers/user/invite.handler.ts
```diff
@@ -15,73 +11,69 @@ import {
...
-    );
-
-    const previousCode = user.data.connections.iv
-      ?.map((c) => c.code)
-      .find((c) => c.match(uuidMatch));
-
-    if (previousCode) {
+    if (user.connections?.length) {
       logger.info(
         `Found a previous invitation code for user ${user.id}, exiting...`,
```
lennym, line 40:
```suggestion
        `Found an existing connection code for user ${user.id}, exiting...`,
```

peterstarling (approved):

markcsims (approved):

PR #3006, opened 2023-04-26
[CRN-999] fix cronjob-sync-orcid

Riccardo-Brown:


apps/crn-server/src/handlers/webhooks/cronjob-sync-orcid.ts
```diff
@@ -35,22 +34,31 @@ const rawHandler = async (): Promise<lambda.Response> => {
...
       orcid: '-',
+      orcidLastSyncDate,
     },
   });
 
+  // orcid rate limit 24 request per second
+  const throttle = pThrottle({
+    limit: 24,
+    interval: 1000,
+  });
```
lennym, line 54: This will likely need to be looked at again once we use contentful data providers since each profile will potentially require several API calls to sync the works, and the contentful management SDK is limited to 7req/s.

It's probably ok for now though.

lennym (approved):

PR #3016, opened 2023-04-28
CT-8 Events data provider

gabiayako:


packages/model/src/event.ts
```diff
@@ -67,12 +67,13 @@ export type EventCreateDataObject = Pick<
   | 'endDateTimeZone'
   | 'status'
   | 'tags'
-  | 'meetingLink'
```
gabiayako, line 70: The `create` method is being used only here https://github.com/yldio/asap-hub/blob/master/packages/server-common/src/utils/sync-google-event.ts#L96 and it does not send `meetingLink`.

apps/crn-server/test/utils/wait-for-updated.test.ts
peterstarling, line 1: why is it here? do you need to rebase?

apps/crn-server/src/data-providers/contentful/event.data-provider.ts
```diff
@@ -0,0 +1,419 @@
...
+        id: filter.teamId,
+      });
+
+      const eventsCollection =
+        teams?.linkedFrom?.eventSpeakersCollection?.items[0]?.linkedFrom
+          ?.eventsCollection;
+
+      return getEventDataObject(eventsCollection);
+    }
+
```
peterstarling, line 138: maybe add another TODO here somewhere so we know we need to add filtering by interest-groups and working-groups?

peterstarling (changes requested):

apps/crn-server/test/utils/wait-for-updated.test.ts
gabiayako: I moved here because I thought this would need to be reused in every update in Contentful. Is it not necessary?

apps/crn-server/test/data-providers/contentful/event.data-provider.test.ts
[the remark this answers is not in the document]
gabiayako: Hey, about this one I've changed the types here https://github.com/yldio/asap-hub/pull/3016/commits/419a90ba60d503dfe417695d8b153781856e5a81 but I still need to check with the GP2 team if this is ok. Without this change in the gp2 model file as well it gave type errors.

Taking a look at the code, it seems that where we filter events by userId, teamId, etc we didn't add more filters:
- groupId: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/event/index-group-events-handler.ts#L48
- teamId: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/event/index-team-events-handler.ts#L49
- externalAuthorId: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/event/index-external-author-events-handler.ts#L50
- userId: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/event/index-user-events-handler.ts#L48
- non exclusive filter (google id + hidden): https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/controllers/events.ts#L43

apps/crn-server/src/data-providers/contentful/event.data-provider.ts
gabiayako: I've added here https://github.com/yldio/asap-hub/blob/CT-8-events-data-provider/apps/crn-server/src/data-providers/contentful/event.data-provider.ts#L397:L398 :)

apps/crn-server/test/utils/wait-for-updated.test.ts
gabiayako: I've reverted it here [fd74f14](https://github.com/yldio/asap-hub/pull/3016/commits/fd74f1485f714f40693ad3ae9c71d6ef17a9021c). We don't need in this case because the update is not made via the hub so we don't need instant update. This is the idea, right?

packages/model/src/event.ts
```diff
@@ -88,15 +89,61 @@ export type EventUpdateDataObject = Partial<
...
-  groupId?: string;
-  userId?: string;
-  externalAuthorId?: string;
-  teamId?: string;
-  googleId?: string;
-  hidden?: boolean;
-};
+type FilterOptions =
+  | {
+      workingGroupId?: string;
```
peterstarling, line 94: might want to consider an intermediary type with all properties marked as `never` and then in each of the unions just override those which are to be given, but that's up to you. It's a little messy for that many properties being mutually exclusive but oh well...

peterstarling (approved):

packages/model/src/event.ts
gabiayako: nice! I've refactored it here https://github.com/yldio/asap-hub/pull/3016/commits/15553e5069779762413fef4cede1a9657925ac45

markcsims (approved):

PR #3019, opened 2023-04-28
Improvements to users data provider `waitForUpdated` implementation

lennym:
* make options passed to `async-retry` explicit
* explicitly handle case where `publishedVersion` is undefined
* update mock default response in tests to be realistic

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/users.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/users.data-provider.ts
@@ -130,10 +130,13 @@ export class UserContentfulDataProvider implements UserDataProvider {
     const environment = await this.getRestClient();
     const user = await environment.getEntry(id);
     const result = await patchAndPublish(user, fields);
-    await this.waitForUpdated(id, result.sys.publishedVersion || Infinity);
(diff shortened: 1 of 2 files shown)


apps/crn-server/src/data-providers/contentful/users.data-provider.ts
```diff
@@ -130,10 +130,13 @@ export class UserContentfulDataProvider implements UserDataProvider {
     const environment = await this.getRestClient();
     const user = await environment.getEntry(id);
     const result = await patchAndPublish(user, fields);
-    await this.waitForUpdated(id, result.sys.publishedVersion || Infinity);
+    await this.waitForUpdated(id, result.sys.publishedVersion);
   }
 
-  private async waitForUpdated(id: string, version: number) {
+  private async waitForUpdated(id: string, version: number | undefined) {
```
peterstarling, line 136:
```suggestion
  private async waitForUpdated(id: string, version?: number) {
```
though to be fair im still not a fan of an optional parameter only to satisfy typescript. I think it should just be 

```suggestion
  private async waitForUpdated(id: string, version: number) {
```

PR #3023, opened 2023-04-28
feat: CT-3 Split Discover Controller

peterstarling:
- create discover data provider
- update the controller


apps/crn-server/src/data-providers/types/discover.data-provider.types.ts
```diff
@@ -0,0 +1,5 @@
+import { DiscoverDataObject } from '@asap-hub/model';
+
+export type DiscoverDataProvider = {
```
gabiayako, line 3: I was wondering if we don't want to add it here https://github.com/yldio/asap-hub/blob/master/packages/model/src/data-providers.ts#L53 :)

apps/crn-server/test/data-providers/discover.data-provider.test.ts
```diff
@@ -0,0 +1,153 @@
...
+  const discoverDataProviderGraphqlMockGraphql =
+    new DiscoverSquidexDataProvider(squidexGraphqlClientMockServer);
+
+  beforeAll(() => {
+    identity();
+  });
+
+  describe('Fetch method', () => {
+    describe('with mock-server', () => {
+      test('Should fetch the dashboard from squidex graphql', async () => {
```
gabiayako, line 34:
```suggestion
      test('Should fetch discover from squidex graphql', async () => {
```

gabiayako (approved):

apps/crn-server/src/data-providers/types/discover.data-provider.types.ts
peterstarling: I think these are the ones that are shared across the apps? Though to be fair I wouldn't mind having all the data provider types in the model package

PR #3032, opened 2023-05-03
CRN-804 - Normalise url and title to support identification of duplicates

andreicovaciu:


apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -79,46 +79,51 @@ export default class ResearchOutputs implements ResearchOutputController {
   async create(
     researchOutputCreateData: ResearchOutputCreateData,
   ): Promise<ResearchOutputResponse | null> {
-    await this.validateResearchOutput(researchOutputCreateData);
+    const normalizedResearchOutputCreateData = this.normaliseResearchOutputData(
+      researchOutputCreateData,
+    ) as ResearchOutputCreateData;
```
peterstarling, line 84: I'm not a big fan of casting, try not typing the return type of the `normaliseResearchOutputData` method - maybe typescript can infer it in such a way that you then don't need to cast. Otherwise you should probably use some conditional magic or overloads to make sure that with `ResearchOutputCreateData` you get back `ResearchOutputCreateData` again.

apps/crn-server/src/controllers/research-outputs.ts
```diff
@@ -79,46 +79,51 @@ export default class ResearchOutputs implements ResearchOutputController {
   async create(
     researchOutputCreateData: ResearchOutputCreateData,
   ): Promise<ResearchOutputResponse | null> {
-    await this.validateResearchOutput(researchOutputCreateData);
+    const normalizedResearchOutputCreateData = this.normaliseResearchOutputData(
```
peterstarling, line 82: you mixed BE and AE spelling here (normalize vs normalise), I would recommend using BE 
```suggestion
    const normalisedResearchOutputCreateData = this.normaliseResearchOutputData(
```

lennym:
I'd be tempted to use a generic type on the normalise function so that the input type and return type are guaranteed to match.

apps/crn-server/test/controllers/research-outputs.test.ts
peterstarling, line 1: you verified that we now look up previous record using normalised input data but there is no test to actually show that the normalised input is passed down to the data provider. I found it confusing - wasn't sure if you only normalise for the lookup but still allow spaces or do we want to make sure that in the future all input data is trimmed (aka normalised) before inserting into the CMS

peterstarling (changes requested):

apps/crn-server/src/controllers/research-outputs.ts
lennym: I'd make it a generic, so `this.normaliseResearchOutputData<ResearchOutputCreateData>(researchOutputCreateData)`

andreicovaciu:
> I'd be tempted to use a generic type on the normalise function so that the input type and return type are guaranteed to match.

I agree

apps/crn-server/test/controllers/research-outputs.test.ts
andreicovaciu: With this implementation, we will create/update normalised research output. That means that the title and the url will be trimmed, which IMO is the correct approach. 
And yes indeed, a new test should be added for the database lookup.

What do you think? @peterstarling

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -232,6 +232,53 @@ describe('ResearchOutputs controller', () => {
         );
       });
 
+      test('Should throw error when a research output with the same type and title (trimmed) already exists', async () => {
```
Riccardo-Brown, line 235: I don't think you need these `(trimmed)` tests as the title and link are always trimmed. I would just have the original tests, include whitespace in the test payload and the `calledWith` values be trimmed.

andreicovaciu: Sounds good too!

apps/crn-server/test/controllers/research-outputs.test.ts
andreicovaciu: Actually, the tests already check that database lookup (with trimmed data). See here https://github.com/yldio/asap-hub/pull/3032/files#diff-9c5ace13bc330fda61ea7f18f63431d9888b21c05a1b6d18e337a286db0d6b5eR200

Riccardo-Brown (approved):

apps/crn-server/test/controllers/research-outputs.test.ts
```diff
@@ -190,6 +190,10 @@ describe('ResearchOutputs controller', () => {
       researchOutputRequest.link = 'https://some-test-link.com';
 
       test('Should throw error when a research output with the same type and title already exists', async () => {
+        const researchOutputWhitespacesRequest = getResearchOutputCreateData();
```
peterstarling, line 193: I don't think I like changing the existing test so it works for your case. We should keep the existing one as is because its a valid scenario of when the title is identical and should add an extra one where it's the same but with extra spaces.

apps/crn-server/test/controllers/research-outputs.test.ts
peterstarling: I don't think it does? it only checks that the fetch method is being called with the normalised data, I was looking for a test which shows that the `create` method from the data provider is called with the normalised values

peterstarling (changes requested):

apps/crn-server/test/controllers/research-outputs.test.ts
andreicovaciu: After discussing with @peterstarling, we came to the conclusion that it might be better to keep all the tests separate. https://github.com/yldio/asap-hub/pull/3032/commits/6ba19cdeb45dd656c7f085fc49a10fcfaab68773#r1187240273
We can have a discussion together with @peterstarling when you're back 🙌

peterstarling (approved):

PR #3039, opened 2023-05-04
feat: CT-38 Enable ORCID sync for Contentful

peterstarling:
### Changes
- refactor sync orcid handler
- add contentful handler to serverless
- adds a clear-on-change Contentful app which resets a field based on a change to another field. 
--> In our case we reset the orcidLastSyncDate when the orcid field is modified, which is how we tell our webhook to sync it.

### OOS
What is missing here is the migration to alter the orcidLastSyncDate to use the extension but that will be done through a follow-up PR, it's testable within this PR though as it can be manually cleared to trigger the sync

### DIFF
--- a/apps/crn-server/serverless.ts
+++ b/apps/crn-server/serverless.ts
@@ -848,7 +848,7 @@ const serverlessConfig: AWS = {
         IS_CONTENTFUL_ENABLED_V2: 'false',
       },
     },
-    cronjobSyncContentful: {
+    cronjobSyncOrcidContentful: {
       handler: './src/handlers/webhooks/cronjob-sync-orcid.handler',
       events: [
         {
--- a/apps/crn-server/serverless.ts
+++ b/apps/crn-server/serverless.ts
@@ -836,7 +836,7 @@ const serverlessConfig: AWS = {
           contentfulWebhookAuthenticationToken,
       },
     },
-    cronjobSyncOrcid: {
+    cronjobSyncOrcidSquidex: {
       handler: './src/handlers/webhooks/cronjob-sync-orcid.handler',
       events: [
         {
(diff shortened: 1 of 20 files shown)


apps/crn-server/serverless.ts
```diff
@@ -814,7 +836,19 @@ const serverlessConfig: AWS = {
...
+        {
+          schedule: 'rate(1 hour)', // run every hour
+        },
+      ],
+      environment: {
+        SENTRY_DSN: sentryDsnHandlers,
+        IS_CONTENTFUL_ENABLED_V2: 'false',
+      },
+    },
+    cronjobSyncContentful: {
```
Riccardo-Brown, line 851:
```suggestion
    cronjobSyncOrcidContentful: {
```

Riccardo-Brown (dismissed):
I don't understand how the contentful extension applies to ORCID sync. Is it meant to be part of the PR ?

peterstarling:
> I don't understand how the contentful extension applies to ORCID sync. Is it meant to be part of the PR ?

@Riccardo-Brown what the extension does is it resets a field based on a change to another field. In our case we reset the `orcidLastSyncDate` when the `orcid` field is modified, that way we tell our webhook to sync it. 

What is missing here is the migration to alter the `orcidLastSyncDate` to use the extension but that will be done through a follow-up PR, it's testable within this PR though as it can be manually cleared to trigger the sync

lennym:
Is there any control in place to stop the ORCID syncing on all user updates in Squidex?

I see the code to test if the ORCID had changed has been removed in favour of the sync data check that works with the new Contentful app, but I don't see any equivalent for Squidex unless I'm missing something.

lennym (approved):

gabiayako (approved):

Riccardo-Brown (approved):

PR #3042, opened 2023-05-05
CT-41 - Contentful App to show related team names on user models

lennym:
Uses a custom instance of `MultipleEntryReferenceEditor` to collect data from the second order related entity and display as part of the card when viewing the top level entity.

In this case we load the team data from the team membership model for display on the user page.


packages/contentful-app-extensions/user-team-memberships/src/locations/Field.tsx
```diff
@@ -0,0 +1,143 @@
...
+  return (
+    <EntryCard
+      contentType="Team"
+      actions={actions}
+      onClick={onClick}
+      status={status}
+    >
+      <Heading marginBottom="none">
+        {team.fields.displayName?.['en-US']}{' '}
+        {isInactive && <Badge variant="secondary">Inactive</Badge>}
```
gabiayako, line 54: Hey, I wonder if it's better to show the inactive date instead of a badge the same way in squidex or leave the badge below the role with a label like "user inactive in this team" because a badge "inactive" beside the team name makes me think that the team is inactive following the hub UI

Screenshot 2023-05-08 at 14 53 48

gabiayako (changes requested):

packages/contentful-app-extensions/user-team-memberships/src/locations/Field.tsx
lennym: I've moved it to be below the role.

PR #3048, opened 2023-05-09
Add contentful delivery api get client and polling

gabiayako:


packages/contentful/src/delivery-api/poll-delivery-api.ts
```diff
@@ -0,0 +1,26 @@
+import { Entry as CDAEntry, EntrySkeletonType, FieldsType } from 'contentful';
+import retry from 'async-retry';
+
+export const pollContentfulDeliveryApi = async <
+  T extends EntrySkeletonType<FieldsType, string>,
+>(
+  fetchEntry: () => Promise<CDAEntry<T>>,
+  version: number,
```
peterstarling, line 8: maybe not to confuse everyone even more let's make it accept a revision so it's clear that this method, as opposed to `pollContentfulGql`, uses a different field to poll?

packages/contentful/src/delivery-api/poll-delivery-api.ts
```diff
@@ -0,0 +1,26 @@
+import { Entry as CDAEntry, EntrySkeletonType, FieldsType } from 'contentful';
+import retry from 'async-retry';
+
+export const pollContentfulDeliveryApi = async <
```
peterstarling, line 4: to be fair the two functions are so similar i would be tempted to write one generic function to fetch and bail and then have these two higher order functions reuse that other one, just pass the right data extraction logic. Basically introduce some reusability maybe? 

Also it feels like both could just be part of one util file

peterstarling (changes requested):

peterstarling (approved):

PR #3053, opened 2023-05-10
[CRN-930] refactor tags to keywords

Riccardo-Brown:
- Switches the payload for Additional Keywords from `tags` to `keywords` field of the research outputs
- refactor `api/research-tags` endpoint to returns all research tags by default


apps/crn-server/src/data-providers/research-tags.data-provider.ts
```diff
@@ -104,8 +106,33 @@ export class ResearchTagSquidexDataProvider implements ResearchTagDataProvider {
       }),
     };
   }
+
+  async fetchAll(options: FetchAllOptions): Promise<ListResearchTagDataObject> {
```
peterstarling, line 110: I am not sure if I would make this a new method, especially since the DataProvider interface clearly shows which 4 methods should be available. Maybe just make `all` a parameter for fetch?

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
```diff
@@ -666,13 +667,11 @@ describe('ResearchOutputs data provider', () => {
         withTeams: true,
       };
 
-      beforeEach(() => {
+      test('Should pass the pagination parameters as expected', async () => {
```
peterstarling, line 670: maybe rather than remove the beforeEach and repeat the mocks all the time just have another nested describe for all tests that do not pass the search parameter and then another one for the ones that do?

peterstarling (commented):
this is not a change request per se as i am not 100% sure you need to make all these changes, though it would be good to at least discuss before marking as resolved

apps/crn-server/src/data-providers/research-tags.data-provider.ts
Riccardo-Brown: I thought the point of the interface was so our data providers implement core CRUD methods, but could also have other methods. If we don't want to have `fetchAll` in the provider then i think we could

- move it to the controller level
- move the fetching to the FE

If we were to use `all` would we pass this from the FE?

peterstarling: controller is my preference, yes

apps/crn-server/test/data-providers/research-outputs.data-provider.test.ts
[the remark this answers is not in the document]
peterstarling: we are not testing a mock though, that's the thing. That `mocked` graphql server instance does a bit more than just return rubbish data, it validates the query against the schema and returns data types as per schema definition. 

In fact - if it wasn't for the query parameters (such as filters and pagination) we could just call it a complete data-provider test, unfortunately because of these query parameters the only ultimate test to see if the data provider works is the integration test which we don't do a lot, but ignore that for now. 

Anyway like I said I think it's always to validate or queries in a data provider against that mocked server instance, aside from just returning data mocks from the client mock directly.

apps/crn-server/src/controllers/research-tags.ts
```diff
@@ -24,4 +30,12 @@ export default class ResearchTags implements ResearchTagController {
       filter,
     });
   }
+
+  async fetchAll(options: FetchAllOptions): Promise<ListResearchTagResponse> {
```
peterstarling, line 34: for this controller is there any point for leaving the `fetch` method as is? i thought we would just put the fetch-all implementation there

peterstarling (approved):

PR #3077, opened 2023-05-15
feat: Use Contentful app for User ORCID field

peterstarling:


packages/contentful/migrations/crn/users/20230511122232-add-token-date-clean-up-app.js
```diff
@@ -0,0 +1,18 @@
+module.exports.description = 'Change field control for tokenDate to app';
+
+module.exports.up = (migration) => {
+  const users = migration.editContentType('users');
+  users.changeFieldControl(
+    'orcidLastSyncDate',
+    'app',
+    '6en82KYFrsxlAIPqypMdOp',
```
lennym, line 8: No objection to this instance, but I wonder if there's a better way to manage these ids so they can be referenced when used rather than appearing as somewhat opaque strings.

I'm thinking maybe a config file in the app directories than can be imported into here, and read at deploy-time.

lennym: It would be trivial to read and parse the app id from a json file in the deploy build step if it was in a json file in the directory. I'm fairly sure I've seen `jq` used in other build steps to pull parameters out from json.

PR #3080, opened 2023-05-16
CRN-1399 - sets patching enabled

andreicovaciu:


peterstarling (approved):
make sure this is tested in dev once merged before running in prod. When testing - make sure to sort records by the updated time to see which ones have been modified to make sure you capture them all

PR #3082, opened 2023-05-16
Added working groups and team drafts Research Outputs on reminders

oratiu:


apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -32,6 +41,29 @@ export type ReminderDataProvider = DataProvider<
   FetchRemindersOptions
 >;
 
+type ResearchOutputQueried = Pick<
```
oratiu, line 44: Would be nice if this could get extracted but since this is deeply nested, I've found no way

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -158,6 +213,40 @@ const getUserTeamIds = (
   return userTeams;
 };
 
+const getUserWorkingGroupIds = (
+  workingGroups: Maybe<
```
oratiu, line 217: How can I avoid having this type here ?

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -214,6 +303,84 @@ const getResearchOutputRemindersFromQuery = (
...
+    return [];
+  }
+
+  return queryResearchOutputsContents.reduce<ResearchOutputDraftReminder[]>(
+    (researchOutputReminders, researchOutput) => {
+      const userName = getUserName(researchOutput);
+      const { associationName, associationType } =
+        getAssociationNameAndType(researchOutput);
+
+      if (
```
oratiu, line 329: Not really happy with this, there must be a better than having this huge if statement

apps/crn-server/src/data-providers/reminders.data-provider.ts
peterstarling: I don't think I understand why you should need it actually. You use it here:
https://github.com/yldio/asap-hub/blob/db2a41aeda360a5ebc8c6c4f314a45d4cea2eeb6/apps/crn-server/src/data-providers/reminders.data-provider.ts#L699
and
https://github.com/yldio/asap-hub/blob/db2a41aeda360a5ebc8c6c4f314a45d4cea2eeb6/apps/crn-server/src/data-providers/reminders.data-provider.ts#L732

The first of the two is then applied here
https://github.com/yldio/asap-hub/blob/db2a41aeda360a5ebc8c6c4f314a45d4cea2eeb6/apps/crn-server/src/data-providers/reminders.data-provider.ts#L327
when you follow the code from the reduce
https://github.com/yldio/asap-hub/blob/db2a41aeda360a5ebc8c6c4f314a45d4cea2eeb6/apps/crn-server/src/data-providers/reminders.data-provider.ts#L323

you can see that we reduce the `queryResearchOutputsContents` variable whose type is declared right above...
https://github.com/yldio/asap-hub/blob/db2a41aeda360a5ebc8c6c4f314a45d4cea2eeb6/apps/crn-server/src/data-providers/reminders.data-provider.ts#L307
...as `FetchReminderDataQuery['draftResearchOutputs']`; so your task is to simply extract the element of the array from that type. If you look at how we extract types in all data providers you should find that the way to achieve that is to via 
`NonNullable<FetchReminderDataQuery['draftResearchOutputs']>[number]`. So that means that your function declaration can be:
```typescript
const getAssociationNameAndType = (
  researchOutput: NonNullable<FetchReminderDataQuery['draftResearchOutputs']>[number],
): {
(...)
```

apps/crn-server/src/data-providers/reminders.data-provider.ts
gabiayako: I think this type is equivalent to
```
type ReferencingWorkingGroupsContents = NonNullable<
  FetchReminderDataQuery['findUsersContent']
>['referencingWorkingGroupsContents'];
```

peterstarling: look above, probably something like `NonNullable<FetchReminderDataQuery['findUsersContent']>['referencingWorkingGroupsContents']`

apps/crn-server/src/data-providers/reminders.data-provider.ts
gabiayako: We could wrap by a `NonNullable` and to get an item from the array we use this `[number]`. So it would be something like

```
type DraftResearchOutput = NonNullable<
  FetchReminderDataQuery['draftResearchOutputs']
>[number];
```

gabiayako (commented):
I'm still reviewing it :) just sending some comments already

oratiu:
> 

And I'm already addressing the comments :D

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -43,25 +75,38 @@ export class ReminderSquidexDataProvider implements ReminderDataProvider {
...
-    });
+    >(
+      FETCH_REMINDER_DATA,
+      {
+        researchOutputFilter,
+        researchOutputDraftFilter,
+        eventFilter: getEventFilter(options.timezone),
+        userId: options.userId,
+      },
+      { includeDrafts: true },
```
peterstarling, line 96: isn't this going to affect the existing queries? they might all now start receiving drafts, maybe you should add the status filters into the other ones? I would write an integration test for the existing research ouput reminders - for instance create an identical test but make sure that the RO that was previously given a reminder is now in a draft state - lets see if it shows up (it should not)

apps/crn-server/test/data-providers/reminders.data-provider.test.ts
```diff
@@ -47,11 +49,31 @@ describe('Reminder Data Provider', () => {
...
+      test('Should fetch the reminders from squidex graphql', async () => {
         const result = await reminderDataProviderMockGraphql.fetch(
           fetchRemindersOptions,
         );
 
+        expect(result).toEqual({
+          total: 2,
+          items: [
+            getResearchOutputPublishedReminder(),
+            getResearchOutputDraftReminder(),
```
peterstarling, line 61: this makes very little sense to me - why do we expect a draft RO for the published reminder test?

peterstarling: this is probably an indication of an error of some sort - our mock server should only return a single RO for each of those two queries (that are part of a singe large query actually) and surely for the published reminders only one of them should be relevant?

peterstarling (changes requested):
no integration tests - and i can tell your unit tests are only there for coverage - i would expect a lot of bugs coming out of this (might be wrong though!) Please make sure that you add a lot of integration tests for it, think about different cases, multiple reminders, mixing up published and drafts, etc

packages/model/src/reminder.ts
```diff
@@ -50,6 +54,20 @@ export interface ResearchOutputPublishedReminder
   };
 }
 
+export interface ResearchOutputDraftReminder extends ResearchOutputReminder {
+  entity: 'Research Output';
+  type: 'Draft';
+  data: {
+    researchOutputId: ResearchOutputDraftDataObject['id'];
+    documentType: ResearchOutputDraftDataObject['documentType'];
```
gabiayako, line 62: I believe the `documentType` and `addedDate` are not necessary 😉

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -527,3 +694,48 @@ const getSortDate = (reminder: ReminderDataObject): DateTime => {
...
+    researchOutput.flatData &&
+    researchOutput.flatData.teams &&
+    researchOutput.flatData.teams.length > 0
+  ) {
+    return {
+      associationType: 'team',
+      associationName:
+        researchOutput.flatData.teams[0]?.flatData.displayName || null,
+    };
+  }
```
gabiayako, line 725: It could be a little simpler here

```suggestion
  if (
    researchOutput.flatData?.workingGroups &&
    researchOutput.flatData.workingGroups[0]
  ) {
    return {
      associationType: 'working group',
      associationName:
        researchOutput.flatData.workingGroups[0].flatData.title || null,
    };
  }
  if (researchOutput.flatData?.teams && researchOutput.flatData.teams[0]) {
    return {
      associationType: 'team',
      associationName:
        researchOutput.flatData.teams[0].flatData.displayName || null,
    };
  }
```

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -214,6 +303,84 @@ const getResearchOutputRemindersFromQuery = (
...
+  return queryResearchOutputsContents.reduce<ResearchOutputDraftReminder[]>(
+    (researchOutputReminders, researchOutput) => {
+      const userName = getUserName(researchOutput);
+      const { associationName, associationType } =
+        getAssociationNameAndType(researchOutput);
+
+      if (
+        !researchOutput.flatData.teams ||
+        !researchOutput.flatData.documentType ||
+        !isResearchOutputDocumentType(researchOutput.flatData.documentType) ||
```
gabiayako, line 332: I wonder if we need these two

```
!researchOutput.flatData.documentType ||
!isResearchOutputDocumentType(researchOutput.flatData.documentType) ||
```

since we're not using `documentType` in this reminder

and if need this

```
!researchOutput.flatData.teams
```

because in `getAssociationNameAndType` we're already checking if this exists

apps/crn-server/src/queries/reminders.queries.ts
```diff
@@ -4,9 +4,23 @@ export const FETCH_REMINDER_DATA = gql`
...
+        id
+        flatData {
+          complete
+          leaders {
+            inactiveSinceDate
+            role
+            user {
+              id
+            }
+          }
```
gabiayako, line 21: we're not using this, right?

packages/model/src/reminder.ts
oratiu: addedDate is necessary since we sort the reminders by that addedDate, documentType indeed is not

apps/crn-server/test/data-providers/reminders.data-provider.test.ts
oratiu: We have the following piece of code in squidex-graphql-client-with-server.mock
`    ApplicationQueries: () => ({
      queryResearchOutputsContents: [...new Array(1)],
    }),
    ` and both the draftROreminders and publishedROreminders use the queryResearchOutputsContents, so it normally returns for both of them.

apps/crn-server/src/data-providers/reminders.data-provider.ts
oratiu: documentType is out, but we do need to check the teams since right below I'm doing this and it complains teams might be missing
`      const researchOutputTeams = researchOutput.flatData.teams.map(
        (team) => team.id,
      );`

PR #3093, opened 2023-05-18
CT-28 Events data migration

gabiayako:


packages/cms-data-sync/src/events/events.data-migration.ts
```diff
@@ -0,0 +1,300 @@
...
+          `Calendar ${calendarId} does not exist in contentful. Event with id ${eventId} is going to be created without a calendar.`,
+          'ERROR',
+        );
+      }
+    }
+
+    return null;
+  };
+
+  const deletePreviousSpeakerReferences = async (contentfulEvent: Entry) => {
```
gabiayako, line 205: Deleting the old speakers so we don't leave a lot of unused entries in the eventSpeakers collection

packages/cms-data-sync/test/events/events.data-migration.test.ts
```diff
@@ -0,0 +1,582 @@
...
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterAll(() => {
+    console.log = consoleLogRef;
+  });
+
+  it('does not call clear contentful entries for events', async () => {
```
peterstarling, line 208: maybe let's enable clearing previous entries conditionally? i feel like by default we should dump them all anyway, at the moment i dont see how we would need to NOT do that at all

packages/cms-data-sync/test/events/events.data-migration.test.ts
peterstarling, line 1: to be fair there are a lot of mocks here to the same mock instance that depend entirely on the order of calls which makes it very difficult to refactor (or understand what is going on). You left the comments next to some of them but i thought maybe you would consider using the `.when` [jest extension](https://github.com/timkindberg/jest-when)?

peterstarling (changes requested):

packages/cms-data-sync/test/events/events.data-migration.test.ts
gabiayako: I've never used this before. It's much cleaner now, thanks

peterstarling (approved):

peterstarling (approved):

PR #3095, opened 2023-05-18
CT-54 Index algolia contentful events

gabiayako:
Example of successful run: https://github.com/yldio/asap-hub/actions/runs/5027058999/jobs/9016090265#step:21:37


apps/crn-server/scripts/export-entity.ts
```diff
@@ -54,8 +55,8 @@ export const exportEntity = async (
 
   do {
     records = await controller.fetch({
-      take: 50,
-      skip: (page - 1) * 50,
+      take: 20,
```
gabiayako, line 58: It was failing due to query complexity. I think this might happen a lot as we add more relationships with these last entities we're migrating. So I decreased to 20, but it was a random number 😛 looking at graphql playground it could be 40. Do you guys think it's ok to leave as 20?
Screenshot 2023-05-19 at 15 13 22

apps/crn-server/scripts/export-entity.ts
```diff
@@ -146,10 +147,13 @@ const getController = (entity: keyof EntityResponses) => {
 
   const assetDataProvider = new AssetSquidexDataProvider(userRestClient);
 
-  const eventDataProvider = new EventSquidexDataProvider(
-    eventsRestClient,
-    squidexGraphqlClient,
-  );
+  const eventDataProvider = isContentfulEnabled
```
peterstarling, line 150: maybe export to `/dependencies` like with the other ones?

apps/crn-server/scripts/export-entity.ts
peterstarling: 20 is fine, why not, just more calls to the CMS is all

apps/crn-server/scripts/export-entity.ts
peterstarling: in fact i can see you have exported it already, just havent used it

peterstarling (approved):

PR #3102, opened 2023-05-22
Confirm Dialog

kmaid:


packages/react-components/src/molecules/Modal.tsx
```diff
@@ -16,6 +16,7 @@ const overlayContainerStyles = css({
   height: '100%',
   top: 0,
   left: 0,
+  zIndex: 1,
```
andreicovaciu, line 19: can we maybe use `createPortal` instead? What are your thoughts on this?

packages/react-components/src/organisms/ConfirmModal.tsx
```diff
@@ -74,54 +83,69 @@ const ConfirmModal: React.FC<ConfirmModalProps> = ({
...
+                {cancelText}
               </Button>
-            </div>
+            )}
+          </div>
+          <div css={saveStyles}>
+            <Button
+              primary
+              enabled={status !== 'isSaving'}
+              onClick={async () => {
```
andreicovaciu, line 129: I know it's not part of these changes but can we maybe extract this?

apps/crn-frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -54,6 +54,7 @@ const DuplicateOutput: FC = () => {
           link: undefined,
           title: `Copy of ${output.title}`,
         }}
+        descriptionUnchangedWarning
```
gabiayako, line 57: Hey, I wonder if we could rename this `descriptionUnchangedWarning` to something like `isDuplicate` or `originatedFromDuplicate` :)

When I saw this truthy boolean I at first I was asking myself if this was ever recalculated to see if the description changed then only after I understood that this `descriptionUnchangedWarning` was only to flag that this comes from a duplicate

gabiayako (commented):
Was this note from the ticket (https://asaphub.atlassian.net/browse/CRN-1396) left out on purpose?

> Note: This popup should only be displayed on the first instance of the user selecting the duplicate button, where the description has not be changed.

packages/react-components/src/templates/ResearchOutputForm.tsx
```diff
@@ -316,153 +324,185 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
...
-                          }).$,
-                        );
-                      }
-                      return researchOutput;
-                    })}
-                    primary={showSaveDraftButton && !showPublishButton}
+              return researchOutput;
+            })();
+          const confirmText = `Keep and ${
+            showDescriptionChangePrompt === 'draft' ? 'draft' : 'publish'
```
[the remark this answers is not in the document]
kmaid: It wasn't specced so I took a stab. I can change it

apps/crn-frontend/src/network/teams/TeamProfile.tsx
kmaid: I feel that its better to describe how the component behaves with the prop than expect the next programmer to have the domain knowledge of thats how a duplicate behaves

packages/react-components/src/templates/ResearchOutputForm.tsx
```diff
@@ -316,153 +324,185 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
...
+            <>
+              {showDescriptionChangePrompt && (
+                <ConfirmModal
+                  title="Keep the same description?"
+                  cancelText="Cancel"
+                  onCancel={() => {
+                    setShowDescriptionChangePrompt(false);
+                  }}
+                  confirmText={confirmText}
+                  onSave={async () => {
```
andreicovaciu, line 363: can we extract?

packages/react-components/src/templates/ResearchOutputForm.tsx
```diff
@@ -316,153 +324,185 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
...
+          const confirmText = `Keep and ${
+            showDescriptionChangePrompt === 'draft' ? 'draft' : 'publish'
+          }`;
+          return (
+            <>
+              {showDescriptionChangePrompt && (
+                <ConfirmModal
+                  title="Keep the same description?"
+                  cancelText="Cancel"
+                  onCancel={() => {
```
andreicovaciu, line 359: (optional): no `{}` needed

packages/react-components/src/templates/ResearchOutputForm.tsx
andreicovaciu: same for all the below

gabiayako (approved):

PR #3104, opened 2023-05-23
CT-55 Migration - Contentful too many requests

gabiayako:


packages/cms-data-sync/src/calendars/calendars.data-migration.ts
```diff
@@ -10,13 +11,16 @@ type CalendarItem = NonNullable<
   FetchCalendarsQuery['queryCalendarsContents']
 >[number];
 
+const limiter = new RateLimiter({ tokensPerInterval: 7, interval: 'second' });
```
Riccardo-Brown, line 14: As the max requests are 7 do we want to give ourselves a buffer and set `tokensPerInterval` to 6 or 5?

packages/cms-data-sync/src/users/users.data-migration.ts
```diff
@@ -73,15 +81,20 @@ export const migrateUsers = async () => {
...
+          limiter,
+        );
       }
     }
 
     const teamLinks = await Promise.all(
       (teams || []).map(async (team) => {
         if (team.id && team.id.length) {
+          // create and publish
+          await limiter.removeTokens(2);
```
Riccardo-Brown, line 96: i think we could easily hit the rate limit here due to the use of `map` as it will execute at the same time. I think using a `for-of` loop will solve the potential issue

packages/cms-data-sync/src/pages/pages.data-migration.ts
```diff
@@ -13,13 +14,16 @@ import { pagesQuery } from './pages.queries';
 
 type PageItem = NonNullable<FetchPagesQuery['queryPagesContents']>[number];
 
+const limiter = new RateLimiter({ tokensPerInterval: 5, interval: 'second' });
```
peterstarling, line 17: i don't think you should be instantiating it every time - instead just export an instance with these settings and make sure to import that instance everywhere you need to make a request. Otherwise each of these requests work independently. Alternatively make sure to create the instance right at the beginning of the run-migrations file and then pass that along

packages/cms-data-sync/src/users/users.data-migration.ts
peterstarling: actually i think that would be the case before the use of the rate limiter, but with that tool now it should control the flow, although all of the async functions will be run at the same time (sort of) each of them will require awaiting of the removal of tokens which means that if we hit the limit that promise will not resolve until there are tokens available again

peterstarling (changes requested):

peterstarling (approved):

packages/cms-data-sync/src/events/events.data-migration.ts
```diff
@@ -77,6 +78,7 @@ export const migrateEvents = async () => {
           const teamId = squidexTeam ? squidexTeam[0].id : null;
           const userId = squidexUser ? squidexUser[0].id : null;
 
+          await contentfulRateLimiter.removeTokens(2);
```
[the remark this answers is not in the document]
gabiayako: because there are two getEntries call below, but I can change to remove one at a time

PR #3109, opened 2023-05-24
fix height bug in login

lctrt:


Riccardo-Brown (approved):

packages/react-components/src/templates/WelcomePage.tsx
```diff
@@ -37,6 +37,12 @@ const defaultValues = {
   },
 };
 
+const rootStyles = {
```
kmaid, line 40: Could this be done in index.html?

PR #3110, opened 2023-05-24
CT-17 Working Groups content model

gabiayako:
This PR 

- adds the migration to add the following content models:
  - workingGroups
  - workingGroupsLeaders
  - workingGroupsMembers
  - workingGroupDeliverables
- adds the working groups data provider

---

What is part of this ticket but it's going to be done in following PRs:
- Working Group TODOs throughout other entities responses
- Deliverables webhook
- Algolia index


apps/crn-server/src/data-providers/contentful/working-groups.data-provider.ts
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  async update(
+    id: string,
+    update: WorkingGroupUpdateDataObject,
+  ): Promise<void> {
+    const environment = await this.getRestClient();
+    const workingGroup = await environment.getEntry(id);
+
+    await Promise.all(
```
Riccardo-Brown, line 115: Do we need to add rate limiting here?

gabiayako: I believe we don't send a lot of update requests in sequence, but I've added it now to be safe: [review-suggestion: add limiter](https://github.com/yldio/asap-hub/pull/3110/commits/b6da28df0028b2d79c4f248cc2e0aa89f514567b)

Riccardo-Brown (approved):

apps/crn-server/src/data-providers/contentful/working-groups.data-provider.ts
```diff
@@ -0,0 +1,261 @@
...
+      workingGroup.fields.deliverables['en-US'].map(
+        async (deliverable: Link<'Entry'>) => {
+          const deliverableId = deliverable.sys.id;
+          await contentfulRateLimiter.removeTokens(1);
+          const deliverableEntry = await environment.getEntry(deliverableId);
+          await contentfulRateLimiter.removeTokens(1);
```
peterstarling, line 121: what's this one for

apps/crn-server/src/data-providers/contentful/working-groups.data-provider.ts
```diff
@@ -0,0 +1,261 @@
...
+    description,
+    membersCollection,
+  } = item;
+
+  const deliverables = (deliverablesCollection?.items || []).reduce(
+    (
+      wgDeliverables: WorkingGroupDataObject['deliverables'][number][],
+      deliverableItem,
+    ) => {
+      if (deliverableItem) {
```
peterstarling, line 181: correct me if i am wrong but aren't you simply filtering out the nulls/undefined and mapping to a slighlty different structure? I would prefer a `.filter` and a `.map` even if its more costly etc, just cause its more readable. Your solution is obvs more optimised (reduce + push rather than spreading) but less readable.

apps/crn-server/src/data-providers/contentful/working-groups.data-provider.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  const { leaders, members } = (membersCollection?.items || []).reduce(
+    (
+      wgMembers: {
+        leaders: WorkingGroupDataObject['leaders'][number][];
+        members: WorkingGroupDataObject['members'][number][];
+      },
+      member,
+    ) => {
+      if (member) {
```
peterstarling, line 204: same as above, can you not just filter these out then map?

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
```diff
@@ -0,0 +1,651 @@
...
+              status: 'In Progress',
+            },
+          ],
+        };
+        contentfulGraphqlClientMock.request.mockResolvedValueOnce(
+          contentfulGraphQLResponse,
+        );
+        const response = await workingGroupDataProvider.fetchById(id);
+
+        expect(response?.deliverables).toEqual([
```
peterstarling, line 435: not a huge deal really but im of an opinion that when dealing with nullable types for properties and variables in tests it is better to assert non-null with `!` rather than do `?` for chaining - reason being: we know we should be getting a response here and if you do `?` this might accidentally default to something you didn't anticipate. Like if you wanted a property to be undefined on an object that should not be undefined then `?` wouldnt really work would it

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
```diff
@@ -0,0 +1,651 @@
...
+                  linkType: 'Entry',
+                  type: 'Link',
+                },
+              },
+            ],
+          },
+        },
+      ]);
+
+      // expect(workingGroupMockUpdated.publish).toHaveBeenCalled();
```
peterstarling, line 648: i think you forgot something here

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
```diff
@@ -0,0 +1,651 @@
...
+
+        const result = await workingGroupDataProvider.fetchById(id);
+
+        expect(result?.members).toHaveLength(1);
+        expect(result?.leaders).toHaveLength(0);
+      });
+    });
+  });
+
+  describe('Update method', () => {
```
peterstarling, line 571: for how complex the method implementation is i think the test coverage is quite low, don't you think? there are a lot of edge cases here like - what happens when there is an error half way through removing old deliverables etc. Not sure if you were planning to add these or are we just going with a simple approach for now until its a problem?

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/working-groups.data-provider.ts
gabiayako: done here [review: refactor to use map+filter instead of reduce](https://github.com/yldio/asap-hub/pull/3110/commits/7ede2e49d24a6a2fc04350324a5dd6387c89536c)

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
gabiayako: done here [review: replace ? by ! in expected response](https://github.com/yldio/asap-hub/pull/3110/commits/37d06671a698a7d18828c8a943d76861ba931601)

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
gabiayako: done here [review: remove comment](https://github.com/yldio/asap-hub/pull/3110/commits/753df173434a994f498cad3804b12a9781aa7c81)

apps/crn-server/test/data-providers/contentful/working-groups.data-provider.test.ts
gabiayako: I've added some try catch blocks. I don't think it's a major issue not deleting the old entries, so I've just added a log when this happens. But when it fails to create a deliverable it would throw [adding try-catch blocks in update method](https://github.com/yldio/asap-hub/pull/3110/commits/aca564ceb11a3b54347bfeb2cc708b8ca974fdc2)

PR #3114, opened 2023-05-24
GP2-395 add contentful event data provider

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/GP2-395


apps/gp2-server/src/data-providers/contentful/event.data-provider.ts
```diff
@@ -0,0 +1,346 @@
...
+    ),
+    thumbnail: thumbnail?.url ?? undefined,
+    meetingLink: meetingLink || undefined,
+    hideMeetingLink: hideMeetingLink || false,
+    status: status as EventStatus,
+    tags: (tags as string[] | undefined | null) ?? [],
+
+    calendar,
+    speakers: parseGraphQLSpeakers(speakersItems),
+    // TODO: projects and workingGroups
```
markcsims, line 325: we should be able to do this now?

apps/gp2-server/src/data-providers/contentful/event.data-provider.ts
```diff
@@ -0,0 +1,346 @@
...
+    meetingMaterials: getContentfulEventMaterial<MeetingMaterial[], []>(
+      meetingMaterials,
+      !!meetingMaterialsPermanentlyUnavailable,
+      isStale,
+      [],
+    ),
+    thumbnail: thumbnail?.url ?? undefined,
+    meetingLink: meetingLink || undefined,
+    hideMeetingLink: hideMeetingLink || false,
+    status: status as EventStatus,
```
markcsims, line 320: maybe we don't need the cast as we have done the isEventStatus test earlier.

apps/gp2-server/src/routes/event.route.ts
```diff
@@ -15,17 +14,10 @@ export const eventRouteFactory = (
...
     async (req, res) => {
-      if (isContentfulEnabled) {
-        res.json({
-          total: 0,
-          items: [],
-        });
-      } else {
-        const query = validateEventFetchParameters(req.query);
-        const result = await eventController.fetch(query);
+      const query = validateEventFetchParameters(req.query);
```
markcsims, line 17: Thanks!

apps/gp2-server/src/data-providers/contentful/event.data-provider.ts
dianaguimaraes: Had to add it otherwise typescript validation complains

apps/gp2-server/test/fixtures/event.fixtures.ts
```diff
@@ -6,9 +7,257 @@ import {
...
+      url: 'https://link.pt/additional-material',
+    },
+  ],
+  calendar: {
+    googleCalendarId: 'c_t92qa82jd702q1fkreoi0hf4hk@group.calendar.google.com',
+    color: '#125A12' as const,
+    name: 'Tech 1 - Sequencing/omics',
+  },
+  thumbnail: {
+    url: 'https://images.ctfassets.net/5v6w5j61tndm/849855829/ac04300efa901ab22278b50d63bfde14/COSA-9900000000079e3c.png',
```
markcsims, line 192: not sure these links contain private information, should it be better to use something like http://example.com?

apps/gp2-server/test/fixtures/event.fixtures.ts
```diff
@@ -33,6 +282,30 @@ export const getEventSpeakerUser = (): gp2Model.EventSpeakerUser => ({
   avatarUrl: undefined,
 });
 
+export const getContentfulEventDataObject = (): gp2Model.EventDataObject => ({
+  ...getEventDataObject(),
+  ...eventUnreadyResponse,
+  notes: '<p>These are the notes from the meeting</p>',
+  presentation:
+    '<p><iframe src="https://drive.google.com/file/d/1euARcFkYJoqj4c_oor_nYIMBDWOfL-3x/preview"/></p>',
```
markcsims, line 290: not sure these links contain private information, should it be better to use something like http://example.com/?

markcsims (approved):

PR #3115, opened 2023-05-25
CRN-1400 - adds markup popup on the description field; fixes the mobile view of the Info component

andreicovaciu:


packages/react-components/src/atoms/Tooltip.tsx
andreicovaciu, line 1: This file was updated to make it work according to the design for all screen sizes. Seems that the tooltip never worked as expected

oratiu (approved):
looks good.

PR #3127, opened 2023-05-30
Crn 1358 reminders for the working group draft outputs on creation

oratiu:


apps/crn-server/src/controllers/reminders.ts
```diff
@@ -50,6 +50,30 @@ export default class Reminders implements ReminderController {
...
+          reminder.type === 'Draft'
+        ) {
+          if (reminder.data.associationType === 'team') {
+            return {
+              id: reminder.id,
+              entity: reminder.entity,
+              href: sharedResearch({}).researchOutput({
+                researchOutputId: reminder.data.researchOutputId,
+              }).$,
+              description: `**${reminder.data.createdBy}** on **${reminder.data.associationName}** created a draft output ${reminder.data.title}.`,
```
gabiayako, line 64: there's a colon missing
```suggestion
              description: `**${reminder.data.createdBy}** on **${reminder.data.associationName}** created a draft output: ${reminder.data.title}.`,
```

apps/crn-server/src/data-providers/reminders.data-provider.ts
```diff
@@ -214,6 +264,81 @@ const getResearchOutputRemindersFromQuery = (
...
+      const isInWorkingGroup = researchOutputWorkingGroups.some(
+        (workingGroup) => userWorkingGroupIds.includes(workingGroup),
+      );
+
+      if (
+        (associationType === 'team' && !isInTeam) ||
+        (associationType === 'working group' && !isInWorkingGroup)
+      ) {
+        return researchOutputReminders;
+      }
```
gabiayako, line 320: hey, reading the ticket again, I've noticed this "**incl. ASAP Hub role=Staff**"

> As a member of the team that is associated with the draft output (PM or Non-PM; **incl. ASAP Hub role=Staff**)

I would ask Tiff before implementing it, but if this requirement above is needed, I should see the reminder even if I don't belong to the working group or team, but I have Asap Role = Staff, so the logic here would change a little bit. Sorry, just noticed this now 😅

oratiu: Done, ty for noticing

apps/crn-server/test/data-providers/reminders.data-provider.test.ts
```diff
@@ -227,6 +261,373 @@ describe('Reminder Data Provider', () => {
...
+        squidexGraphqlClientMock.request.mockResolvedValueOnce(
+          squidexGraphqlResponse,
+        );
+
+        const result = await reminderDataProvider.fetch(fetchRemindersOptions);
+
+        expect(result).toEqual({ items: [], total: 0 });
+      });
+
+      test('Should return an empty result when research-output referencingTeamsContents property is null', async () => {
```
gabiayako, line 571:
```suggestion
      test('Should return an empty result when draft research-output team property is null', async () => {
```

gabiayako (approved):
Nice 🎉

apps/crn-server/test/integration/reminders.integration-test.ts
peterstarling, line 1: remember when I said that we enabled drafts for all reminder queries by doing this 
https://github.com/yldio/asap-hub/blob/3877efb8c161565e5481aaf4a0fffd4edb26c49a/apps/crn-server/src/data-providers/reminders.data-provider.ts#L70
?
I think you should add a test to the events and regular (published) data provider reminder to see if they come up when there is a draft that falls into the same query parameters. You already explained that for research-outputs we filter by added-date which the drafts should not have, but think about when admins would want to quickly disable an RO to hide it from the site - the reminder would still show up. Same goes for events, and I believe it's a common practice for them to turn something into draft (i could be wrong though)

peterstarling (changes requested):

apps/crn-server/test/integration/reminders.integration-test.ts
```diff
@@ -242,6 +251,253 @@ describe('Reminders', () => {
...
+        creatorId,
+      );
+      const researchOutputId = await researchOutputDataProvider.create(
+        researchOutputInput,
+        { publish: false },
+      );
+
+      const researchOutput = await researchOutputDataProvider.fetchById(
+        researchOutputId,
+      );
```
lennym, line 292: Because this data is created inside the test you will want to move the `fetchById` inside the `retryable` call in order to make sure it is not subject to a race condition with the graphql cache.

apps/crn-server/test/fixtures/reminders.fixtures.ts
```diff
@@ -41,6 +42,27 @@ export const getResearchOutputPublishedReminder =
     };
   };
 
+export const getResearchOutputDraftReminder = (
+  teamAssociation: boolean = true,
```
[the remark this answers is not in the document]
oratiu: Mutating the property in the test would look a bit ugly since the data is nested, thinking of just creating two functions

lennym (changes requested):
There are a few places in the integration tests where you create data and then immediately `fetchById` with the returned id.

This will cause a race condition with the graphql cache making the test liable to failure. To resolve always fetch data that has been created inside the test inside the `retryable` call.

apps/crn-server/test/integration/reminders.integration-test.ts
oratiu: Added test where I'm changing the published status from Published to Draft on multiple entities.

peterstarling (approved):

lennym (approved):

PR #3128, opened 2023-05-30
GP2 400 data provider outputs

markcsims:
- refactor output contoller to use DateProvider type
- starts moving over output data provider


packages/contentful/migrations/gp2/outputs/20230531101009-date-fields.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+  outputs.changeFieldControl('lastUpdatedPartial', 'builtin', 'datePicker', {
+    ampm: '12',
+    format: 'time',
+    helpText: 'Does not include changes to Publish Date and Admin notes',
+  });
+};
+
+module.exports.down = (migration) => {
+  const outputs = migration.editContentType('outputs');
```
dianaguimaraes, line 20: I'm confused with this down function 🤔
Why are we changing the validations on documentType and type since we're just updating the field control for dates on the up function?

markcsims: good point. I've removed them

PR #3129, opened 2023-05-31
[CT-36] feat: add sync all calendars github action

Riccardo-Brown:


packages/cms-data-sync/scripts/run-calendar-sync-all.ts
```diff
@@ -0,0 +1,12 @@
...
+import { getCalendarDataProvider } from '@asap-hub/crn-server/src/dependencies/calendars.dependencies';
+import { syncCalendars } from '../src/calendars/sync-all';
+
+const logLevel = process.env.LOG_LEVEL || 'info';
+const logEnabled = !!process.env.LOG_ENABLED || true;
+
+const logger = getPrettyLogger({ logEnabled, logLevel });
+const dataProvider = getCalendarDataProvider();
+
+syncCalendars({ dataProvider, logger }).catch(console.error);
```
lennym, line 12: This should probably exit with a non-zero code if there's an error so that the build job fails.

packages/cms-data-sync/src/calendars/sync-all.ts
```diff
@@ -0,0 +1,46 @@
...
+    logger.info('No calendars to sync');
+    return;
+  }
+
+  const promises = await Promise.allSettled(
+    calendars.map(async (calendar) => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      await dataProvider.update(calendar.sys.id, { googleApiMetadata: {} });
+      await rateLimiter.removeTokens(3);
```
lennym, line 31: This approach to rate limiting is probably redundant given the client is already [applying the same rate limits in the API adapter](https://github.com/yldio/asap-hub/blob/master/packages/cms-data-sync/src/utils/setup.ts#L21-L26).

However, if you are going to do it this way then you definitely need to remove the tokens _before_ making the update call, otherwise this map will fire off all the updates simultaneously without any rate limiting, and only rate limit the promise resolution _afterwards_.

packages/cms-data-sync/src/calendars/sync-all.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  promises.forEach((promise, i) => {
+    if (promise.status === 'rejected') {
+      const calendar = calendars[i];
+      logger.error(
+        promise.reason,
+        `Failed to sync calendar: ${calendar.sys.id}`,
+      );
+    }
+  });
```
lennym, line 43: Again we probably do want to throw errors here if syncing fails rather than swallowing them.

packages/cms-data-sync/package.json
```diff
@@ -12,11 +12,14 @@
     "build:babel": "../../scripts/build-babel.sh build",
     "watch:babel": "../../scripts/build-babel.sh watch",
     "migrate": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' ts-node ./src/index.ts",
-    "schema:update:crn": "SQUIDEX_BASE_URL=$SQUIDEX_BASE_URL SQUIDEX_APP_NAME=$CRN_SQUIDEX_APP_NAME SQUIDEX_CLIENT_ID=$CRN_SQUIDEX_CLIENT_ID SQUIDEX_CLIENT_SECRET=$CRN_SQUIDEX_CLIENT_SECRET ./scripts/schema-update.sh"
+    "schema:update:crn": "SQUIDEX_BASE_URL=$SQUIDEX_BASE_URL SQUIDEX_APP_NAME=$CRN_SQUIDEX_APP_NAME SQUIDEX_CLIENT_ID=$CRN_SQUIDEX_CLIENT_ID SQUIDEX_CLIENT_SECRET=$CRN_SQUIDEX_CLIENT_SECRET ./scripts/schema-update.sh",
+    "calendar:sync:all": "ts-node -T ./scripts/run-calendar-sync-all.ts"
```
peterstarling, line 16: i actually think we should do type-checking when running these

packages/cms-data-sync/src/calendars/sync-all.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function syncCalendars({
+  dataProvider,
+  logger,
+}: {
+  dataProvider: CalendarDataProvider;
+  logger: Logger;
+}) {
+  const { contentfulEnvironment } = await getContentfulClient();
+
+  const calendars = await fetchContentfulEntries(
```
peterstarling, line 16: can't you use the data provider here too?

peterstarling (changes requested):

packages/cms-data-sync/src/calendars/sync-all.ts
Riccardo-Brown: Would it make sense to log all errors then if there has been an error throw an error. So we get all errors if there are multiple?

packages/cms-data-sync/src/calendars/sync-all.ts
Riccardo-Brown: This function ensures we get all the calendar, i don't think it makes sense to duplicate the function just so it uses the data provider.

peterstarling (approved):

PR #3130, opened 2023-06-01
CT-17 Working Groups webhook

gabiayako:
This PR updates the handler `update-deliverables-handler` so it is able to process events from squidex and contentful and update the deliverables from working groups.

Testing:
- Create a working group in Contentful with 
  - complete equals false
  - a deliverable with status `Pending`
  - a deliverable with status `In progress`
- Publish this working group entry
- Change the complete to true and publish the entry again
- Check that the deliverables are now with status `Not started` and `Incomplete` respectively

### DIFF
--- a/apps/crn-server/src/handlers/working-group/update-deliverables-handler.ts
+++ b/apps/crn-server/src/handlers/working-group/update-deliverables-handler.ts
@@ -3,11 +3,6 @@ import 'source-map-support/register';
 import { Handler } from 'aws-lambda/handler';
 import { EventBridgeHandler } from '@asap-hub/server-common';
 import {
-  RestWorkingGroup,
-  SquidexGraphql,
-  SquidexRest,
-} from '@asap-hub/squidex';
-import {
   WorkingGroupDeliverable,
   DeliverableStatus,
   WorkingGroupEvent,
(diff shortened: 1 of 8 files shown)


apps/crn-server/src/handlers/working-group/update-deliverables-handler.ts
```diff
@@ -29,19 +22,15 @@ export const workingGroupUpdateHandler =
     dataProvider: WorkingGroupDataProvider,
   ): EventBridgeHandler<WorkingGroupEvent, WorkingGroupPayload> =>
   async (event) => {
+    const workingGroupId = event.detail.resourceId;
     logger.info(
-      `Received ${event.detail.type} event for ${event.detail.payload.data?.title?.iv}`,
```
gabiayako, line 33: I'm changing from `title` to `id` because the structure of the event is different in Squidex and Contentful.

In contentful we would get the title by doing something like `event.detail.fields.title['en-US']`. Since the advantage of displaying the title instead of the id is not significantly greater, I've changed to id, that we can get in the same way in both squidex and contentful.

apps/crn-server/test/handlers/working-group/update-deliverables-handler.test.ts
```diff
@@ -1,31 +1,36 @@
...
 
   afterEach(() => jest.clearAllMocks());
 
-  test('updates Pending/In Progress deliverable statuses to Not Started/Incomplete if working group is completed', async () => {
+  test(`updates Pending/In Progress deliverable statuses to Not Started/Incomplete if working group is completed for a ${cms} event`, async () => {
     const deliverables: WorkingGroupDeliverable[] = [
       { description: 'A pending deliverable', status: 'Pending' },
       { description: 'An in progress deliverable', status: 'In Progress' },
     ];
-    const event = getWorkingGroupEvent({ complete: true, deliverables });
```
gabiayako, line 20: we don't really use the value of `complete` and `deliverables`  from the event. We get those values from data provider entry. So I'm removing them from the tests because they don't make a difference in what we're expecting.

Riccardo-Brown (approved):

PR #3131, opened 2023-06-01
Add CMS migration for interest groups

lennym:


packages/cms-data-sync/src/interest-groups/interest-groups.data-migration.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    let calendar;
+    if (calendars && calendars.length && calendars[0].id) {
+      calendar = {
+        sys: {
+          type: 'Link',
+          linkType: 'Entry',
+          id: calendars[0].id,
+        },
+      };
```
gabiayako, line 113: It shouldn't happen if migrated in the right order, but in the events migration https://github.com/yldio/asap-hub/blob/CT-14/interest-group-migration/packages/cms-data-sync/src/events/events.data-migration.ts#L190 I've checked if the calendar exists first then I would create a link. I don't think this is super necessary, but I just wonder what would happen if we link to a calendar that does not exist.

gabiayako (approved):
I'm approving because I don't believe the comments above are blockers 😉

PR #3136, opened 2023-06-05
feat: CT-20 Create Tutorials Content Model

peterstarling:


apps/crn-server/src/app.ts
```diff
@@ -468,6 +464,27 @@ export const appFactory = (libs: Libs = {}): Express => {
...
+      new TutorialsSquidexDataProvider(squidexGraphqlClient),
+    'IS_CONTENTFUL_ENABLED_V2',
+    false,
+  );
+  featureFlagDependencySwitch.setDependency(
+    'tutorials',
+    libs.tutorialsContentfulDataProvider ||
+      new TutorialsContentfulDataProvider(contentfulGraphQLClient),
+    'IS_CONTENTFUL_ENABLED_V2',
+    true,
```
gabiayako, line 479: It is still necessary to add the header here https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/discover/tutorials/api.ts#L9 so we can see the data coming from Contentful using the `ASAP_CONTENTFUL` cookie.

gabiayako (changes requested):
Looks good! Just requesting changes because the header that makes `ASAP_CONTENTFUL` cookie work is missing

apps/crn-server/src/app.ts
lennym: Done.

gabiayako (approved):

PR #3145, opened 2023-06-06
feat: CT-48 Add polling to Contentful webhook

peterstarling:


Riccardo-Brown (approved):

lennym:
My only slight concern is that a fairly common pattern is to unpublish and then delete immediately, which would cause the polling to throw "Not Found" if the entry had been deleted before the webhook was processed.

gabiayako (approved):

lennym (approved):

packages/server-common/src/handlers/webhooks/contentful.handler.ts
```diff
@@ -25,44 +29,86 @@ const getDetailFromRequest = (
...
+      environment: config.environment,
+    });
+    const fetchEntryById = async () => {
+      try {
+        const entry = await cpaClient.getEntry(detail.resourceId);
+        return entry;
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message.match(/The resource could not be found/)
```
lennym, line 71: We might also want to check that the event being handled is an `Unpublish` before continuing here.

lennym (approved):

PR #3152, opened 2023-06-07
CT-62 Fix conflict issue due to same asset/media in multiple entries

gabiayako:
When several events have the same thumbnail or media attached we could have conflict errors like in this action run https://github.com/yldio/asap-hub/actions/runs/5187380563/jobs/9374300439#step:6:8913 although we're already checking if the asset or entry exist.

So I'm wrapping the createById in a try catch block. Another error that appears is when we call `processForAllLocales` for the same entry in sequence, so I'm also wrapping it in a try catch block.

Example of successful run after the changes in this PR: https://github.com/yldio/asap-hub/actions/runs/5211680138/jobs/9404274226


packages/cms-data-sync/src/utils/assets-and-media.ts
```diff
@@ -0,0 +1,197 @@
...
+    try {
+      console.log(`Creating ${type} with id ${id}.`);
+
+      const entry = await createFn();
+      await processAndPublishEntry(entry);
+    } catch (error) {
+      if (error instanceof Error) {
+        const errorParsed = JSON.parse(error?.message);
+        if (errorParsed.status === 409) {
+          handleErrorFn(contentfulEnvironment, id);
```
Riccardo-Brown, line 82: Do we need to wait for the `handleErrorFn`?
```suggestion
          await handleErrorFn(contentfulEnvironment, id);
```

packages/cms-data-sync/src/utils/assets-and-media.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+    throw error;
+  }
+};
+
+export const createAndPublishIfNonExistent = async (
+  type: 'asset' | 'media',
+  contentfulEnvironment: Environment,
+  id: string,
+  checkIfExistsFn: (
```
peterstarling, line 60: why make this a dependency? any legitimate case for when this isn't the `checkIfAssetAlreadyExistsInContentful` from above?

peterstarling (approved):
just a small comment. This PR is actually much smaller than what it looks like, its just that you moved some functionality AND added bits in between which in result looks like it is a one big addition and removal of functions

packages/cms-data-sync/src/utils/assets-and-media.ts
gabiayako: because for the media we need to check if the `entry` exists not the `asset`, so we pass `checkIfEntryAlreadyExistsInContentful` here

PR #3157, opened 2023-06-08
CT-22: add content model for discover ASAP page

lennym:


packages/contentful/src/crn/queries/discover.queries.ts
```diff
@@ -0,0 +1,96 @@
...
+    alumniSinceDate
+    createdDate
+    degree
+    email
+    firstName
+    institution
+    jobTitle
+    lastName
+    onboarded
+    role
```
gabiayako, line 26: looking at the query from crn (squidex) it seems that we're not retrieving some of these fields
```suggestion
    email
    firstName
    institution
    jobTitle
    lastName
```

gabiayako (changes requested):
I've just pointed out some tweaks in the query, other than that, it looks really good :)

packages/contentful/src/crn/queries/discover.queries.ts
lennym: Yeah, I got thrown off by [the unit test fixtures](https://github.com/yldio/asap-hub/blob/master/apps/crn-server/test/fixtures/discover.fixtures.ts#L117-L177) defining _everything_.

I've removed the extra properties now.

gabiayako (approved):

PR #3159, opened 2023-06-08
create research output dropdown

kmaid:
Mobile size of menu. I think they want the dropdown menu to match button size.
Dropdown menu button colour (Dark grey for team lighter grey for output type)

Consider users who don't have permission to do anything anywhere lol. Button should be hidden.
Push back on half a row thing.

Unit testing. Start with Dropdown changes and then move to the new component


oratiu (approved):
Looks good

PR #3160, opened 2023-06-09
Feat/crn 1406

oratiu:


packages/react-components/src/templates/ResearchOutputForm.tsx
```diff
@@ -336,6 +336,7 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
                 const { id } = researchOutput;
                 const savePath = sharedResearch({}).researchOutput({
                   researchOutputId: id,
+                  draftCreated: draftSave && !researchOutputData?.created,
```
gabiayako, line 339: I've noticed that when we create a draft from a duplicate the toast does not appear because `researchOutputData?.created` is populated when it comes from a duplicate

lctrt (approved):

packages/react-components/src/templates/ResearchOutputForm.tsx
oratiu: @gabiayako do we know if this is the intended action from duplicate ? Should Research Output B which is duplicated from Research Output A, also duplicate the created date of Research Output A, because Research Output B gets created at that specific moment.

gabiayako: > do we know if this is the intended action from duplicate ? 
No, I don't know for sure. It's better to ask Tiff 🤔 I was assuming the banner should appear when a draft is created from an existing output because for me it's a draft like any other one. But it's better to ask Tiff

oratiu: nvm, that's not the issue, we do have the right created date on the new copied RO, but we also have the created date on researchOutputData

packages/validation/src/permissions/research-output.ts
```diff
@@ -2,9 +2,42 @@ import { User } from '@asap-hub/auth';
...
+): boolean => {
+  if (user === null) {
+    return false;
+  }
+  return user[association].some(
+    (teamOrWorkingGroup) =>
+      associationIds?.includes(teamOrWorkingGroup.id) &&
+      teamOrWorkingGroup.role !== 'Project Manager',
+  );
+};
```
gabiayako, line 36: To be honest, I'm not a big fan of this refactor because we're not using `isUserProjectManager` separated from asap role = staff anywhere now and this might make it easier to use only `isUserProjectManager` and lead to error in the permission.

Also, I think it's a little misleading having `teamOrWorkingGroup.role !== 'Project Manager'` in `isUserMember` because a PM is still a member of the team/wg, but he/she is also a PM 😝 

I would remove `isUserProjectManager` and `isUserMember` and leave `getUserRole` the same as before

packages/react-components/src/templates/SharedResearchOutput.tsx
```diff
@@ -133,27 +184,57 @@ const SharedResearchOutput: React.FC<SharedResearchOutputProps> = ({
...
+                  {actionIcon} Ready for PM Review
+                </Button>
+              </div>
             )}
           </div>
         )}
+        {displayModal && (
+          <ConfirmModal
+            title="Output Ready for PM Review?"
+            description="All team members listed on this output will be notified and PMs will be able to review and publish this output."
```
gabiayako, line 227: since we're already checking if the association is team or working group above to show the text in the toast, couldn't we check here as well? :) 

So we would have `All team members...` or `All working group members...` and then this ticket would address team and working groups already, right?

gabiayako (changes requested):

packages/validation/src/permissions/research-output.ts
oratiu: Ok I will drop the changes done here

packages/validation/src/permissions/research-output.ts
```diff
@@ -36,6 +36,9 @@ export const getUserRole = (
   return 'None';
 };
 
+export const hasReadyDraftForReviewPermission = (userRole: UserRole): boolean =>
+  isEnabled('DRAFT_RESEARCH_OUTPUT') && userRole === 'Member';
```
gabiayako, line 40:
```suggestion
export const hasRequestForReviewPermission = (userRole: UserRole): boolean =>
  isEnabled('DRAFT_RESEARCH_OUTPUT') && userRole === 'Member';
```

gabiayako (approved):
Nice! 🎉

PR #3162, opened 2023-06-09
Add flags to make migrations optional

lennym:
Most of the time when working on a migration you are only interested in migrating the one entity type that you're working on. This provides a quick way to selectively migrate only one model type without having to comment out lots of code.

Omitting any flags defaults to migrating everything.

Example:

```
yarn workspace @asap-hub/cms-data-sync migrate --teams
```

```
yarn workspace @asap-hub/cms-data-sync migrate --teams --users
```

### DIFF
--- a/packages/cms-data-sync/src/run-migrations.ts
+++ b/packages/cms-data-sync/src/run-migrations.ts
@@ -12,7 +12,23 @@ import { migrateTutorials } from './tutorials/tutorials.data-migration';
 import { logger } from './utils';
 import { contentfulRateLimiter } from './contentful-rate-limiter';
 
-export const runMigrations = async (flags: string[] = []) => {
+export const models = [
+  'teams',
+  'externalAuthors',
+  'calendars',
+  'labs',
+  'users',
+  'events',
+  'interestGroups',
+  'workingGroups',
+  'tutorials',
+];
+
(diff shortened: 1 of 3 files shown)


Riccardo-Brown (approved):

packages/cms-data-sync/src/run-migrations.ts
```diff
@@ -12,7 +12,7 @@ import { migrateTutorials } from './tutorials/tutorials.data-migration';
 import { logger } from './utils';
 import { contentfulRateLimiter } from './contentful-rate-limiter';
 
-export const runMigrations = async () => {
+export const runMigrations = async (flags: string[] = []) => {
```
peterstarling, line 15: I would do
```suggestion
export const runMigrations = async (flags: Flag[] = []) => {
```
then then somewhere else:
```typescript
const flags = [
 "teams",
 (....)
] as const;

export type Flag = (typeof flags)[number];
```
and then in your index file you could add a very simple validation after https://github.com/yldio/asap-hub/blob/f50fdd36bd4ace1bf9efda0b6a60cab623954cdb/packages/cms-data-sync/src/index.ts#L4 to make sure that the string is then recognised as an element from the union

peterstarling (approved):

PR #3177, opened 2023-06-13
refactor: add working and interest group to user response

Riccardo-Brown:


packages/contentful/src/crn/queries/users.queries.ts
```diff
@@ -57,6 +57,17 @@ export const usersContentQueryFragment = gql`
           }
           displayName
           inactiveSince
+          linkedFrom {
+            interestGroupsCollection {
```
lennym, line 61: I think the default limit is 100 anyway, but for consistency with working groups.

```suggestion
            interestGroupsCollection(limit: 100) {
```

PR #3181, opened 2023-06-13
CT-25 - Add content model for research tags

lennym:


packages/cms-data-sync/test/research-tags/research-tags.data-migration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      queryResearchTagsContentsWithTotal: null,
+    });
+
+    await migrateResearchTags();
+
+    expect(contentfulEnv.createEntryWithId).not.toHaveBeenCalled();
+  });
+
+  describe('creates contentful entries', () => {
+    it('for a tutorial', async () => {
```
peterstarling, line 133: tutorial?

packages/cms-data-sync/test/run-migrations.test.ts
```diff
@@ -131,6 +132,16 @@ jest.mock('../src/discover/discover.data-migration', () => {
   };
 });
 
+var mockMigrateResearchTags: jest.MockedFunction<typeof migrateResearchTags>;
```
peterstarling, line 135: have you tried using const for this? i tried in here https://github.com/yldio/asap-hub/blob/25f049fcd703e11c0ec19023ddef05d8d604b4d4/packages/server-common/test/handlers/webhooks/contentful.handler.test.ts#L17 and it worked

peterstarling (approved):

packages/cms-data-sync/test/run-migrations.test.ts
lennym: See above re. wholesale copy-pasting. I'll probably change them all because there's no good reason to use `var` that I'm aware of.

Riccardo-Brown (approved):

packages/cms-data-sync/test/run-migrations.test.ts
Riccardo-Brown: nostalgia

lennym: So apparently there is a good reason. If I either swap out the `var` directly for a `let`, or `const` with the assignment to `jest.fn()` moved inline then I get the following error.

```
ReferenceError: Cannot access 'mockMigrateEvents' before initialization
```

I think it works in the example you linked because the mock is the return value of an imported function rather than the import itself.

PR #3186, opened 2023-06-14
[CT-57] refactor: add working and interest group to user response

Riccardo-Brown:
This ensure the profile dropdown get populated when using Contentful CMS.


apps/crn-server/src/data-providers/contentful/users.data-provider.ts
```diff
@@ -340,3 +391,75 @@ const parseOrcidWorksContentful = (
...
+        role: (user as GroupLeaderItem).role || 'Member',
+        active: workingGroup?.complete ? false : !isAlumni && !isInActive,
+      } as WorkingGroupMembership);
+
+      return workingGroups;
+    },
+    [],
+  );
+
+const parseToInterestGroups = (teams: TeamMembership[]) =>
```
peterstarling, line 419: I would prefer to see the return types on these functions (same as above) - reason being - I have a feeling TypeScript might often infer it to something that is not quite what we are looking for and then when spreading that onto the end object we might never realise it was invalid

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -51,7 +51,35 @@ describe('User data provider', () => {
           contentfulRestClientMock,
         );
       const result = await userDataProviderWithMockServer.fetchById('123');
-      const expectation = getUserDataObject();
+      const expectation = {
+        ...getUserDataObject(),
+        workingGroups: [
```
peterstarling, line 56: looks like this is part of the standard return fixture now, why don't you just extend the USerDataObject fixture then so you don't have to repeat it all the time?

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
peterstarling, line 1: with all the logic added to the data provider i would expect a few more tests to cover all the different cases, at a minimum you should test the alumni logic, i think

packages/contentful/src/crn/queries/users.queries.ts
```diff
@@ -49,14 +49,25 @@ export const usersContentQueryFragment = gql`
     avatar {
       url
     }
-    teamsCollection(limit: 100) {
+    teamsCollection(limit: 30) {
```
peterstarling, line 52: meh - I don't think anyone should have more than 5, maybe 10

peterstarling (changes requested):

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
Riccardo-Brown: i think i did thay before and ended up breaking at lot of test as this fixture is used across different entities.

peterstarling (approved):

PR #3187, opened 2023-06-14
CT-43 Contenful Extension - Date fields as firstPublishedAt

gabiayako:
This PR

- Adds an app extension that updates a date field to be equal to `firstPublishedAt` when this field isn't populated previously
- Adds the migration for 
  - `news`: changes `publishDate` field control to use this new app extension and change it from required to not required, as the user could not publish the entry if this field was required, then the field would never be populated
  - `users`: changes `createdAt` field control to use this new app extension


packages/contentful-app-extensions/date-field-as-first-published-at/src/locations/Field.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [field, setField] = useState(sdk.field.getValue());
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  let unsubscribe = () => {};
+
+  unsubscribe = sdk.entry.onSysChanged(async (sys: EntrySys) => {
+    if (!field && sys.firstPublishedAt) {
+      setField(sys.firstPublishedAt);
+      await sdk.field.setValue(sys.firstPublishedAt);
+      await sdk.entry.publish();
```
Riccardo-Brown, line 18: Why are we publishing the entry? I thought the user had to press the publish button on the form

gabiayako: The flow would be something like
- The user publishes the entry
- The field is set with the value `firstPublishedAt`
- The entry status becomes changed

if I don't publish in this line 18 the status remains as changed. With this `publish` the status becomes `published` without the user having to publish two times in a row :)

Riccardo-Brown (approved):

PR #3218, opened 2023-06-21
CRN-1078 add guides to contentful

lctrt:


apps/crn-frontend/src/dashboard/__tests__/Dashboard.test.tsx
```diff
@@ -34,6 +36,7 @@ const mockGetDashboard = getDashboard as jest.MockedFunction<
 const mockGetReminders = getReminders as jest.MockedFunction<
   typeof getReminders
 >;
+// const mockGetGuides = getGuides as jest.MockedFunction<typeof getGuides>;
```
lennym, line 39: You _probably_ want to clean out these comments if they're not being used.

apps/crn-frontend/src/guides/api.ts
```diff
@@ -0,0 +1,24 @@
...
+import { API_BASE_URL } from '../config';
+
+export const getGuides = async (
+  authorization: string,
+): Promise<ListGuideResponse> => {
+  const resp = await fetch(`${API_BASE_URL}/guide`, {
+    headers: {
+      authorization,
+      ...createSentryHeaders(),
+      ...createFeatureFlagHeaders(),
```
lennym, line 15: You won't need the feature flag because guides won't be implemented in squidex.

apps/crn-server/src/data-providers/contentful/guides.data-provider.ts
```diff
@@ -0,0 +1,47 @@
...
+});
+
+export const parseGraphQLGuide = (guide: GuideItem): GuideDataObject => ({
+  title: guide?.title ? guide.title : '',
+  content: guide?.contentCollection
+    ? guide.contentCollection.items.map((content: GuideContentItem | null) =>
+        parseGuideContent(content),
+      )
+    : [],
+});
```
lennym, line 44: I think you'll want to remove any null items from the response rather than populating with all empty strings.

peterstarling: that's actually debatable, for some reason this has been the way we have done things in CRN although I would say that unless its a single value normalised to an empty string it should throw some sort of exception (As in when all values are undefined we should throw)

peterstarling (approved):

lennym (approved):

PR #3220, opened 2023-06-21
GP2 import scripts

markcsims:
- allows for updates to user members
- refactors update project members
- fixes broken tests
- adds update members to working groups

### DIFF
--- a/apps/gp2-server/src/controllers/project.controller.ts
+++ b/apps/gp2-server/src/controllers/project.controller.ts
@@ -47,7 +47,9 @@ export default class Projects implements ProjectController {
     update: gp2.ProjectUpdateRequest,
     loggedInUserId: string,
   ): Promise<gp2.ProjectResponse> {
-    await this.projectDataProvider.update(id, update);
+    await this.projectDataProvider.update(id, {
+      ...update,
+    });
     return this.fetchById(id, loggedInUserId);
   }
 }
--- a/apps/gp2-server/src/data-providers/contentful/utils/common.ts
+++ b/apps/gp2-server/src/data-providers/contentful/utils/common.ts
@@ -1,0 +1,29 @@
+import { Entry, Environment } from '@asap-hub/contentful';
+
+export const deleteEntities = async (
+  idsToDelete: string[],
+  environment: Environment,
+) =>
+  Promise.all(
+    idsToDelete.map(async (id) => {
+      const deletable = await environment.getEntry(id);
+      await deletable.unpublish();
+      return deletable.delete();
+    }),
+  );
+export const getIdsToDelete = (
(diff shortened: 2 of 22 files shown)


dianaguimaraes (approved):

apps/gp2-server/src/controllers/project.controller.ts
```diff
@@ -47,7 +47,9 @@ export default class Projects implements ProjectController {
     update: gp2.ProjectUpdateRequest,
     loggedInUserId: string,
   ): Promise<gp2.ProjectResponse> {
-    await this.projectDataProvider.update(id, update);
+    await this.projectDataProvider.update(id, {
+      ...update,
```
jhonny-m, line 51: whats the difference?

jhonny-m (approved):

apps/gp2-server/src/data-providers/contentful/utils/common.ts
```diff
@@ -0,0 +1,27 @@
...
+) =>
+  Promise.all(
+    idsToDelete.map(async (id) => {
+      const deletable = await environment.getEntry(id);
+      await deletable.unpublish();
+      return deletable.delete();
+    }),
+  );
+export const getIdsToDelete = (
+  previous: { id?: string }[] | undefined,
```
peterstarling, line 15:
```suggestion
  previous: string[]
```
?

apps/gp2-server/src/data-providers/contentful/utils/members.ts
```diff
@@ -1,41 +1,130 @@
...
       if (!user?.onboarded) {
         return membersList;
       }
-      const groupMember = parseMember<T>(user, member.role);
+      const groupMember = parseMember<T>(member.sys.id, user, member.role);
       return [...membersList, groupMember];
     }, []) || [];
+
+const addNextMember = async <T extends string>(
+  environment: Environment,
```
peterstarling, line 51: looks to me like all these could be members of a class which takes environment as a dependency, or they could be part of the data provider which has that environment in the scope

packages/server-common/src/utils/csv-parse.ts
```diff
@@ -14,7 +14,7 @@ export const parse =
       pump(
         createReadStream(src),
         csvParse({ from_line: 2 }),
-        through.obj({ maxConcurrency: 10 }, async (chunk, _, callback) => {
+        through.obj({ maxConcurrency: 1 }, async (chunk, _, callback) => {
```
peterstarling, line 17: is this used anywhere else? are we ok to reduce it to 1 like this?

peterstarling (approved):

PR #3236, opened 2023-06-27
CT-21 Contentful Reminders Data Provider (Events)

gabiayako:
Reminders from Contentful Events appearing in the hub
Screenshot 2023-07-05 at 13 31 42


apps/crn-server/test/data-providers/contentful/reminders.data-provider.test.ts
```diff
@@ -0,0 +1,1264 @@
...
+                items: [eventMockResponse],
+              },
+              users,
+            });
+
+            const result = await remindersDataProvider.fetch(
+              fetchRemindersOptions,
+            );
+
+            expect(result.items.map((r) => r.type)).not.toContain(
```
peterstarling, line 368: you sometimes use this pattern for your expectation but the other time you just assert that no results were returned. I would lean towards the second one although it might leave less room for extension later, question is if maybe you should only use one of the two?

apps/crn-server/test/data-providers/contentful/reminders.data-provider.test.ts
```diff
@@ -0,0 +1,1264 @@
...
+        const result = await remindersDataProvider.fetch(fetchRemindersOptions);
+
+        expect(result.items.map((r) => r.type)).not.toContain(
+          'Share Presentation',
+        );
+      });
+    });
+
+    describe('Publish Material Reminder', () => {
+      describe('When less than 72 hours have passed since the event ended', () => {
```
peterstarling, line 514: i think this describe is unnecessary (or simply wrong) because later you test cases for more than 72 hours have passed

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/reminders.data-provider.ts
```diff
@@ -0,0 +1,463 @@
...
+};
+
+const inLast24Hours = (date: string, zone: string) => {
+  const convertedDate = convertEventDate(date, zone);
+  const { last24HoursISO, now } = getReferenceDates(zone);
+
+  return convertedDate >= last24HoursISO && convertedDate <= now;
+};
+
+const getEventHappeningRemindersFromQuery = (
```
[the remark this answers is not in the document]
gabiayako: I called only `Happening` because it returns `Happening Today` and `Happening Now` 😛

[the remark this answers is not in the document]
gabiayako: But I can change to `getEventHappeningNowOrTodayRemindersFromQuery`

apps/crn-server/test/fixtures/reminders.fixtures.ts
```diff
@@ -375,3 +379,108 @@ export const getSquidexReminderEventsContents = (): NonNullable<
     },
   };
 };
+
+export const getContentfulRemindersGraphqlResponse =
+  (): FetchRemindersQuery => ({
+    eventsCollection: {
+      items: [getContentfulReminderEventsCollectionItem()],
+    },
```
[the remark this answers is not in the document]
gabiayako: hmm I've ended up never using this fixture, in the tests since I change either the content of the events or users I've ended up writing this in the test itself. I think maybe it's better to remove `getContentfulRemindersGraphqlResponse` then

apps/crn-server/test/data-providers/contentful/reminders.data-provider.test.ts
```diff
@@ -0,0 +1,1264 @@
...
+
+        expect(result.items.map((r) => r.type)).not.toContain(
+          'Happening Today',
+        );
+      });
+
+      test('Should not fetch the reminder if it is a future event', async () => {
+        const startDate = '2023-01-01T08:00:00Z';
+        const endDate = '2023-01-01T10:00:00Z';
+        jest.setSystemTime(DateTime.fromISO('2022-12-31T08:01:00Z').toJSDate());
```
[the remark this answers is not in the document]
gabiayako: The dates are different 😉 
Screenshot 2023-06-29 at 09 18 42

apps/crn-server/src/data-providers/contentful/reminders.data-provider.ts
```diff
@@ -0,0 +1,463 @@
...
+          eventId: event.sys.id,
+          title: event.title || '',
+          endDate: event.endDate,
+        },
+      }))
+    : [];
+
+const getUploadPresentationRemindersFromQuery = (
+  eventsEndedInLast72Hours: EventItem[],
+  users: User,
```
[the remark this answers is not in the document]
gabiayako: Solved here
- [Update apps/crn-server/src/data-providers/contentful/reminders.data-p…](https://github.com/yldio/asap-hub/pull/3236/commits/0c7c7fdcd1d27d785d4c13bf74525cdfd3fbfa22)
- [Update apps/crn-server/src/data-providers/contentful/reminders.data-p…](https://github.com/yldio/asap-hub/pull/3236/commits/0c7c7fdcd1d27d785d4c13bf74525cdfd3fbfa22)

apps/crn-server/src/data-providers/contentful/reminders.data-provider.ts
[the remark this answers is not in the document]
gabiayako: Solved here [review: rename event happening to event happening now or today](https://github.com/yldio/asap-hub/pull/3236/commits/eed0cfe587d1ea31f54b5abff0e65d5c3276d427)

apps/crn-server/test/fixtures/reminders.fixtures.ts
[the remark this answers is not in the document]
gabiayako: solved here [remove fixture getContentfulRemindersGraphqlResponse](https://github.com/yldio/asap-hub/pull/3236/commits/d055b1047af5eb0c18cee780666d562ce448d84b)

apps/crn-server/test/data-providers/contentful/reminders.data-provider.test.ts
gabiayako: I prefer to use `not.toContain` because sometimes like in the test below, I am testing the this type of reminder shouldn't appear if the event hasn't ended. But given the conditions, other type of reminder could appear, like "Happening Now". So the response does not come empty
Screenshot 2023-06-29 at 09 37 51

I've changed all the empty responses to `not.toContain` [review: change expect empty result to not.toContain reminder type](https://github.com/yldio/asap-hub/pull/3236/commits/db37835885729c3e4c544dffc9a4b813db85d695)

apps/crn-server/test/data-providers/contentful/reminders.data-provider.test.ts
gabiayako: removed the describes here [remove nested describes indicating time since reference date](https://github.com/yldio/asap-hub/pull/3236/commits/5bc3a9f2789a14eb5138bdfe61d1c836a628df13)

apps/crn-server/test/integration/common/reminders.integration-test.ts
```diff
@@ -0,0 +1,583 @@
...
+    app = appFactory({
+      authHandler: (req, _res, next) => {
+        req.loggedInUser = loggedInUser as UserResponse;
+        next();
+      },
+    });
+  });
+
+  afterAll(async () => {
+    jest.useRealTimers();
```
peterstarling, line 82: don't we want to teardown everything, ie use ```await fixtures.teardown();```

apps/crn-server/test/integration/common/reminders.integration-test.ts
```diff
@@ -0,0 +1,583 @@
...
+        expressApp,
+      )
+        .get(`/reminders?timezone=${timezone}`)
+        .expect(200);
+
+      expect(response.body.items.map((reminder) => reminder.id)).toContain(id);
+    });
+  };
+
+  const expectNotToContainingReminderWithId = async (
```
peterstarling, line 565:
```suggestion
  const expectNotToContainReminderWithId = async (
```

peterstarling (changes requested):

apps/crn-server/test/integration/common/reminders.integration-test.ts
gabiayako: committed here [review: rename expectNotToContainingReminderWithId -> expectNotToCont…](https://github.com/yldio/asap-hub/pull/3236/commits/bd0cd42ef60e1afad3e15b1c35daf6b617b3b24f) :)

apps/crn-server/test/integration/common/reminders.integration-test.ts
gabiayako: committed here [add teardown](https://github.com/yldio/asap-hub/pull/3236/commits/04bd546215dda035793a31f6a812b6b09314ade0)

apps/crn-server/test/helpers/teardown.ts
```diff
@@ -23,9 +23,11 @@ export const teardownHelper = (
   }
 
   return async () => {
-    for (const record of toDelete) {
-      await record.client.delete(record.id);
-    }
+    try {
```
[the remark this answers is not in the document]
gabiayako: hmm the issue with this line is that since I'm deleting the events while the tests are running, in the teardown the events ids are still in the array, so it gives an error because it can't find the record to delete it. Do you think it's a major problem to leave it like that?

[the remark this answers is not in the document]
peterstarling: im not sure i fully understand, technically this shouldn't be a huge deal cause the most likely reason this fails is because the records has already been deleted or is missing, but maybe at least log a warning for some visibility?

peterstarling (approved):

apps/crn-server/test/helpers/teardown.ts
```diff
@@ -23,9 +23,11 @@ export const teardownHelper = (
...
+      for (const record of toDelete) {
+        await record.client.delete(record.id);
+      }
+    } catch {}
```
lennym, line 30: Shouldn't the try-catch be inside the loop? This way it will stop iterating if it fails on any one deletion, but if you move the try-catch to inside the for-of then it will still (attempt to) delete all records.

apps/crn-server/test/helpers/teardown.ts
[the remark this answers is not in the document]
gabiayako: I think I only understood your comment now after this comment from Lenny https://github.com/yldio/asap-hub/pull/3236#discussion_r1254459362 😅 
The try catch was indeed out of place. I added the log here [review: add log message when deleting fixture fails](https://github.com/yldio/asap-hub/pull/3236/commits/16fa09ffd7de0c77e3f51b0426d167cb294a8ded)

apps/crn-server/test/integration/common/reminders.integration-test.ts
```diff
@@ -0,0 +1,580 @@
...
+      jest.setSystemTime(new Date('2022-08-10T11:05:00.0Z'));
+      await expectNotToContainReminderWithId(`event-happening-now-${event.id}`);
+    });
+
+    test('Should not see the reminder when the event is happening but its status is Draft', async () => {
+      // setting system time to 10:05AM in UTC
+      // event happening at 10PM in UTC and ending at 11AM UTC
+      jest.setSystemTime(new Date('2022-08-10T10:05:00.0Z'));
+      await fixtures.publishEvent(event.id, 'Draft');
+      await delay(1000);
```
lennym, line 208: This is to give time for the cache to clear so that the negative assertion doesn't trivially pass, right? It might be worth just adding a comment to explain it because it looks redundant at first glance.

apps/crn-server/test/helpers/teardown.ts
```diff
@@ -23,11 +23,14 @@ export const teardownHelper = (
...
 
   return async () => {
-    try {
-      for (const record of toDelete) {
+    for (const record of toDelete) {
+      try {
         await record.client.delete(record.id);
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.log(`Fail to delete entry ${record.id}`);
```
lennym, line 31: Why does the deletion ever fail?

Do we want to also log the error message to give some hope of debugging it?

gabiayako: Because I'm deleting the events while I'm running the tests. So when it comes to the teardown some entries don't exist anymore, so it fails to delete.

I'm deleting the events while running the tests so the reminders don't conflict with each other. And I'm not running teardown after each test so I don't need to re-create teams and user every time.

gabiayako: About the log to debug it, I'm not really sure if it will become very polluted because it will fail for every event deleted during the test, do you think it's better to add it? :)

apps/crn-server/test/integration/common/reminders.integration-test.ts
gabiayako: actually I've added this because this tests was failing randomly, I believe it fails sometimes because it of the time it takes to convert the event to draft

apps/crn-server/test/helpers/teardown.ts
lennym: Well, personally if the failures are routine and expected then I'd probably not log _anything_.

It might be worth checking the error and silently swallowing "Not Found" errors, and logging some useful output if there's any other non-expected errors.

apps/crn-server/test/integration/common/reminders.integration-test.ts
gabiayako: I've added a comment here to explain that [add comment explain delay inside test](https://github.com/yldio/asap-hub/pull/3236/commits/9c1687ecd9c0e459cabfc134456e7697627e6993) :)

lennym (approved):

PR #3239, opened 2023-06-28
Increase default test timeout

lennym:
We occasionally get test timeouts in React component tests with the default 5000ms timeout, to which the "solution" is generally to simply brute force the test by restarting.

Increase timeout to 10000ms to allow these to run reliably.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -64,5 +64,6 @@ module.exports = {
     ...packagePaths,
     ...lintConfigs,
   ],
+  testTimeout: 10000,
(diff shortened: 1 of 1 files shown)


jhonny-m:
are you sure is the tests that times out or is it a waitFor that's timing out?

lennym:
@jhonny-m In the case I have seen a lot lately ([example](https://github.com/yldio/asap-hub/actions/runs/5401199418/jobs/9810670974)) it is definitely the test.

PR #3267, opened 2023-07-05
update news in cms

jhonny-m:


peterstarling (approved):

packages/fixtures/src/gp2/news.ts
dianaguimaraes, line 1: do we need to add `type` here? or will it be added separately?

packages/model/src/gp2/news.ts
dianaguimaraes, line 1: do we need to add `type` here as well? or will it be added separately?

PR #3279, opened 2023-07-06
Related Events card and stories

kmaid:


packages/react-components/src/organisms/RelatedEventsCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <div
+        css={[
+          container,
+          ...(displayShowMoreButton ? [{ paddingBottom: 0 }] : []),
+        ]}
+      >
+        <Headline2 noMargin>Related CRN Hub Events</Headline2>
+        <div css={descriptionStyles}>
+          <Paragraph noMargin accent="lead">
+            Find out all CRN Hub events that are related to this output.
```
lennym, line 100: "Find out all..." doesn't seem correct. Should it be "Find all..."?

packages/react-components/src/organisms/RelatedEventsCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+          </Paragraph>
+        </div>
+        {relatedEvents.length === 0 ? (
+          <Paragraph noMargin accent="lead">
+            <b>No related CRN hub event available.</b>
```
lennym, line 105:
```suggestion
            <b>No related CRN Hub event available.</b>
```

oratiu (approved):

PR #3283, opened 2023-07-07
refactor: change reviewRequestedBy to StatusChangedBy

Riccardo-Brown:


apps/crn-server/test/integration/common/reminders.integration-test.ts
```diff
@@ -600,7 +604,8 @@ describe('Reminders', () => {
           .put(`/research-outputs/${draftOutputId}`)
           .send({
             ...draftOutput,
-            reviewRequestedById: loggedInUser.id,
+            statusChangedById: loggedInUser.id,
+            isInReview: true,
```
lennym, line 608: Rather than adding this everywhere - if it isn't pertinent to the test - then I'd add it to the default data [here](https://github.com/yldio/asap-hub/blob/master/apps/crn-server/test/integration/fixtures/research-output.ts#L23-L41).

apps/crn-server/test/integration/common/research-outputs.integration-test.ts
```diff
@@ -199,14 +199,14 @@ describe('research outputs', () => {
...
           .expect(200);
 
-        expect(response.body.reviewRequestedBy.id).toEqual(loggedInUser.id);
-        expect(response.body.reviewRequestedBy.firstName).toEqual(
+        expect(response.body.statusChangedBy.id).toEqual(loggedInUser.id);
+        expect(response.body.statusChangedBy.firstName).toEqual(
           loggedInUser.firstName,
         );
-        expect(response.body.reviewRequestedBy.lastName).toEqual(
+        expect(response.body.statusChangedBy.lastName).toEqual(
```
lennym, line 209: Should this payload also include `isInReview`? It seems like it should.

If it's inferred on the server then we _probably_ want to assert that it is inferred correctly.

apps/crn-server/test/integration/common/reminders.integration-test.ts
Riccardo-Brown: the default value for this field is false, so needs to be overwritten for these tests

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -247,14 +247,14 @@ export const getResearchOutputFilter = (
...
-          { reviewRequestedBy_exists: false },
+          { statusChangedBy_exists: false },
         ],
       },
       {
         AND: [
           { sys: { publishedVersion_exists: false } },
           { addedDate_exists: false },
-          { reviewRequestedBy_exists: true },
+          { statusChangedBy_exists: true },
```
gabiayako, line 257: Hey, I believe the condition is wrong here. It should be the same as the change made in squidex

Screenshot 2023-07-20 at 06 39 41

so here we should also check if `isInReview` is true and above if it is false

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -634,7 +634,7 @@ const getDraftResearchOutputRemindersFromQuery = (
     (researchOutputReminders, researchOutput) => {
       const userName = getUserName(researchOutput);
       const isPublished = !!researchOutput.sys.publishedAt;
-      const isInReview = !!researchOutput.reviewRequestedBy;
+      const isInReview = !!researchOutput.statusChangedBy;
```
gabiayako, line 637: the same here, we should check if the `isInReview` field is true

gabiayako (changes requested):

apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -89,7 +89,9 @@ export const researchOutputToCSV = (
   created: output.created,
   lastModifiedDate: output.lastModifiedDate,
   published: output.published,
-  reviewRequestedBy: `${output.reviewRequestedBy?.firstName} ${output.reviewRequestedBy?.lastName}`,
+  statusChangedBy: `${output.statusChangedBy?.firstName} ${output.statusChangedBy?.lastName}`,
+  statusChangedAt: output.statusChangedAt,
+  isInReview: true,
```
lennym, line 94: This should probably be the actual value rather than being hardcoded to `true`

apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -89,7 +89,9 @@ export const researchOutputToCSV = (
   created: output.created,
   lastModifiedDate: output.lastModifiedDate,
   published: output.published,
-  reviewRequestedBy: `${output.reviewRequestedBy?.firstName} ${output.reviewRequestedBy?.lastName}`,
+  statusChangedBy: `${output.statusChangedBy?.firstName} ${output.statusChangedBy?.lastName}`,
```
lennym, line 92: This outputs `undefined undefined` for things which have not been in review (which will be 100% of research outputs initially).

lennym (changes requested):

apps/crn-server/src/data-providers/contentful/research-output.data-provider.ts
```diff
@@ -416,8 +418,8 @@ const prepareInput = (
     ),
     labs: getLinkEntities(labIds),
     teams: getLinkEntities(teamIds),
-    relatedResearch: getLinkEntities(relatedResearchIds),
-    relatedEvents: getLinkEntities(relatedEventIds),
+    relatedResearch: relatedResearchIds && getLinkEntities(relatedResearchIds),
+    relatedEvents: relatedEventIds && getLinkEntities(relatedEventIds),
```
lennym, line 422: What's the reason for this change? There are several other calls to `getLinkEntities` here that _don't_ have the check, which seems odd. Surely it should be all of them or none of them?

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -719,7 +719,7 @@ const getInReviewResearchOutputRemindersFromQuery = (
       !researchOutput.documentType ||
       !isResearchOutputDocumentType(researchOutput.documentType) ||
       isPublished ||
-      !researchOutput.reviewRequestedBy
+      !researchOutput.statusChangedBy
```
gabiayako, line 722: I believe we should use `!researchOutput.isInReview` here as well. This is the `getInReviewResearchOutputRemindersFromQuery` and although `statusChangedBy` can be populated, this will not guarantee that the status is "in review", right?

gabiayako (changes requested):

apps/crn-server/src/data-providers/contentful/research-output.data-provider.ts
Riccardo-Brown: Fix Typescript error 
```
Argument of type 'string[] | undefined' is not assignable to parameter of type 'string[]'.
  Type 'undefined' is not assignable to type 'string[]'
```

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
gabiayako: No @andreicovaciu, because the filter we're using here is this 

```
export const getResearchOutputFilter = (
  zone: string,
): ResearchOutputsFilter => {
  const { last24HoursISO } = getReferenceDates(zone);
  return {
    OR: [
      {
        AND: [
          { addedDate_gte: last24HoursISO },
          { sys: { publishedVersion_exists: true } },
        ],
      },
      {
        AND: [
          { createdDate_gte: last24HoursISO },
          { sys: { publishedVersion_exists: false } },
          { addedDate_exists: false },
          { reviewRequestedBy_exists: false },
        ],
      },
      {
        AND: [
          { sys: { publishedVersion_exists: false } },
          { addedDate_exists: false },
          { reviewRequestedBy_exists: true },
        ],
      },
    ],
  };
};
```

It fetches all 3 types of research outputs reminders and in the functions `getPublishedResearchOutputRemindersFromQuery`, `getDraftResearchOutputRemindersFromQuery` and `getInReviewResearchOutputRemindersFromQuery` we separate them

packages/squidex/schema/crn/schemas/research-outputs.json
```diff
@@ -243,7 +243,7 @@
         }
       },
       {
-        "name": "reviewRequestedBy",
```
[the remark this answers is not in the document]
Riccardo-Brown: not 100% sure so instead of renaming i have just added `statusChangedBy`. We can remove `reviewRequestedBy` in another PR

apps/crn-server/src/data-providers/contentful/research-output.data-provider.ts
lennym: How come that only applies to `relatedResearch` and `relatedEvents` and not also the labs, teams, methods, keywords, etc...?

Riccardo-Brown: So i reverted this change as it was a rebasing issue `relatedResearchIds` and `relatedEventIds` was optional when i first started working on this now they are required.

gabiayako (approved):
🎉

lennym (approved):

PR #3291, opened 2023-07-11
Add handlers for indexing data from contentful

lennym:
* Add dependency resolver for research tag and research output data providers
* Make research output handler load data providers from dependency resolver to be platform agnostic
* Add missing index handlers to serverless config
* Rewrite team research outputs handler which was based on defunct data models
* Modify handlers to use `details.resourceId` to reference items
* Fix algolia index in some handlers in serverless config which were referencing the incorrect (non-contentful) index


lennym:
@gabiayako Sorted now, thanks.

apps/crn-server/scripts/export-entity.ts
gabiayako, line 1: I haven't noticed before, but this file is using `labDataProvider` as Squidex lab provider only

apps/crn-server/src/handlers/event/algolia-index-external-author-events-handler.ts
gabiayako, line 1: we could change from `event: EventBridgeEvent<ExternalAuthorEvent, ExternalAuthorSquidexPayload>` to `event: EventBridgeEvent<ExternalAuthorEvent, ExternalAuthorPayload>` :)

apps/crn-server/scripts/export-entity.ts
lennym: It does, and I don't really know _why_ because it's never used. I'm just going to remove all mention of labs from that file.

apps/crn-server/src/handlers/event/algolia-index-external-author-events-handler.ts
lennym: Done.

gabiayako (approved):
nice! yeah, labs were never called, good catch 🎉

PR #3302, opened 2023-07-13
GP2-534/GP2-535 - update Working Groups and Projects cards UI

marianasrv:
remove role and change copy of Members column in WG card
GP2-534 - remove role from Projects Card


packages/gp2-components/src/icons/user-circle-icon.tsx
markcsims, line 1: Maybe this should have a title?

markcsims (changes requested):

packages/gp2-components/src/molecules/CollapsibleTable.tsx
```diff
@@ -67,23 +67,25 @@ type CollapsibleTableRow = {
 type CollapsibleTableProps = {
   headings: string[];
   children: CollapsibleTableRow[];
+  tableStyles?: SerializedStyles;
```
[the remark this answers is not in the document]
marianasrv: Sure. But this is supposed to be a reusable component. however, it's fixed to 3 columns...

packages/gp2-components/src/organisms/CardTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <CollapsibleTable
+      headings={
+        <div css={[tableStyles, gridTitleStyles]}>
+          {headings.map((heading, idx) => (
+            <Subtitle
+              key={`heading-${idx}`}
+              noMargin
+              data-testid={`heading-desktop-${idx}`}
```
markcsims, line 73: Is there a way to avoid using data-testid. We should be trying to use getByRole, then getByText etc with the last option getByTestId

packages/gp2-components/src/organisms/__tests__/CardTable.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+          },
+        ]}
+      </CardTable>,
+    );
+    expect(screen.getByTestId('heading-desktop-0')).toBeVisible();
+    expect(screen.getByTestId('heading-desktop-1')).toBeVisible();
+    expect(screen.getByTestId('heading-desktop-2')).toBeVisible();
+    expect(screen.getByText('val1')).toBeVisible();
+    expect(screen.getByText('val2')).toBeVisible();
+    expect(screen.getByText('val3')).toBeVisible();
```
markcsims, line 22: Is there a way to avoid using data-testid. We should be trying to use getByRole, then getByText etc with the last option getByTestId?  Same comment as before.

markcsims (changes requested):

markcsims (approved):

PR #3311, opened 2023-07-13
GP2 485 update dependancies

markcsims:


peterstarling (approved):

apps/gp2-server/src/dependencies/event.dependency.ts
```diff
@@ -1,22 +1,15 @@
...
-    environment: contentfulEnvId,
-  });
+import {
+  getContentfulGraphQLClientFactory,
+  getContentfulRestClientFactory,
+} from './clients.dependency';
 
-  return new EventContentfulDataProvider(
+export const getEventDataProvider = (
+  contentfulGraphQLClient: GraphQLClient = getContentfulGraphQLClientFactory(),
```
peterstarling, line 10: this might mean people are gonna use the defaults soon plus may lead to unexpected results, is there a reason why we pass in the default at all?

PR #3334, opened 2023-07-17
Add in-memory cache for Contentful API client

lennym:
Rather than creating a new client for every request that uses the Contentful Management API, maintain an in-memory cache of environment instances per accesskey/space/env tuple so that the same client instance is re-used across multiple requests within the same process.

This will reduce the number of requests made to Contentful by 2 per POST/PUT, and so should reduce the risk of hitting API rate limits.


lennym:
@peterstarling I did consider that. It created a circular dependency between `@asap-hub/contentful` and `@asap-hub/server-common`, which for something which was nothing more than a no-op wrapper around `Map` seemed not worth unpicking.

PR #3338, opened 2023-07-17
CT-68 Research Output Reminders

gabiayako:
This PR adds the Research Output reminders to the existent contentful reminder data provider with its unit tests and integration tests. With this, it was possible to delete the squidex reminder integration tests.

---

##### Published
reminder-ro-published

##### Draft
reminder-ro-draft

##### In Review
reminder-ro-in-review


apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -123,15 +163,11 @@ export class ReminderContentfulDataProvider implements ReminderDataProvider {
...
-    ReminderDataObject,
-    // TODO: add these types back when implementing RO reminders
-    | ResearchOutputPublishedReminder
-    | ResearchOutputDraftReminder
-    | ResearchOutputInReviewReminder
-  >,
-): DateTime => {
+const getSortDate = (reminder: ReminderDataObject): DateTime => {
+  if (reminder.entity === 'Research Output') {
+    return DateTime.fromISO(reminder.data.addedDate);
```
lennym, line 168: This will only exist for published reminders, so if we're also going to have reminders related to draft outputs in review then we might want to have a fallback of `reminder.data.lastUpdatedPartial`
```suggestion
    return DateTime.fromISO(reminder.data.addedDate || reminder.data.lastUpdatedPartial);
```

gabiayako: We were using `createdDate` in squidex (https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/reminder.data-provider.ts#L365) but good point, I've renamed it here [review: rename addedDate to createdDate for draft/in review reminders](https://github.com/yldio/asap-hub/pull/3338/commits/6b42db582688c573f5f36edee4d2b3b16c116cd3) so it makes more sense 🙂

lennym (approved):

PR #3344, opened 2023-07-18
Add feature flag headers to remaining api calls

lennym:
Add the feature flag cookies to all API calls where it wasn't previously there, so all data can be written to and loaded from Contentful.


gabiayako (approved):
🎉

PR #3346, opened 2023-07-18
Small refactor of groups to interest groups

oratiu:


apps/crn-server/serverless.ts
```diff
@@ -846,10 +846,10 @@ const serverlessConfig: AWS = {
             pattern: {
               source: [eventBusSourceSquidex],
               'detail-type': [
-                'GroupsPublished',
-                'GroupsUpdated',
-                'GroupsUnpublished',
-                'GroupsDeleted',
+                'InterestGroupsPublished',
```
gabiayako, line 849: I think this will break the hook as the name is still "groups" in the schema although its label is "Interest Groups"

interest-groups-name

So when you update a interest group and go check the webhook call in `Rules` -> `Logs`, we can see that the name of the event is still `GroupsUpdated`

interest-groups-hook

I believe we would need to change the name https://github.com/yldio/asap-hub/blob/CRN-1228-refactor-all-code-components-from-groups-to-interest-groups/packages/squidex/schema/crn/schemas/interest-groups.json#L358 to `interest-groups` as well. But it would be nice to check this flow after updating the name to see if it works :)

packages/react-components/src/organisms/__tests__/InterestGroupTools.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render, fireEvent } from '@testing-library/react';
 import { ComponentProps } from 'react';
 
-import GroupTools from '../GroupTools';
+import GroupTools from '../InterestGroupTools';
```
gabiayako, line 4:
```suggestion
import InterestGroupTools from '../InterestGroupTools';
```

apps/crn-server/src/routes/interest-group.route.ts
```diff
@@ -3,15 +3,15 @@ import { validateFetchOptions } from '@asap-hub/server-common';
 import { Router } from 'express';
 import InterestGroupController from '../controllers/interest-group.controller';
 import { validateEventFetchParameters } from '../validation/event.validation';
-import { validateGroupParameters } from '../validation/group.validation';
+import { validateGroupParameters } from '../validation/interest-group.validation';
 
 export const interestGroupRouteFactory = (
   groupController: InterestGroupController,
```
gabiayako, line 9:
```suggestion
  interestGroupController: InterestGroupController,
```

gabiayako (commented):
Sending more some tiny suggestions (I haven't had looked at everything yesterday)

apps/crn-server/serverless.ts
```diff
@@ -871,10 +871,10 @@ const serverlessConfig: AWS = {
...
               source: [eventBusSourceContentful],
               'detail-type': [
-                'InterestGroupsPublished',
-                'InterestGroupsUpdated',
-                'InterestGroupsUnpublished',
-                'InterestGroupsDeleted',
+                'GroupsPublished',
+                'GroupsUpdated',
+                'GroupsUnpublished',
+                'GroupsDeleted',
```
gabiayako, line 877: This isn't right 😅 in Contentful the content type is called InterestGroups so here it should remain as it was

packages/model/src/webhook.ts
```diff
@@ -7,8 +7,7 @@ export type EntityEventAction =
 
 export type EventEvent = `Events${EntityEventAction}`;
 export type ExternalAuthorEvent = `ExternalAuthors${EntityEventAction}`;
-export type GroupEvent = `Groups${EntityEventAction}`;
-export type InterestGroupEvent = `InterestGroups${EntityEventAction}`;
```
gabiayako, line 11: This should be reverted as well I believe when we undo the changes in serverless

apps/crn-server/serverless.ts
oratiu: Whoops, I changed the wrong file

packages/model/src/calendar.ts
```diff
@@ -1,17 +1,20 @@
...
-import { GroupDataObject, GroupResponse } from './group';
+import {
+  InterestGroupDataObject,
+  InterestGroupResponse,
+} from './interest-group';
 import { WorkingGroupDataObject, WorkingGroupResponse } from './working-group';
 
 export interface CalendarDataObject extends BasicCalendar {
-  groups?: Pick<GroupDataObject, 'id' | 'active'>[];
+  groups?: Pick<InterestGroupDataObject, 'id' | 'active'>[];
```
gabiayako, line 10: Just saw this now, I think it would be nice to change it here as well
```suggestion
  interestGroups?: Pick<InterestGroupDataObject, 'id' | 'active'>[];
```

gabiayako (approved):
Nice! 🎉

PR #3349, opened 2023-07-18
Display related research on an event page

kmaid:
Refactor RelatedResearch to RelatedResearchCard.
Misc fixes
Add related research to event response
Add event card


apps/crn-server/src/data-providers/contentful/event.data-provider.ts
```diff
@@ -475,6 +475,7 @@ export const parseGraphQLEvent = (item: EventItem): EventDataObject => {
     status,
     hidden: hidden || false,
     tags: tags ?? [],
+    relatedResearch: [],
```
Riccardo-Brown, line 478: Should `relatedResearch` always be an empty array?

apps/crn-server/test/data-providers/contentful/event.data-provider.test.ts
```diff
@@ -415,6 +417,7 @@ describe('Events Contentful Data Provider', () => {
       const result = await eventDataProviderMockGraphql.fetchById(eventId);
 
       const expectedResult = getContentfulEventDataObject();
+      // expectedResult.relatedResearch = result!.relatedResearch;
```
peterstarling, line 420: ?

apps/crn-server/test/data-providers/event.data-provider.test.ts
```diff
@@ -54,8 +53,11 @@ describe('Event data provider', () => {
       const result = await eventDataProviderMockGraphql.fetch({
         before: 'before',
       });
-
-      expect(result).toMatchObject(getSquidexGraphqlEvents());
+      const { relatedResearch, ...event } = getEventResponse();
```
peterstarling, line 56: why are you not able to test it with related-research included?

apps/crn-server/test/fixtures/events.fixtures.ts
```diff
@@ -47,6 +47,28 @@ export const getContentfulUserSpeakerTeams = () => ({
   role: 'Lead PI (Core Leadership)',
 });
 
+export const getContentfulRelatedResearch = () => ({
```
peterstarling, line 50: could you not have used this instead https://github.com/yldio/asap-hub/blob/e4e5b499ff81c04a40b3db3cce5d68b3b60bf643/apps/crn-server/test/fixtures/research-output.fixtures.ts#L505

i think the mock server doesn't mind returning extra properties

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -87,6 +87,7 @@ export const getSquidexGraphqlResearchOutput = (): NonNullable<
             },
           },
         ],
+        workingGroups: [],
```
peterstarling, line 90: i see you've added the working groups to the data object and now you keep passing empty array in all fixtures, which is unlike what we have been doing for the rest of them. I think it's better if you add some data to it by default as thats the pattern we have been using

peterstarling (changes requested):

apps/crn-server/test/data-providers/event.data-provider.test.ts
kmaid: Its returning multiple WG and Teams rather than one so i just gave up and removed the object from the comparison because I don't know how to get it to work this deep.

apps/crn-server/test/data-providers/contentful/research-output.data-provider.test.ts
```diff
@@ -354,6 +354,48 @@ describe('Research Outputs Data Provider', () => {
       expect(result!.documentType).toEqual('Grant Document');
     });
 
+    test('Should default missing working group reference to an empty array of teams', async () => {
```
peterstarling, line 357: didnt you mean 
```suggestion
    test('Should default an empty teams collection to an empty array of teams', async () => {
```
or something like that?

peterstarling (approved):

Riccardo-Brown (approved):

PR #3377, opened 2023-07-21
Migrate Research Output data from Squidex to Contentful

lennym:
## To-do:

- [x] Test against prod data
- [x] Add new fields for review state
- [x] Migrate usage notes from rich text to markdown
- [x] Write tests
- [ ] Handle proposals on Teams (OOS)

## Failures

A total of 4 ROs failed to migrate, and will need manual attention.

3 ROs failed to parse due to unsupported tags in the rich text content (2 x `<table>`, 1 x `<section>`)

- 57b12373-fe17-4657-819d-6e66b7156d01
- 5af07fd4-b017-4360-9b83-a89fb58dda69
- f354ec08-c7e2-48bd-9fd8-816b44ac5bc0

1 RO failed to publish due to containing a reference to itself in related research.

- d49495f6-9678-44a6-aa1d-280f2241b41e


packages/contentful/migrations/crn/researchOutputs/20230724090200-unique-urls.js
peterstarling, line 1: I thought contentful offer a built-in url validation?

packages/contentful/migrations/crn/researchOutputs/20230724090200-unique-urls.js
```diff
@@ -0,0 +1,17 @@
+module.exports.description = 'Remove uniqueness constraint on link field';
```
peterstarling, line 1:
```suggestion
module.exports.description = 'Add URL validation on link field';
```

peterstarling (approved):

packages/contentful/migrations/crn/researchOutputs/20230724090200-unique-urls.js
lennym: The URL validation is there before this migration. What this migration is doing is setting the validation to _only_ that, and without the uniqueness validator.

gabiayako, line 1: do we really want to remove that? I wonder because even in the frontend we have the url constraint

Screenshot 2023-07-27 at 08 09 26

so I wonder if we don't want to enforce that in Contentful as well

But even it is only for the migration because in old times we didn't have this constraint and people were able to create duplicate links in prod, maybe it would be nice to create a ticket to add this constraint back once the migration is done?

packages/cms-data-sync/test/run-migrations.test.ts
```diff
@@ -143,6 +144,20 @@ jest.mock('../src/research-tags/research-tags.data-migration', () => {
...
+jest.mock('../src/research-outputs/research-outputs.data-migration', () => {
+  mockMigrateResearchOutputs = jest.fn().mockReturnValue({});
+  return {
+    ...jest.requireActual(
+      '../src/research-outputs/research-outputs.data-migration',
+    ),
+    migrateResearchOutputs: mockMigrateResearchOutputs,
+  };
+});
+
```
gabiayako, line 160: very minor, but don't we want to check if `migrateResearchOutputs` were called or not called in the tests?

gabiayako (approved):

packages/contentful/migrations/crn/researchOutputs/20230724090200-unique-urls.js
lennym: If we enforce in contentful then _a lot_ of existing data from prod won't migrate. 

It was there originally, but there's a lot of data in Squidex prod with non-unique links, so I removed it.

packages/cms-data-sync/test/run-migrations.test.ts
lennym: Yes, but I'm currently writing tests for the next bit of the migration so will do it there so they don't conflict.

packages/contentful/migrations/crn/researchOutputs/20230724090200-unique-urls.js
gabiayako: hmm ok, and if we add this back after the migration is done if people edit these research outputs with duplicate links they won't be able to publish. Yeah, it makes sense to remove it :)

lennym: That's how it currently is in Squidex, so I doubt it would have _significant_ impact to add it back post-migration.

PR #3393, opened 2023-07-24
CRN-1442 Add new version to RO in the CMS

lctrt:


packages/contentful/migrations/crn/researchOutputs/20230724000000-version.js
```diff
@@ -0,0 +1,63 @@
+module.exports.description = 'Change usageNotes type';
+
+module.exports.up = function (migration) {
+  const outputVersions = migration
+    .createContentType('outputVersions')
+    .name('Output Versions')
```
peterstarling, line 6: I think I would prefer to use the `Research Output` prefix so that we know it belongs to that, unless there is a limit on the name length? I was looking at the content models and i couldn't find it until i opened the definition of the field inside Research Output content model

packages/contentful/src/crn/queries/research-outputs.queries.ts
```diff
@@ -195,6 +195,20 @@ export const researchOutputContentQueryFragment = gql`
         endDate
       }
     }
+    versions {
+      versionsCollection {
```
peterstarling, line 199: although this comment could go anywhere else just as well i am going to leave it here - i found it hard to understand your implementation for the versions, and I don't think this aligns with what we discussed. 

So if I understand correctly you would like to have research output point at the output versions element to then point back to RO? so basically each version would just be a different research output? See, this would make things a lot more complicated later.

What we agreed to, at least to what I can remember, is that the versions content model would STORE the title, link and would not point back at RO. So this would just be a way to track the changes of these fields without an actual version history.

peterstarling (changes requested):

apps/crn-server/test/fixtures/research-output.fixtures.ts
```diff
@@ -248,6 +248,7 @@ export const getResearchOutputDataObject =
     keywords: ['Keyword1'],
     published: true,
     isInReview: false,
+    versions: [],
```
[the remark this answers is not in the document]
lctrt: At the moment it's weird to do anything else with this fixture, because it's used by Squidex but squidex doesn't have versions.

apps/crn-server/test/data-providers/contentful/research-output.data-provider.test.ts
```diff
@@ -1643,6 +1648,20 @@ describe('Research Outputs Data Provider', () => {
   });
 });
 
+describe('mapOutputVersions ', () => {
```
peterstarling, line 1651: i think you should add one test to just show how this mapping function really works - basic input and basic output, what do you think?

apps/crn-server/test/fixtures/research-output.fixtures.ts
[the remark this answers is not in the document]
peterstarling: even better - in squidex tests make sure that you modify the expected output to `expected.versions = []` with a note that this feature is not implemented in squidex. That way we have test evidence for this too.

And we definitely need to have a test to verify that versions are coming out of the data provider correctly, just testing the mapper in isolation is not going to cut it

peterstarling (changes requested):

peterstarling (approved):

PR #3395, opened 2023-07-24
[CRN-1428] add switch to draft reminders

Riccardo-Brown:


apps/crn-server/src/controllers/reminder.controller.ts
```diff
@@ -74,6 +74,20 @@ export default class ReminderController {
...
+          reminder.entity === 'Research Output' &&
+          reminder.type === 'Switch To Draft'
+        ) {
+          return {
+            id: reminder.id,
+            entity: reminder.entity,
+            href: sharedResearch({}).researchOutput({
+              researchOutputId: reminder.data.researchOutputId,
+            }).$,
+            description: `**${reminder.data.statusChangedBy}** on ${reminder.data.associationType} **${reminder.data.associationName}** switched to draft a ${reminder.data.associationType} ${reminder.data.documentType} output: ${reminder.data.title}.`,
```
lennym, line 87:
```suggestion
            description: `**${reminder.data.statusChangedBy}** on ${reminder.data.associationType} **${reminder.data.associationName}** switched a ${reminder.data.associationType} ${reminder.data.documentType} output to draft: ${reminder.data.title}.`,
```

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
gabiayako, line 1: It's an edge case but it would be nice to talk to Tiff to see if we want to fix this: if I create a draft and it is switched to draft in last than 24h of the creation of the draft, both of these reminders appear:

Screenshot 2023-07-28 at 06 30 01

But I'm not sure if we would want to show them both

gabiayako (approved):

gabiayako (approved):
🎉

PR #3409, opened 2023-07-26
Draft, added condition for active membership on teams and working gro…

oratiu:
…up permission handler


gabiayako (commented):
Really nice! Just a question before I accept, if the team/working group is inactive should we display the draft tab?
Screenshot 2023-07-26 at 12 45 46

gabiayako (approved):
Approving based on the conversation during the daily :)

PR #3411, opened 2023-07-26
Add `--forceExit` flag to integration tests

lennym:
This ensures that jest exists once tests are complete when repeated rate limit errors cause tests to time out and otherwise keep an open async operation.

Also add a 20 minute timeout to integration test runs as a fallback.


gabiayako (approved):
🎉

PR #3429, opened 2023-07-28
GP2HUB 34 output search and filters

markcsims:
- starts adding output indexer
- adds output indexer to serverless
- adds test coverage for output indexer
- extends AlgoliaSearchClient for use in GP2
- fixes typings for AlgoliaSearchClient


packages/algolia/src/client.ts
```diff
@@ -2,103 +2,103 @@ import { SearchOptions, SearchResponse } from '@algolia/client-search';
...
+    type: ResponsesKey;
   };
+};
+export type ClientSearchResponse<
+  App extends Apps,
+  ResponsesKey extends keyof EntityResponses[App],
+> = SearchResponse<
+  DistributeToEntityRecords<EntityResponses[App], ResponsesKey>
+>;
+export type ClientSearch<
```
peterstarling, line 49: this type and its naming is a little hard to understand, from what I can tell you are defining a method separately and then applying that within the client interface. Have you considered having that definition only as part of the client interface and then referencing that particular method in your fixtures/etc? You can add a generic to the interface if thats something you need

apps/gp2-server/serverless.ts
```diff
@@ -59,6 +60,11 @@ const apiUrl = `https://${apiHostname}`;
 const currentRevision = process.env.CI_COMMIT_SHA;
 const nodeEnv = 'production';
 const sesRegion = process.env.GP2_SES_REGION!;
+const envRef = ['production', 'dev'].includes(stage) ? envAlias : `CI-${stage}`;
+
+const algoliaIndex = process.env.GP2_ALGOLIA_INDEX
```
peterstarling, line 65: can you do...
```suggestion
const algoliaIndex = process.env.GP2_ALGOLIA_INDEX ?? `gp2-hub_${envRef}`;
```
?
I'm not sure I know what `'${env:....}'` is but it looks like it's a serverless way of referencing env vars which you are also doing with process.env so why complicate it

peterstarling (approved):
you should consider switching from CRN suffix to Crn (and GP2 to Gp2)

packages/algolia/src/client.ts
markcsims: I wanted to do

```type ClientSearchResponse = ReturnType<SearchClient['search']>;```

But I was getting `Type ClientSearchResponse is not generic`

PR #3430, opened 2023-07-28
CT-33 Event custom validation

gabiayako:
This PR adds an extension to add custom validation to `events` following this model https://github.com/contentful/apps/tree/master/examples/custom-validation as suggested by contentful support (read this thread https://contentful-community.slack.com/archives/CBR24P3T4/p1688145410897879?thread_ts=1687960592.599549&cid=CBR24P3T4).

The validations are listed here in this ticket https://asaphub.atlassian.net/browse/CT-33 similar of what we have in squidex, only 3 of them (requirements 1, 2, 4 of the ticket) make the entry unpublishable. Below there is the code of the script we use to do these validations in Squidex:

Screenshot 2023-07-28 at 14 25 19

This requirement

> When a CMS user updates an event and they create a speaker and select a team and a user that is not part of that team then that user should not be displayed.

does not make the entry unpublishable in Squidex, so I made it behave the same way in Contentful. And this is covered in the response parser of the data providers
- squidex: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/transformers/event.ts#L121:L135
- contentful: https://github.com/yldio/asap-hub/blob/CT-33-create-custom-validation-for-events/apps/crn-server/src/data-providers/contentful/event.data-provider.ts#L332:L346

But I also left a warning in Contentful when this happens:
Screenshot 2023-07-28 at 14 30 31

--- 

#### Example of an error that makes the error unpublishable
Screenshot 2023-07-28 at 14 19 06

### DIFF
--- a/packages/contentful-app-extensions/event-custom-validation/src/test/locations/Field.test.tsx
+++ b/packages/contentful-app-extensions/event-custom-validation/src/test/locations/Field.test.tsx
@@ -1,0 +1,368 @@
+import { FieldExtensionSDK } from '@contentful/app-sdk';
+import { useSDK, useAutoResizer } from '@contentful/react-apps-toolkit';
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+
+import Field, {
+  VALID_ENTRY_MESSAGE,
+  DUPLICATE_SPEAKERS_MESSAGE,
+  EMPTY_SPEAKER_MESSAGE,
+  EXTERNAL_AUTHOR_WITH_TEAM_MESSAGE,
+} from '../../locations/Field';
+
+jest.mock('@contentful/react-apps-toolkit', () => ({
+  useSDK: jest.fn(),
+  useAutoResizer: jest.fn(),
+}));
+
+const speakerFields = {
+  team: {
+    'en-US': {
+      sys: {
+        id: 'team-1',
+      },
+    },
+  },
+  user: {
+    'en-US': {
+      sys: {
+        id: 'user-1',
+      },
+    },
+  },
+};
+
+const unsubscribeFn = () => {};
+const mockBaseSdk = () => ({
+  field: {
+    setValue: jest.fn(),
+  },
+  entry: {
+    getSys: jest.fn(),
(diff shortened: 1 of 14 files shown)


packages/contentful-app-extensions/event-custom-validation/src/test/locations/Field.test.tsx
peterstarling, line 1: any way we could test that the entry cannot be published for when there errors appear? I haven't been able to find it in the implementation either so I was wondering if errors actually forbid us from publishing?

packages/contentful-app-extensions/event-custom-validation/src/test/locations/Field.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+    await waitFor(() => {
+      expect(
+        screen.getByText(EXTERNAL_AUTHOR_WITH_TEAM_MESSAGE),
+      ).toBeInTheDocument();
+      expect(screen.getByText(EMPTY_SPEAKER_MESSAGE)).toBeInTheDocument();
+      expect(screen.queryByText(VALID_ENTRY_MESSAGE)).toBeNull();
+    });
+  });
+
+  it('displays warning message when there is an user associated with a team that he/she does not belong', async () => {
```
peterstarling, line 278: does this mean the entry can still be published? can you test that?

Also - do you know what happens when we publish a speaker with a user and a team that they do not belong to?

peterstarling (changes requested):

packages/contentful-app-extensions/event-custom-validation/src/test/locations/Field.test.tsx
gabiayako: I've added these checks [adding expect set value to have been called tests](https://github.com/yldio/asap-hub/pull/3430/commits/1bc258ce77e2b74941e396b92c634e3297655e83). Not sure if it's what you had in mind though

peterstarling: this is fine, any other test would have to validate submission of the entire entry, and i dont even know if thats possible so

peterstarling (approved):

PR #3440, opened 2023-08-01
GP2HUB-8 Add self administration of keywords in contentful

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/GP2HUB-8


apps/gp2-server/test/fixtures/keyword.fixtures.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const getListKeywordsDataObject =
+  (): gp2Model.ListKeywordsDataObject => ({
+    total: 1,
+    items: [getKeywordDataObject()],
+  });
+
+export const getKeywordCreateDataObject =
+  (): gp2Model.KeywordCreateDataObject => {
+    const { id: _id, name } = getKeywordDataObject();
```
markcsims, line 36:
```suggestion
    const { name } = getKeywordDataObject();
```

packages/contentful/migrations/gp2/events/20230801150001-add-keywords.js
markcsims, line 1: Should there be a max/min number of keywords you can add?

packages/contentful/migrations/gp2/outputs/20230801150001-add-keywords.js
```diff
@@ -0,0 +1,30 @@
+module.exports.description = 'Add tags to outputs content model';
+
+module.exports.up = (migration) => {
+  const outputs = migration.editContentType('outputs');
+
+  outputs
+    .createField('tags')
```
markcsims, line 7: The name tags is not consistent with the field name in Events.

Should there be a min/max number added?

packages/contentful/migrations/gp2/projects/20230801150001-add-keywords.js
```diff
@@ -0,0 +1,39 @@
+module.exports.description = 'Add tags to projects content model';
+
+module.exports.up = (migration) => {
+  const projects = migration.editContentType('projects');
+
+  projects
+    .createField('tags')
```
markcsims, line 7: The name tags is not consistent with the field name in Events.

packages/contentful/migrations/gp2/users/20230801150001-add-keywords.js
```diff
@@ -0,0 +1,39 @@
+module.exports.description = 'Add tags to users content model';
+
+module.exports.up = (migration) => {
+  const users = migration.editContentType('users');
+
+  users
+    .createField('tags')
```
markcsims, line 7: The name tags is not consistent with the field name in Events.

packages/contentful/migrations/gp2/workingGroups/20230801150001-add-keywords.js
```diff
@@ -0,0 +1,30 @@
+module.exports.description = 'Add tags to workingGroups content model';
+
+module.exports.up = (migration) => {
+  const workingGroups = migration.editContentType('workingGroups');
+
+  workingGroups
+    .createField('tags')
```
markcsims, line 7: The name tags is not consistent with the field name in Events.

markcsims: Should there be a min/max

packages/contentful/migrations/gp2/events/20230801150001-add-keywords.js
marianasrv: We copied the validations that were in the CMS previously

packages/contentful/migrations/gp2/outputs/20230801150001-add-keywords.js
marianasrv: Yes, because in Projects, WG, outputs and users the existing field is keywords, and in events is tags. So, to complete the migration, we needed to create another field with a different name since we need to keep them both until it's done. We can align on one single name to keep consistency for the last PR.
Regarding validations, we copied the validations that were in the CMS previously

markcsims (approved):

PR #3469, opened 2023-08-07
Add wrapping `<p/>` around rich text fields populated with plain text

lennym:
Some description fields on research outputs in Squidex return content as plain text strings rather than rich text html. This causes them to migrate as empty because the parser removes top-level `text` nodes.

When a rich text field contains no html tags then wrap the text with a single `<p/>` before parsing so that the top level node is a `paragraph` as expected.

### DIFF
--- a/packages/cms-data-sync/src/utils/rich-text.ts
+++ b/packages/cms-data-sync/src/utils/rich-text.ts
@@ -41,6 +41,14 @@ export const wrapIframeWithPTag = (html: string): string => {
   return output ?? html;
 };
 
+export const wrapPlainTextWithPTag = (html: string): string => {
+  const $ = cheerio.load(html);
+  if ($('body').children().length === 0 && $('body').text()) {
+    return `<p>${html}</p>`;
+  }
+  return html;
+};
+
 export const clearParsedHtmlOutput = (htmlDocument: Document) => ({
   ...htmlDocument,
   content: htmlDocument?.content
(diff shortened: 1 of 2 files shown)


packages/cms-data-sync/src/utils/rich-text.ts
```diff
@@ -111,6 +119,7 @@ export const convertHtmlToContentfulFormat = (html: string) => {
   processedHtml = removeStylingTagsWrappingIFrameTags(processedHtml);
   processedHtml = removeStylingTagsWrappingImgTags(processedHtml);
   processedHtml = wrapIframeWithPTag(processedHtml);
+  processedHtml = wrapPlainTextWithPTag(processedHtml);
 
   // processedHtml = addPTagsToIframeTagNotWrappedByAnything(processedHtml);
```
gabiayako, line 124: I think I've left this comment by mistake 🙈 could you erase before merging?

gabiayako (approved):
Nice 🎉

packages/cms-data-sync/src/utils/rich-text.ts
lennym: Do you mean uncomment, or remove the line completely?

gabiayako: remove it completely :)

lennym, line 124:
```suggestion
```

gabiayako (approved):

PR #3474, opened 2023-08-07
CT-84 Adjust limit to linked related research query

gabiayako:
There is this research output https://hub.asap.science/shared-research/78210d68-bfa4-45d7-8719-22233ae25f75 with 73 related research outputs. Since we don't have access to production squidex it's kind of hard to confirm how many of those 73 are related research added to this research output and how many are from references. But I'm assuming the maximum of related research output items is 15, so I've increased the limit of ROs from references to 60. And they will only come in the fetch by id request method due to query complexity.


packages/contentful/src/crn/queries/research-outputs.queries.ts
```diff
@@ -166,8 +166,8 @@ export const researchOutputContentQueryFragment = gql`
         }
       }
     }
-    linkedFrom {
-      researchOutputsCollection(limit: 20) {
+    linkedFrom @include(if: $fetchLinkedRelatedResearch) {
```
lennym, line 169: I would add this same include filter to the relation in the other direction as well (at L151) so we always either get all of the related outputs, or none of them. Getting _some_ is liable to be a bit confusing.

gabiayako: it makes sense! I've changed it here [condition to fetch related research added to relatedResearchCollection](https://github.com/yldio/asap-hub/pull/3474/commits/ba0c9a11533bc4b28d88331299b03ccc34ed52bc)

lennym (approved):

PR #3476, opened 2023-08-07
Added empty algolia response for non onboarded users

oratiu:


apps/crn-frontend/src/events/state.ts
```diff
@@ -117,6 +117,10 @@ export const useEvents = (options: GetEventListOptions, user?: User | null) => {
   const [events, setEvents] = useRecoilState(eventsState(options));
   const { client } = useAlgolia();
 
+  if (user && !user.onboarded) {
```
lennym, line 120: If it's _possible_ that `user` is not defined then surely we'd want to treat that case as the same as a non-onboarded user?
```suggestion
  if (!user || !user.onboarded) {
```

packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -340,13 +340,16 @@ const UserProfileHeader: React.FC<UserProfileHeaderProps> = ({
             <TabLink href={tabRoutes.about({}).$}>Background</TabLink>
             <TabLink href={tabRoutes.outputs({}).$}>
               Shared Outputs
-              {` (${sharedOutputsCount})`}
+              {sharedOutputsCount !== undefined && ` (${sharedOutputsCount})`}
```
lctrt, line 343: for these tab count display, might be worth setting a default value in props instead?

apps/crn-frontend/src/events/state.ts
oratiu: We use this hook in different components where we don't send the user as the 2nd param

apps/crn-frontend/src/network/users/UserProfile.tsx
```diff
@@ -120,9 +124,13 @@ const UserProfile: FC<UserProfileProps> = ({ currentTime }) => {
...
-      sharedOutputsCount: researchOutputsResult.total,
-      upcomingEventsCount: upcomingEventsResult.total,
-      pastEventsCount: pastEventsResult.total,
+      sharedOutputsCount: isUserOnboarded
+        ? researchOutputsResult.total
+        : undefined,
+      upcomingEventsCount: isUserOnboarded
+        ? upcomingEventsResult.total
+        : undefined,
+      pastEventsCount: isUserOnboarded ? pastEventsResult.total : undefined,
```
lennym, line 133: These checks seem slightly redundant since we're already checking at the point of making the request. Is there a reason for doing effectively the same thing twice?

lennym: So it looks like the reason is to have the tab label be `Upcoming Events` rather than `Upcoming Events (0)`.

Seems a little arbitrary, but fine.

lennym: 🤷‍♂️

apps/crn-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
```diff
@@ -154,6 +154,22 @@ describe('Fetch-user-by-code handler', () => {
       expect(JSON.parse(result.body)).toEqual(getUserResponse());
     });
 
+    test('should return an empty string as the algolia key if the user is not onboarded', async () => {
```
peterstarling, line 157: shouldn't it be null or undefined or something? although then i reckon you would need to assert that the key is present which imho is correct

peterstarling (changes requested):

apps/crn-server/test/handlers/webhooks/fetch-by-code/fetch-by-code.test.ts
```diff
@@ -154,6 +154,22 @@ describe('Fetch-user-by-code handler', () => {
       expect(JSON.parse(result.body)).toEqual(getUserResponse());
     });
 
+    test('should return undefined as the algolia key if the user is not onboarded', async () => {
```
peterstarling, line 157:
```suggestion
    test('should return null as the algolia key if the user is not onboarded', async () => {
```

peterstarling (approved):

PR #3480, opened 2023-08-08
CRN-1454 - fix allow empty links

andreicovaciu:


apps/crn-server/test/controllers/research-output.controller.test.ts
```diff
@@ -500,6 +500,18 @@ describe('ResearchOutputs controller', () => {
...
+        const researchOutputEmptyLinkRequest = getResearchOutputCreateData();
+        researchOutputEmptyLinkRequest.documentType = 'Lab Resource';
+        researchOutputEmptyLinkRequest.link = undefined;
+
+        await researchOutputs.create(researchOutputEmptyLinkRequest);
+        expect(researchOutputDataProviderMock.fetch).not.toHaveBeenCalledWith({
+          filter: { link: expect.any(String) },
+          includeDrafts: true,
+        });
+      });
```
lennym, line 514: You will want to test that the payload sends the `link` as `null` or `undefined` and not `''` as well.

apps/crn-server/test/controllers/research-output.controller.test.ts
```diff
@@ -500,6 +500,32 @@ describe('ResearchOutputs controller', () => {
...
+          await researchOutputs.create(researchOutputEmptyLinkRequest);
+          expect(researchOutputDataProviderMock.fetch).not.toHaveBeenCalledWith(
+            {
+              filter: { link: expect.any(String) },
+              includeDrafts: true,
+            },
+          );
+        });
+
+        test('Should create the research output when no link is passed', async () => {
```
lennym, line 519: I would personally test this as an integration test rather than (or as well as) a unit test, since the functionally important thing is that when no value is provided in the browser, then the value passed to Contentful is null/undefined and not an empty string. If the data provider were to add a default value of `''` then this test would still pass, but functionally it would fail.

I would also probably be clear in the test description that the critical thing being tested here is that the value does _not_ default to an empty string (and not simply that the `create` is called at all).

lennym (approved):

PR #3483, opened 2023-08-08
Migration/update entries

gabiayako:
This PR adds an "update mode" to the on-demand migration action which is disabled as default.

Screenshot 2023-08-08 at 13 23 26

When enabled (checkbox checked) it will first try to update the entry but if it gives an error because the entry is not found it will create the entry. Also, it prevents from deleting all entries at the beginning of the script.

Example of a run: https://github.com/yldio/asap-hub/actions/runs/5799477437/job/15719558065

### DIFF
--- a/packages/cms-data-sync/src/utils/migration.ts
+++ b/packages/cms-data-sync/src/utils/migration.ts
@@ -30,43 +30,49 @@ export const migrateFromSquidexToContentfulFactory =
 
         const { id, updateEntry, ...payload } = parsed;
 
-        try {
-          const contentfulEntry = await contentfulEnvironment.getEntry(id);
+        const createEntry = async () => {
+          const createdEntry = await contentfulEnvironment.createEntryWithId(
+            contentTypeId,
+            id,
+            {
+              fields: addLocaleToFields(payload),
+            },
+          );
+          await contentfulRateLimiter.removeTokens(1);
 
-          if (contentfulEntry && (updateEntry || isUpdateModeEnabled)) {
-            updateEntryFields(contentfulEntry, payload);
-            const updatedEntry = await contentfulEntry.update();
-            await contentfulRateLimiter.removeTokens(1);
-
-            n += 1;
-            logger(
-              `Updated entry with id ${id}. (${n}/${data.length})`,
-              'INFO',
-            );
-            return updatedEntry;
+          n += 1;
+          logger(`Created entry with id ${id}. (${n}/${data.length})`, 'INFO');
+          if (item.status === 'PUBLISHED') {
+            return createdEntry;
           }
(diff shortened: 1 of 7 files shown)


packages/cms-data-sync/src/utils/migration.ts
```diff
@@ -57,9 +43,39 @@ export const migrateFromSquidexToContentfulFactory =
...
+
+        const shouldUpdateEntry = updateEntry || isUpdateModeEnabled;
+
+        try {
+          return shouldUpdateEntry
+            ? await updateExistingEntry()
+            : await createEntry();
         } catch (err) {
+          if (err instanceof Error) {
+            const errorParsed = JSON.parse(err.message);
```
lennym, line 70: These message parsing lines should be wrapped in a try-catch so that if the parse fails then the original error is thrown.

See [here](https://github.com/yldio/asap-hub/blob/master/packages/cms-data-sync/src/utils/assets-and-media.ts#L83-L91) for example.

gabiayako: hmm it makes sense, thanks for the reference, I'll change that in a bit

packages/cms-data-sync/src/utils/migration.ts
```diff
@@ -57,9 +43,39 @@ export const migrateFromSquidexToContentfulFactory =
...
+            // this is a fallback when it should have updated the entry
+            // but it does not exist
+            if (shouldUpdateEntry && errorParsed.status === 404) {
```
lennym, line 73: I'm not 100% sure that we want to fall back to creation if the `parseData` function returns an `updateEntry` property. The previous implementation did not. 

```suggestion
            if (isUpdateModeEnabled && errorParsed.status === 404) {
```

gabiayako: good catch, I'll change that to your suggestion!

packages/cms-data-sync/src/utils/setup.ts
```diff
@@ -18,6 +18,9 @@ import { rateLimiter } from './rate-limiter';
 export const isVerbose = () =>
   process.env.VERBOSE_DATA_SYNC && process.env.VERBOSE_DATA_SYNC === 'true';
 
+export const upsertInPlace =
+  process.env.UPSERT_IN_PLACE && process.env.UPSERT_IN_PLACE === 'true';
```
lennym, line 22: Not essential, but it would be nice to be able to easily specify it via a CLI flag when running locally.

```suggestion
  (process.env.UPSERT_IN_PLACE && process.env.UPSERT_IN_PLACE === 'true') ||
  process.argv.includes('--upsert');
```

PR #3489, opened 2023-08-09
feat: CRN-1447 Replace diff with tsc for the verify-step

peterstarling:


peterstarling:
it moans when a field we use is changed from string to boolean type:
image

Riccardo-Brown (approved):

PR #3492, opened 2023-08-09
Crn 1444

lctrt:


apps/storybook/src/OutputVersions.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import { ResearchOutputResponse } from '@asap-hub/model';
+import { OutputVersions } from '@asap-hub/react-components';
+import { NoPaddingDecorator } from './layout';
+
+export default {
+  title: 'Organisms / OutputVersions',
+  component: OutputVersions,
+  decorators: [NoPaddingDecorator],
+};
+const versions: Pick<
```
kmaid, line 10: use ComponentProps to get the type of the versions prop. Will work better

apps/crn-frontend/src/shared-research/state.ts
```diff
@@ -229,6 +229,7 @@ export const useResearchOutputPermissions = (
       userRole,
       published ?? false,
     ),
+    canVersionResearchOutput: hasPublishResearchOutputPermission(userRole),
```
kmaid, line 232: Is this behind the contentful feature flag?

PR #3493, opened 2023-08-09
GP2HUB-65 - update FilterSearchExport placeholder

AkosuaA:


packages/gp2-components/src/templates/DashboardPageBody.tsx
```diff
@@ -93,11 +93,11 @@ const DashboardPageBody: React.FC<DashboardPageBodyProps> = ({
         <div css={contentCardsStyles}>
           <InfoCard
             icon={<LabIcon color="#00202C" size={40} />}
-            title="Samples Completed"
+            title="Samples shared w/AMP PD"
```
marianasrv: where is this copy? in figma, this didn't change.

marianasrv (changes requested):

packages/gp2-components/src/templates/DashboardPageBody.tsx
AkosuaA: It's in the spreadsheet attached to the ticket. It's been changed to 'Samples Processed & Shared' now which I've just updated it to

marianasrv (approved):

PR #3500, opened 2023-08-10
migration: remove throw if fallback create gives an error

gabiayako:


packages/cms-data-sync/src/utils/migration.ts
```diff
@@ -81,13 +81,6 @@ export const migrateFromSquidexToContentfulFactory =
                 return null;
               }
             }
-
-            // if in create mode we could still use
-            // fallbackParseData func so just throw when
-            // not in create mode
-            if (updateEntry || upsertInPlace) {
```
gabiayako, line 88: I haven't added to the fallbackParseData condition because I don't think there's a harm having a fallback parse when we're using update in place. There's no current use case for that now.

lennym (approved):

PR #3502, opened 2023-08-10
add 5xx api gateway cloudwatch alarm

Riccardo-Brown:


apps/crn-server/serverless.ts
peterstarling, line 1: have you considered using serverless syntax for handler SNS subscription? Check it out:
https://www.serverless.com/framework/docs/providers/aws/events/sns

I think that way you could simply define the alarm and topic without permission or subscription in cloudformation. I would give it a go

peterstarling (changes requested):

apps/crn-server/serverless.ts
```diff
@@ -1189,6 +1192,13 @@ const serverlessConfig: AWS = {
         IS_CONTENTFUL_ENABLED: 'true',
       },
     },
+    sendSlackAlert: {
+      handler: './src/handlers/send-slack-alert.handler',
+      condition: 'IsProd',
```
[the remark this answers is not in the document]
Riccardo-Brown: Why would we run the lambda in dev or branch environment to do nothing, when we can not create it to begin with?

apps/crn-server/serverless.ts
Riccardo-Brown: I have got it working using serverless syntax. i haven't used the plugin as it only has one maintainer and we are only creating one alarm. https://github.com/ACloudGuru/serverless-plugin-aws-alerts/issues/147#issuecomment-759429876

apps/crn-server/serverless.ts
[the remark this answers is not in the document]
peterstarling: because then we can actually test our infrastructure elsewhere. IMHO its dangerous to have such conditions in place, there is no reason why the lambda shouldn't run in dev or PR.

peterstarling (changes requested):
the template looks good except i still stand by what i said about not setting the IsProd condition, feel free to bring it up in a channel for discussion if needed

apps/crn-server/serverless.ts
```diff
@@ -1189,6 +1193,21 @@ const serverlessConfig: AWS = {
...
+      events: [
+        {
+          sns: {
+            arn: { Ref: 'TopicCloudwatchAlarm' },
+            topicName: '${self:custom.apiGateway5xxTopic}',
+          },
+        },
+      ],
+      environment: {
+        SLACK_WEBHOOK: SLACK_WEBHOOK!,
```
peterstarling, line 1207: what if you just pass an empty URL here instead, it will error out yeah? maybe thats what we should do? the aim is to have functions work the same way across all environments. So not only should we not conditionally disable parts of the infrastrcture, we should also not have anything like "if (env === "prod") { ..."

peterstarling (changes requested):

apps/crn-server/serverless.ts
Riccardo-Brown: i don't think we should use an empty url to configure if we send alerts to slack for the following reasons

- devs have remember that this is the way to disable it, compared to setting a boolean environment var
- only you have access to retrieve the hook url from slack 

Also i'm not sure i understand what's wrong with setting a flag based on the environment. But i can pass the value down from the workflow file.

peterstarling (approved):

PR #3507, opened 2023-08-10
Crn 974 block algolia access to non onboarded users

oratiu:


kmaid (approved):

packages/algolia/src/index.ts
```diff
@@ -1,23 +1,32 @@
...
+  | AlgoliaSearchClient<App>
+  | NoTokenAlgoliaClient<App>;
+
 type AlgoliaSearchClientNativeFactoryParams = {
-  algoliaApiKey: string;
+  algoliaApiKey: string | null;
   algoliaAppId: string;
 };
 
 export const algoliaSearchClientNativeFactory = ({
```
peterstarling, line 21: i dont think this is used anywhere outside of BE and i guess there is no point returning a client without a key, maybe throw an error instead?

peterstarling (changes requested):

packages/algolia/src/index.ts
oratiu: Oki

peterstarling (approved):

PR #3545, opened 2023-08-17
CT-94 Filter non onboarded users to display team members

gabiayako:
In Contentful we’re not filtering non-onboarded users, so the number of members is bigger in Contentful than in Prod (Squidex)

non-onboarded

This PR adds a filter in Contentful teams data provider so we don't bring non-onboarded users as members.


lennym:
Would it be simpler to apply the filter in the graphql query itself?

lennym (approved):

gabiayako:
> Would it be simpler to apply the filter in the graphql query itself?

I couldn’t find a way to filter in the query :sweat_smile: we can’t add “where” directly in the userCollections and if I try to do a nested query from the top there isn’t a “linkedFrom” option. I even tried “onboarded” directly but it doesn’t allow me
Screenshot 2023-08-17 at 07 36 31
Screenshot 2023-08-17 at 07 36 44

PR #3548, opened 2023-08-17
Remove broken link

kmaid:
Hopefully pass through undefined links


apps/crn-server/src/controllers/research-output.controller.ts
```diff
@@ -144,6 +146,20 @@ export default class ResearchOutputController {
...
+        link: currentResearchOutput?.link,
+        type: currentResearchOutput?.type,
+        documentType:
+          currentResearchOutput?.documentType &&
+          isResearchOutputDocumentType(currentResearchOutput?.documentType)
+            ? currentResearchOutput?.documentType
+            : 'Grant Document',
+      };
+    }
+
```
Riccardo-Brown, line 162: i think the flow of the code would be better if this code was below the  `currentResearchOutput` check. 

Also should we be checking the RO is published before creating the version

apps/crn-server/src/controllers/research-output.controller.ts
```diff
@@ -144,6 +146,20 @@ export default class ResearchOutputController {
     const currentResearchOutput =
       await this.researchOutputDataProvider.fetchById(id);
 
+    let version: ResearchOutputVersionPostRequest | undefined;
+    if (researchOutputUpdateData.createVersion) {
+      version = {
+        title: currentResearchOutput?.title || '',
```
gabiayako, line 152: shouldn't we send the new data here?

kmaid: No, when we make a new version that will replace the existing RO and the existing RO is transformed into the version.

oratiu: I knew I'd find the answer somehow

PR #3549, opened 2023-08-17
CT-92 Fix working group last updated at

gabiayako:
Ticket: https://asaphub.atlassian.net/browse/CT-92

This PR:
* Adds a migration to add a new field `lastUpdated` to `WorkingGroups` content type in contentful
* Adds `lastUpdated` to the migration with its value being squidex's `lastModified`
* This field uses as its editor the app-extension `partial-last-updated` so it gets updated when some value changes
* Changes the Contentful data provider parser to return `lastUpdated` in `lastModifiedDate`


packages/contentful/migrations/crn/workingGroups/20230817113657-add-last-updated-field.js
```diff
@@ -0,0 +1,30 @@
...
+    .omitted(false);
+
+  workingGroups.changeFieldControl(
+    'lastUpdated',
+    'app',
+    'v97H7wgmtstfxNheYWy0G', // last partial update app
+    {
+      // this is a hack, we dont want to exclude any field
+      // but exclude is required
+      exclude: '"non-existent"',
```
lennym, line 22: I have changed the app definition to make this not be required. 
```suggestion
      exclude: '',
```

lennym: We probably want to update the app code so that it still works if the parameter is undefined. I think right now it will throw because we call `.split(',)` on the value without checking it is a string.

peterstarling (approved):

PR #3559, opened 2023-08-18
CT-93/CT-95 Fix interest group and users last updated

gabiayako:
This is very similar to #3549 

Tickets: https://asaphub.atlassian.net/browse/CT-93 / https://asaphub.atlassian.net/browse/CT-95

This PR:

- Adds a migration to add a new field `lastUpdated` to `interestGroups` and `users` content type in contentful
- Adds `lastUpdated` to the `interestGroups` and `users` migrations with its value being squidex's `lastModified`
- This field uses as its editor the app-extension `partial-last-updated` so it gets updated when some value changes
- Changes the `interestGroups` and `users`  Contentful data provider parser to return `lastUpdated` in `lastModifiedDate`


lennym (approved):

gabiayako:
rebasing to master before merge

PR #3570, opened 2023-08-21
GP2HUB-74 gp2 google tag manager

AkosuaA:


.github/workflows/reusable-build.yml
```diff
@@ -192,6 +192,7 @@ jobs:
           REACT_APP_GP2_AUTH0_DOMAIN: ${{ steps.setup.outputs.gp2-auth0-domain }}
           REACT_APP_GP2_SENTRY_DSN: ${{ steps.setup.outputs.react-app-gp2-sentry-dsn }}
           REACT_APP_GTM_CONTAINER_ID: ${{ steps.setup.outputs.react-app-gtm-container-id }}
```
markcsims, line 194: I think this should be renamed to REACT_APP_CRN_GTM_CONTAINER_ID to reflect that it is CRN only.

.github/actions/setup-environment/action.yml
```diff
@@ -233,6 +233,9 @@ outputs:
   react-app-gtm-container-id:
```
markcsims, line 233: I think this should be renamed to react-app-crn-gtm-container-id to reflect that it is CRN only.

markcsims (changes requested):

.github/workflows/reusable-build.yml
AkosuaA: updated

.github/actions/setup-environment/action.yml
AkosuaA: updated

markcsims (approved):

PR #3582, opened 2023-08-22
Linked notion page in README.md

oratiu:


marianasrv (approved):

peterstarling (changes requested):
what about aws-vault?

oratiu:
> 

After reading about AWS-vault, I think it's pretty self explanatory and there isn't much more that I can add besides a link to their github

PR #3590, opened 2023-08-23
CT-104 Interest groups not appearing in User page

gabiayako:
In prod (right side of screenshot) 

interest-groups

when the user does not belong to any team, the API was still returning all interest groups in this call

```
this.interestGroupDataProvider.fetch({ filter: {  teamId: teamIds } })
```

because the Squidex data provider was getting called with an empty filter here (https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/interest-group.data-provider.ts#L63-L89) and since it was not filtering anything it returned all the interest groups which is wrong.

I've changed the controller so it does not even make a call to the data provider if the user does not belong to any team.

Although it should not happen, it seems there are 8 Grantees without teams in Prod. I've ran the following query in Squidex Prod:

```
  queryUsersContentsWithTotal(filter: "empty(data/teams/iv) and data/role/iv eq 'Grantee'") {
    total
    items {
      id
      flatData {
        onboarded
        role
      }
    }
  }
```

### DIFF
--- a/apps/crn-server/src/controllers/interest-group.controller.ts
+++ b/apps/crn-server/src/controllers/interest-group.controller.ts
@@ -62,12 +62,17 @@ export default class InterestGroupController {
     options: FetchPaginationOptions,
   ): Promise<ListInterestGroupResponse> {
     const teamIds = Array.isArray(teamId) ? teamId : [teamId];
-    const { total, items } = await this.interestGroupDataProvider.fetch({
-      filter: {
-        teamId: teamIds,
-      },
-      ...options,
-    });
+    const { total, items } = teamIds.length
+      ? await this.interestGroupDataProvider.fetch({
+          filter: {
+            teamId: teamIds as [string, ...string[]],
+          },
+          ...options,
+        })
+      : {
+          total: 0,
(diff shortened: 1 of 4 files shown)


peterstarling (changes requested):
although this is correct, i think the broken logic lives in the data provider itself. We should require at least one element in this array:

https://github.com/yldio/asap-hub/blob/880905f429d6f9ee55ed6c6f9971e0298ab3975e/packages/model/src/interest-group.ts#L62

which you can do by changing that line to 
```typescript
  teamId?: [string, ...string[]];
```
then TypeScript will forbid you from sending across a string[] which means in your controller you will be forced to check if the array contains at least one element.

We could, of course, continue to allow any number of elements in the array and skip filtering by teamId if the array has 0 elements (instead of filtering by undefined which is what happens now, I believe) but then you will still end up with the same problem and you might accidentally call it with [] and get all results. Alternatively we could return 0 results from the data provider if the array is empty but i think that's silly.

gabiayako:
rebased to master

peterstarling (approved):

PR #3594, opened 2023-08-24
Allow more complex queries for single entities

lennym:
Fetching all the data for an entity on the list endpoint often results in exceeding query complexity limits in Contentful where there are multiple related entities. In particular for research output authors and event speakers.

Where these are not rendered into the list view, or used for searching this can be solved by only loading these relations when querying for a single entity, and not for list endpoints.

Add an extra recoil state key for the list representations so that the full data is always fetched from the server when clicking from a list entry into the full detail page.

Adds an integration tests for users, research outputs and events to fetch the same page size as used in the bulk indexer so we can know as part of a CI pipeline if the indexer will be able to run successfully without hitting complexity limits.


apps/crn-frontend/src/network/interest-groups/state.ts
```diff
@@ -48,18 +48,12 @@ export const interestGroupsState = selectorFamily<
         newInterestGroups === undefined ||
         newInterestGroups instanceof DefaultValue
       ) {
-        const oldInterestGroups = get(interestGroupIndexState(options));
-        if (!(oldInterestGroups instanceof Error)) {
-          oldInterestGroups?.ids?.forEach((id) =>
-            reset(interestGroupState(id)),
-          );
-        }
```
lennym, line 56: This code had no test coverage, nobody knew what it was for, and never seems to be called at runtime, so I removed it.

apps/crn-server/test/controllers/event.controller.test.ts
```diff
@@ -129,7 +129,7 @@ describe('Event controller', () => {
         filter: { googleId, hidden: true },
         take: 1,
       });
-      expect(result).toEqual(getEventResponse());
+      expect(result).toEqual(getEventDataObject(true));
```
peterstarling, line 132: Pretty sure the controller should return a Response type?

apps/crn-server/test/data-providers/event.data-provider.test.ts
```diff
@@ -92,7 +92,7 @@ describe('Event data provider', () => {
           top: 10,
         },
       );
-      expect(result).toEqual(getListEventResponse());
+      expect(result).toEqual(getListEventResponse(false));
```
peterstarling, line 95: - And this should be a DataObject type actually no?

peterstarling (approved):

PR #3604, opened 2023-08-25
GP2HUB-51-add working groups membership validation

AkosuaA:


lennym:
This doesn't appear to actually enforce validation, only display a warning when the users are in an invalid state.

This looks like it's based on a direct copy of the event custom validation app, except that one sets the value of the field to `true` when valid, and `false` when invalid. and adds a validation to enforce that the value must be `true`, but that code seems to be missing here.

Is that intentional?

lctrt (approved):

PR #3606, opened 2023-08-25
fix: add last updated to event when it already exists

gabiayako:
I forgot to add `lastUpdated` to the migration when it is updating an existing event, now it only fills `lastUpdated` if it is creating an event. So this PR fixes it.


peterstarling (approved):

gabiayako:
rebased to master

PR #3626, opened 2023-08-30
Copy settings when running partial algolia sync

lennym:
Removing the settings from the scope of the partial algolia sync was causing the settings on the index to be completely removed when the temporary index was restored (because the temporary index was created with empty settings).

When creating the temporary index for partial indexing runs, do not limit the scope to only synonyms, and allow all data to be copied, and the settings therefore to be preserved when restoring back to the main index.


.github/workflows/reusable-crn-algolia-sync.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           ALGOLIA_INDEX: ${{ steps.setup.outputs.crn-algolia-index }}
           ALGOLIA_INDEX_TEMP: '${{ steps.setup.outputs.crn-algolia-index }}_temp'
           # Do not copy the records if all entities are being synced
-          SCOPE: ${{ inputs.entity == 'all' && '-s settings,synonyms' || '-s synonyms' }}
```
kmaid, line 86: can we just copy everything over regardless of selective sync or not?

kmaid (approved):

PR #3629, opened 2023-08-30
FE for event search

lctrt:


apps/gp2-frontend/src/events/api.ts
```diff
@@ -37,6 +38,43 @@ export const getEvents = async (
...
+  return client
+    .search(['event'], options.searchQuery, {
+      filters,
+      page: options.currentPage ?? 0,
+      hitsPerPage: options.pageSize ?? 10,
+    })
+    .catch((error: Error) => {
+      throw new Error(`Could not search:${error.message}`);
+    });
+};
```
markcsims, line 76: try {
  ...
} catch

I'm not sure why we are catching the error and then throwing another one.

PR #3632, opened 2023-08-30
CRN-1458 - edit version

andreicovaciu:


apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
```diff
@@ -122,7 +122,7 @@ const WorkingGroupOutput: React.FC<WorkingGroupOutputProps> = ({
           this period you will not be able to create or update research outputs
           on the hub. Normal service will resume on 11th September.
         </Toast>
-        {createVersion && (
+        {versionAction && (
```
gabiayako, line 125: Looking at the figma, it seems to me that this should only appear when the person is creating a new version. But it would be nice to confirm with Tiff or João.

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
```diff
@@ -122,7 +122,7 @@ const WorkingGroupOutput: React.FC<WorkingGroupOutputProps> = ({
           this period you will not be able to create or update research outputs
```
gabiayako, line 122: We don't want to show this after the migration is over 😅 so it wasn't necessary to rebase it

gabiayako (changes requested):

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
gabiayako: are you sure? the figma link attached to the card https://www.figma.com/file/JgcEOEbozf8CzVjA0SLsaq/CRN-%2F-Designs-%2F-Shared-Research?type=design&node-id=22282-192837&mode=design&t=ZOSAca3tMJa6BXu8-4 does not have the toast "The previous output page will..." it only has the version output card

PR #3639, opened 2023-08-31
CT-117 Events created via webhook not being published

gabiayako:
Events synced via webhook are not being published because the new field `lastUpdated` is required and it is not being filled in the sync.

In this PR it was added a migration to make `lastUpdated` not required and if the value is null it returns the `publishedAt` from Contentful.


peterstarling (approved):

gabiayako:
rebased to master

PR #3640, opened 2023-09-01
Enable Contentful feature flag by default

lennym:
* update the default deployment config to enable feature flag
* set feature flag to be enabled in the frontend
* disregard any headers sent to the server when checking feature flag
* pass incorrect squidex url to deployments so squidex requests from backend fail


apps/crn-server/test/integration/fixtures/contentful.ts
```diff
@@ -389,12 +389,12 @@ export class ContentfulFixture implements Fixture {
       content_type: 'events',
     });
 
-    events.forEach(async (entry) => {
+    for (const entry of events) {
```
peterstarling, line 392: if this is a fix not part of the switch then maybe best to take it out in case we were revert this commit/PR at some point or whatever

peterstarling (approved):

PR #3648, opened 2023-09-04
CRN-1479-add-icons-to-the-guides-card

kmaid:


apps/crn-frontend/src/tags/Tags.tsx
```diff
@@ -0,0 +1,19 @@
+import { TagsPage, TagsPageBody } from '@asap-hub/react-components';
+import { useSearch } from '../hooks';
+
+const Tags: React.FC<Record<string, never>> = () => {
+  const { tags, setTags } = useSearch();
+  return (
+    <TagsPage tags={tags} setTags={setTags}>
+      <TagsPageBody
+        currentPage={0}
```
kmaid, line 9: This is all hardcoded temporatily because we haven't done the search part yet

PR #3651, opened 2023-09-05
Map empty string values in user profile to null

lennym:
Passing an empty string for a non-required field with pattern or "allowed values" validators throws an error. In particular when onboarding a new user, the "Degree" field is marked as optional on the frontend, but not selecting a value causes an error because `""` is not one of the allowed values in Contentful.

Map any empty string values to `null` in the data provider to prevent the validator throwing an error.

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/user.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/user.data-provider.ts
@@ -217,6 +217,9 @@ const cleanUser = (userToUpdate: UserUpdateDataObject) =>
         connections: connections.map(({ code }) => code),
       };
     }
+    if (typeof value === 'string') {
+      return { ...acc, [key]: value.trim() === '' ? null : value };
+    }
     return { ...acc, [key]: value };
   }, {} as { [key: string]: unknown });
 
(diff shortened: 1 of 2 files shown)


apps/crn-server/src/data-providers/contentful/user.data-provider.ts
```diff
@@ -217,7 +217,7 @@ const cleanUser = (userToUpdate: UserUpdateDataObject) =>
         connections: connections.map(({ code }) => code),
       };
     }
-    return { ...acc, [key]: value };
+    return { ...acc, [key]: value === '' ? null : value };
```
gabiayako, line 220: do you think maybe we could leave as `value.trim() === ''`? not sure if the user could leave with empty space by mistake in this flow

gabiayako (approved):

apps/crn-server/src/data-providers/contentful/user.data-provider.ts
lennym: For this particular issue - the degree field in onboarding - they cannot because it's a drop-down menu, but generally it's probably worth it.

PR #3654, opened 2023-09-06
CRN-1475 - adds try-catch blocks on events and users searches

andreicovaciu:


apps/crn-frontend/src/events/__tests__/api.test.ts
```diff
@@ -58,6 +58,17 @@ describe('getEvents', () => {
...
   });
 
+  it('should return an error and not retry the search', async () => {
+    search.mockRejectedValueOnce(new Error('Network error'));
+    await expect(
+      getEvents(
+        algoliaSearchClient,
+        getEventListOptions(new Date('2021-01-01T12:00:00'), { past: true }),
+      ),
+    ).rejects.toThrowError('Failed to fetch events.');
```
peterstarling, line 68: before you made your changes this tests would fail only because the error message would be different, but the promise would still reject with an Error and search would be called once, so I don't think this really proves anything,

 May I ask where the retry logic is located? I think it is where you want to be showing that you fixed the problem. I would suggest TDDing these type of bugs, start by replicating an issue through a failed test that you then fix.

peterstarling (changes requested):

PR #3656, opened 2023-09-06
Scroll to top on all RO pages

kmaid:


lennym:
Is it possible that this will be incredibly annoying if it loses scroll position when navigating "back" on long pages?

packages/react-components/src/atoms/ScrollToTop.tsx
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+
+const ScrollToTop: React.FC = () => {
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
```
Riccardo-Brown, line 8: Should this component be able to be configured to set the scroll behaviour to smooth?

kmaid: I think the user just wants to be at the top of the page when the path changes. They shouldn't have been at the bottom of the next page to start with

lennym (approved):

Riccardo-Brown (approved):

PR #3658, opened 2023-09-07
first pass on algolia user search

lctrt:


apps/gp2-frontend/src/users/UserDirectory.tsx
gabiayako, line 1: It seems the export does not work properly when we add a keywords filter because it is still fetching from our BE using `getUsers` and before it was fetching by name of the tag and now it uses the id, so the response comes empty

before

after

gabiayako (changes requested):

apps/gp2-frontend/src/users/UserDirectory.tsx
markcsims: It should be using the api to get the users response. Non-onboarded users are not in Algolia. 

The filter wasn't correctly implemented. This is now fixed.

gabiayako (approved):
Nice! 🎉

PR #3677, opened 2023-09-11
feat: Remove Squidex Data Providers

peterstarling:
- remove data providers
- remove some parsers


apps/crn-server/src/utils/fetch-orcid.ts
```diff
@@ -1,6 +1,8 @@
+import { OrcidWork } from '@asap-hub/model';
 import Got from 'got';
 import get from 'lodash.get';
-import { CMSOrcidWork } from '../data-providers/user.data-provider';
+
+export type CMSOrcidWork = OrcidWork;
```
lennym, line 5: This doesn't seem to be `import`ed anywhere else. Is there a reason to `export` it?

peterstarling, line 5:
```suggestion
type CMSOrcidWork = OrcidWork;
```

lennym (approved):

PR #3683, opened 2023-09-12
Changed display mode of the Pill component to flex

oratiu:
There was a bug with the text inside the Pill component. It wasn't centered.
This fix centers the text inside the pill component

### DIFF
--- a/packages/react-components/src/atoms/Pill.tsx
+++ b/packages/react-components/src/atoms/Pill.tsx
@@ -6,7 +6,8 @@ import { lineHeight, perRem } from '../pixels';
 
 const borderWidth = 1;
 const styles = css({
-  display: 'inline-block',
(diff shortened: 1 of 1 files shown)


lctrt (approved):

packages/react-components/src/atoms/Pill.tsx
```diff
@@ -6,7 +6,8 @@ import { lineHeight, perRem } from '../pixels';
 
 const borderWidth = 1;
 const styles = css({
-  display: 'inline-block',
+  display: 'flex',
+  alignItems: 'center',
```
gabiayako, line 10: I think we need to constraint the width. It's looking like this on News
Screenshot 2023-09-12 at 08 15 55

gabiayako (approved):

PR #3686, opened 2023-09-12
fix project search hiding bar when loading

lctrt:


apps/gp2-frontend/src/projects/ProjectDirectory.tsx
```diff
@@ -1,4 +1,6 @@
-import { ProjectsBody } from '@asap-hub/gp2-components';
+import { Frame } from '@asap-hub/frontend-utils';
+import { ProjectPageList, ProjectsBody } from '@asap-hub/gp2-components';
+// import { ProjectPageList } from '@asap-hub/gp2-components/build/templates';
```
marianasrv, line 3: remove this line

marianasrv (changes requested):

packages/gp2-components/src/templates/ProjectPageList.tsx
marianasrv, line 1: To keep it consistent, may you add mainStyles to <main> in ProjectsPage? Just like we have in OutputsPage

marianasrv (changes requested):

marianasrv (approved):

PR #3689, opened 2023-09-13
[CRN-1495] fix(version-reminders): use output's current title

Riccardo-Brown:
Ensure version reminders use the current title of the related research output

JIRA: https://asaphub.atlassian.net/browse/CRN-1495


apps/crn-server/test/fixtures/reminders.fixtures.ts
```diff
@@ -674,12 +674,12 @@ export const getContentfulReminderResearchOutputVersionCollectionItem =
         id: 'ec3086d4-aa64-4f30-a0f7-5c5b95ffbcca',
         publishedAt: '2020-09-23T16:34:26.842Z',
       },
-      documentType: 'Bioinformatics',
-      title: 'test-research-output-version',
```
peterstarling, line 678: you might want to keep a different title here, that way the before and after would be more visible since you would simply show that previously it used this title and now its going to get one from the relationship

peterstarling (approved):

apps/crn-server/test/fixtures/reminders.fixtures.ts
Riccardo-Brown: To keep the title of the version would require keeping that field in the graphql query. This would go against graphql as we are requesting data we don't need.

PR #3690, opened 2023-09-13
GP2HUB-22 Add meeting link to event series

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/GP2HUB-22

---

This PR adds a new field called `copyMeetingLink` to the Events schema

Screenshot 2023-09-14 at 09 10 48

If the user clicks yes in this `copyMeetingLink`, adds a meeting link and publishes the event, the meeting link will be copied to all events that belong to the series (the ones that share a common googleId prefix).


peterstarling (changes requested):
I think there might a missed edge case where if you update one event and set the copyMeetingLink to true and then you have another event from the series with that property set then it will end up staying in an endless loop? might want to write a test that shows that it will never go into an endless loop - after all we are listening to event updates and making updates to events so it sounds possible

gabiayako:
> I think there might a missed edge case where if you update one event and set the copyMeetingLink to true and then you have another event from the series with that property set then it will end up staying in an endless loop? might want to write a test that shows that it will never go into an endless loop - after all we are listening to event updates and making updates to events so it sounds possible

I'm not updating `copyMeetingLink`, just `meetingLink` so it wouldn't fall into an infinite loop. I've added this test https://github.com/yldio/asap-hub/pull/3690/commits/4dd76b53a6646cbbbe79db6501402d687ab31856

Or was it a requirement to update the `copyMeetingLink`?

apps/gp2-server/src/handlers/event/update-events-meeting-link-handler.ts
```diff
@@ -0,0 +1,78 @@
...
+      log.info(`Found ${seriesEvents.length} series events`);
+
+      for (const seriesEvent of seriesEvents) {
+        log.info(
+          `Updating event ${seriesEvent.id} with meeting link ${calendarEvent.meetingLink}`,
+        );
+
+        try {
+          await eventController.update(seriesEvent.id, {
+            meetingLink: calendarEvent.meetingLink,
```
peterstarling, line 57: you may want to set the `copyMeetingLink` here to false as well

peterstarling (changes requested):

apps/gp2-server/test/handlers/event/update-events-meeting-link-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+import {
+  getCommonGoogleId,
+  meetingLinkUpdateHandler,
+} from '../../../src/handlers/event/update-events-meeting-link-handler';
+import { getEventEvent, getEventResponse } from '../../fixtures/event.fixtures';
+import { eventControllerMock } from '../../mocks/event.controller.mock';
+import { loggerMock } from '../../mocks/logger.mock';
+
+describe('getCommonGoogleId', () => {
+  test('works when it receives a series googleId event', () => {
```
peterstarling, line 12: im not a big fan of the the use of words like `works` in tests - it should say what the function does, ie `should extract commong google ID from the event ID` or something like that

apps/gp2-server/test/handlers/event/update-events-meeting-link-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    const webhookEvent = getEventEvent(seriesEvent1.id, 'EventsPublished');
+
+    await handler(webhookEvent);
+
+    expect(eventControllerMock.fetchById).toHaveBeenCalledWith(seriesEvent1.id);
+    expect(eventControllerMock.fetch).toHaveBeenCalledWith({
+      filter: { googleId: 'googleId' },
+    });
+
+    expect(eventControllerMock.update).toHaveBeenNthCalledWith(
```
peterstarling, line 66: maybe you want to show that it has not been called with `seriesEvent1` data? i think otherwise it would pass even without filtering that event out, or maybe not because of the use of `nthCalled` but still

peterstarling (approved):

apps/gp2-server/test/handlers/event/update-events-meeting-link-handler.test.ts
peterstarling: in fact i dont think you need the nth-called expectations right? what difference does it make which order they are being updated in, for all i care they could be updated in parallel, as long as we skip the one that caused the change

PR #3692, opened 2023-09-13
[CRN-1492] refactor(research-output indexer): populate _tags field

Riccardo-Brown:
JIRA: https://asaphub.atlassian.net/browse/CRN-1492

### DIFF
--- a/apps/crn-server/scripts/export-entity.ts
+++ b/apps/crn-server/scripts/export-entity.ts
@@ -84,8 +84,10 @@ const getController = (entity: keyof EntityResponsesCRN) => {
 };
 
 const transformRecords = <T extends EntityResponsesCRN, K extends keyof T>(
-  record: T[K] & { id: string },
-  type: K,
+  record: T[K] extends EntityResponsesCRN[keyof EntityResponsesCRN]
+    ? T[K] & { id: string }
+    : never,
+  type: K extends keyof EntityResponsesCRN ? K : never,
 ) => {
   const payload = {
     ...record,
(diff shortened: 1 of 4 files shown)


apps/crn-server/scripts/export-entity.ts
```diff
@@ -86,10 +86,29 @@ const getController = (entity: keyof EntityResponsesCRN) => {
 const transformRecords = <T extends EntityResponsesCRN, K extends keyof T>(
```
peterstarling, line 86: I tried to: 
1. Get rid of casting
2. Remove unnecessary ts-ignore

I've achieved it by doing this:
```typescript
const transformRecords = <T extends EntityResponsesCRN, K extends keyof T>(
  record: T[K] extends EntityResponsesCRN[keyof EntityResponsesCRN]
    ? T[K] & { id: string }
    : never,
  type: K extends keyof EntityResponsesCRN ? K : never,
) => {
  const payload = {
    ...record,
    objectID: record.id,
    __meta: {
      type,
    },
  };

  if (type === 'research-output' && 'subtype' in record) {
    const subtype = record.subtype;

    return {
      ...payload,
      _tags: [
        ...record.methods,
        ...record.organisms,
        ...record.environments,
        ...(subtype ? [subtype] : []),
        ...record.keywords,
      ],
    };
  }
```

What's changed:
* added conditional type checks for function arguments to narrow the types down
* removed ts-ignore comment and replace mutation with a simple return
* added extra condition to recognise the type of `record`, then removed all casts

apps/crn-server/test/handlers/research-output/algolia-index-research-output-handler.test.ts
```diff
@@ -30,7 +30,25 @@ describe('Research Output index handler', () => {
     await indexHandler(createEvent('ro-1234'));
 
     expect(algoliaSearchClientMock.save).toHaveBeenCalledWith({
-      data: researchOutputResponse,
+      data: expect.objectContaining(researchOutputResponse),
+      type: 'research-output',
+    });
+  });
+
+  test('Should populate _tags field before saving the research-output to Algolia', async () => {
```
peterstarling, line 38: I think it's best if you explicitly declare a tag on each of the properties (environments, subtype, etc) to make it easier to understand (and more future proof).

peterstarling (changes requested):

peterstarling (approved):

PR #3695, opened 2023-09-14
Make news search feature use algolia

lctrt:


markcsims (approved):

marianasrv (commented):
I see that you are no longer getting data from Contentful - is it supposed to be like that?

PR #3697, opened 2023-09-15
GP2HUB-92 bugfix: Events not being created when calendar is added

gabiayako:
An error was happening during sync google events due to a field that is present in CRN events but not present in GP2 events.

Log in AWS: https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fgp2-hub-3690-eventsUpdated/log-events/2023%2F09%2F13%2F[%24LATEST]f7949a7fcf76436ba97bf7cac4b356f6?start=PT72H

Screenshot 2023-09-14 at 09 37 35

Somehow typescript didn't catch this. So I'm using type narrowing now to ensure tags won't be sent to GP2 events. 

--- 

After the changes in this PR, I've manually added a calendar to both CRN and GP2 and checked that the events are syncing properly

CRN: [https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fasap-hub-3697-gcalEventsUpdatedContentful/log-events/2023%2F09%2F15%2F[%24LATEST]56e402e041e347d7bdd705b02c77a191?start=PT3H](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fasap-hub-3697-gcalEventsUpdatedContentful/log-events/2023%2F09%2F15%2F%5B%24LATEST%5D56e402e041e347d7bdd705b02c77a191?start=PT3H)

GP2: [https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fgp2-hub-3697-eventsUpdated/log-events/2023%2F09%2F15%2F[%24LATEST]4a332d31edef4fbb855df2262ae2720d?start=PT3H](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fgp2-hub-3697-eventsUpdated/log-events/2023%2F09%2F15%2F%5B%24LATEST%5D4a332d31edef4fbb855df2262ae2720d?start=PT3H)

### DIFF
--- a/packages/server-common/src/utils/sync-google-event.ts
+++ b/packages/server-common/src/utils/sync-google-event.ts
@@ -4,6 +4,7 @@ import {
   GoogleEvent,
   validateGoogleEvent,
 } from '../validation/sync-google-event.validation';
+import { isCRNEventController } from './type-narrowing';
 import { Logger } from './logger';
 
 export type SyncEvent = (
@@ -79,16 +80,34 @@ export const syncEventFactory =
         return eventsController.update(existingEvent.id, eventToUpdate);
       }
 
-      const eventToCreate = {
-        ...newEvent,
-        hidden: newEvent.status === 'Cancelled',
-        tags: [],
-      };
+      const hidden = newEvent.status === 'Cancelled';
+      if (isCRNEventController(eventsController)) {
+        const crnEventToCreate = {
+          ...newEvent,
+          hidden,
+          tags: [],
+        };
+
+        logger.info(
+          {
+            id: googleEvent.id,
+            event: crnEventToCreate,
+          },
+          'Event not found. Creating.',
+        );
+        return eventsController.create(crnEventToCreate);
+      }
+
+      const gp2EventToCreate = { ...newEvent, hidden };
+
       logger.info(
-        { id: googleEvent.id, event: eventToCreate },
+        {
+          id: googleEvent.id,
+          event: gp2EventToCreate,
+        },
         'Event not found. Creating.',
       );
-      return eventsController.create(eventToCreate);
+      return eventsController.create(gp2EventToCreate);
     } catch (err) {
       logger.error(err, 'Error syncing event');
       throw err;
--- a/packages/server-common/src/utils/type-narrowing.ts
+++ b/packages/server-common/src/utils/type-narrowing.ts
@@ -1,0 +1,6 @@
+/* istanbul ignore file */
+import { EventController, gp2 } from '@asap-hub/model';
+
+export const isCRNEventController = (
+  controller: EventController | gp2.EventController,
+): controller is EventController => 'tags' in controller.create;
--- a/packages/server-common/test/utils/sync-google-event.test.ts
+++ b/packages/server-common/test/utils/sync-google-event.test.ts
@@ -1,5 +1,6 @@
 import { calendar_v3 as calendarV3 } from 'googleapis';
 import { syncEventFactory } from '../../src';
+import * as typeNarrowing from '../../src/utils/type-narrowing';
 import { getEventResponse } from '../fixtures/events.fixtures';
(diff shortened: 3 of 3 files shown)


packages/server-common/src/utils/sync-google-event.ts
```diff
@@ -79,16 +80,34 @@ export const syncEventFactory =
...
-        ...newEvent,
-        hidden: newEvent.status === 'Cancelled',
-        tags: [],
-      };
+      const hidden = newEvent.status === 'Cancelled';
+      if (isCRNEventController(eventsController)) {
+        const crnEventToCreate = {
+          ...newEvent,
+          hidden,
+          tags: [],
```
peterstarling, line 88: so you say we can only add `tags` when it's a CRN event but if that was the case we would never be able to add to the input object for GP2 - can you make sure gp2 types reflect it?

peterstarling (changes requested):

packages/server-common/src/utils/sync-google-event.ts
gabiayako: (I've left the same message on Slack but I believe it's better to have it here as well)

To be honest I don’t know why typescript didn’t catch that tags can’t be added if the controller can have the type gp2.EventController  🤔 

The types are already correct, the CRN controller with type EventController accepts this EventCreateDataObject  (with tags) https://github.com/yldio/asap-hub/blob/master/packages/model/src/event.ts#L82:L97 as the event type in create while the GP2 controller accepts this EventCreateDataObject (without tags) https://github.com/yldio/asap-hub/blob/master/packages/model/src/gp2/event.ts#L44:L59 as the event type

The only way I was able to make TS display an error was with this type guard

gabiayako: I've spent some time trying to understand why it does not give an error the way it is on master now. I wrote this example in my VSCode

```
interface CreatePropsDog {
  age: number;
  name: string;
  owner: string;
  birthdate: string;
  breed: string;
}
interface Dog {
  create: (pet: CreatePropsDog) => string;
}

interface CreatePropsCat {
  age: number;
  name: string;
  owner: string;
  birthdate: string;
  color: string;
}

interface Cat {
  create: (pet: CreatePropsCat) => string;
}

let pet: Dog | Cat = {
  create: () => 'hey',
};

pet.create({
  age: 12,
  name: 'Pet',
  owner: 'Owner',
  birthdate: '28/01/2021',
  breed: 'Chihuahua',
  color: 'black',
});
```

And it seemed odd to me that I could pass `breed` only present in DogProps and `color` only present in CatProps without any TS errors and it turns out when I hover `create` it gives me this

Screenshot 2023-09-15 at 14 05 31

It seems Typescript makes a intersection of arguments when there's a union type https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-3.html#improved-behavior-for-calling-union-types

That's why it doesn't give an error the way it is now on master when we have `tags: []` in the create payload, because even though GP2 events don't have this field CRN events do.

peterstarling: yeah I was obviously wrong when I said it - TypeScript will always accept an object with tags property because of reasons you said (intersection when using unions). There is no excess property checking either. So yeah we would need to find a way to somehow link this with a passed generic and enforce type checking when the factory is invoked (which is the either crn or gp2 handler at which point we know exactly which input type we are dealing with) but lets leave it for now. This works so i am approving

peterstarling (approved):

gabiayako:
rebased to master

PR #3713, opened 2023-09-20
GP2HUB-89 Add first section fields to research output

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/GP2HUB-89

---

This PR
- Adds a migration to add the fields description, gp2Supported and sharingStatus to the Outputs schema in Contentful
- Adds description, gp2Supported and sharingStatus to the output query and the data provider response
- Adds the inputs for these description, gp2Supported, sharingStatus and publishDate to the OutputForm in the FE
- Adds the contraints below written in the ticket to the OutputForm in the FE
    - [GP2 Supported] Note: (Blog and GP2 Reports document types - are always set to Yes)
    - [GP2 Supported] Note: This is not required for Training Materials and Procedural Form, do not display this question)
    - [Sharing Status] Note: (Training Materials - always public)
    - [Publish Date] Note: (Publish Date only required if sharing status is Public)


packages/model/src/research-output.ts
```diff
@@ -182,6 +182,11 @@ export const researchOutputToIdentifierType: Record<
   Report: [],
 };
 
+export const gp2SharingStatuses = ['GP2 Only', 'Public'] as const;
+
+export type GP2ResearchOutputSharingStatus =
+  (typeof gp2SharingStatuses)[number];
+
```
markcsims, line 189: This is GP2 specific and should be under the gp2 model. Probably in outputs.

packages/gp2-components/src/utils/research-output-form.ts
markcsims, line 1: We have outputs in GP2 not research outputs.

peterstarling: and i wonder whether that is a good thing or a bad thing that they are different but only a little different. Like... the idea is to eventually nearly align them two so should we also not name them as research outputs?

packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -83,14 +96,36 @@ const OutputForm: React.FC<OutputFormType> = ({
...
+  const [newGp2Supported, setGp2Supported] = useState<
+    gp2Model.DecisionOption | undefined
+  >(isGP2SupportedAlwaysTrue ? 'Yes' : gp2Supported);
+  const [newSharingStatus, setSharingStatus] =
+    useState<gp2Model.OutputSharingStatus>(
+      isAlwaysPublic ? 'Public' : sharingStatus || 'GP2 Only',
+    );
+
+  const [newPublishDate, setPublishDate] = useState<Date | undefined>(
+    getPublishDate(publishDate) || undefined,
```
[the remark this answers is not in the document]
gabiayako: done! committed here [review: refactor to remove output-form.ts file](https://github.com/yldio/asap-hub/pull/3713/commits/3537aac09a2915ce95245573908e43f69cf939bb)

markcsims (approved):

PR #3717, opened 2023-09-20
GP2HUB-93 - additional information section in outputs

marianasrv:


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -134,6 +154,25 @@ const OutputForm: React.FC<OutputFormType> = ({
...
+    tags || [],
+  );
+
+  const identifierType: gp2Model.OutputIdentifierType = doi
+    ? gp2Model.OutputIdentifierType.DOI
+    : rrid
+    ? gp2Model.OutputIdentifierType.RRID
+    : accessionNumber
+    ? gp2Model.OutputIdentifierType.AccessionNumber
+    : title // if it's editing
```
gabiayako, line 167: I'm curious about this `title` here. If it's editing and the values of doi, rrid and accessionNumber are undefined, shouldn't be none directly instead of this title?
Screenshot 2023-09-27 at 13 07 02

gabiayako (approved):

packages/gp2-components/src/templates/OutputForm.tsx
marianasrv: yes, but if it's creating a new output, we want to have nothing selected (empty). in that case, title is not defined.

gabiayako: I see, it makes sense

PR #3726, opened 2023-09-27
fix: Sync ORCID handler query complexity error

peterstarling:
* Moves integration tests out of 'common' into root dir
* Fixes the issue with query complexity in cronjob-sync-orcid handler
* Moves away from Promise.all for processing users in cronjob-sync-orcid handler
* Adds an integration test to prevent regression


gabiayako (approved):
nice 🎉

PR #3728, opened 2023-09-27
GP2HUB-109 gp2Supported as not required

gabiayako:
This PR makes the field `gp2Supported` to be not required since for document types Training Materials and Procedural Form this field does not even appear in the form in the hub. It also makes the default value to be "Don't Know" when it doesn't have a previous value in the form.

### DIFF
--- a/packages/gp2-components/src/templates/OutputForm.tsx
+++ b/packages/gp2-components/src/templates/OutputForm.tsx
@@ -25,6 +25,11 @@ import { EntityMappper } from './CreateOutputPage';
 
 const { rem } = pixels;
 
+const DOC_TYPES_GP2_SUPPORTED_NOT_REQUIRED = [
+  'Training Materials',
+  'Procedural Form',
+];
+
 const getBannerMessage = (
   entityType: 'workingGroup' | 'project',
   documentType: gp2Model.OutputDocumentType,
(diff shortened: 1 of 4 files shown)


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -253,12 +260,10 @@ const OutputForm: React.FC<OutputFormType> = ({
...
               }
             />
-            {!['Training Materials', 'Procedural Form'].includes(
-              documentType,
-            ) ? (
+            {!DOC_TYPES_GP2_SUPPORTED_NOT_REQUIRED.includes(documentType) ? (
               <LabeledRadioButtonGroup<gp2Model.DecisionOption>
                 title="Has this output been supported by GP2?"
-                subtitle="(required)"
+                subtitle="(optional)"
```
marianasrv, line 266: This field should be filled when it is shown, so for the user, we should say it's required, as it even has a default value "Don't Know". If it was optional, we would have nothing selected as default. does it make sense?

marianasrv (changes requested):

packages/gp2-components/src/templates/OutputForm.tsx
gabiayako: Ok! it makes sense

peterstarling (commented):
if we are no longer requiring it in the CMS would it make sense to add some validation for this in the BE?

packages/gp2-components/src/templates/OutputForm.tsx
gabiayako: committed here [change gp2 supported subtitle to required again](https://github.com/yldio/asap-hub/pull/3728/commits/547256297716ad63e79e9f48b7c729b9f9c216c3)

gabiayako:
> if we are no longer requiring it in the CMS would it make sense to add some validation for this in the BE?

I've added a test here [adds a test in BE to check that gp2Supported is not required](https://github.com/yldio/asap-hub/pull/3728/commits/7c937de631b0755b3cb3ab9945957ba007696f63)

marianasrv (approved):

PR #3739, opened 2023-09-29
[CRN-1488] refactor: reduce data indexed in algolia

Riccardo-Brown:


packages/model/src/research-output.ts
```diff
@@ -404,3 +423,51 @@ export const convertBooleanToDecision = (bool?: boolean): DecisionOption => {
...
+    isInReview,
+    methods,
+    organisms,
+    environments,
+    subtype,
+    keywords,
+  } = researchOutput;
+
+  return {
+    id,
```
gabiayako, line 452: I think we need to add "publishingEntity" otherwise it appears like this
Screenshot 2023-11-15 at 09 49 42

gabiayako (changes requested):
It seems "Export as CSV" currently uses algolia data, so this breaks the export as it is today.

apps/crn-frontend/src/shared-research/api.ts
```diff
@@ -145,6 +145,32 @@ export const getDraftResearchOutputs = async (
   };
 };
 
+// As algolia has reduced data we need to fetch data from cms for `Export as CSV` feature
+export const getResearchOutputsFromCMS = async (
```
gabiayako, line 149: The export using source filters (`teams`/`working groups`) is not working properly
Screenshot 2023-11-29 at 11 11 05

gabiayako (changes requested):

PR #3742, opened 2023-10-02
feat: Add ORCID sync to GP2 user model

peterstarling:
Add ORCID live and scheduled sync.


packages/server-common/src/utils/fetch-orcid.ts
```diff
@@ -0,0 +1,103 @@
+import { OrcidWork } from '@asap-hub/model';
+import Got from 'got';
+import get from 'lodash.get';
```
markcsims, line 3: Is there some other alternative to using lodash?

PR #3744, opened 2023-10-03
adds related outputs migration

markcsims:


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -75,6 +91,11 @@ type OutputFormType = {
     typeof AuthorSelect
   >['loadOptions'];
   keywordSuggestions: gp2Model.TagDataObject[];
+  getRelatedOutputSuggestions?: NonNullable<
```
gabiayako, line 94: Does it give an error somewhere if we make this props required?
```suggestion
  getRelatedOutputSuggestions: NonNullable<
```
I saw an error in CRN once where we forgot to pass the suggestion function to Working Groups but we passed to Teams but the card appeared in both so it was working for Teams but not for Working Groups. And if the suggestion function was required in the form we would find this error with Typescript

packages/gp2-components/src/organisms/OutputRelatedResearchCard.tsx
```diff
@@ -0,0 +1,47 @@
+import { ResearchOutputRelatedResearchCard } from '@asap-hub/react-components';
+import { EmotionJSX } from '@emotion/react/types/jsx-namespace';
+import { ComponentProps } from 'react';
+import {
+  outputArticle,
+  outputCode,
+  outputDataset,
+  outputForm,
+  outputMaterial,
+  outputReport,
```
gabiayako, line 10: The icons are bigger than the space they should fit into

Screenshot 2023-10-06 at 07 52 46

I think adding viewBox to the svgs will fix it :)

packages/contentful/src/gp2/queries/output.queries.ts
```diff
@@ -69,6 +69,17 @@ export const outputsContentQueryFragment = gql`
         name
       }
     }
+    relatedOutputsCollection(limit: 10) {
```
gabiayako, line 72: I've noticed that it's possible to add more than 10 related outputs in the hub (typing the title of an output after adding all available suggestions). I wonder if there's a way to block adding only 10 in the UI, it was not mentioned in the ticket though. So not sure it's a big issue.

gabiayako (approved):

PR #3745, opened 2023-10-03
GP2HUB-76 Update keywords to tags - Part 1

gabiayako:
This PR

- [x] Creates tags content-type
- [x] Adds tags field to event content-type (it was called keywords before)
- [x] Adds tags as valid linked content-type to tags fields in outputs, projects, users, working groups (it was already called tags, but accepted as linked content type only Keywords) 
- [x] Adds a script to copy all data from keywords to tags, with the same id + suffix  `-tag`. It can’t have the same id as Contentful throws a conflict error.
- [x] Adds a script to recreate links for outputs, projects, users, working groups and events but using tags, removes keywords
- [x] Makes the code changes related to the backend. Some things are still named “”keywords”, for example, the filters as this would require a lot of frontend changes as well. This will be changed in a future PR.

In the next PR I intend to
- [ ] Delete the field keywords from events
- [ ] Delete keywords content-type
- [ ] Change the filters in the backend from keywords to tags
- [ ] More renames in FE not covered in this PR

---

Notes:

It's not possible to rename the content-type id only its name that's why it was necessary to create a new content type and migrate the data. I've asked in contentful telegram recently:
Screenshot 2023-10-03 at 12 55 26
Another reference: https://www.contentfulcommunity.com/t/is-there-a-way-to-modifiy-a-content-type-id/1023

---

Manually tested in the hub

- [x] It’s possible to add tags to my profile
- [x] It’s possible to export users as CSV filtering by tags


apps/gp2-frontend/src/shared/__tests__/api.test.ts
```diff
@@ -12,10 +12,10 @@ describe('getKeywords', () => {
     nock.cleanAll();
   });
   it('returns a successfully fetched keywords list', async () => {
-    const keywordsResponse: gp2Model.ListKeywordsResponse =
-      gp2Fixtures.createKeywordsResponse();
+    const keywordsResponse: gp2Model.ListTagsResponse =
```
marianasrv, line 15: maybe we should also rename variables. in this case, tagsResponse. and also update the test name

apps/gp2-frontend/src/shared/state.ts
```diff
@@ -10,7 +10,7 @@ const keywordsSelector = selector({
     return getKeywords(authorization);
   },
 });
-const keywordsState = atom<gp2.ListKeywordsResponse>({
+const keywordsState = atom<gp2.ListTagsResponse>({
```
marianasrv, line 13: same with state and selector above

apps/gp2-frontend/src/shared/api.ts
```diff
@@ -4,8 +4,8 @@ import { API_BASE_URL } from '../config';
 
 export const getKeywords = async (
```
marianasrv, line 5: change this to getTags, as well as all the mocks

packages/gp2-components/src/organisms/KeywordsModal.tsx
```diff
@@ -18,7 +18,7 @@ const KeywordsModal: React.FC<KeywordsModalProps> = ({
   tags,
   suggestions,
 }) => {
-  const [newKeywords, setNewKeywords] = useState<gp2.KeywordDataObject[]>(
+  const [newKeywords, setNewKeywords] = useState<gp2.TagDataObject[]>(
```
marianasrv, line 21: same here

packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -74,7 +74,7 @@ type OutputFormType = {
   readonly getAuthorSuggestions?: ComponentPropsWithRef<
     typeof AuthorSelect
   >['loadOptions'];
-  keywordSuggestions: gp2Model.KeywordDataObject[];
+  keywordSuggestions: gp2Model.TagDataObject[];
```
marianasrv, line 77: here

packages/gp2-components/src/templates/__tests__/UsersPageList.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('UsersPageList', () => {
       const { items: projects } = gp2Fixtures.createProjectsResponse();
       const { items: workingGroups } =
         gp2Fixtures.createWorkingGroupsResponse();
-      const { items: keywords } = gp2Fixtures.createKeywordsResponse();
+      const { items: keywords } = gp2Fixtures.createTagsResponse();
```
marianasrv, line 79: here

apps/gp2-frontend/src/users/__tests__/UserDirectory.test.tsx
```diff
@@ -69,7 +69,7 @@ const renderUserDirectory = async ({
   listUserResponse?: gp2Model.ListUserResponse;
   listProjectResponse?: ClientSearchResponse<'gp2', 'project'>;
   listWorkingGroupResponse?: gp2Model.ListWorkingGroupResponse;
-  listKeywordResponse?: gp2Model.ListKeywordsResponse;
+  listKeywordResponse?: gp2Model.ListTagsResponse;
```
marianasrv, line 72:
```suggestion
  listTagResponse?: gp2Model.ListTagsResponse;
```

apps/gp2-frontend/src/onboarding/__tests__/Background.test.tsx
```diff
@@ -97,9 +97,7 @@ describe('Background', () => {
   });
 
   it('saves the keywords modal', async () => {
```
marianasrv, line 99: I don't know if the modal will change to Tag modal, but if so, the test name should be updated as well

marianasrv (changes requested):

apps/gp2-frontend/src/shared/state.ts
gabiayako: About these changes in files that are not tests, I would rather to do it a future PR, because this change would cause a lot of changes in other parts of FE and this PR is already too big.

gabiayako: I've added a checkbox "More renames in FE not covered in this PR" in the description along with the things I intend to do in the next PR 😉 

 and I've added this commit https://github.com/yldio/asap-hub/pull/3745/commits/b0773ef5174bfb4282847cf0a26b328dc74dc505 to this PR with the changes that are only affected in is own file (that I don't need to make changes in a series of other files).

marianasrv (approved):

gabiayako:
I'm adding a script to delete all keywords because it is necessary before the next PR that will remove keywords content type

Screenshot 2023-10-05 at 06 05 03

PR #3747, opened 2023-10-04
GP2HUB-103 Update algolia handlers to inject tags in _tags

lctrt:


apps/gp2-server/src/handlers/event/algolia-index-handler.ts
```diff
@@ -27,8 +27,16 @@ export const indexEventHandler =
         const calendarEvent = await eventController.fetchById(id);
         log.debug(`Fetched calendar event ${calendarEvent.id}`);
 
+        const tags = calendarEvent.keywords?.map((tag) => tag.name);
```
marianasrv, line 30: Gabi is working on changing this to tags, so I think you should either wait for her to merge the [PR](https://github.com/yldio/asap-hub/pull/3745) or we will need a 2nd PR afterwards to fix this.

Same for the export script.

lctrt: Agreed, I'll wait for her PR to be merged :)

peterstarling (changes requested):
I really think you should consider removing the _tags property from the EventResponse type before doing all this

peterstarling (approved):

PR #3756, opened 2023-10-09
[GP2HUB-110] addes relatedEvents to schema

markcsims:


packages/model/src/gp2/output.ts
```diff
@@ -133,16 +135,26 @@ export type AuthorUpsertDataObject =
...
-export type OutputCreateDataObject = OutputCoreObject & {
+export type OutputCreateDataObject = Omit<
+  OutputCoreObject,
+  'relatedOutputs' | 'relatedEvents'
+> & {
   authors: AuthorUpsertDataObject[];
   createdBy: string;
   workingGroupId?: string;
   projectId?: string;
+  relatedOutputs: string[];
```
peterstarling, line 146: this is an array of strings, these strings being.. IDs, right? In which case I think you should follow our normaln convention and call them `relatedOutputIds` or `relatedEventIds`

peterstarling (changes requested):

peterstarling (approved):

PR #3769, opened 2023-10-10
GP2HUB-118 Create Output Detail Page

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/GP2HUB-118


packages/gp2-components/src/organisms/OutputDetailPageHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+  | 'link'
+  | 'mainEntity'
+  | 'projects'
+  | 'subtype'
+  | 'title'
+  | 'type'
+  | 'workingGroups'
+>;
+
+const OutputDetailPageHeader = ({
```
marianasrv, line 62: Why not using the OutputCard here? There are just a few differences. we would avoid repeated code.

packages/gp2-components/src/organisms/OutputCard.tsx
marianasrv, line 1: We should remove the edit button from the output card, shouldn't we? Now that we have it on the output detail page.

packages/gp2-components/src/molecules/AssociationList.tsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+
+  return (
+    <div>
+      {inline && <Indicator type={type} />}
+      <ul css={[containerStyles, inline && inlineContainerStyles]}>
+        {associations.map(({ id, title }, index) => (
+          <li key={id} css={[itemStyles, inline && inlineItemStyles]}>
+            {inline || <Indicator type={type} />}
```
marianasrv, line 113: i believe the order here matters (we may double check with Joao), but the main entity type (WG or Projects) should come first, just like in the output. check below, when it's a project (1st pill), projects come first, otherwise, groups come first. I did this in the output card.

Screenshot 2023-10-11 at 10 19 09
ist.

packages/gp2-components/src/organisms/OutputDetailPageHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+  mainEntity,
+  projects,
+  subtype,
+  title,
+  type,
+  workingGroups,
+}: OutputDetailPageHeaderProps) => (
+  <Card padding={false}>
+    <div css={containerStyles}>
+      <SharedResearchMetadata
```
marianasrv, line 78: I know this was done before, but I noticed that the pills are under <small> element, and the text seems to be smaller than the designs

FIGMA
Screenshot 2023-10-11 at 10 32 57


WITH SMALL (font-size 80%)
Screenshot 2023-10-11 at 10 32 22

WITHOUT SMALL (font-size 100%)
Screenshot 2023-10-11 at 10 33 38

marianasrv (changes requested):

packages/react-components/src/atoms/Pill.tsx
```diff
@@ -89,12 +89,7 @@ const Pill: React.FC<PillProps> = ({
   accent = 'default',
 }) => (
   <span
-    css={({ components }) => [
-      styles,
-      components?.Pill?.styles,
```
gabiayako, line 94: removing this `components?.Pill?.styles,`
fixes the issue below where the tags are over each other

this is before
dev gp2 asap science_outputs(Samsung Galaxy S20 Ultra)

this is after
localhost_4000_outputs_4elOfmpAGi24ce2XebHvAp(Samsung Galaxy S20 Ultra)


apparently it didn't affect CRN, I've ran it locally and they still seem good

marianasrv: That's great, but those styles were setting the correct font size and other props, but it had` margin: 0` which i believe was the root cause of that issue. so i think it would be nicer to just remove the margin:0 from Pill.styles in `asap-hub/packages/gp2-components/src/templates/components.tsx`

marianasrv (approved):

apps/gp2-frontend/src/outputs/ShareOutput.tsx
```diff
@@ -31,25 +41,33 @@ const ShareOutput: React.FC<Record<string, never>> = () => {
...
-        cohortSuggestions={cohortSuggestions}
-        workingGroupSuggestions={workingGroupSuggestions}
-        projectSuggestions={projectSuggestions}
-        mainEntityId={output.mainEntity.id}
-        projects={output.projects}
-        workingGroups={output.workingGroups}
-      />
-    </CreateOutputPage>
+    <Switch>
+      <Route exact path={path}>
```
peterstarling, line 45: are we sure we want to nest the detail page inside the ShareOutput component?

PR #3787, opened 2023-10-12
GP2HUB-131 - DashboardPageBody past event section

AkosuaA:
ref: https://asaphub.atlassian.net/jira/software/c/projects/GP2HUB/boards/12?modal=detail&selectedIssue=GP2HUB-131

This pr adds the 'Past Events' section to the Dashboard. We conditionally show the 'View All' button & allow navigation to the Event Detail page when the 'ASAP_DISPLAY_EVENTS' cookie is set.

NB: to set cookie - document.cookie = 'ASAP_DISPLAY_EVENTS=true';
       to unset cookie - cookieStore.delete('ASAP_DISPLAY_EVENTS');


packages/react-components/src/organisms/PastEventsDashboardCard.tsx
```diff
@@ -64,7 +69,14 @@ const PastEventsDashboardCard: React.FC<PastEventsDashboardCardProps> = ({
         ({ id, title, startDate, notes, presentation, videoRecording }) => (
           <div key={id} css={[rowStyles]}>
             <span css={[titleStyles, rowTitleStyles]}>Event</span>
-            <Link ellipsed href={eventsRoute({}).event({ eventId: id }).$}>
+            <Link
```
gabiayako, line 72: I wonder if it wouldn't be better to have something like

```
{linksEnabled ? (
  <Link ellipsed href={eventsRoute({}).event({ eventId: id }).$}>
    {title}
  </Link>
) : (
  <span>{title}</span>
)}
```

because with the `Link` the title is still blue and has an underline on hover so it gives the impression that can be clickable

AkosuaA: Yea, that sounds better. Updating...

gabiayako (approved):

PR #3790, opened 2023-10-12
[CRN-1518] feat: add auto populate app extension

Riccardo-Brown:
https://asaphub.atlassian.net/browse/CRN-1518

Creates auto-populating non-editable field extension app

### DIFF
--- a/.github/environment/Base
+++ b/.github/environment/Base
@@ -4,6 +4,7 @@ aws-region=us-east-1
 cache-prefix-definitions=definitions-ouput
 cache-prefix-frontend=frontend-output
 cache-prefix-transpile=transpile-output
+contentful-app-auto-populate-date-field=5k7jEBXizWrEHPIebdDTJN
 contentful-app-clear-on-change=6en82KYFrsxlAIPqypMdOp
 contentful-app-date-field-as-first-published-at=6AEbdOZc7KICbwW6KszNlp
 contentful-app-disabled-fields=2finDNk15g5UtOq4DaLNxv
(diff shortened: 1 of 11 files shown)


peterstarling (approved):

packages/contentful/migrations/crn/researchOutputs/20231012113418-change-created-date-to-auto-populate-field.js
```diff
@@ -0,0 +1,20 @@
+module.exports.description = 'Change createdDate field editor';
+
+module.exports.up = (migration) => {
+  const researchOutputs = migration.editContentType('researchOutputs');
+  researchOutputs.changeFieldControl(
```
gabiayako, line 5: The field still appears in the UI though and it seems it's a requirement of the ticket to hide it. Maybe we'll need to select this "Hide field when editing"

Screenshot 2023-10-17 at 06 31 15

It seems what it changes in the field is that it changes this

```
"disabled": true,
```

gabiayako (approved):
Approved, although reading the messages on slack, maybe this change to disabled true might be in another PR

packages/contentful/migrations/crn/researchOutputs/20231012113418-change-created-date-to-auto-populate-field.js
Riccardo-Brown: If you hide it while editing the field is not in the form, therefore it does not auto populate so the `createdDate ` field is set to null which results in the FE showing date as 01-01-1970

PR #3805, opened 2023-10-16
feat: Add redirect to RO detail page

peterstarling:


AkosuaA (approved):
LGTM

markcsims (approved):

marianasrv (changes requested):
You must change the `setBannerMessage` function to show the success notification on the output detail page instead of the outputs list page. Also, right now the output detail page doesn't have a notification context

packages/gp2-components/src/templates/OutputDetailPage.tsx
```diff
@@ -67,32 +68,43 @@ const OutputDetailPage: React.FC<OutputDetailPageProps> = ({
...
-        <div css={leftButtons}>
-          <Link
-            noMargin
-            href={
-              gp2Routing.outputs({}).output({ outputId: output.id }).edit({}).$
-            }
-            buttonStyle
-            small
-            primary
+  <PageNotifications page="outputs">
```
marianasrv, line 71: this page is not exactly `outputs`. for more clarity, maybe change it to "output" or "output-detail"?

marianasrv (changes requested):

marianasrv (approved):

PR #3815, opened 2023-10-16
create pr tenant for auth0

kmaid:
Uses a new PR environment auth0 when a label is added


.github/actions/setup-environment/action.yml
```diff
@@ -248,6 +251,23 @@ runs:
         if [ "$ENVIRONMENT_NAME" = "Branch" ]; then
           PR_NUMBER=${{ github.event.number || github.event.inputs.pr-number || inputs.pr-number }}
           HAS_CONTENTFUL_LABEL=${{ contains(github.event.pull_request.labels.*.name, 'contentful') }}
+          AUTH0_LABEL=${{contains(github.event.pull_request.labels.*.name, 'use-pr-auth0-tenant') }}
+          echo "crn-auth0-label=$AUTH0_LABEL" >> $GITHUB_OUTPUT
```
peterstarling, line 255: is this crn-only label? doesn't align with what is inside the condition below (looks like we set the gp2 client IDs of the PR tennant for when the label is on)

.github/workflows/reusable-deployment.yml
```diff
@@ -376,6 +382,20 @@ jobs:
         uses: ./.github/actions/cache-unplugged
         with:
           environment-name: ${{ inputs.environment-name }}
+
+      - name: Deploy CRN PR
```
peterstarling, line 386: I would be tempted to merge the two. For conditions it looks like we want to skip if it's a branch but the label isn't set. As for the config I think where it varies you could probably have them passed into this reusable workflow

apps/auth0/src/post-login-add-metadata.ts
```diff
@@ -92,6 +94,9 @@ export const onExecutePostLogin = async (
 ) => {
   try {
     const [apiUrl, redirect_uri] = getApiUrls(event);
+    console.log(
```
peterstarling, line 97: you might want to drop it

peterstarling (changes requested):

apps/auth0/src/post-login-add-metadata.ts
kmaid: Its logs on auth0 so if its comprimised its A only data we attach to the id token and B we have bigger problems

.github/workflows/reusable-deployment.yml
kmaid: The problem is that the PR CI secret is different and needs to be to support both flows. Its really a different case. I tried to make it one job initially.

.github/actions/setup-environment/action.yml
kmaid: The label will give you an auth0 PR env for both apps.

PR #3820, opened 2023-10-17
[GP2HUB-115] add tooltip to disabled radio buttons in output form

marianasrv:


packages/react-components/src/atoms/Label.tsx
```diff
@@ -16,15 +17,24 @@ interface LabelProps {
...
+  onLeave = noop,
 }) => {
   const contentId = useRef(uuidV4());
   return (
-    <div css={containerStyles}>
+    <div
+      css={containerStyles}
+      onMouseOver={onHover}
+      onMouseLeave={onLeave}
+      data-testid={`label-${contentId.current}`}
```
markcsims, line 36: We should avoid using data-testid, use only in a last resort.

But I can't see the test where it's been used.

markcsims (changes requested):

packages/react-components/src/atoms/Label.tsx
marianasrv: Yes, I understand. it is used here `packages/react-components/src/molecules/__tests__/LabeledRadioButton.test.tsx `. I added it here because the over functionality is in the label component to detect the over on the text and the input. 
if I looked for the text, it wouldn't be found because the over would not be detected

markcsims (approved):

markcsims (approved):

PR #3828, opened 2023-10-18
[GP2HUB-123] add additional information card to output detail page

marianasrv:


packages/gp2-components/src/templates/__tests__/OutputDetailPage.test.tsx
```diff
@@ -24,41 +24,6 @@ describe('OutputDetailPage', () => {
     expect(queryByTitle('Edit')).not.toBeInTheDocument();
   });
 
-  it('displays tags with source, documentType, type, subtype and source name', () => {
```
marianasrv, line 27: I deleted this test because this is already tested in Output card component

packages/gp2-components/src/organisms/OutputAdditionalInformationCard.tsx
AkosuaA, line 1: Noticed you created a new component rather than modifying the crn version. Was there an issue?

marianasrv: CRN has other fields - gp2 supported vs asap funded, and even the values are different in this case. also, for example they call `accession` and gp2 `accessionNumber` to the same field

AkosuaA (approved):

PR #3841, opened 2023-10-20
[GP2-129] Sync related outputs on the output detail page

markcsims:


apps/gp2-server/test/handlers/output/algolia-index-handler.test.ts
```diff
@@ -31,10 +32,43 @@ describe('Output index handler', () => {
       type: 'output',
     });
   });
+  test('Should fetch the output and create three records in Algolia when research-output is created and own and foreign related ROs are created', async () => {
```
peterstarling, line 35: i think you meant `two`?

apps/gp2-server/test/handlers/output/algolia-index-handler.test.ts
```diff
@@ -31,10 +32,43 @@ describe('Output index handler', () => {
...
+    ];
+
+    outputControllerMock.fetchById
+      .mockResolvedValueOnce(outputResponse)
+      .mockResolvedValueOnce(relatedOutputResponse);
+
+    await indexHandler(createEvent('ro-1234'));
+
+    expect(algoliaSearchClientMock.save).toHaveBeenCalledTimes(2);
+    expect(algoliaSearchClientMock.save).toHaveBeenNthCalledWith(1, {
```
peterstarling, line 58: I don't think we are that bothered about the order of calls are we?

peterstarling (changes requested):

peterstarling (approved):

PR #3851, opened 2023-10-24
Crn 1519 remove remaining references to squidex

oratiu:


apps/crn-server/src/config.ts
Riccardo-Brown, line 1: If these values are squidex base, should they not be removed?

apps/crn-server/test/fixtures/users.fixtures.ts
```diff
@@ -17,7 +17,7 @@ import { EventBridgeEvent } from 'aws-lambda';
 import { createEventBridgeEventMock } from '../helpers/events';
 
 export const fetchUserResponseDataObject = (): UserDataObject => ({
-  avatarUrl: `${baseUrl}/api/assets/${appName}/squidex-asset-id`,
+  avatarUrl: `${baseUrl}/api/assets/${appName}/contentful-asset-id`,
```
Riccardo-Brown, line 20: Does this test fixture need to use `baseUrl` and `appName`, why can it not just be fake? eg `https://test-avatar-asset-id`

gabiayako (commented):
This file https://github.com/yldio/asap-hub/blob/CRN-1519-remove-remaining-references-to-squidex/packages/server-common/src/validation/calendar-handler.squidex.validation.ts can also be removed

When we search for squidex in VSCode we can still find some references to squidex. Not sure if this is the last task related to removing squidex from the codebase. If it is, maybe we should rename the rest of things :)

Screenshot 2023-10-27 at 09 08 10

Riccardo-Brown (approved):

PR #3852, opened 2023-10-24
CRN-1002-skeleton-loading-states

kmaid:


packages/react-components/src/templates/LoadingLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }),
+  animation(rem(100), rem(18)),
+];
+
+const searchIconContainerStyles = css({
+  display: 'flex',
+  padding: rem(24),
+});
+
+export const LoadingUserButton: React.FC<Record<string, never>> = () => (
```
gabiayako, line 108: Shouldn't be 
```suggestion
export const LoadingUserButton: React.FC = () => (
```
since it doesn't receive any props?

The same thing for `LoadingMenu` and `LoadingLayout`

gabiayako (approved):

PR #3853, opened 2023-10-24
[GP2HUB-146] fix output filtering for working groups and projects

marianasrv:


packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json
```diff
@@ -9,14 +9,14 @@
...
-    "project.id",
+    "projects.id",
     "projectIds",
     "region",
     "speakers.user.id",
     "status",
     "type",
     "tagIds",
-    "workingGroup.id",
+    "workingGroups.id",
```
peterstarling, line 19: i think some elements (maybe events?) are still using a workingGroup property - are we not filtering them by this attribute anywhere? if we are you are going to need to have both of them here

marianasrv: actually we are not using WG to filter events

peterstarling (approved):

PR #3856, opened 2023-10-25
CRN-1545 Add users to tag search

gabiayako:
tags

This PR:
- adds `_tags` to user data in algolia handlers
- adds `tags/state` and `tags/api` files to query more than one entity ("research-output" | "user" | "external-author" | "event") in algolia using tags. So for the next entities we'll just need to update the data we sent to algolia (using `_tags`), update `entities` in [here](https://github.com/yldio/asap-hub/blob/CRN-1545-populate-user-tags-field/apps/crn-frontend/src/tags/Routes.tsx#L15) and add the card to `TagsPageBody`.

### DIFF
--- a/apps/crn-frontend/src/tags/Routes.tsx
+++ b/apps/crn-frontend/src/tags/Routes.tsx
@@ -1,4 +1,4 @@
-import { EntityResponses } from '@asap-hub/algolia';
+import { CRNTagEntitiesList } from '@asap-hub/algolia';
 import { Switch, Route, useRouteMatch } from 'react-router-dom';
 import { NotFoundPage, TagsPage } from '@asap-hub/react-components';
 import { Frame } from '@asap-hub/frontend-utils';
@@ -12,10 +12,7 @@ const Routes: React.FC<Record<string, never>> = () => {
   const { client } = useAlgolia();
   const { tags, setTags } = useSearch();
 
-  const entities: Array<keyof EntityResponses['crn']> = [
-    'research-output',
-    'user',
-  ];
+  const entities: CRNTagEntitiesList = ['research-output', 'user'];
 
   return (
     <Switch>
--- a/apps/crn-frontend/src/tags/Routes.tsx
+++ b/apps/crn-frontend/src/tags/Routes.tsx
@@ -1,3 +1,4 @@
+import { EntityResponses } from '@asap-hub/algolia';
 import { Switch, Route, useRouteMatch } from 'react-router-dom';
 import { NotFoundPage, TagsPage } from '@asap-hub/react-components';
 import { Frame } from '@asap-hub/frontend-utils';
@@ -11,6 +12,11 @@ const Routes: React.FC<Record<string, never>> = () => {
   const { client } = useAlgolia();
   const { tags, setTags } = useSearch();
 
+  const entities: Array<keyof EntityResponses['crn']> = [
+    'research-output',
+    'user',
+  ];
+
   return (
     <Switch>
       <Route exact path={path}>
@@ -19,7 +25,7 @@ const Routes: React.FC<Record<string, never>> = () => {
           setTags={setTags}
           loadTags={async (tagQuery) => {
             const searchedTags = await client.searchForTagValues(
-              ['research-output'],
+              entities,
               tagQuery,
               { tagFilters: tags },
             );
@@ -30,7 +36,7 @@ const Routes: React.FC<Record<string, never>> = () => {
           }}
         >
           <Frame title="Search">
-            <Tags />
+            <Tags entities={entities} />
           </Frame>
         </TagsPage>
       </Route>
--- a/apps/crn-frontend/src/tags/Routes.tsx
+++ b/apps/crn-frontend/src/tags/Routes.tsx
@@ -7,12 +7,12 @@ import Tags from './TagsList';
 import { useAlgolia } from '../hooks/algolia';
 import { useSearch } from '../hooks';
 
+export const entities: CRNTagSearchEntitiesList = ['research-output', 'user'];
+
 const Routes: React.FC<Record<string, never>> = () => {
   const { path } = useRouteMatch();
   const { client } = useAlgolia();
   const { tags, setTags } = useSearch();
-
-  const entities: CRNTagSearchEntitiesList = ['research-output', 'user'];
 
   return (
     <Switch>
--- a/apps/crn-frontend/src/tags/Routes.tsx
+++ b/apps/crn-frontend/src/tags/Routes.tsx
@@ -1,4 +1,4 @@
-import { CRNTagEntitiesList } from '@asap-hub/algolia';
+import { CRNTagSearchEntitiesList } from '@asap-hub/algolia';
 import { Switch, Route, useRouteMatch } from 'react-router-dom';
 import { NotFoundPage, TagsPage } from '@asap-hub/react-components';
 import { Frame } from '@asap-hub/frontend-utils';
@@ -12,7 +12,7 @@ const Routes: React.FC<Record<string, never>> = () => {
   const { client } = useAlgolia();
   const { tags, setTags } = useSearch();
 
-  const entities: CRNTagEntitiesList = ['research-output', 'user'];
+  const entities: CRNTagSearchEntitiesList = ['research-output', 'user'];
 
   return (
     <Switch>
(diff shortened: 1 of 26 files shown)


packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -56,11 +66,25 @@ const TagsPageBody: React.FC<TagsPageBodyProps> = ({
...
   >
-    {results.map((data) => (
-      <div key={data.id}>
-        <SharedResearchCard {...data} />
-      </div>
-    ))}
+    {results.map((data) => {
+      if ('documentType' in data) {
+        return (
+          <div key={data.id}>
```
kmaid, line 72: This could be outside

apps/crn-frontend/src/tags/TagsList.tsx
```diff
@@ -1,18 +1,23 @@
...
+const Tags: React.FC<TagsProps> = ({ entities }) => {
   const { tags, searchQuery, filters } = useSearch();
   const { currentPage, pageSize } = usePaginationParams();
-  const { items, total } = useResearchOutputs({
+  const { items, total } = useTagSearch(entities, {
     searchQuery,
     filters,
     currentPage,
     pageSize,
     noResultsWithoutCriteria: true,
```
kmaid, line 17: We probablly don't need this as a prop because it will be the behaviour for all tag searches

apps/crn-frontend/src/tags/__tests__/Routes.test.tsx
```diff
@@ -101,8 +103,9 @@ it('Will search algolia using selected tag', async () => {
   userEvent.click(screen.getByRole('textbox'));
   userEvent.click(screen.getByText('LGW'));
   await waitFor(() =>
-    expect(mockGetResearchOutputs).toHaveBeenCalledWith(
+    expect(mockGetTagSearch).toHaveBeenCalledWith(
       expect.anything(),
+      ['research-output', 'user'],
```
kmaid, line 108: This seems like it will be anoying to update for each entity type. Could we test it once and then set the rest to anything?

apps/crn-frontend/src/tags/Routes.tsx
```diff
@@ -11,6 +12,11 @@ const Routes: React.FC<Record<string, never>> = () => {
   const { client } = useAlgolia();
   const { tags, setTags } = useSearch();
 
+  const entities: Array<keyof EntityResponses['crn']> = [
```
kmaid, line 15: Maybe this should be exported from model or something? It doesn't feel like a prop unless it changes because of a filter i guess

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -56,11 +66,25 @@ const TagsPageBody: React.FC<TagsPageBodyProps> = ({
...
+            <SharedResearchCard {...data} />
+          </div>
+        );
+      }
+
+      if ('expertiseAndResourceTags' in data) {
```
kmaid, line 78: Should this be a type discriminator?

kmaid (changes requested):
See comments

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -56,11 +66,25 @@ const TagsPageBody: React.FC<TagsPageBodyProps> = ({
       />
     }
   >
-    {results.map((data) => (
-      <div key={data.id}>
-        <SharedResearchCard {...data} />
-      </div>
-    ))}
+    {results.map((data) => {
+      if ('documentType' in data) {
```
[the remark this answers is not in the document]
gabiayako: I'm not sure if I understood the suggestion. I thought about doing something like

```
const tagFieldByEntity = {
    'research-output': 'keywords',
    'user': 'expertiseAndResourceTags'
  }
  
  const EntityCard: React.FC<TagsPageBodyProps['results'][number]> = ({
    ...data
  }) => {
    if (tagFieldByEntity['research-output'] in data) {
      return <SharedResearchCard {...data} />;
    }
  
    if (tagFieldByEntity['user'] in data) {
      return <PeopleCard {...data} />;
    }
  
    return null;
  };
```

But then TS give me an error because it can't make the type narrowing properly as `tagFieldByEntity['research-output']` is recognized as a string not `keywords` exactly

Screenshot 2023-10-30 at 07 21 54

packages/react-components/src/templates/TagsPageBody.tsx
gabiayako: I think this is related to https://github.com/yldio/asap-hub/pull/3856#discussion_r1375980474 right? if so it was not done properly for similar reasons. But I'm happy to accept suggestion of how to do it properly if there's a nice way to make TS work :)

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -29,8 +34,27 @@ const MessageBody: React.FC<{ title: string; body: string }> = ({
...
+}) => {
+  if ('keywords' in data) {
+    return <SharedResearchCard {...data} />;
+  }
+
+  if ('expertiseAndResourceTags' in data) {
+    return <PeopleCard {...data} />;
+  }
+
+  return null;
```
kmaid, line 48: Could we type this against the type in the model and never returns null? What I am looking to do is make sure if a new entity is added to algolia then its either handled or excluded from the type and search response. Right now it silently fail and we would get pages of 10 showing less results which would be harder to pickup on.

apps/crn-frontend/src/tags/__tests__/TagsList.test.tsx
```diff
@@ -15,7 +15,7 @@ const renderTags = async (): Promise<RenderResult> =>
           <Suspense fallback="Loading...">
             <MemoryRouter initialEntries={['/']}>
               <Route path="/">
-                <TagList />
+                <TagList entities={['research-output', 'user']} />
```
kmaid, line 18: Could we export this from the file so the test cases always match the use

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -29,8 +41,44 @@ const MessageBody: React.FC<{ title: string; body: string }> = ({
...
+  if (TagFieldByEntity.event in data) {
+    return <EventCard {...eventMapper(data)} />;
+  }
+
+  return <ErrorCard />;
+};
+
 interface TagsPageBodyProps {
-  readonly results: ListResearchOutputResponse['items'];
+  readonly results: (
```
kmaid, line 77: Could we use the type that we exported from the model here?

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -29,8 +41,44 @@ const MessageBody: React.FC<{ title: string; body: string }> = ({
...
+
+  if (TagFieldByEntity.user in data) {
+    return <PeopleCard {...data} />;
+  }
+
+  if (TagFieldByEntity.event in data) {
+    return <EventCard {...eventMapper(data)} />;
+  }
+
+  return <ErrorCard />;
```
kmaid, line 73: Do we need this case? Shouldn't the types prevent this from happening?

kmaid (approved):

packages/react-components/src/templates/TagsPageBody.tsx
gabiayako: You're right! This isn't necessary, I've removed it here [review: remove unnecessary error card](https://github.com/yldio/asap-hub/pull/3856/commits/674e1b8e74b778d5051ef9e62dcb2f5be7bbe1d1)

apps/crn-frontend/src/tags/__tests__/TagsList.test.tsx
gabiayako: committed here [review: use same entities in the test as the ones used in the non-test files](https://github.com/yldio/asap-hub/pull/3856/commits/5dcaaeaffc9e9f8746237ea4755b2c853549ec6d)

packages/react-components/src/templates/TagsPageBody.tsx
gabiayako: committed here [review: use TagSearchResponse from model in TagsPageBody](https://github.com/yldio/asap-hub/pull/3856/commits/5991aaf54f35559f84380018178e877d07d7be3f)

PR #3861, opened 2023-10-26
[GP2HUB-112] add related outputs to events page

marianasrv:


packages/contentful/src/gp2/queries/event.queries.ts
```diff
@@ -2,6 +2,34 @@
...
+        ... on Projects {
+          sys {
+            id
+          }
+          title
+        }
+      }
+    }
+  }
+`;
```
gabiayako, line 31: I think we can import `relatedOutputQueryFragment` to this file instead of creating the same fragment 😉

packages/react-components/src/templates/EventPage.tsx
```diff
@@ -78,9 +100,15 @@ const EventPage: React.FC<EventPageProps> = ({
       {eventConversation}
       {relatedResearch && relatedResearch?.length > 0 && (
         <RelatedResearchCard
-          description="Find out all shared research outputs that are related to this event."
+          title={titleOutputs}
```
gabiayako, line 103: I believe this requirement from https://asaphub.atlassian.net/browse/GP2HUB-112

> Display 3 rows of events, if there are more than 3 rows of related outputs then display show more/show less functionality

is missing

gabiayako (changes requested):

gabiayako (approved):

PR #3863, opened 2023-10-26
fix: team-user handler error

gabiayako:


apps/crn-server/src/data-providers/contentful/user.data-provider.ts
```diff
@@ -139,11 +139,12 @@ export class UserContentfulDataProvider implements UserDataProvider {
...
-        teamMembershipCollection?.items?.map(
-          (item) => item?.linkedFrom?.usersCollection?.items[0],
-        ) || [];
+        teamMembershipCollection?.items
+          ?.map((item) => item?.linkedFrom?.usersCollection?.items[0])
+          .filter((item) => !!item) || [];
+
       return {
-        total: teamMembershipCollection?.total || 0,
+        total: users?.length || 0,
```
Riccardo-Brown, line 147: We don't need the optional chaining as users will always be an array because of `|| []` 
```suggestion
        total: users.length
```

Riccardo-Brown (approved):

PR #3867, opened 2023-10-31
[ASAP-46] Update all tags to link

kmaid:


kmaid:
Don't forget to remove that bypass thing you did on one of the RO cards

packages/react-components/src/molecules/TagList.tsx
marianasrv, line 1: There are components, such as this one, that are used for both apps. As the tag route is the same for both, I decided to use the crn route for all tags. The alternative would be adding a flag or having repeated code

lctrt (approved):

PR #3871, opened 2023-11-01
fix: PR indexes & api setup and teardown

Riccardo-Brown:
PR adds back setting up api and index based on `crn-create-environment` or `gp2-create-environment` label. That was added in https://github.com/yldio/asap-hub/pull/3862 and reverted

Also fixes the following issue

### Problem
PR that did not have dedicated environment were deleting `asap-hub_dev` and `gp2-hub_dev` indexes during PR clean up workflow

### Solution
only run the delete index step if the index is not dev or prod
only remove webhook if it is PR environment
only copy index if it is a PR environment

### DIFF
--- a/.github/workflows/on-remove.yml
+++ b/.github/workflows/on-remove.yml
@@ -83,6 +83,7 @@ jobs:
           CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
           CONTENTFUL_ENV_ID: ${{ steps.setup.outputs.crn-contentful-env-id }}
       - name: Contentful Webhook Teardown
+        if: ${{ steps.setup.outputs.crn-contentful-env-id != 'Development' }}
         id: contentful-webhook-teardown
         shell: bash
         run: yarn workspace @asap-hub/contentful space:remove-webhook
@@ -122,6 +123,7 @@ jobs:
           CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
           CONTENTFUL_ENV_ID: ${{ steps.setup.outputs.gp2-contentful-env-id }}
       - name: Contentful Webhook Teardown
+        if: ${{ steps.setup.outputs.gp2-contentful-env-id != 'Development' }}
         id: contentful-webhook-teardown
         shell: bash
         run: yarn workspace @asap-hub/contentful space:remove-webhook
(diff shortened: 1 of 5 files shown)


.github/workflows/on-remove.yml
```diff
@@ -163,6 +163,7 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-default-region: ${{ steps.setup.outputs.aws-default-region }}
       - name: Delete the index
+        if: ${{ steps.setup.outputs.crn-algolia-index != 'asap-hub_dev' }}
```
peterstarling, line 166: isn't this a little too specific, shouldn't we rely on the environment or label or something instead? like if we ever change the naming convention it would be good if this still worked as expected.

Plus - if you really are going to be excluding the dev index you might as well add prod in there too just to be safe

peterstarling (changes requested):

peterstarling (changes requested):
i think you should have a look at how we currently set up contentful environment in on-push-branch workflow - you will see that we already have a way to decide if a dedicated env is used there and i think it should be expanded for algolia. Your changes would suggest that contentful never work on branches but it did

peterstarling:
Check out how its currently done for Contentful:

https://github.com/yldio/asap-hub/blob/708ab7a97ba2267bc3e24ec19a455e8ffca1d627/.github/workflows/on-push-branch.yml#L31-L40

...is how we get the current state of the environment

https://github.com/yldio/asap-hub/blob/708ab7a97ba2267bc3e24ec19a455e8ffca1d627/.github/workflows/on-push-branch.yml#L96

... and that's how it is then passed on to create-environment workflow. You should follow this approach for algolia.

.github/workflows/on-remove.yml
gabiayako, line 1: Shouldn't we add the condition to run if it is a PR env on `Contentful Environment Teardown` in both `crn-contentful-remove` and `gp2-contentful-remove` so we don't delete `Development` environment?

.github/workflows/reusable-create-environment.yml
```diff
@@ -218,7 +218,7 @@ jobs:
           ALGOLIA_ADMIN_API_KEY: ${{ steps.algolia-keys.outputs.algolia-api-key }}
           ALGOLIA_INDEX: ${{ steps.setup.outputs.gp2-algolia-index }}
       - name: Copy index from Dev
-        if: ${{ inputs.environment-name=='Branch' && steps.index-exists.outputs.index-exists == 'false' }}
+        if: ${{ inputs.environment-name=='Branch' && steps.index-exists.outputs.index-exists == 'false' && inputs.crn-on-branch-env == 'true' }}
```
gabiayako, line 221:
```suggestion
        if: ${{ inputs.environment-name=='Branch' && steps.index-exists.outputs.index-exists == 'false' && inputs.gp2-on-branch-env == 'true' }}
```

.github/workflows/on-remove.yml
Riccardo-Brown: the condition is in the script https://github.com/yldio/asap-hub/blob/master/packages/contentful/scripts/teardown.sh

gabiayako: hmm ok, I thought it was deleted before when the PR got merged the first time, but I think I was wrong then :)

gabiayako (approved):

markcsims (approved):

peterstarling (approved):

PR #3876, opened 2023-11-02
CRN-1398 Filter research output by source

gabiayako:

### DIFF
--- a/packages/algolia/schema/crn/algolia-schema.json
+++ b/packages/algolia/schema/crn/algolia-schema.json
@@ -42,7 +42,8 @@
     "startDateTimestamp",
     "interestGroup.id",
     "workingGroup.id",
-    "workingGroups.id"
+    "workingGroups.id",
+    "publishingEntity"
   ],
   "numericAttributesToIndex": null,
   "attributesToRetrieve": null,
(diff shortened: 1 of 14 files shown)


packages/algolia/schema/crn/algolia-schema.json
```diff
@@ -63,7 +64,8 @@
     "interestGroup.id",
     "workingGroup.id",
     "workingGroups.id",
-    "_tags"
+    "_tags",
+    "source"
```
kmaid, line 68: Source? Dont we have meta data saying what type it is?

gabiayako: We don't have it. We check workingGroups to see if it is a working groups RO or a team RO https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/organisms/SharedResearchCard.tsx#L60

apps/crn-server/src/handlers/research-output/algolia-index-research-output-handler.ts
```diff
@@ -33,6 +33,9 @@ export const indexResearchOutputHandler =
 
         const data = {
           ...researchOutput,
+          source: researchOutput.workingGroups?.length
```
kmaid, line 36: This should belong in the modal if it isn't there already. We don't want to duplicate this logic in lots of places incase it changes

kmaid (approved):

PR #3879, opened 2023-11-03
GP2HUB 142 reject unexpected sync notifications and reduce concurrency

markcsims:
- divides calendar handlers into 2
- update project handlers for calendar
- adds event handlers to serverless

### DIFF
--- a/packages/server-common/src/handlers/webhooks/events-updated-process.handler.ts
+++ b/packages/server-common/src/handlers/webhooks/events-updated-process.handler.ts
@@ -1,0 +1,78 @@
+import { CalendarDataProvider, gp2 } from '@asap-hub/model';
+import { SQSEvent } from 'aws-lambda';
+import { Logger, SyncCalendar } from '../../utils';
+
+export const webhookEventUpdatedProcessHandlerFactory = (
+  calendarDataProvider: CalendarDataProvider | gp2.CalendarDataProvider,
+  syncCalendar: SyncCalendar,
+  logger: Logger,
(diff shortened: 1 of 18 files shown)


packages/server-common/src/handlers/webhooks/events-updated-process.handler.ts
```diff
@@ -0,0 +1,86 @@
...
+      const calendar = await getCalendar(resourceId);
+
+      const {
+        googleCalendarId,
+        id: cmsCalendarId,
+        syncToken,
+        channelId: currentChannelId,
+      } = calendar;
+
+      if (channelId !== currentChannelId) {
```
peterstarling, line 54: i would log a warning or a debug line here. I guess this is not really something undesired in a sense that we know that we don't unsubscribe old subscriptions so its expected, but since we have a live bug that is probably to do with multiple subscriptions i would like to know when this kicks in. So yeah, debug i guess

packages/server-common/test/handlers/webhooks/events-updated-process.handler.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      total: 0,
+    });
+    const event = getGoogleCalenderEventProcessPayload();
+
+    const { statusCode } = await handler(event);
+
+    expect(statusCode).toStrictEqual(500);
+  });
+
+  test('Should return 200 when the channel ids are different but not sync the calendar', async () => {
```
peterstarling, line 90: and log

peterstarling (changes requested):

peterstarling (approved):

PR #3885, opened 2023-11-06
CRN-1546 Add events to tag search

gabiayako:


apps/crn-server/src/handlers/event/helper.ts
```diff
@@ -0,0 +1,12 @@
+import { Payload } from '@asap-hub/algolia';
+import { EventResponse } from '@asap-hub/model';
+
+export const addTagsToEvents = <T extends Payload>(
```
Riccardo-Brown, line 4: Maybe we should turn this into a generic function for the remaining entities that have a `tags` field we are going index later?

Riccardo-Brown (approved):

PR #3892, opened 2023-11-08
Remove is contentful enabled references

gabiayako:
There's an unused `IS_CONTENTFUL_ENABLED` and `X-Contentful-Enabled` header left on the codebase. This PR removes both of them.

### DIFF
--- a/packages/frontend-utils/src/api-util/api-util.ts
+++ b/packages/frontend-utils/src/api-util/api-util.ts
@@ -1,5 +1,4 @@
 import { ErrorResponse, ValidationErrorResponse } from '@asap-hub/model';
-import { isEnabled } from '@asap-hub/flags';
 import { configureScope } from '@sentry/react';
 
 export type GetListOptions = {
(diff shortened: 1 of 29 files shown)


packages/frontend-utils/src/api-util/api-util.ts
```diff
@@ -40,10 +39,6 @@ export const createSentryHeaders = (): {
   };
 };
 
-export const createFeatureFlagHeaders = () => ({
```
Riccardo-Brown, line 43: Should we not keep this for future feature flags

peterstarling (approved):

packages/frontend-utils/src/api-util/api-util.ts
gabiayako: This has been added specifically for this case to switch between squidex and contentful https://github.com/yldio/asap-hub/pull/2770/files#diff-5a88dd4b45d008f2791dc6fa0772cde8df72c13c63df9adedd19fd5f3e949bbb. I think we could just remove this one and if we ever need another feature flag maybe the person could implement a new header or way to switch a feature. What do you think?

PR #3895, opened 2023-11-09
ASAP-39 Add teams to tag search

gabiayako:


apps/crn-server/test/handlers/teams/algolia-index-team-users-handler.test.ts
peterstarling, line 1: i guess this is fine although im not a huge fan of mapping expectations, I would prefer to have a fixture for that maybe modify it a bit when needed, having to do some funky mapping for the expectation could technically mean that you should also test your test :D

peterstarling: but i know you only copied that pattern so

peterstarling (approved):

PR #3903, opened 2023-11-14
ASAP-45 Add working groups to tag search

gabiayako:


apps/crn-server/serverless.ts
```diff
@@ -757,6 +757,26 @@ const serverlessConfig: AWS = {
...
+    algoliaIndexWorkingGroups: {
+      handler:
+        './src/handlers/working-group/algolia-index-working-group-handler.handler',
+      events: [
+        {
+          eventBridge: {
+            eventBus: 'asap-events-${self:provider.stage}',
+            pattern: {
+              source: [eventBusSourceContentful],
+              'detail-type': ['WorkingGroupsPublished'],
```
marianasrv, line 769: Why only on publish?

apps/crn-server/src/handlers/working-group/algolia-index-working-group-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+  (
+    workingGroupController: WorkingGroupController,
+    algoliaClient: AlgoliaClient<'crn'>,
+  ): EventBridgeHandler<WorkingGroupEvent, WorkingGroupPayload> =>
+  async (event) => {
+    logger.debug(`Event ${event['detail-type']}`);
+    const workingGroupId = event.detail.resourceId;
+
+    try {
+      const workingGroup = await workingGroupController.fetchById(
```
Riccardo-Brown, line 28: Should we not use the dataprovider directly not via the controller?

Riccardo-Brown (changes requested):
I think we should be using the detail-type `WorkingGroupsPublished` or `WorkingGroupsUnpublished` to determine if the handler should add or remove data from the index, instead of using the result of fetching the record

gabiayako:
@Riccardo-Brown I've followed the pattern in the other webhooks fetching the content using the controller instead of the data provider 

- https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/event/algolia-index-event-handler.ts#L24 
- https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/external-author/algolia-index-external-author-handler.ts#L22
- https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/handlers/user/algolia-index-user-handler.ts#L26

About using `WorkingGroupsPublished` or `WorkingGroupsUnpublished` as we're already fetching the record to get the parsed content we will save on algolia, I believe it's consistent to check it the content exists or not to determine if we will save it or remove it. If we add `[Entity]Delete` this would work as well. Also it's consistent with the existing handlers.

marianasrv (approved):

Riccardo-Brown (approved):

PR #3904, opened 2023-11-14
calendar subscriber handler uses sdk

markcsims:


apps/crn-server/src/utils/get-calendar-subscription-id.ts
```diff
@@ -1,4 +1,4 @@
 import { contentfulEnvId } from '../config';
 
-export const getCalendarSubscriptionId = (id: string): string =>
+export const getCalendarSubscriptionId = (id: string) =>
```
peterstarling, line 3: i never understood why people prefer typescript to infer the return type. 

I start writing my function declaration from defining the name and input/output types, so when I am done I know I got just what I wanted. When you remove this then sure, you still see the definition in your IDE etc, but say someone comes in and changes the implementation of the function - you may never even spot the issue that might arise from it. It might pop up as an error where the function is used but what if your function was returning an array before and it was changed to return a string instead? if we were to take the output of this function and use the first element of the array previously - now that will still work albeit with completely different results. 

Best to keep the return types imho

peterstarling (approved):

apps/crn-server/src/utils/get-calendar-subscription-id.ts
markcsims: You can infer the return types in most cases. It speeds up development, less code to maintain and the real value of what you are returning in the code - the source of truth.

In C# the best practise was not to give explicit data types to variables when the type can be inferred. They had a var datatype to help.

You'll probably want to be explicit on a return type when you are returning something complex. 

Here we are returning a string.

PR #3907, opened 2023-11-15
[ASAP-169] remove used CMS env from integration tests API key

marianasrv:


packages/contentful/scripts/remove-environment-from-key.ts
```diff
@@ -0,0 +1,37 @@
...
+const app = async () => {
+  const space = await client.getSpace(spaceId);
+
+  const keys = await space.getApiKeys();
+  const apiKey = keys.items.find(
+    (key) => key.accessToken === contentfulAccessToken,
+  );
+
+  if (!apiKey) {
+    throw new Error('Could not find API key');
```
peterstarling, line 22: maybe there is no point throwing here? just log and exit

peterstarling (changes requested):

peterstarling (approved):

PR #3908, opened 2023-11-16
(refactor): adapt webhook types to contentful

gabiayako:
As discussed in this [thread](https://yld.slack.com/archives/G01A6J33G8H/p1700045781352039), currently we're only triggering `Publish` and `Unpublish` events on Contentful.

Screenshot 2023-11-15 at 07 54 27

So the `EntityEventAction` type was changed from

```
export type EntityEventAction =
  | 'Created'
  | 'Published'
  | 'Updated'
  | 'Unpublished'
  | 'Deleted';
```

to 

```
export type EntityEventAction = 'Published' | 'Unpublished';
```

and it was added a type with all possible actions from the print above

```
type ContentfulActions =
  | 'publish'
  | 'unpublish'
  | 'create'
  | 'save'
  | 'autosave'
  | 'archive'
  | 'unarchive'
  | 'delete'
  | 'complete';
```

even though we're not processing all of them.

### DIFF
--- a/packages/server-common/src/handlers/webhooks/contentful.handler.ts
+++ b/packages/server-common/src/handlers/webhooks/contentful.handler.ts
@@ -6,11 +6,22 @@ import { APIGatewayProxyResult } from 'aws-lambda';
 import { Logger } from '../../utils';
 import { validateContentfulRequest } from '../../utils/validate-contentful-request';
 
+type ContentfulActions =
+  | 'publish'
+  | 'unpublish'
+  | 'create'
+  | 'save'
+  | 'autosave'
+  | 'archive'
+  | 'unarchive'
+  | 'delete'
+  | 'complete';
+
 const getActionFromRequest = (
   request: lambda.Request<ContentfulWebhookPayload>,
-): 'publish' | 'unpublish' => {
+): ContentfulActions => {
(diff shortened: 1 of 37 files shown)


packages/server-common/src/handlers/webhooks/contentful.handler.ts
```diff
@@ -5,11 +5,22 @@ import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
...
 const getActionFromRequest = (
   request: lambda.Request<ContentfulWebhookPayload>,
-): 'publish' | 'unpublish' => {
+): ContentfulActions => {
   const actions = request.headers['x-contentful-topic']?.split('.') ?? [];
-  return actions[actions.length - 1] as 'publish' | 'unpublish';
+  return actions[actions.length - 1] as ContentfulActions;
 };
 
 const getDetailTypeFromRequest = (
```
peterstarling, line 26: because of the cast in the last line (and generally how its written) we could still associate any type of event in contentful here and I don't see actions other than publish and unpublish being rejected as unsupported which I think they should be. You might want to play around with this function implementation to see if there is a way to do it without casting so that you are forced to say something like if actions i not publish or unpublish then throw

peterstarling (changes requested):

peterstarling (approved):

PR #3909, opened 2023-11-16
[GP2] new version route

lctrt:


apps/gp2-frontend/src/outputs/OutputDetail.tsx
```diff
@@ -36,12 +38,15 @@ const OutputDetail: FC = () => {
...
     return <NotFoundPage />;
   }
-
   return (
     <Switch>
       <Route exact path={path}>
         <Frame title="Output">
-          <OutputDetailPage isAdministrator={isAdministrator} {...output} />
+          <OutputDetailPage
+            canVersion={isEnabled('ASAP_VERSION_RESEARCH_OUTPUT')}
```
gabiayako, line 46: I believe there is a problem here, the button still appears even when the feature flag is false
Screenshot 2023-11-20 at 08 42 00

I believe it's not working properly because of this line
https://github.com/yldio/asap-hub/blob/master/packages/react-context/src/flags.tsx#L22

we remove `ASAP_` from the feature flag name, so maybe if you change to 

```suggestion
            canVersion={isEnabled('VERSION_RESEARCH_OUTPUT')}
```

it will work

packages/gp2-components/src/templates/__tests__/OutputDetailPage.test.tsx
```diff
@@ -26,6 +26,28 @@ describe('OutputDetailPage', () => {
...
+      <OutputDetailPage
+        isAdministrator
+        canVersion
+        {...gp2Fixtures.createOutputResponse()}
+      />,
+    );
+    expect(queryByTitle('Version Icon')).toBeInTheDocument();
+  });
+
+  it('ddoes not isplays version button if canVersion is false', () => {
```
gabiayako, line 40:
```suggestion
  it('does not display version button if canVersion is false', () => {
```

gabiayako (changes requested):

gabiayako (approved):
Nice! 🎉

PR #3917, opened 2023-11-17
ASAP-62 Reduce data from external-user

gabiayako:
This PR:
- Removes some unused data from `external-authors.queries`
- Removes `orcid` from external author event speaker as this is never used by the frontend
- It was not possible to remove `orcid` for now from research output external author as it is being used here https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/shared-research/export.ts#L46


AkosuaA (approved):
Looks good

PR #3929, opened 2023-11-21
ASAP-62 Reduce teams data

gabiayako:


apps/crn-server/src/data-providers/contentful/team.data-provider.ts
```diff
@@ -1,11 +1,12 @@
 import {
   LabResponse,
-  ListTeamDataObject,
   TeamCreateDataObject,
   TeamDataObject,
   TeamTool,
   TeamUpdateDataObject,
   TeamMember,
+  TeamReducedDataObject,
+  ListTeamReducedDataObject,
```
kmaid, line 9: Why do we have a list team reduced response? Surely this is now the response for all list endpoints?

gabiayako: yeah, you're right! I've made the change here [rename ListReducedTeamResponse -> ListTeamResponse](https://github.com/yldio/asap-hub/pull/3929/commits/a2d7da1fbe458a850c637a5f9648b879812eb4db)

apps/crn-server/src/routes/team.route.ts
```diff
@@ -25,10 +25,7 @@ export const teamRouteFactory = (
 
     const options = validateFetchOptions(query);
 
-    const result = await teamController.fetch({
-      ...options,
-      showTeamTools: req.loggedInUser?.teams.map((team) => team.id),
```
kmaid, line 30: Slightly concerned we could be breaking team tools.

apps/crn-server/test/fixtures/teams.fixtures.ts
```diff
@@ -139,7 +172,19 @@ export const getTeamDataObject = (): TeamDataObject => ({
   tools: [],
 });
 
+export const getTeamReducedDataObject = (): TeamReducedDataObject => ({
```
kmaid, line 175: I don't like the name reduced. I think its better to be specific that this is the list response data type. Reduced could be for anything.

apps/crn-server/src/handlers/teams/algolia-index-team-handler.ts
```diff
@@ -28,7 +28,10 @@ export const indexTeamHandler =
 
       if (team) {
         await algoliaClient.save({
-          data: addTagsFunction(team) as TeamDataObject,
+          data: {
+            ...addTagsFunction(team),
+            members: team.members.length,
+          } as TeamReducedDataObject,
```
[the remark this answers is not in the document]
gabiayako: I've ended up renaming it to `TeamsDataObject` ([rename: TeamReduce -> Teams](https://github.com/yldio/asap-hub/pull/3929/commits/a160c8ed57d2e1a73014ec843f55a7e23cbc102b)) because we already have `ListTeamDataObject` so this

```
export type ListTeamDataObject = ListResponse<TeamListDataObject>;
```

would be confusing

apps/crn-server/test/fixtures/teams.fixtures.ts
gabiayako: ref: https://github.com/yldio/asap-hub/pull/3929#discussion_r1408057063

apps/crn-server/src/routes/team.route.ts
gabiayako: It is still working in https://3929.hub.asap.science/ :)

For a team I'm part of -> tools appear
Screenshot 2023-11-28 at 13 30 08 (2)

For a team I'm not part of -> tools not appearing
Screenshot 2023-11-28 at 13 30 22 (2)

apps/crn-server/src/controllers/team.controller.ts
```diff
@@ -46,16 +46,7 @@ export default class TeamController {
 
     return {
       total,
-      items: items.map((team) => {
```
gabiayako, line 49: For other reviewers that may come here, please look at this https://github.com/yldio/asap-hub/pull/3929#discussion_r1408059312

I've removed the `showTeamTools` only for `fetch` method, the `fetchById` remains the same, so the tools feature is still working

apps/crn-server/src/handlers/teams/algolia-index-team-handler.ts
```diff
@@ -28,7 +28,10 @@ export const indexTeamHandler =
 
       if (team) {
         await algoliaClient.save({
-          data: addTagsFunction(team) as TeamDataObject,
+          data: {
+            ...addTagsFunction(team),
```
Riccardo-Brown, line 32: Do we not need to save more data to algolia like the team name?

gabiayako: actually we're saving more data than we need here 😅 these addTagsFunction returns all data + tags https://github.com/yldio/asap-hub/blob/861d4637a47fe17cc8341c8b94fdbdcbae6bd3ca/apps/crn-server/src/handlers/helper.ts#L7

Riccardo-Brown: We might want to refactor this function to just return tags

gabiayako: I've ended up removing this `addTagsFunction` from here, as we were sending more data than we need anyway. 
Screenshot 2023-11-29 at 10 53 08

Maybe down the road we can remove `addTagsFunction` everywhere. But now I don't think it makes sense to refactor this function to just return `_tags` as it is being used in several places to return data + _tags

apps/crn-server/src/controllers/team.controller.ts
peterstarling: So functionality may remain the same but are you sure we are not surfacing the tools for each time in the response now? As we are not meant to do that either

apps/crn-server/src/data-providers/contentful/team.data-provider.ts
```diff
@@ -172,7 +186,59 @@ export class TeamContentfulDataProvider implements TeamDataProvider {
...
+      const memberLabIds = labsCollection?.items
+        .map((labItem) => labItem?.sys.id)
+        .filter((x): x is string => x !== undefined);
+
+      memberLabIds?.forEach((labId) => labIdsSet.add(labId));
+
+      return [memberCount + 1, labIdsSet];
+    },
+    [0, new Set() as Set<string>],
+  );
```
peterstarling, line 226: this seems a little complicated, in a sense that something similar could probably be achieved with some flatMap and length, even if you were to do it twice (the optimisation seems unnecessary anyway, even if we are processing this for each team). Leave it but in the future I would start off simple and then optimise like this when really needed.

packages/model/src/team.ts
```diff
@@ -65,14 +65,22 @@ export type TeamCreateDataObject = {
...
+
+export type TeamListItemDataObject = Pick<
+  TeamDataObject,
+  | 'id'
+  | 'displayName'
+  | 'inactiveSince'
+  | 'projectTitle'
+  | 'expertiseAndResourceTags'
+  | 'labCount'
+> & { members: number };
```
peterstarling, line 80: `members` implies a list of members (which it was previously), not a count. If we do not need a list that's fine but rename the property to, say, `memberCount` or something like that?

apps/crn-server/test/routes/team.route.test.ts
```diff
@@ -150,50 +150,7 @@ describe('/teams/ route', () => {
...
-      const expectedParams: FetchTeamsOptions = {
-        ...(params as Required<FetchOptions>),
-        showTeamTools: [loggedUser.teams[0]!.id],
-      };
-
-      expect(teamControllerMock.fetch).toBeCalledWith(expectedParams);
-    });
-
-    describe('Team tools', () => {
-      test('Should select the team tools for the teams the user is a member of', async () => {
```
peterstarling, line 162: we are not showing team tools for each team now, right, the tools are just no longer part of the list response?

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -251,97 +253,291 @@ describe('Teams data provider', () => {
...
           },
         });
 
-        const result = await teamDataProvider.fetch({
-          take: 8,
-          skip: 0,
+        const result = await teamDataProvider.fetch({});
+
+        expect(result).toEqual({
+          items: [expect.objectContaining({ members: 2 })],
```
peterstarling, line 362: i wonder if maybe you should not just rely on the count here but actual items returned. What if someone comes in, changes the implementation so something else gets filtered out and the count still matches so its still all green but a different team is actually returned - thats not what you want right?

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
```diff
@@ -251,97 +253,291 @@ describe('Teams data provider', () => {
...
+                  linkedFrom: {
+                    usersCollection: {
+                      total: 1,
+                      items: [
+                        {
+                          ...getContentfulGraphqlTeamMembers(),
+                          labsCollection: {
+                            items: [
+                              { sys: { id: 'lab-1' } },
+                              { sys: { id: 'lab-2' } },
```
peterstarling, line 498: I'd use a `lab-4` and a count of 4, that way you also test that the count does not just come from a single team membership

apps/crn-server/src/routes/team.route.ts
peterstarling: yeah i think we never really needed tools to be present in the list anyway so now that it was dropped all of this functionality was moved to fetch-by-id endpoint and it just works

apps/crn-server/src/handlers/teams/algolia-index-team-handler.ts
```diff
@@ -27,8 +26,25 @@ export const indexTeamHandler =
...
       if (team) {
+        const {
+          id,
+          displayName,
+          inactiveSince,
+          projectTitle,
+          labCount,
+          expertiseAndResourceTags,
+        } = team;
         await algoliaClient.save({
```
peterstarling, line 37: i am not a big fan of the mapping above (or plucking properties really), it feels like we are constructing a new object just for algolia while previously it would have been the exact `team` object that comes out of the `fetchById` method (+ tags). The reason I am leaving my comment here is just for you to make sure this is type safe in a sense that - can you make sure that we are unable to save any extra properties or different properties. This still leaves more room for errors as people might come in and remap properties and its gonna be a mess but I would hope the types would at least give us some safety here

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/team.data-provider.ts
```diff
@@ -205,33 +271,28 @@ export const parseContentfulGraphQlTeams = (item: TeamItem): TeamDataObject => {
...
         labsCollection,
         alumniSinceDate,
       } = membership.linkedFrom?.usersCollection?.items[0] || {};
-      const id = sys?.id;
-
-      if (!id) {
-        return userList;
-      }
-
-      if (!role || !isTeamRole(role)) {
```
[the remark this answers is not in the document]
gabiayako: because of this discussion https://yld.slack.com/archives/G01A6J33G8H/p1701088240393909

Screenshot 2023-11-30 at 07 30 10

in Contentful the team role is a dropdown so it is really impossible to select an invalid role and I thought it wasn't worth it the time to discuss if we could keep it or not 😅

apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts
gabiayako: But this is the `fetch` method test, so it does return only the number, not the items. The `fetchBy` method test still checks the items returned https://github.com/yldio/asap-hub/blob/ASAP-62-reduce-data-teams/apps/crn-server/test/data-providers/contentful/teams.data-provider.test.ts#L694:L696

apps/crn-frontend/src/network/teams/api.ts
```diff
@@ -39,7 +39,7 @@ export const getTeam = async (
 export const getTeams = async (
   options: GetListOptions,
   authorization: string,
-): Promise<ListTeamResponse> => {
+): Promise<ListTeamDataObject> => {
```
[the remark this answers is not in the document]
gabiayako: committed here [review: replace ListTeamDataObject by ListTeamResponse in FE](https://github.com/yldio/asap-hub/pull/3929/commits/51ac684e3c07e94b3b2fe3a9680f135f417080e7)

apps/crn-server/src/handlers/teams/algolia-index-team-handler.ts
gabiayako: I've added this `WithAlgoliaTag` here [review: guarantee team data type sent to algolia is correct](https://github.com/yldio/asap-hub/pull/3929/commits/951e3eb0412a623510ef662019b30880e23a9325) and removed type casting

So now if I add a property that is not required by team card (`TeamListItem`) I get a Typescript error

Screenshot 2023-11-30 at 10 50 22


and if I remove a property required by TeamListItem + _tags I get a Typescript error as well

Screenshot 2023-11-30 at 10 51 12


I can add this `WithAlgoliaTag` to the rest of entities in `SavePayload` but I think it's better to do it in a separate PR

peterstarling (approved):

gabiayako:
rebasing to master before merging it

PR #3939, opened 2023-11-22
ASAP-219-add-filter-to-tags-page-and-update-tag-icon

kmaid:


gabiayako (changes requested):
Looks good! But the requirement to change the copy and the icon (empty view) is missing

pr-copy

packages/react-components/src/templates/TagsPageHeader.tsx
```diff
@@ -22,44 +23,71 @@ const textStyles = css({
...
+  setTags,
+  filters,
+  filterOptions,
+  onChangeFilter,
   loadTags = noop,
 }) => (
   <header>
     <div css={visualHeaderStyles}>
       <Display styleAsHeading={2}>Tags Search</Display>
       <div css={textStyles}>
```
gabiayako, line 59: this `maxWidth` from `textStyles` is making it look a little bit weird in the desktop view

Screenshot 2023-11-28 at 09 23 07

packages/react-components/src/templates/TagsPageHeader.tsx
```diff
@@ -22,44 +23,71 @@ const textStyles = css({
...
         <Paragraph accent="lead">
-          Search for research outputs that include selected tags.
+          Search for all CRN Hub areas that include selected tags (research
+          outputs, events, people, teams, interest groups, working groups and
```
gabiayako, line 62: I know it's not in the copy but `tutorials` is missing here

gabiayako: there's also a dot missing at the end

PR #3941, opened 2023-11-23
feat: ASAP-211 Convert auth-frontend to Vite

peterstarling:
- basic setup
- add dist to gitignore
- use dist output to upload to auth0 and s3
- add preset to auth-frontend to support import.meta


markcsims (approved):

kmaid:
image
I wanted to look at the build output but this seems to have broken the artifacts created

apps/crn-auth-frontend/index.html
```diff
@@ -48,5 +39,6 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script type="module" src="/src/index.tsx"></script>
```
kmaid, line 42:
```suggestion
    <script type="module" src="/src/index.tsx" />
```
Seems odd that its a tsx file and that this was added without removing another

apps/crn-auth-frontend/package.json
```diff
@@ -5,24 +5,27 @@
...
     "@babel/runtime-corejs3": "7.23.4",
+    "@vitejs/plugin-react": "^4.2.0",
     "history": "4.10.1",
     "path": "npm:path-browserify@1.0.1",
     "react": "17.0.2",
     "react-app-polyfill": "3.0.0",
     "react-dom": "17.0.2",
-    "react-router-dom": "5.3.4"
+    "react-router-dom": "5.3.4",
+    "vite": "^5.0.2"
```
kmaid, line 28: Should these be dependencies or dev dependencies?

apps/gp2-auth-frontend/index.html
```diff
@@ -24,13 +24,13 @@
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+    <link rel="manifest" href="/manifest.json" />
     <!--
-      Notice the use of %PUBLIC_URL% in the tags above.
+      Notice the use of  in the tags above.
```
kmaid, line 29: Should these comments be removed like CRN

apps/gp2-auth-frontend/package.json
```diff
@@ -42,10 +44,11 @@
     "@types/react-router-dom": "5.3.3",
     "@types/testing-library__jest-dom": "5.14.9",
     "@types/testing-library__react": "10.0.1",
+    "babel-preset-vite": "^1.1.0",
```
kmaid, line 47: Same as CRN

apps/gp2-auth-frontend/vite.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+
+export default defineConfig({
+  base:
+    process.env.GP2_AUTH_FRONTEND_BASE_URL || 'https://dev.hub.asap.science',
```
kmaid, line 6: Probably shouldn't be crn

apps/crn-auth-frontend/index.html
peterstarling: So, yeah, with regards to pointing the script tag to a tsx file - that's just how Vite works - check out the output files (now in a fixed artifact) to see what it gets converted to. More on this [here](https://vitejs.dev/guide/#index-html-and-project-root). The way Vite works in dev is that they serve transpiled code from memory while replacing the path to a standard .js file for production builds. 

As for not replacing a different script - react scripts append the script tag to the head tag afak

apps/crn-auth-frontend/package.json
peterstarling: regular dependencies according to docs

apps/crn-auth-frontend/index.html
kmaid: Shouldn't this be in the head with the "defer" attribute for best performance (https://gomakethings.com/when-should-you-add-the-defer-attribute-to-the-script-element/#so-when-should-you-use-defer)

kmaid:
image
This error also seems due to your changes

peterstarling:
> image This error also seems due to your changes

its not 
image

PR #3959, opened 2023-11-27
[ASAP-247] refactor(TagsPageBody): card rendering

Riccardo-Brown:
https://asaphub.atlassian.net/browse/ASAP-247

The conditional logic we use to determine which card to render will not work as we reduce the fields we index in algolia
https://asaphub.atlassian.net/browse/ASAP-62

### DIFF
--- a/packages/react-components/src/templates/TagsPageBody.tsx
+++ b/packages/react-components/src/templates/TagsPageBody.tsx
@@ -1,5 +1,12 @@
-import { CRNTagSearchEntities } from '@asap-hub/algolia';
-import { TagSearchResponse } from '@asap-hub/model';
+import {
+  TagSearchResponse,
+  ResearchOutputResponse,
+  UserResponse,
+  EventResponse,
+  WorkingGroupResponse,
+  TutorialsResponse,
+  TeamResponse,
+} from '@asap-hub/model';
 import { css } from '@emotion/react';
 import { Headline3, Paragraph } from '../atoms';
 import { charcoal } from '../colors';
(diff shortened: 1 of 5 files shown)


packages/react-components/src/templates/__tests__/TagsPageBody.test.tsx
```diff
@@ -34,20 +35,41 @@ it('renders a list of cards', () => {
...
+            ...createWorkingGroupResponse(),
+            __meta: { type: 'working-group' },
+            title: 'Comparative Neuroanatomy Working Group',
+          },
+          {
+            ...createTutorialsResponse({ key: '1' }),
+            __meta: { type: 'tutorial' },
+            title: 'Tutorial 1',
+          },
+        ] as unknown as TagSearchResponse[]
```
gabiayako, line 71: with the change above in the model I believe we won't need this `unknown`

packages/react-components/src/templates/TagsPageBody.tsx
```diff
@@ -1,5 +1,13 @@
-import { CRNTagSearchEntities } from '@asap-hub/algolia';
-import { TagSearchResponse } from '@asap-hub/model';
+/* eslint-disable no-underscore-dangle */
```
gabiayako, line 1: I don't think this `/* eslint-disable no-underscore-dangle */` is necessary, is it?

gabiayako (changes requested):

packages/react-components/src/templates/TagsPageBody.tsx
Riccardo-Brown: https://github.com/yldio/asap-hub/actions/runs/7018626145/job/19094459725?pr=3959

gabiayako (approved):

PR #3969, opened 2023-11-28
[ASAP-43] add news to algolia and tag search page

marianasrv:


apps/crn-server/src/handlers/news/algolia-index-news-handler.ts
```diff
@@ -0,0 +1,79 @@
...
+export const indexNewsHandler =
+  (
+    newsController: NewsController,
+    algoliaClient: AlgoliaClient<'crn'>,
+  ): EventBridgeHandler<NewsEvent, NewsPayload> =>
+  async (event) => {
+    const eventType = event['detail-type'] as Extract<
+      NewsEvent,
+      'NewsPublished' | 'NewsUnpublished'
+    >;
```
gabiayako, line 24: Removing this 

```
as Extract<
      NewsEvent,
      'NewsPublished' | 'NewsUnpublished'
    >;
```
it seems that Typescript still understands that `eventType` should be `'NewsPublished' | 'NewsUnpublished'`

Screenshot 2023-11-29 at 09 30 24

So I think we can remove it :)

packages/contentful/src/crn/types/webhook.ts
```diff
@@ -9,7 +9,8 @@ export type ContentfulWebhookPayloadType =
   | 'teams'
   | 'users'
   | 'workingGroups'
-  | 'tutorials';
+  | 'tutorials'
+  | 'news';
```
gabiayako, line 13: we already have `news` here :)
Screenshot 2023-11-29 at 09 35 05

gabiayako (approved):
nice! 🎉

PR #3974, opened 2023-11-29
[ASAP-175] - add skeleton loading state (header)

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-175

### DIFF
--- a/packages/react-components/src/templates/LoadingLayout.tsx
+++ b/packages/react-components/src/templates/LoadingLayout.tsx
@@ -6,7 +6,7 @@ import {
   menuButtonStyles,
   styles as menuButtonContainer,
 } from '../organisms/MenuHeader';
-import { rem, perRem } from '../pixels';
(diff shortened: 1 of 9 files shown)


packages/react-components/src/templates/LoadingLayout.tsx
```diff
@@ -60,6 +61,20 @@ const userButtonContainerStyles = css({
   columnGap: rem(16),
   paddingRight: rem(24),
 });
+
+const contentHeaderContainerStyles = css({
+  display: `grid`,
+  width: '100%',
+  boxSizing: 'border-box',
+  gridColumn: '1 / -1',
+  padding: `${36 / perRem}em ${contentSidePaddingWithNavigation(8)} ${
```
marianasrv, line 70: why don't we use `rem(36)` and `rem(48)` here?

marianasrv (approved):

PR #4001, opened 2023-12-05
ASAP-254 Calendar not found

gabiayako:
#### Problem

We have a handler that resubscribes the calendar expiring in 24h and it subscribes again afterwards. The problem is that when we resubscribe we update the calendar with resourceId: null. This triggers the subscribe webhook with the CalendarsPublished event. And in the resubscribe code we try to subscribe as well.

So we have 2 subscribes and 2 events sync happening more or less at the same time. One of them can fall into the error calendar not found if the resourceId is still null.

Screenshot 2023-12-05 at 14 37 46

Some evidence that this happens:

- Calendar not found in [gcalEventsUpdatedContentfulProcess](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:logs-insights$3FqueryDetail$3D~(end~'2023-12-04T06*3a00*3a00.000Z~start~'2023-12-01T05*3a00*3a00.000Z~timeType~'ABSOLUTE~tz~'LOCAL~editorString~'fields*20*40timestamp*2c*20*40message*2c*20*40logStream*2c*20*40log*0a*7c*20sort*20*40timestamp*20desc*0a*7c*20limit*20200*0a*7c*20filter*20*40message*20like*20*2fnot*20found*2f~queryId~'67a37abd75b1ba8a-7af2108-46b4a43-d7be12d6-9c39d6c294413b2966cbfb35~source~(~'*2faws*2flambda*2fasap-hub-production-gcalEventsUpdatedContentfulProcess)))
Screenshot 2023-12-05 at 14 42 59

- Conflict error when subscribing to calendar in [gcalResubscribeCalendarsContentful](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fasap-hub-production-gcalResubscribeCalendarsContentful/log-events/2023$252F12$252F04$252F$255B$2524LATEST$255Dced07438aa564ef88cd7aaee688838d0)
conflict

---

#### Proposed solution

Since updating resourceId with null triggers subscribe, I believe we can just remove the subscribe part from the resubscribe handler.

Screenshot 2023-12-05 at 14 39 25

So I've renamed it to `unsubscribe` only.

---

#### Test

- In the Contentful environment created in this PR (crn-4001) I've made the `Google Api Metadata` from calendar to be editable and changed the expiration date to now.
- Then I've changed the `gcalResubscribeCalendarsContentful` cron to `cron(50 15 * * ? *)` in serverless
- Then I could see that `gcalEventsUpdatedContentfulProcess` ran without any problems: [cloudwatch link](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fasap-hub-4001-gcalEventsUpdatedContentfulProcess/log-events/2023$252F12$252F05$252F$255B$2524LATEST$255Db5661edb2afc4c03bdf2bc73103daf09)
- After some time I've updated the title of one event in Google Calendar UI and I could see that it was synced in the CMS so subscribe worked


packages/server-common/src/handlers/webhooks/events-updated-process.handler.ts
```diff
@@ -40,6 +40,11 @@ export const webhookEventUpdatedProcessHandlerFactory = (
       const resourceId = record.messageAttributes.ResourceId?.stringValue;
       const channelId = record.messageAttributes.ChannelId?.stringValue;
 
+      if (!resourceId) {
+        logger.debug('Empty resourceId');
```
peterstarling, line 44: ok first of all i think there is a check just underneath that sort of handles this already, I would probably refactor it if we want to bail out without errors for when `resourceId` is missing

second - how are we ever receiving payload from google with an empty resourceId? I would have thought this would handle it https://github.com/yldio/asap-hub/blob/95be5f71aabfe7b1eb0eaeb9bd76817203de57b7/packages/server-common/src/handlers/webhooks/events-updated.handler.ts#L42-L49

finally - why do you think that upon receiving an empty resourceId we could log and return? To me that's a legitimate error, I think we want to throw here?

packages/server-common/test/handlers/calendar/unsubscribe-calendars.test.ts
```diff
@@ -0,0 +1,190 @@
...
+  test('Should get the list of calendars and unsubscribe each of them', async () => {
+    const calendarDataObject1 = {
+      ...getCalendarDataObject(),
+      channelId: undefined,
+    };
+    const calendarDataObject2 = {
+      ...getCalendarDataObject(),
+      id: 'uuid2',
+      resourceId: 'resource-id-2',
+      channelId: undefined,
```
peterstarling, line 81: I realise this is copied from the tests before but I was wondering if it's actually correct - we are meant to unsubscribe fully working subscriptions here right? So channelId should be present. We could examine an edge case for when it's not but again this would be an edge case. What do you think?

packages/server-common/test/handlers/calendar/unsubscribe-calendars.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      calendarDataObject1.resourceId,
+      `cms:${calendarDataObject1.id}`,
+    );
+    expect(unsubscribeMock).toHaveBeenCalledWith(
+      calendarDataObject2.resourceId,
+      `cms:${calendarDataObject2.id}`,
+    );
+  });
+
+  test('Should remove the resourceId after unsubscribing', async () => {
```
peterstarling, line 129: I actually wonder if we should only drop the resourceId when unsubscribing or should we just drop all the metadata, what do you think?

packages/server-common/test/handlers/webhooks/events-updated-process.handler.test.ts
```diff
@@ -90,16 +90,21 @@ describe('Event Webhook', () => {
     );
   });
 
-  test('Should throw when the resourceId is not found', async () => {
+  test('Should log when the resourceId is not empty', async () => {
```
peterstarling, line 93: see my comment above - i dont really understand why should should be the case

peterstarling (changes requested):

packages/server-common/test/handlers/calendar/unsubscribe-calendars.test.ts
gabiayako: yeah, I think we can drop all metadata 👍

apps/crn-server/src/data-providers/contentful/calendar.data-provider.ts
```diff
@@ -160,6 +160,12 @@ export class CalendarContentfulDataProvider implements CalendarDataProvider {
 
     updateEntryFields(calendar, otherFields);
 
+    if ('googleApiMetadata' in update) {
```
peterstarling, line 163: why do we need this exactly? looks like we are bailing out of the regular update if there is any metadata to update, but why do you think we should do that? technically someone might want to update all fields if they feel like it, and looking at the interface it seems perfectly fine. We would need to update the typings if we want to say 'you can update google metadata or the remaining fields but not both'.

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/calendar.data-provider.ts
gabiayako: I thought about using this `googleApiMetadata` field only if we want to set it to null
Screenshot 2023-12-06 at 12 55 53


> looks like we are bailing out of the regular update if there is any metadata to update
I thought that it would be nice to return if `googleApiMetadata` is present because for me it doesn't make sense to update `googleApiMetadata` and `recordId` at the same time since recordId is inside `googleApiMetadata`

PR #4020, opened 2023-12-08
ASAP-62 Reduce user data (data-provider version)

gabiayako:


packages/model/src/user.ts
```diff
@@ -100,13 +100,46 @@ export interface UserDataObject extends Invitee {
...
+  _tags: string[];
+  teams: UserListItemTeam[];
+};
+export type ListUserDataObject = ListResponse<UserListItemDataObject>;
+
+export type UserListItemResponse = UserListItemDataObject;
+
+export type UserAlgoliaResponse = UserListItemDataObject;
+
+export type UserListAlgoliaResponse = ListResponse<UserAlgoliaResponse>;
```
peterstarling, line 141: I think we still don't want to expose engine specific response types. Check out how this was done for Teams:

https://github.com/yldio/asap-hub/blob/91758e7b2438325d6bb579f6b23cbaf9724f0750/packages/model/src/team.ts#L86

btw you were the one to propse it. It might sound like just a naming convention but it really isn't.

apps/crn-server/src/controllers/user.controller.ts
```diff
@@ -32,9 +32,7 @@ export default class UserController {
   }
 
   async fetch(options: FetchUsersOptions): Promise<ListUserResponse> {
```
peterstarling, line 34: pretty sure that the fetch method in the controller should change too?

packages/model/src/user.ts
```diff
@@ -199,3 +232,55 @@ export type FetchUsersFilter =
 export type FetchUsersOptions = Omit<FetchOptions<FetchUsersFilter>, 'search'>;
 
 export type UserRole = 'Staff' | 'Member' | 'None';
+
+export const toAlgoliaUserItem = (user: UserResponse): UserAlgoliaResponse => {
```
peterstarling, line 236: we are not doing this for algolia - we are splitting the list response - period. That way both Algolia indexers and API responses need to be adjusted (you haven't done the latter).

peterstarling (changes requested):

packages/model/src/user.ts
gabiayako: I've renamed it here [review: rename things with Algolia to UserItem](https://github.com/yldio/asap-hub/pull/4020/commits/22969e0ea8949b58dbba01630b2c9f591a51e206) + [fix rename mistake](https://github.com/yldio/asap-hub/pull/4020/commits/333767be894dce993026378c49ba9021305b93d9)

packages/model/src/user.ts
gabiayako: I've renamed it here [review: rename things with Algolia to UserItem](https://github.com/yldio/asap-hub/pull/4020/commits/22969e0ea8949b58dbba01630b2c9f591a51e206) + [fix rename mistake](https://github.com/yldio/asap-hub/pull/4020/commits/333767be894dce993026378c49ba9021305b93d9)

apps/crn-server/src/controllers/user.controller.ts
gabiayako: But I think it's already with the right type since https://github.com/yldio/asap-hub/blob/asap-62-users-2/packages/model/src/user.ts#L194

```
export type ListUserResponse = ListResponse<UserListItemResponse>
```

packages/model/test/user.test.ts
```diff
@@ -20,4 +26,88 @@ describe('User', () => {
       expect(isUserDegree('not-a-degree')).toEqual(false);
     });
   });
+
+  describe('toUserListItem', () => {
+    it('should convert user response to algolia user item response', () => {
```
peterstarling, line 31:
```suggestion
    it('should convert user response to list user response', () => {
```

packages/server-common/src/handlers/utils.ts
```diff
@@ -26,7 +26,8 @@ export const createProcessingFunction =
     type: Type,
     logger: Logger,
     filterFunction: (item: T['data']) => boolean = () => true,
-    addTagsFunction?: (item: T['data']) => T['data'] & { _tags: string[] },
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    addTagsFunction?: (item: any) => SavePayload['data'],
```
peterstarling, line 30: this is strange, why are you having to do this?

peterstarling (approved):

packages/server-common/src/handlers/utils.ts
gabiayako: Actually I think I can undo this change. It was necessary when the data provider was returning the old UserResponse and I was calling the function to convert into the UserListItemResponse here. So the function couldn't be types as `(item: T['data']) => T['data'] & { _tags: string[] }` because the input would be `UserResponse` and output `UserListItemResponse`. And the input now could be anything. 

I was calling `toUserListItem` in `algolia-index-team-users-handler.ts` but now that the data provider sends exactly UserListItemResponse. I could just remove it.

PR #4042, opened 2023-12-14
add version history card in gp2 output form

lctrt:


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -461,6 +478,7 @@ const OutputForm: React.FC<OutputFormProps> = ({
             )}
 
             <div css={containerStyles}>
+              {createVersion && <OutputVersions versions={versions} />}
```
gabiayako, line 481: I think we need to pass `versionAction` to have the correct style and copy. The left side is figma and right side the hub

Screenshot 2023-12-18 at 11 51 48

gabiayako (changes requested):

gabiayako (approved):
nice! 🎉

PR #4069, opened 2023-12-19
feat: ASAP-234 Use Esbuild for packaging serverless files

peterstarling:
Swaps serverless-webpack for serverless-esbuild. Notable changes:
* introduces S3_SYNC_ENABLED flag for conditionally switching on the s3 sync plug. This is a plugin we need for deployment only. during packaging however it freezes the zipping process which is why it was disabled
* removes `googleapis` package in favour of `@googleapis/calendar` and  `google-auth-library` for easier bundling


lctrt (approved):
Looks good! Just curious about why S3_SYNC_ENABLED is set to false?

peterstarling:
> Looks good! Just curious about why S3_SYNC_ENABLED is set to false?

good point, I added an explanation to the description

PR #4086, opened 2023-12-26
ASAP-268 add version handling for GP2

lctrt:


apps/gp2-server/test/controllers/output.controller.test.ts
```diff
@@ -357,10 +357,17 @@ describe('outputs controller', () => {
...
+      const outputUpdateDataObject = {
+        ...getOutputUpdateDataObject(),
+        versions: ['1'],
+      };
+      expect(outputDataProviderMock.update).toHaveBeenCalledWith(
+        outputId,
+        {
+          ...outputUpdateDataObject,
+        },
+        { newVersion: undefined },
```
peterstarling, line 369: what about testing the opposite scenario, when the newVersion field is populated?

apps/gp2-server/test/controllers/output.controller.test.ts
```diff
@@ -357,10 +357,17 @@ describe('outputs controller', () => {
...
+      );
     });
 
     describe('Validation', () => {
```
peterstarling, line 373: new validation method was added but i don't see it being tested anywhere?

peterstarling (changes requested):

peterstarling (approved):

PR #4091, opened 2024-01-02
[ASAP-229]-remove entities field in researchTags

AkosuaA:


packages/contentful/migrations/crn/researchTags/20240102105424-remove-entities-field.js
```diff
@@ -0,0 +1,27 @@
+module.exports.description = '<Put your description here>';
```
gabiayako, line 1:
```suggestion
module.exports.description = 'Remove field entities from researchTags';
```

gabiayako (approved):
Nice job! 🎉  

We can remove it from this integration test as well

https://github.com/yldio/asap-hub/blob/ASAP-229-remove-research-tags-entities/apps/crn-server/test/integration/research-outputs.integration-test.ts#L40

PR #4098, opened 2024-01-04
feat: ASAP-16 Display version history when editing outputs

peterstarling:


packages/gp2-components/src/templates/OutputForm.tsx
```diff
@@ -514,12 +514,10 @@ const OutputForm: React.FC<OutputFormProps> = ({
...
-              {createVersion && (
-                <OutputVersions
-                  versions={versionList}
-                  versionAction={createVersion && 'edit'}
-                />
-              )}
+              {(versionList.length > 0 && (
+                <OutputVersions versions={versionList} versionAction="edit" />
+              )) ||
+                ''}
```
gabiayako, line 520: Do we need this `|| ''`?

packages/gp2-components/src/templates/__tests__/OutputForm.test.tsx
```diff
@@ -1932,15 +1932,48 @@ describe('OutputForm', () => {
...
+        {
+          addedDate: '2021-12-28T14:00:00.000Z',
+          documentType: 'Article',
+          id: '12345',
+          title: 'Previous Version Test',
+        },
+      ];
+      renderWithCreateVersion(undefined, false, versions);
+      expect(screen.getByText('Version History')).toBeInTheDocument();
+      expect(screen.getByText('Previous Version Test')).toBeInTheDocument();
```
gabiayako, line 1968: maybe we could also add 

```
expect(screen.getByText('test versioning')).not.toBeInTheDocument();
```

because the current title does not appear as a new version when editing (not creating a new version)

gabiayako (approved):
nice! 🎉

PR #4099, opened 2024-01-05
Asap 262 update view export buttons

AkosuaA:
ref: https://asaphub.atlassian.net/jira/software/c/projects/ASAP/boards/13?selectedIssue=ASAP-262

The isAdministrator prop was being used in the ResultList component to display different UIs for the 'Export as CSV' section. This is no longer needed and so has been removed. It also caused a bug in gp2 where the 'Export as CSV' section was rendered differently depending on whether the user was an admin or not.

Screenshot 2024-01-04 at 16 48 00
Screenshot 2024-01-04 at 16 55 53

### DIFF
--- a/apps/gp2-frontend/src/tags/ResultList.tsx
+++ b/apps/gp2-frontend/src/tags/ResultList.tsx
@@ -10,7 +10,6 @@ import {
   EventCard,
   ResultList as ResultListComponent,
 } from '@asap-hub/react-components';
-import { useCurrentUserGP2 } from '@asap-hub/react-context';
 import { eventMapper } from '../events/EventsList';
 import { usePagination, usePaginationParams } from '../hooks/pagination';
 import { useTagSearchResults } from './state';
@@ -21,8 +20,6 @@ export type ResultListProps = {
 };
 const ResultList: React.FC<ResultListProps> = ({ filters = new Set() }) => {
   const { currentPage, pageSize } = usePaginationParams();
-  const currentUser = useCurrentUserGP2();
-  const isAdministrator = currentUser?.role === 'Administrator';
 
   const { tags } = useSearch();
   const { items, total } = useTagSearchResults({
@@ -39,7 +36,6 @@ const ResultList: React.FC<ResultListProps> = ({ filters = new Set() }) => {
       numberOfPages={numberOfPages}
       currentPageIndex={currentPage}
       renderPageHref={renderPageHref}
-      isAdministrator={isAdministrator}
     >
       {items.map((result) => {
         // eslint-disable-next-line no-underscore-dangle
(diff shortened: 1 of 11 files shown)


packages/react-components/src/organisms/__tests__/ResultList.test.tsx
```diff
@@ -69,62 +69,46 @@ it('renders page controls', () => {
...
     </ResultList>,
   );
-  userEvent.click(screen.getByText(/export/i));
+  expect(screen.getByText(/export as:/i)).toBeVisible();
+  userEvent.click(screen.getByText(/csv/i));
   expect(mockExport).toHaveBeenCalled();
 });
 
-it('renders export csv as admin', () => {
+it('renders view as dropdown when view and card links are available', () => {
```
gabiayako, line 91:
```suggestion
it('renders view as dropdown when list and card links are available', () => {
```

packages/react-components/src/molecules/ListControls.tsx
```diff
@@ -1,43 +1,24 @@
...
-  fontSize: `${18 / perRem}em`,
-  svg: {
-    stroke: lead.rgb,
+  alignItems: 'center',
+  gap: rem(15),
+  'span + svg': {
+    width: '11px',
+  },
+  button: {
+    padding: `${rem(8)} ${rem(24)}`,
```
gabiayako, line 14: I think this padding `rem(24)` is not matching what we see in figma. It seems there is an extra space on the left side.

Screenshot 2024-01-05 at 13 14 32

apps/gp2-frontend/src/tags/ResultList.tsx
```diff
@@ -39,7 +36,6 @@ const ResultList: React.FC<ResultListProps> = ({ filters = new Set() }) => {
       numberOfPages={numberOfPages}
       currentPageIndex={currentPage}
       renderPageHref={renderPageHref}
-      isAdministrator={isAdministrator}
```
gabiayako, line 42: Are you sure we want to remove this? Just asking because it's not written in the ticket that we want to remove the condition to enable only admins to export outputs

AkosuaA: I removed the isAdministrator prop because it wasn't controlling access to export functionality but rather controlling the ui displayed (whether the export functionality is rendered as a link or button). Currently output exports can be done whether the user is an administrator or not. Confirming with Tiffany if this shouldn't be the case.

AkosuaA: Confirmed that all users can export outputs

gabiayako (approved):
Nice! 🎉

This CSV button has an extra left space in prod already but if it's easy enough I would fix the space here too 😄  if it's not, maybe it's worth it to create a ticket to fix this visual issue
Screenshot 2024-01-09 at 07 43 45

PR #4108, opened 2024-01-09
feat: ASAP-288 Add middle name and nickname to user models

peterstarling:


lctrt (changes requested):
Would be great to remove duplication of the displayName function :) Otherwise looks great!

lctrt (approved):

apps/crn-server/test/routes/user.route.test.ts
```diff
@@ -379,70 +406,59 @@ describe('/users/ route', () => {
...
+            expect(response.status).toBe(200);
+          });
+
+          test('Should be able to provide an empty string for the %s parameter ', async () => {
+            const response = await supertest(appWithMockedAuth)
+              .patch(`/users/${userId}`)
+              .send({ [parameter]: '' });
+
+            expect(response.status).toBe(200);
+          });
```
gabiayako, line 449: I believe these two are the same

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -97,6 +97,21 @@ describe('User data provider', () => {
       expect(result).toEqual(null);
     });
 
+    test.each(['middleName', 'nickname'] satisfies Array<keyof UserDataObject>)(
+      'Should not return a field when its an emptry string for %s',
```
gabiayako, line 101:
```suggestion
      'Should not return a field when its an empty string for %s',
```

apps/gp2-server/test/data-providers/user.data-provider.test.ts
```diff
@@ -62,6 +62,23 @@ describe('User data provider', () => {
       expect(result).toEqual(getUserDataObject());
     });
 
+    test.each(['middleName', 'nickname'] satisfies Array<
+      keyof gp2Model.UserDataObject
+    >)(
+      'Should not return a field when its an emptry string for %s',
```
gabiayako, line 68:
```suggestion
      'Should not return a field when its an empty string for %s',
```

PR #4118, opened 2024-01-11
ASAP-290 Fix prettier version

gabiayako:
Updating prettier version to the last version available so it uses the fix below introduced in `3.0.3` release: //github.com/prettier/prettier/blob/main/CHANGELOG.md#303

> Prettier v3 uses dynamic imports, user https://github.com/yarnpkg/berry/pull/5411#issuecomment-1523502224 when Yarn's PnP mode is enabled, add [preferUnplugged: true](https://yarnpkg.com/configuration/manifest#preferUnplugged) to package.json, so Yarn will install Prettier as unplug by default.

Also `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` were updated accordingly to this PR https://github.com/yldio/asap-hub/pull/4047/files

---

Note: It's necessary to reload window after changing to this branch to see it working as the video below

https://github.com/yldio/asap-hub/assets/16595804/e2c087e0-1011-4477-8dc7-6b2583e572d8


apps/crn-frontend/src/welcome/Welcome.tsx
```diff
@@ -50,7 +50,7 @@ const Welcome: React.FC<Record<string, never>> = () => {
       allowSignup
       onClick={() =>
         // Effect should populate this ref before a click can occur
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
gabiayako, line 53: I couldn't find in the changelog but I believe this `@typescript-eslint/no-non-null-assertion` and `@typescript-eslint/no-empty-function` was changed in one of the versions in between the one we had and the last one because now we get this `Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-non-null-assertion')` and ` Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-empty-function')` in several files


Screenshot 2024-01-12 at 07 58 12

.github/workflows/reusable-deployment.yml
```diff
@@ -171,6 +171,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - run: yarn cache clean
```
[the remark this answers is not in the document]
peterstarling: so the reason you are having this issue is because you unplugged a couple of packages - that means you told yarn to not store them as zips inside `.yarn/cache` but rather unpack it and store in `.yarn/unplugged` and that folder (as opposed to the `cache` one) is not checked into git meaning that we have to do `yarn install` do get the contents of those unplugged packages. 

The way we go about those packages is that we run yarn install once and then cache it and then we use this step
```
       - name: Cache build typecheck output
        uses: ./.github/actions/cache-unplugged
        with:
          environment-name: ${{ inputs.environment-name }}
```

to load this cache into your current job. So just try and add this instead of cache clean and install.

Would be keen to understand why you had to unplug those packages?

packages/react-components/src/molecules/TabbedContent.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
peterstarling, line 1: out of curiosity - why do we no longer need these? are non-null assertions now allowed?

package.json
```diff
@@ -74,10 +74,17 @@
...
+    "@typescript-eslint/eslint-plugin": "6.18.1",
+    "@typescript-eslint/parser": "6.18.1",
+    "eslint-plugin-import": "2.29.1"
   },
   "dependenciesMeta": {
+    "@typescript-eslint/eslint-plugin@6.18.1": {
+      "unplugged": true
+    },
+    "@typescript-eslint/parser@6.18.1": {
+      "unplugged": true
```
peterstarling, line 86: any particular reason why you need to unplug this and the one above?

peterstarling (changes requested):

.github/workflows/reusable-deployment.yml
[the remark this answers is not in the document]
gabiayako: From what is written in prettier changelog I understood version 3 needs to be unplugged. Here in the description of this PR I've left the link to the changelog https://github.com/yldio/asap-hub/pull/4118#issue-2076592651

peterstarling (approved):

PR #4126, opened 2024-01-17
feat: ASAP-284 Use turborepo to speed up builds

peterstarling:
- add basic turbo setup for build jobs
- add caching to the CI


turbo.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build:babel": {
+      "outputs": ["build/**/*", "build-cjs/**/*"]
+    },
+    "//#build:typecheck:tsc": {
```
peterstarling, line 7: unlike other commands which are all run once per app/package, this one is run from the root. That explains the weird syntax and the fact we had to define the `inputs` as turbo wasn't able to figure it out so easily and this caused issues with caching. 

Would be nice to break this job down so each app and package has its own typecheck but for now that should do.

turbo.json
```diff
@@ -0,0 +1,26 @@
...
+      "outputs": [
+        "{apps,packages}/**/build/**/*.d.ts",
+        "{apps,packages}/**/*.tsbuildinfo"
+      ],
+      "cache:": false
```
AkosuaA, line 12:
```suggestion
      "cache": false
```

AkosuaA (approved):
LGTM

PR #4131, opened 2024-01-24
[ASAP-313] - update gp2 algolia schema

AkosuaA:
ref: https://asaphub.atlassian.net/jira/software/c/projects/ASAP/boards/13?selectedIssue=ASAP-313

### DIFF
--- a/packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json
+++ b/packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json
@@ -32,6 +32,12 @@
     "speakers.team.displayName",
     "speakers.user.displayName",
     "title",
+    "role",
+    "degrees",
+    "city",
+    "country",
+    "region",
+    "contributingCohorts.name",
     "workingGroup.title"
   ]
 }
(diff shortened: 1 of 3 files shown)


gabiayako (commented):
Don't we need to add the same thing to `packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json`? 

Also, it would be nice to add the tag gp2-create-environment so it's possible to test it 😉

AkosuaA:
> Don't we need to add the same thing to `packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json`?
> 
> Also, it would be nice to add the tag gp2-create-environment so it's possible to test it 😉

It seems we don't use the reverse index in gp2? I didn't find a query using it so I wasn't sure if we needed to add it there as well. There were two fields in the algolia-schema.json that weren't in the algolia-reverse-timestamp-schema.json as well so I wasn't sure what the criteria was. I guess to keep it uniform I could just add them

gabiayako (approved):
To be honest I couldn't find if we use the `reverse` timestamp, but I think it is nice to have them consistent 😅

PR #4133, opened 2024-01-25
ASAP-287 Active Campaign

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-287

---

#### How to test it works

I've made changes to my user profile in both https://4133.hub.asap.science/ and https://4133.gp2.asap.science/ and checked that the changes appeared in https://asapyldtesting.activehosted.com/app/contacts/17

Screenshot 2024-01-26 at 07 39 34

Also, I've set `AlumniSinceDate` on Contentful `crn-4133` environment https://app.contentful.com/spaces/5v6w5j61tndm/environments/crn-4133/views/entries then I checked that `AlumniStatus` changed on https://asapyldtesting.activehosted.com/app/contacts/17


apps/crn-server/src/controllers/user.controller.ts
```diff
@@ -31,6 +36,42 @@ export default class UserController {
     await this.userDataProvider.update(id, update, { suppressConflict });
     return this.fetchById(id);
   }
+  async createActiveCampaignContact(user: UserResponse): Promise<void> {
```
peterstarling, line 39: extracting this and the other active campaign methods is good, but I am not sure if this should become controller responsibility. I think we intentionally removed controllers from the handlers that do not interfere with the `Response` objects (so our public API + algolia indexers use them) because controllers responsibility was supposed to be just returning data for the FE app. Maybe just have them as functions next to your handler?

peterstarling: and to be fair these methods should just be part of some ActiveCampaign client or something like that, interaction with that third party API or generating API specific payload mixed with all the remaining functionality here is imho not a great practice

apps/crn-server/test/controllers/user.controller.test.ts
```diff
@@ -10,6 +15,25 @@ import {
 } from '../fixtures/users.fixtures';
 import { getDataProviderMock } from '../mocks/data-provider.mock';
 
+jest.mock('@asap-hub/server-common', () => ({
```
peterstarling, line 18: I think in BE we try to move away from jest.mock in favour of a DI

apps/crn-server/test/controllers/user.controller.test.ts
```diff
@@ -472,4 +496,150 @@ describe('Users controller', () => {
       );
     });
   });
+
+  describe('ActiveCampaign', () => {
```
peterstarling, line 500: how about when the API returns a 500 or maybe other status codes?

peterstarling (changes requested):

packages/server-common/src/handlers/user/sync-active-campaign-contact-factory.ts
```diff
@@ -0,0 +1,136 @@
...
+export interface UserController {
+  fetchById(id: string): Promise<gp2.UserResponse | UserResponse>;
+  update(
+    id: string,
+    update: UserUpdateRequest | gp2.UserUpdateRequest,
+  ): Promise<gp2.UserResponse | UserResponse>;
+}
+
+export const syncActiveCampaignContactFactory =
+  <Controller extends UserController>(
```
peterstarling, line 28: are you sure you need this generic here? can you not just expect userController which implements the UserController interface?

peterstarling (approved):

packages/server-common/src/handlers/user/sync-active-campaign-contact-factory.ts
gabiayako: Yeah, I think this is not necessary, I'll remove it :)

gabiayako:
pushing it without ACTIVE_CAMPAIGN_TOKEN set, so I can check if it gives an error

peterstarling (approved):

PR #4137, opened 2024-01-29
Asap 235 users

lctrt:


packages/model/src/user.ts
```diff
@@ -178,6 +181,7 @@ export type UserCreateDataObject = {
   role: Role;
   social?: Omit<UserSocialLinks, 'orcid'>;
   teams?: Pick<UserTeam, 'id' | 'role' | 'inactiveSinceDate'>[];
+  tags?: Pick<ResearchTagDataObject, 'id' | 'name'>[];
```
peterstarling, line 184: although it might seem like it's correct because of the `teams` definition above, I actually think this field should just be `tagIds` similarly to how we have defined the input for the relationships elsewhere where the is no extra data on the relationship (`teams` has data on the 'linking model': `role` and `inactiveSinceDate`, they both need to be defined along with the `id`). 

Let me ask you this, could a user use an existing tag `id` and give it any arbitrary `name`?

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -781,6 +821,22 @@ describe('User data provider', () => {
           },
         });
       });
+      test('map tag value to a linked resource', async () => {
+        await userDataProvider.update('123', {
+          tags: [{ id: '1', name: '1' }],
```
peterstarling, line 826: check out my comment related to the type definition for the `UserCreateObject`- you can see here that the `name` is being ignored

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
```diff
@@ -680,6 +694,32 @@ describe('User data provider', () => {
...
+
+        expect(contentfulGraphqlClientMock.request).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.objectContaining({
+            limit: 8,
+            order: ['lastName_ASC'],
+            skip: 0,
+            where: {
+              AND: [
+                { OR: [{ expertiseAndResourceDescription_contains: 'Tag' }] },
```
peterstarling, line 715: I don't understand this one, aren't we removing `expertiseAndResources` field and switching to research-tags?

apps/crn-server/test/data-providers/contentful/research-tags.data-provider.test.ts
```diff
@@ -117,11 +129,70 @@ describe('Research Tags Data Provider', () => {
...
+    });
+
+    test('can create a research tag', async () => {
+      const publish = jest.fn();
+      environmentMock.createEntry.mockResolvedValue({
+        sys: { id: '1' },
+        publish,
+      } as unknown as Entry);
+      const response = await researchTagsDataProvider.create('tag');
+      expect(environmentMock.createEntry).toHaveBeenCalled();
```
peterstarling, line 189: no need to assert this when you have the expectation below

apps/crn-server/src/data-providers/contentful/user.data-provider.ts
```diff
@@ -433,6 +462,14 @@ export const parseContentfulGraphQlUserListItem = (
     onboarded: typeof item.onboarded === 'boolean' ? item.onboarded : true,
     role: item.role && isUserRole(item.role) ? item.role : 'Guest',
     teams: userTeams,
+    tags:
+      item.tagsCollection?.items
```
peterstarling, line 466: you might want to extract this and reuse across two parsers for detail and list view

peterstarling (changes requested):
I think you also need to validate the research-tag IDs when they are part of the user input - currently there is only standard ajv validation added to make sure the ID is present) but what happens when the client sends an invalid (non-existent) ID?

apps/crn-server/test/data-providers/contentful/users.data-provider.test.ts
lctrt: It's for the script to be working. Once everything is migrated we would be able to remove the whole field (including that search).

apps/crn-server/scripts/cli.ts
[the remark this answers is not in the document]
lctrt: I was struggling a bit to find a good place for this that seemed the most sensible place. We won't repeat the process (normally) but we can reuse the script for the other entities.

peterstarling (approved):

PR #4142, opened 2024-01-31
ASAP-319 Add state/province to the hub

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-319

---

This PR in both CRN and GP2:
- Adds the migrations to add the field stateOrProvince
- Adds this field stateOrProvince to the GraphQL user queries
- Adapts the user data provider to return the stateOrProvince field
- Adds the stateOrProvince to user profile edit form
- Displays the stateOrProvince in the user profile header and card
- Updates algolia schema so stateOrProvince is searchable in the People list


apps/gp2-server/scripts/import-users.ts
```diff
@@ -82,6 +82,7 @@ const app = async () => {
         firstName: data[0]!,
         lastName: data[1]!,
         country: data[4]!,
+        stateOrProvince: 'Unknown',
```
gabiayako, line 85: I'm not sure if this script will ever be used again, but it was giving a TS error without this line. So I've added it. Maybe we could delete this in a following PR?

peterstarling (approved):

gabiayako:
rebase to master to fix conflicts

PR #4148, opened 2024-02-06
ASAP-328 Adjustments to ActiveCampaign push data

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-328

### DIFF
--- a/apps/crn-server/src/handlers/teams/sync-active-campaign-team-member-status.ts
+++ b/apps/crn-server/src/handlers/teams/sync-active-campaign-team-member-status.ts
@@ -1,11 +1,5 @@
 import { NotFoundError } from '@asap-hub/errors';
 import { TeamEvent } from '@asap-hub/model';
-import { isBoom } from '@hapi/boom';
-
-import type { TeamPayload } from '../event-bus';
-
-import TeamController from '../../controllers/team.controller';
-import UserController from '../../controllers/user.controller';
 import {
   createContact,
   EventBridgeHandler,
@@ -15,12 +9,19 @@ import {
   syncUserActiveCampaignData,
   updateContact,
 } from '@asap-hub/server-common';
-import { activeCampaignAccount, activeCampaignToken } from '../../config';
+import { isBoom } from '@hapi/boom';
+
+import type { TeamPayload } from '../event-bus';
+
 import {
   getContactPayload,
   getFieldIdByTitle,
   updateContactLists,
 } from '../user/sync-active-campaign-contact';
+
+import { activeCampaignAccount, activeCampaignToken } from '../../config';
+import TeamController from '../../controllers/team.controller';
+import UserController from '../../controllers/user.controller';
 
 export const syncActiveCampaignTeamMemberStatusFactory =
   (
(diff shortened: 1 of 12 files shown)


apps/crn-server/src/handlers/teams/sync-active-campaign-team-member-status.ts
```diff
@@ -0,0 +1,101 @@
+/* istanbul ignore file */
```
peterstarling, line 1: i think there is far too much logic to just ignore this file, in fact I don't think we should ever ignore any handlers unless maybe there are using some modules which fire immediate side effects and then we should make sure to only do basic DI and setup in there and move all the logic elsewhere where it is properly tested. Here this is not the case

apps/crn-server/src/handlers/user/sync-active-campaign-contact.ts
```diff
@@ -23,32 +26,39 @@ const userDataProvider = getUserDataProvider();
...
-    activeCampaignAccount,
-    activeCampaignToken,
-  );
-
-  const fieldIdByTitle = getCustomFieldIdByTitle(customFields);
+): ContactPayload => {
+  const teams = user.teams.length
+    ? user.teams.flatMap((team, index) => [
+        {
+          field: fieldIdByTitle[`CRN Team ${index + 1}`]!, // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
peterstarling, line 36: this is a very strange assertion, how are you so sure that the value in the array with this index actually exists?

apps/crn-server/src/handlers/user/sync-active-campaign-contact.ts
```diff
@@ -76,22 +86,54 @@ const getContactPayload = async (
...
   };
 };
 
+export const updateContactLists = async (contactId: string) => {
+  const listIdByName = await getListIdByName(
+    activeCampaignAccount,
+    activeCampaignToken,
+  );
+
+  const masterListId = listIdByName['Master List']!; // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
peterstarling, line 113: either make sure that these aren't represented as generic string indexes or do a check each time you use them.

packages/server-common/src/handlers/user/sync-active-campaign-contact-factory.ts
```diff
@@ -24,8 +36,144 @@ export interface UserController {
   ): Promise<gp2.UserResponse | UserResponse>;
 }
 
+export const syncUserActiveCampaignData = async (
```
peterstarling, line 39: this function's signature is insane, make sure to use an object with config for some of the parameters like tokens, etc, you should also consider not passing all functions as dependencies (especially if they are pure with no side effects

peterstarling (changes requested):

apps/crn-server/src/handlers/user/sync-active-campaign-contact.ts
gabiayako: Because I added this field manually in Active Campaign. The idea is to create a type for it then?

peterstarling (approved):

PR #4150, opened 2024-02-06
[ASAP-332] - implement researchTags fetchById

AkosuaA:
- implements researchTags fetchById
- modifies the onboarding validation process to use the tags field instead of expertiseAndResourceTags


apps/crn-server/test/data-providers/contentful/research-tags.data-provider.test.ts
```diff
@@ -166,8 +168,22 @@ describe('Research Tags Data Provider', () => {
   });
 
   describe('Fetch-by-id', () => {
-    test('should throw an error', async () => {
-      await expect(researchTagsDataProvider.fetchById()).rejects.toThrow();
+    test('Should return null when user is not found', async () => {
```
peterstarling, line 171:
```suggestion
    test('Should return null when tag is not found', async () => {
```

packages/validation/src/__tests__/user.test.ts
```diff
@@ -69,32 +69,32 @@ describe('isUserOnboardable validation', () => {
...
     };
 
     expect(isUserOnboardable(userIncompleteResponse)).toEqual({
       isOnboardable: false,
-      expertiseAndResourceTags: { valid: false },
+      tags: { valid: false },
     });
   });
 
   it('Should fail if fewer than 5 Expertise and Resources are provided in user profile', async () => {
```
peterstarling, line 84:
```suggestion
  it('Should fail if fewer than 5 Tags are provided in user profile', async () => {
```

peterstarling (approved):

PR #4151, opened 2024-02-07
[ASAP-17] - new version reminder

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-17


apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -879,6 +879,15 @@ const getPublishedResearchOutputVersionRemindersFromQuery = (
     return [];
   }
 
+  items.sort((reminderA, reminderB) => {
```
peterstarling, line 882: we are already doing the exact same sort here: https://github.com/yldio/asap-hub/blob/f9cffb6533da1de710d124646a0387ffb5762d3e/apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts#L193-L198
any reason why this needs to occur again earlier?

apps/gp2-server/test/data-providers/reminder.data-provider.test.ts
```diff
@@ -41,209 +42,568 @@ describe('Reminders data provider', () => {
...
+          total: 3,
         });
-      },
-    );
+      });
 
-    test('Should sort the published output reminders based on added date showing most recent first', async () => {
-      const date = '2024-01-05T08:00:00Z';
-      setNowToLast24Hours(date);
+      test('Should only return the most recent published output version reminder for an output', async () => {
```
gabiayako, line 531: nice 🎉

apps/gp2-server/src/controllers/reminder.controller.ts
```diff
@@ -18,7 +18,16 @@ export default class ReminderController {
...
         }).$,
-        description: `**${reminder.data.statusChangedBy}** in ${reminder.data.associationType} **${reminder.data.associationName}** published a **${reminder.data.documentType}** output: "${reminder.data.title}".`,
+        description:
+          reminder.entity === 'Output'
+            ? `**${reminder.data.statusChangedBy}** in ${reminder.data.associationType} **${reminder.data.associationName}** published a **${reminder.data.documentType}** output: "${reminder.data.title}".`
+            : `**${reminder.data.statusChangedBy}** on **${
+                reminder.data.associationName
+              }** published a new ${
+                reminder.data.associationType
+              } ${reminder.data.documentType.toLowerCase()} output version: "${
```
gabiayako, line 28: I believe this document type is in bold as well

Screenshot 2024-02-08 at 07 34 42

packages/contentful/src/gp2/queries/reminder.queries.ts
```diff
@@ -65,5 +69,45 @@ export const FETCH_REMINDERS = gql`
...
+              sys {
+                id
+              }
+              title
+              documentType
+              createdBy {
+                firstName
+                lastName
+              }
+              relatedEntitiesCollection(limit: 1) {
```
gabiayako, line 90: I've noticed now reading the spreadsheet with the reminders instructions (https://docs.google.com/spreadsheets/d/1dDB2wboG6cS7FW3suPa7K2k9PCTCIZRcY8le31nKyUA/edit#gid=0)

Screenshot 2024-02-08 at 07 44 52

that there's an issue here and with the published outputs. We are fetching only the main entity not all of them and the reminder should be displayed to all of them 😅

apps/gp2-server/src/data-providers/reminder.data-provider.ts
```diff
@@ -134,8 +161,98 @@ const getPublishedOutputRemindersFromQuery = (
...
+      outputVersionReminders.push({
+        id: `output-version-published-${outputVersion.sys.id}`,
+        entity: 'Output Version',
+        type: 'Published',
+        data: {
+          outputId: output.sys.id,
+          documentType: output.documentType,
+          title: output.title,
+          publishedAt: outputVersion.sys.publishedAt,
+          statusChangedBy: userName,
```
gabiayako, line 237: In the published reminder we can use the name of the person who created the output because since we don't have drafts, the person who created is the one who published it.

But in version reminders, we can't guarantee that the person who created the output is the one who created a version. In CRN it seems that we don't display the name of the person who created the version

Screenshot 2024-02-08 at 07 48 37

Maybe you could ask Tiff if the name is necessary. Otherwise we would need to add a new field on Outputs to track who was the last person to modify it

gabiayako (changes requested):

peterstarling (approved):

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
AkosuaA: The earlier sort is so that if multiple versions have been created for one output within the time limit we return a reminder for the latest only. I sort the output versions and then when a reminder is generated for an output version keep track of its related output. Subsequent versions with the same related output will be ignored.

The later sort sorts all reminders (eg: output publish, output version publish, events etc).

packages/contentful/src/gp2/queries/reminder.queries.ts
AkosuaA: I've raised this with Tiffany. She's noted it down and will discuss with João when he's back.

apps/gp2-server/src/data-providers/reminder.data-provider.ts
AkosuaA: Discussed this with Tiffany. For now she's agreed to removing the user name as we do in CRN.

gabiayako (approved):
Nice! 🎉

PR #4154, opened 2024-02-12
feat: ASAP-134 Event Page related research copy change

peterstarling:


packages/react-components/src/templates/EventPage.tsx
```diff
@@ -106,7 +106,7 @@ const EventPage = <
           title={titleOutputs}
           description={
             descriptionOutput ||
-            'Find out all shared research outputs that are related to this event.'
+            'Find all shared research outputs related to this event.'
```
gabiayako, line 109: according to the ticket https://asaphub.atlassian.net/browse/ASAP-134 the copy should be "Find all related Research" but in figma is like this one above 🤔 I think maybe the ticket is right?

gabiayako (approved):

PR #4155, opened 2024-02-12
ASAP-314 Tags searchable in WG list view

gabiayako:
Jira: https://asaphub.atlassian.net/browse/ASAP-314

---

wg-tags

### DIFF
--- a/packages/algolia/schema/crn/algolia-schema.json
+++ b/packages/algolia/schema/crn/algolia-schema.json
@@ -44,8 +44,7 @@
     "interestGroup.id",
     "workingGroup.id",
     "workingGroups.id",
-    "publishingEntity",
-    "complete"
+    "publishingEntity"
   ],
   "numericAttributesToIndex": null,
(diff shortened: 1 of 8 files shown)


packages/algolia/schema/crn/algolia-schema.json
```diff
@@ -44,7 +44,8 @@
     "interestGroup.id",
     "workingGroup.id",
     "workingGroups.id",
-    "publishingEntity"
+    "publishingEntity",
+    "complete"
```
AkosuaA, line 48: Just wondering why we add the complete field to the attributesToIndex.

AkosuaA (approved):

packages/algolia/schema/crn/algolia-schema.json
gabiayako: nice! you're right. I thought it was necessary to add it to `attributesToIndex` as well so the filter could work, but it is not necessary

PR #4157, opened 2024-02-14
ASAP-335 Active Campaign - Remove alumni from lists

gabiayako:


packages/server-common/test/handlers/user/sync-active-campaign-contact-factory.test.ts
```diff
@@ -183,6 +187,48 @@ describe('Sync ActiveCampaign Contact Factory', () => {
...
+  });
+
+  it('should not call unsubscribe from lists if user is not alumni', async () => {
+    const userId = getUserDataObject().id;
+    const user = {
+      ...getUserResponse(),
+      alumniSinceDate: undefined,
+    } as UserResponse;
+    userController.fetchById.mockResolvedValue(user);
+    mockActiveCampaign.getContactIdByEmail.mockResolvedValue(null);
```
peterstarling, line 220: what about when they were not an alumni but have become one, surely this would then return something and it would be another valid use case that you should test, no? I feel like this test is just a repetition of the one from line 95

gabiayako: We always fetch the user from Contentful which is the source of truth. 

Screenshot 2024-02-14 at 10 32 51

I believe it doesn't matter in this case what comes back from `getContactIdByEmail` from Active Campaign 🤔 because even if the data is not synced yet there, we always have the data from Contentful. I can change the mock in this test though to seem more realistic

gabiayako: Committed here [adjustment in test mock](https://github.com/yldio/asap-hub/pull/4157/commits/bcbc666db1dbd50947cc7f2db4ecf870d2678591) :)

peterstarling (approved):

PR #4166, opened 2024-02-23
ASAP-70 Re-enable integration tests

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-70

#### Note
It was created an environment called “Empty” in Contentful (CRN space) with only migrations and some research tags left.

#### This PR
- Re-enables integration tests in CI, cloning the new “Empty” environment instead of “Development”
- Removes test/integration/contentful folder that was only useful when we still had squidex. All tests are in test/integration now. The content of apps/crn-server/test/integration/contentful/research-outputs.integration-test.ts was moved to apps/crn-server/test/integration/research-outputs.integration-test.ts 
- Removes teardown to reduce the test duration since we’re going to delete the integration test env anyway and we’re cloning from an empty env.
- Adds a step to “on-push-master” to run the migrations following the requirement from Jira ticket -> “make sure to run content model migrations with every master merge (same as prod and dev)”


.github/workflows/on-push-master.yml
```diff
@@ -102,6 +102,30 @@ jobs:
       GP2_CONTENTFUL_WEBHOOK_AUTHENTICATION_TOKEN: ${{ secrets.GP2_CONTENTFUL_WEBHOOK_AUTHENTICATION_TOKEN }}
       SLACK_WEBHOOK: 'n-a'
 
+  empty-integration-test-env-contentful-migration:
```
peterstarling, line 105: I think I would prefer to stick to the similar naming convention of that within the reusable-deployment, something like:
```suggestion
  crn-contentful-migrate-empty:
```

peterstarling (approved):

gabiayako:
rebasing to master

PR #4169, opened 2024-02-26
update data providers to use research tags instead of plain text tags

lctrt:


apps/crn-server/test/fixtures/events.fixtures.ts
```diff
@@ -402,6 +402,16 @@ export const getContentfulEventDataObject = (
...
+  tags: [
+    //    {
+    //     "id": "Hello World",
+    //           "name": "Hello World",
+    //    },
+    // {
+    //     "id": "Hello World",
+    //           "name": "Hello World",
+    //    }
+  ],
```
gabiayako, line 414: it's possible to delete this, right?

apps/crn-server/src/data-providers/contentful/team.data-provider.ts
```diff
@@ -71,9 +72,6 @@ export class TeamContentfulDataProvider implements TeamDataProvider {
             ...searchTerms.map((term) => ({
               projectTitle_contains: term,
             })),
-            {
```
gabiayako, line 74: We currently list teams fetching from the API. I think if we remove this query, it will break the search by tags
Screenshot 2024-02-27 at 06 38 00

apps/crn-server/test/fixtures/teams.fixtures.ts
```diff
@@ -79,7 +81,9 @@ export const getContentfulGraphqlTeam = (): NonNullable<
   inactiveSince: null,
   projectTitle:
     'The genome-microbiome axis in the cause of Parkinson disease: Mechanistic insights and therapeutic implications from experimental models and a genetically stratified patient population.',
-  expertiseAndResourceTags: ['Animal resources'],
+  researchTagsCollection: {
+    items: [],
```
gabiayako, line 85: Maybe we could leave a tag here?
```suggestion
    items: [
    { sys: { id: 'tag-1' }, name: 'Animal resources 1' },
    ],
```

packages/model/src/team.ts
```diff
@@ -46,7 +47,7 @@ export interface TeamMember {
 
 export type TeamDataObject = Omit<TeamCreateRequest, 'applicationNumber'> & {
   id: string;
-  expertiseAndResourceTags: string[];
+  tags?: Pick<ResearchTagDataObject, 'id' | 'name'>[];
```
gabiayako, line 50: isn't it required? (the same question to the other models)

apps/crn-server/test/fixtures/interest-groups.fixtures.ts
```diff
@@ -228,7 +227,7 @@ export const getContentfulGraphqlInterestGroup = (): NonNullable<
   },
   active: true,
   name: "JT's Group",
-  tags: ['coding'],
+  researchTagsCollection: { items: [] },
```
gabiayako, line 230:
```suggestion
  researchTagsCollection: { items: [
    { sys: {id: 'tag-1'}, name: 'coding'}
  ] },
```

gabiayako (changes requested):

apps/crn-server/src/data-providers/contentful/team.data-provider.ts
peterstarling: interesting. If searching by tags after we add this relationship is difficult (or impossible) we might want to consider adding a follow up ticket to move this search to algolia

peterstarling (changes requested):

apps/crn-server/src/data-providers/contentful/research-tag.data-provider.ts
```diff
@@ -120,3 +125,18 @@ const parseResearchTag = (item: ResearchTagItem): ResearchTagDataObject => {
       : undefined,
   };
 };
+
+export const parseResearchTags = (
```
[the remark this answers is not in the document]
lctrt: It's the parse function that comes from the user data provider, I extracted it to avoid copy pasting to all data providers. I could use the parseResearchTag function but that would mean querying data that we don't actually use for tag display so it seems wasteful.

peterstarling (approved):

gabiayako (approved):

PR #4204, opened 2024-03-13
ASAP-341 Team leadership and membership in Algolia

gabiayako:
JIRA ticket: https://asaphub.atlassian.net/browse/ASAP-341

---

#### Frontend fetching analytics data from algolia
fe-analytics

---

#### Example of a successful on-demand sync

Action link: https://github.com/yldio/asap-hub/actions/runs/8294600796/job/22699930025
Screenshot 2024-03-15 at 10 33 09


apps/asap-cli/src/scripts/algolia/set-settings.ts
```diff
@@ -33,15 +35,18 @@ export const setAlgoliaSettings = async ({
   await index
     .setSettings({
       ...indexSchema,
-      replicas: [replicaIndexName],
+      ...(hasReverseTimestampReplica ? { replicas: [replicaIndexName] } : {}),
```
gabiayako, line 38: I think we won't need the reverse timestamp replica for analytics because we don't want to order teams based on created at or something like that

.github/actions/setup-environment/action.yml
```diff
@@ -305,10 +308,12 @@ runs:
           if [ "$CRN_CREATE_CONTENTFUL_PR_ENV" = true ]; then
             echo "setting up dedicated PR env"
             echo "crn-algolia-index=asap-hub_CI-$PR_NUMBER" >> $GITHUB_OUTPUT
+            echo "analytics-algolia-index=asap-hub-analytics_CI-$PR_NUMBER" >> $GITHUB_OUTPUT
```
peterstarling, line 311: don't you think it is probably wise to add the app prefix (crn/gp2) already? I can imagine at some point the new requirement might be to introduce analytics to gp2

.github/environment/Development
```diff
@@ -1,3 +1,4 @@
+analytics-algolia-index=asap-hub-analytics_dev
```
peterstarling, line 1: same here (and probably in many other places in the CI) - add the crn_ prefix

.github/workflows/on-remove.yml
```diff
@@ -172,6 +172,54 @@ jobs:
           CONTENTFUL_MANAGEMENT_ACCESS_TOKEN: ${{ secrets.CONTENTFUL_MANAGEMENT_TOKEN }}
           CONTENTFUL_SPACE_ID: ${{ steps.setup.outputs.gp2-contentful-space-id }}
 
+  analytics-algolia-index:
```
peterstarling, line 175: it feels like a lot of setup just to delete the index and then do it again for the crn-algolia-index one. I'm guessing even the algolia api key you are passing into it has got access to both this one and regular crn-algolia-index so why not just add one line to the the last step of `crn-algolia-index`?

.github/workflows/reusable-create-environment.yml
```diff
@@ -101,6 +101,69 @@ jobs:
...
+          fi
+        env:
+          ALGOLIA_APPLICATION_ID: ${{ steps.algolia-keys.outputs.algolia-app-id }}
+          ALGOLIA_ADMIN_API_KEY: ${{ steps.algolia-keys.outputs.algolia-api-key }}
+          ALGOLIA_INDEX: ${{ steps.setup.outputs.analytics-algolia-index }}
+      - name: Copy index from Dev
+        if: ${{ inputs.environment-name=='Branch' && steps.index-exists.outputs.index-exists == 'false' && inputs.crn-on-branch-env == 'true' }}
+        shell: bash
+        run: |
+          algolia indices copy -y -w $ALGOLIA_INDEX_DEV $ALGOLIA_INDEX
```
peterstarling, line 160: I'm also not sure if we should keeps these two jobs separate... I think creating indexes in general could be done all within one job. The setup is identical right?

.github/workflows/reusable-deployment.yml
```diff
@@ -166,6 +166,49 @@ jobs:
...
+        uses: ./.github/actions/algolia-keys
+        with:
+          algolia-api-key-name: ${{ steps.setup.outputs.crn-algolia-api-key-name }}
+          algolia-app-id-name: ${{ steps.setup.outputs.crn-algolia-app-id-name }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-default-region: ${{ steps.setup.outputs.aws-default-region }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Synchronize Schema
+        run: |
+          yarn algolia:set-settings -a $ALGOLIA_APP_ID -k $ALGOLIA_API_KEY -n $ALGOLIA_INDEX -p analytics
```
peterstarling, line 206: same goes for settings - you could just rename ALGOLIA_INDEX to something like ALGOLIA_ENTITY_INDEX and add ALGOLIA_ANALYTICS_INDEX or something like that and have them run step after step

apps/asap-cli/src/cli.ts
```diff
@@ -261,9 +261,9 @@ yargs(hideBin(process.argv))
         })
         .option('appName', {
           alias: 'p',
-          description: 'Name of App - crn or gp2',
+          description: 'Name of App - crn, gp2 or analytics',
```
peterstarling, line 264: what if we have crn and gp2 analytics?

peterstarling (changes requested):

lctrt (approved):

peterstarling (approved):

PR #4213, opened 2024-03-21
ASAP-361 Adjust parameters to reduce revision errors

gabiayako:
### Why do we have a polling script?

From https://www.contentful.com/faq/webhooks/#if-i-received-a-publish-event-why-do-i-get-an-old-version-in-the-cda
Screenshot 2024-03-21 at 10 38 41

### Retry parameters

The `retry` had `maxRetryTime` as 8s. So we would try to fetch fresh data during only 8 seconds. More about parameters here https://github.com/vercel/async-retry?tab=readme-ov-file#api.

Running it locally, it would give us most of times 5 trials

Screenshot 2024-03-21 at 10 28 33

which might be not enough if the cache is taking a long time to be purged. 

This is not happening so frequently, though. From Sentry it seems that we had a peak over March 6th ~ 8th but it is not happening so often now

Screenshot 2024-03-21 at 10 29 27

Link to sentry: https://yld.sentry.io/issues/5038500144/?referrer=slack&notification_uuid=9aa914de-ff4c-414b-a453-a4b8c950514c&environment=production&alert_rule_id=14694571&alert_type=issue

### Proposed solution

The `maxRetryTime` was increased to 1 minute. The parameter `randomize` was set to false, so now, the delay between attempts will always increase by a factor of 2, so we have more time between attempts than before.

### DIFF
--- a/packages/contentful/src/utils/poll-content.ts
+++ b/packages/contentful/src/utils/poll-content.ts
@@ -65,7 +65,12 @@ const pollContentful = async <T extends EntrySkeletonType<FieldsType, string>>(
         }
       }
     },
-    { minTimeout: 200, maxRetryTime: 8_000 },
+    {
+      minTimeout: 200,
+      maxRetryTime: 60_000,
+      factor: 2,
+      randomize: false,
+    },
   );
 
 export const pollContentfulDeliveryApi = async <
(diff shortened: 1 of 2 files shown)


AkosuaA (approved):
LGTM!

PR #4214, opened 2024-03-22
ASAP-379

gabiayako:


peterstarling (changes requested):
I'm not sure all this casting is a way to achieve type safety here. The issue we had was that the `record` was not being recognised as a specific element from the union, instead we were relying on the value of `type` input argument to decide which entity type we are dealing with, but that is misleading (we could even receive 'user' and a completely different entity under `record` argument).

If you want to recognise the value of `record` correctly we should probably do something like this:
image

this way we are able to determine that the type of the record is `TeamListItemResponse`. What I think we should do is drop `type` entirely and try to rely on the properties of the `record` to recognise which type we are dealing with.

Although.. it is still tempting to just make sure that the `tags` property lives on that object and then pass that on to `getTagNames` without having to differentiate them (but thats exactly what got us to this place).

peterstarling (approved):

PR #4217, opened 2024-03-26
feat: ASAP-382 Leadership and membership count update

peterstarling:
- **update membership counting method**
- **count leaders as members**


apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -52,7 +62,17 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
             ),
             interestGroupPreviousMemberCount:
               (team.inactiveSince &&
-                team.linkedFrom?.interestGroupsCollection?.total) ||
+                getUniqueIdCount([
+                  ...(team.linkedFrom?.teamMembershipCollection?.items.flatMap(
+                    (item) =>
+                      item?.linkedFrom?.usersCollection?.items
+                        .filter(filterOutAlumni)
```
gabiayako, line 69: if the team is inactive, does it matter if the user is alumni or not?

gabiayako (approved):

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
peterstarling: I don't think that part of the expression is being evaluated in that case, when the team is inactive it should immediately break out of the expression at line 64

gabiayako: Would it? if the team is inactive, `team.inactiveSince` is truthy and it would continue, right?
Screenshot 2024-03-28 at 07 13 29

peterstarling: Sorry you're right I misunderstood. This is actually a bug and should be fixed in the follow up ticket ASAP-359

PR #4221, opened 2024-03-29
feat: ASAP-394 Inactive interest groups

peterstarling:
- **add IG leadership inactive check**
- **add IG memebership inactive check**


apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -32,76 +34,94 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
-                    (interestGroup) => interestGroup?.sys.id,
+                    (teamMembershipItem) =>
+                      teamMembershipItem?.linkedFrom?.usersCollection?.items
+                        .flatMap(flattenInterestGroupLeaders)
+                        .filter(
+                          (item) =>
+                            item.interestGroupActive &&
+                            item.userIsAlumni === false,
+                        )
+                        .map((item) => item.interestGroupId),
```
gabiayako, line 59: This is the same as `interestGroupLeadershipRoleCount`, right? maybe it's better to extract it because if we need to change the way it is calculated we don't need to remember to change it in two separate places

packages/contentful/src/crn/queries/analytics.queries.ts
gabiayako, line 1: I believe there's some things missing in this query. We have some possibilities:

- The user can be inactive
- The interest group can be inactive
- The team can be inactive
- The user can be inactive in the interest group
- The user can be inactive in the team


So we're not checking if the user is inactive in the interest group
Screenshot 2024-03-29 at 06 47 42


```
linkedFrom {
  interestGroupLeadersCollection(limit: 3) {
    items {
      inactiveSinceDate        <-----------
      linkedFrom {
        interestGroupsCollection(limit: 1) {
          items {
            sys {
              id
            }
            active
          }
        }
      }
    }
  }
}
```

and we're not checking if the user is inactive in the team

Screenshot 2024-03-29 at 06 55 06

```
teamMembershipCollection(limit: 20) {
  inactiveSinceDate     <------
  items {
    linkedFrom {
      usersCollection(limit: 1) {
        ...
      }
    }
  }
}
```

gabiayako (changes requested):

packages/contentful/src/crn/queries/analytics.queries.ts
peterstarling: I think this would be another req and another PR but agreed

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
peterstarling: do you know what I think this is still such a small duplication that given the amount of requirements that we still have to implement I would hold off with such tiny refactoring, what do you think?

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -32,76 +34,94 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
             displayName: team.displayName || '',
             interestGroupLeadershipRoleCount: getUniqueIdCount(
```
gabiayako, line 35: I think it's missing the `!team.inactiveSince` here like there is for `interestGroupMemberCount`

apps/crn-server/test/data-providers/contentful/analytics.data-provider.test.ts
```diff
@@ -786,7 +1021,7 @@ describe('Analytics Data Provider', () => {
           expect(result.items[0]!.interestGroupPreviousMemberCount).toBe(3);
         });
 
-        test('Should return interestGroupPreviousMemberCount of 2 when the client returns a total of 2 interest groups associated with the team and one leader of the same interest group and the team is inactive', async () => {
+        test('Should return interestGroupPreviousMemberCount of 2 when the client returns a total of 2 activ interest groups associated with the team and one leader of the same interest group and the team is inactive', async () => {
```
gabiayako, line 1024:
```suggestion
        test('Should return interestGroupPreviousMemberCount of 2 when the client returns a total of 2 active interest groups associated with the team and one leader of the same interest group and the team is inactive', async () => {
```

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -32,76 +34,94 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
+                    .flatMap((interestGroup) => interestGroup?.sys.id) || []),
                 ])) ||
               0,
             interestGroupPreviousLeadershipRoleCount: getUniqueIdCount(
               team.linkedFrom?.teamMembershipCollection?.items.flatMap(
-                (item) =>
-                  item?.linkedFrom?.usersCollection?.items
-                    .filter(filterAlumni)
-                    .flatMap(flattenInterestGroupLeaders),
+                (teamMembershipItem) =>
```
gabiayako, line 68: I believe the `!team.inactiveSince` verification is also missing here. I'll point it out in some of the tests

apps/crn-server/test/data-providers/contentful/analytics.data-provider.test.ts
```diff
@@ -176,14 +176,15 @@ describe('Analytics Data Provider', () => {
           expect(result.items[0]!.interestGroupLeadershipRoleCount).toBe(0);
         });
 
-        test('Should return interestGroupLeadershipRoleCount of 2 when two users are leaders of two different interest groups', async () => {
+        test('Should return interestGroupLeadershipRoleCount of 2 when two users are leaders of two different active interest groups', async () => {
```
gabiayako, line 179: If the team is inactive (when we add `inactiveSince` in the fixture) like the screenshot below

Screenshot 2024-03-29 at 12 54 17

This test still passes although it should be 0

apps/crn-server/test/data-providers/contentful/analytics.data-provider.test.ts
```diff
@@ -510,11 +569,112 @@ describe('Analytics Data Provider', () => {
...
+          );
+
+          const result = await analyticsDataProvider.fetchTeamLeadership({});
+
+          expect(
+            result.items[0]!.interestGroupPreviousLeadershipRoleCount,
+          ).toBe(2);
+        });
+
+        test('Should return interestGroupPreviousLeadershipRoleCount of 1 when 2 active users are leaders of 2 different interest groups one of which is inactive', async () => {
```
gabiayako, line 608: The same thing here, if the team is inactive (if we add `inactiveSince` in the fixture), the test still passes

Screenshot 2024-03-29 at 12 54 17

but the `interestGroupPreviousLeadershipRoleCount` should return 2 in this case.

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
gabiayako: ok :)

gabiayako (changes requested):

gabiayako (approved):

PR #4225, opened 2024-04-03
feat: ASAP-395 Complete Working Groups 

peterstarling:
- **check for incomplete working groups**
- **extract reusable logic**


apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -24,6 +25,87 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
+
+    const getPreviousWorkingGroupIdsFromTeamLeaders = (team: Team): string[] =>
+      team.linkedFrom?.teamMembershipCollection?.items.flatMap(
+        (teamMembershipItem) =>
+          teamMembershipItem?.linkedFrom?.usersCollection?.items
+            .flatMap(flattenWorkingGroupLeaders)
+            .filter((item) => item.userIsAlumni || item.workingGroupComplete)
+            .map((item) => item.workingGroupId) || [],
+      ) || [];
+
```
gabiayako, line 99: What if we use a reduce to get the current and previous in one loop something like 


```
type EntityIdsByStatus = { current: string[]; previous: string[] };
const emptyEntityIdsByStatus: EntityIdsByStatus = {
  current: [],
  previous: [],
};

const getWorkingGroupIdsLeadership = (team: Team): EntityIdsByStatus =>
  team.linkedFrom?.teamMembershipCollection?.items.reduce(
    ({ current, previous }: EntityIdsByStatus, teamMembershipItem) => {
      teamMembershipItem?.linkedFrom?.usersCollection?.items
        .flatMap(flattenWorkingGroupLeaders)
        .forEach((item) => {
          if (
            item.userIsAlumni ||
            item.workingGroupComplete ||
            team.inactiveSince
          ) {
            previous.push(item.workingGroupId);
          } else {
            current.push(item.workingGroupId);
          }
        });
      return { current, previous };
    },
    emptyEntityIdsByStatus,
  ) || emptyEntityIdsByStatus;

(...)

    return {
      total: teamsCollection?.total || 0,
      items:
        teamsCollection?.items
          .filter((team): team is Team => team !== null)
          .map((team) => {
            const workingGroupIdsLeadership =
              getWorkingGroupIdsLeadership(team);

            return {
              id: team.sys.id,
              (...)
              workingGroupLeadershipRoleCount: team.inactiveSince
                ? 0
                : getUniqueIdCount(workingGroupIdsLeadership.current),
              workingGroupPreviousLeadershipRoleCount: getUniqueIdCount(
                workingGroupIdsLeadership.previous,
              ),
}
```

This way we go through this loop once instead of 3 times. I was wondering if we could do the same for working group members and interest group leaders as well. But it's just a suggestion.

apps/crn-server/test/data-providers/contentful/analytics.data-provider.test.ts
```diff
@@ -1248,6 +1248,19 @@ describe('Analytics Data Provider', () => {
         });
 
         test('Should return workingGroupLeadershipRoleCount of 0 when 1 user is a leader of a working group but the team is inactive', async () => {
```
gabiayako, line 1250:
```suggestion
        test('Should return workingGroupLeadershipRoleCount of 0 when 1 user is a leader of a working group but the working group is complete', async () => {
```

apps/crn-server/test/data-providers/contentful/analytics.data-provider.test.ts
```diff
@@ -1958,6 +2079,19 @@ describe('Analytics Data Provider', () => {
           expect(result.items[0]!.workingGroupPreviousMemberCount).toBe(1);
         });
 
+        test('Should return workingGroupPreviousMemberCount of 1 when the member of a working groupis not an alumni but the working group is complete', async () => {
```
gabiayako, line 2082:
```suggestion
        test('Should return workingGroupPreviousMemberCount of 1 when the member of a working groups not an alumni but the working group is complete', async () => {
```

gabiayako (approved):

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -35,131 +117,48 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
                 ),
             workingGroupPreviousLeadershipRoleCount: getUniqueIdCount([
               ...((team.inactiveSince &&
-                team.linkedFrom?.teamMembershipCollection?.items.flatMap(
-                  (teamMembershipItem) =>
-                    teamMembershipItem?.linkedFrom?.usersCollection?.items
-                      .filter(filterOutAlumni)
-                      .flatMap(flattenWorkingGroupLeaders),
-                )) ||
+                getCurrentWorkingGroupIdsFromTeamMembers(team)) ||
```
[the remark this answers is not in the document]
peterstarling: you would think that wouldn't you - I tried removing it but the issue is that if you use the `&&` expression then if its falsy it returns the value of the first element which is then spread in the array, and since the value of the `inactiveSince` is not iterable it throws an error, even if we do `!!inactiveSince` that's still the same case

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -35,131 +117,48 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
-                    teamMembershipItem?.linkedFrom?.usersCollection?.items
-                      .flatMap(flattenInterestGroupLeaders)
-                      .filter(
-                        (item) =>
-                          item.interestGroupActive &&
-                          item.userIsAlumni === false,
-                      )
-                      .map((item) => item.interestGroupId),
-                )) ||
+                getCurrentInterestGroupIdsFromTeamLeaders(team)) ||
```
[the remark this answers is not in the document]
peterstarling: same as above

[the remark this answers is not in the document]
peterstarling: you would think that wouldn't you - I tried removing it but the issue is that if you use the && expression then if its falsy it returns the value of the first element which is then spread in the array, and since the value of the inactiveSince is not iterable it throws an error, even if we do !!inactiveSince that's still the same case

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -24,144 +25,153 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
+              interestGroupPreviousMemberCount: getUniqueIdCount([
+                ...previousInterestGroupIdsFromTeamLeaders,
+                ...previousInterestGroupIdsFromInterestGroupCollection,
+              ]),
+              workingGroupLeadershipRoleCount: team.inactiveSince
+                ? 0
+                : getUniqueIdCount(currentWorkingGroupIdsFromTeamLeaders),
+              workingGroupPreviousLeadershipRoleCount: getUniqueIdCount([
+                ...((team.inactiveSince &&
+                  currentWorkingGroupIdsFromTeamMembers) ||
```
gabiayako, line 159: isn't it
```suggestion
                  currentWorkingGroupIdsFromTeamLeaders) ||
```

gabiayako (approved):

PR #4239, opened 2024-04-10
ASAP-400 (GP2) Make it possible to save user without orcid

gabiayako:
This bug was introduced in this PR https://github.com/yldio/asap-hub/pull/4160. 

The issue was that before the bug was introduced we were not sending `orcid` if its value is not truthy. 
-before

After the bug was introduced, we were always sending `orcid` and when it doesn't have a value, it's initialized as `''` so it was giving a 400 error in the BE.
-after

BE error in gp2.dev
Screenshot 2024-04-10 at 12 00 59

So this PR makes the `orcid` field to be sent only when it's value is truthy as it was before the bug was introduced.

### DIFF
--- a/packages/gp2-components/src/organisms/__tests__/KeyInformationModal.test.tsx
+++ b/packages/gp2-components/src/organisms/__tests__/KeyInformationModal.test.tsx
@@ -111,6 +111,45 @@ describe('KeyInformationModal', () => {
     expect(screen.getByText(/Please add your city/i)).toBeVisible();
   });
 
+  it('does not call onSave with orcid when orcid is not provided', async () => {
+    const onSave = jest.fn();
+    renderKeyInformation({
+      onSave,
+      social: {
+        ...defaultProps.social,
+        orcid: undefined,
+      },
+    });
+    const saveButton = getSaveButton();
+    userEvent.click(saveButton);
+    const {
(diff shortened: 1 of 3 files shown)


packages/gp2-components/src/organisms/__tests__/KeyInformationModal.test.tsx
```diff
@@ -111,6 +111,45 @@ describe('KeyInformationModal', () => {
     expect(screen.getByText(/Please add your city/i)).toBeVisible();
   });
 
+  it('does not call onSave with orcid when orcid is not provided', async () => {
```
peterstarling, line 114: will that not be a problem if someone has an orcid but then wants to remove it?

peterstarling (changes requested):

packages/gp2-components/src/organisms/__tests__/KeyInformationModal.test.tsx
gabiayako: I've tested it here and it still works because the data provider deals with it here https://github.com/yldio/asap-hub/blob/master/apps/gp2-server/src/data-providers/user.data-provider.ts#L225:L242

peterstarling (approved):

PR #4240, opened 2024-04-10
[ASAP-411] - playwright-e2e-test-issues

AkosuaA:
Although I'm not sure what changed exactly, the issue seems to be compatibility with the container's Node version. The Playwright [documentation](https://playwright.dev/docs/ci-intro#via-containers) indicates setting up node 18, which seems to solve the issue
Screenshot 2024-04-11 at 13 46 36

I left the playwright version update as well since it doesn't seem to break anything.


peterstarling (approved):
clearly this fixes it but using node 20 should not cause errors (obviously), as I mentioned to you could you raise this issue with playwright?

PR #4242, opened 2024-04-12
ASAP-407 (GP2) Add short description field

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-407

--- 

**Notes:**
- It's not necessary to add short description to the output detail view
- The link mentioned in the third requirement was removed since we don't have the correct link yet.
(source: https://yld.slack.com/archives/CV4UY4411/p1712937377468709)

### DIFF
--- a/apps/gp2-server/src/validation/output.validation.ts
+++ b/apps/gp2-server/src/validation/output.validation.ts
@@ -103,6 +103,7 @@ const outputPostRequestValidationSchema: JSONSchemaType<gp2Model.OutputPostReque
       shortDescription: {
         type: 'string',
         nullable: true,
+        maxLength: 250,
       },
       gp2Supported: { type: 'string', nullable: true },
       sharingStatus: { type: 'string' },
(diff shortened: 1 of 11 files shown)


apps/gp2-server/src/validation/output.validation.ts
```diff
@@ -100,6 +100,10 @@ const outputPostRequestValidationSchema: JSONSchemaType<gp2Model.OutputPostReque
         type: 'string',
         nullable: true,
       },
+      shortDescription: {
+        type: 'string',
+        nullable: true,
```
peterstarling, line 105: what about maximum length?

peterstarling (changes requested):

apps/gp2-server/src/validation/output.validation.ts
gabiayako: nice catch, I've added it here [add missing maxLength validation](https://github.com/yldio/asap-hub/pull/4242/commits/0603e2d1f314d630fb06b45c544ef58193e2426d)

peterstarling (approved):

PR #4245, opened 2024-04-16
Asap 374 - Add Filtering for User/Team Productivity Analytics

lctrt:
image

### DIFF
--- a/packages/react-components/src/molecules/AnalyticsControls.tsx
+++ b/packages/react-components/src/molecules/AnalyticsControls.tsx
@@ -1,0 +1,70 @@
+import { css } from '@emotion/react';
+import { dropdownChevronIcon } from '../icons';
+import DropdownButton from './DropdownButton';
+import { rem, tabletScreen } from '../pixels';
+
+const containerStyles = css({
+  display: 'flex',
+  alignItems: 'center',
+  gap: rem(15),
+  'span + svg': {
+    width: '11px',
+    marginRight: '10px',
+  },
+  button: {
+    padding: `${rem(8)} ${rem(16)}`,
+  },
+
+  [`@media (max-width: ${tabletScreen.min}px)`]: {
+    flexDirection: 'column',
+    alignItems: 'flex-start',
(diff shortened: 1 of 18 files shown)


packages/react-components/src/molecules/AnalyticsControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <span css={{ marginRight: rem(10) }}>
+            {timeRangeOptions[timeRange]}
+          </span>
+          {dropdownChevronIcon}
+        </>
+      )}
+    >
+      {Object.keys(timeRangeOptions).map((key) => ({
+        item: <>{timeRangeOptions[key as TimeRangeOption]}</>,
+        href: `${href}?range=${key}`,
```
gabiayako, line 64: I've noticed that if I'm at a specific page then I apply a filter, the pagination gets lost. I think maybe we should keep the pagination, but I'm not sure this is a requirement, maybe we could ask João or Tiff about that?

packages/react-components/src/molecules/AnalyticsControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+    width: '100%',
+  },
+});
+
+export type TimeRangeOption =
+  | '30d'
+  | '90d'
+  | 'current-year'
+  | 'last-year'
+  | 'all';
```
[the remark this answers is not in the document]
lctrt: Ah yeah forgot that one, built FE first :)

packages/react-components/src/molecules/AnalyticsControls.tsx
lctrt: At the moment it would make sense yeah, but as soon as we add custom ordering selection it's gonna break down. What does page 5 means if you order by amount of outputs? 

I can see how useful it could be in the current state since it allows to get to the user we want to see, but as soon as filtering by name is added it won't be useful anymore.

gabiayako (approved):

PR #4247, opened 2024-04-18
ASAP-362 Re-enabling integration tests

gabiayako:
This PR re-enables integration tests and adds a log to errorHandler so we can see what failed in integration tests instead of having just the generic "Internal Server Error"


apps/crn-server/test/data-providers/contentful/event.data-provider.test.ts
```diff
@@ -1033,7 +1033,7 @@ describe('Events Contentful Data Provider', () => {
           status: {
             'en-US': 'Confirmed',
           },
-          tags: {
+          researchTags: {
```
peterstarling, line 1036: is there a type for all of this input? if yes then you should consider adding `satisfies ....` at the end of this embedded object fixture

apps/crn-server/test/fixtures/events.fixtures.ts
```diff
@@ -513,7 +513,7 @@ export const getEventCreateDataObject = (): EventCreateDataObject => ({
   endDateTimeZone: 'Europe/Lisbon',
   calendar: 'calendar-id',
   status: 'Confirmed' as EventStatus,
-  tags: [],
+  researchTags: [],
```
peterstarling, line 516: do we still accept `tags` in our `EventCreateDataObject`?

peterstarling (approved):

PR #4248, opened 2024-04-19
feat: ASAP-388 Amends to analytics

peterstarling:
This PR accomplishes the following:
1. Removes PMs from leadership (or membership) - only Chairs are accounted for
2. Removes previous membership/leadership for when there is an existing current membership/leadership for the same IG/WG - this means that if there is an active and past leadership or membership for a group then it's counted as current only.


apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
```diff
@@ -207,17 +210,29 @@ export class AnalyticsContentfulDataProvider implements AnalyticsDataProvider {
...
                 ...((team.inactiveSince &&
                   currentWorkingGroupIdsFromTeamMembers) ||
                   []),
-                ...previousWorkingGroupIdsFromTeamMembers,
+                ...removeDuplicates(
+                  [
+                    ...previousWorkingGroupIdsFromTeamMembers,
+                    ...previousWorkingGroupIdsFromTeamLeaders,
+                  ],
+                  currentWorkingGroupIdsFromTeamMembers,
```
gabiayako, line 234: I was trying to think about an example and I came up with this

working-groups

In this case above, the `workingGroupPreviousMemberCount` should be 0 because there isn't a working group where all the users associated with this team are past members, right? But it's giving 1

lctrt (approved):

PR #4249, opened 2024-04-22
ASAP-432 Fix analytics published at reference date

gabiayako:
It seems we have migrated also the addedDate from squidex. References: 
- Query https://github.com/yldio/asap-hub/blob/09952308b9d994137a586b79ddca9131b77a7004/packages/cms-data-sync/src/research-outputs/research-outputs.queries.ts#L18
- Migration (addedDate is among fields) https://github.com/yldio/asap-hub/blob/09952308b9d994137a586b79ddca9131b77a7004/packages/cms-data-sync/src/research-outputs/research-outputs.data-migration.ts#L106

So for outputs that were created in Squidex we have the addedDate as the day the output was published.
Screenshot 2024-04-19 at 14 14 32

For drafts we currently show as “Date added” the created date https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/organisms/SharedResearchCard.tsx#L133

Screenshot 2024-04-19 at 14 11 18

And the added date is only populated when the draft is published. Reference: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/controllers/research-output.controller.ts#L193

Description

So we can use added Date as published date for all cases (outputs created before the migration, outputs that began as published and outputs that began as draft).


packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -146,11 +145,10 @@ export const FETCH_TEAM_PRODUCTIVITY = gql`
         displayName
         inactiveSince
         linkedFrom {
-          researchOutputsCollection(limit: 100) {
+          researchOutputsCollection(limit: 600) {
```
gabiayako, line 148: These are the teams that have most research outputs in prod
Screenshot 2024-04-23 at 08 24 55

The max currently is 470, but I've made the limit to be 600 randomly.
Screenshot 2024-04-23 at 08 25 02

packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -146,11 +145,10 @@ export const FETCH_TEAM_PRODUCTIVITY = gql`
         displayName
         inactiveSince
         linkedFrom {
-          researchOutputsCollection(limit: 100) {
+          researchOutputsCollection(limit: 1000) {
```
gabiayako, line 148: These are the teams that have most research outputs in prod
Screenshot 2024-04-23 at 08 24 55

The max currently is 470, but I've made the limit to be 1000 randomly. Not sure if it needs to be bigger.
Screenshot 2024-04-23 at 08 25 02

peterstarling (approved):

PR #4251, opened 2024-04-24
feat: ASAP-413 Add GP2 output public endpoint

peterstarling:
* test old app for cors
* refactor logger to cater for more basic usage 
* add more filtering capabilities to output controller and data provider
* add public app and basic output endpoint


gabiayako (changes requested):

apps/gp2-server/src/data-providers/output.data-provider.ts
```diff
@@ -433,6 +433,8 @@ export const parseContentfulGraphQLOutput = (
       .map(parseTag) ?? [];
   return {
     id: data.sys.id,
+    systemPublishedVersion: data.sys.publishedVersion ?? undefined,
```
gabiayako, line 436: oh I thought we would just name the variable `systemPublishedVersion` not assign it to contentful publishedVersion 😛 

I wonder if by version they meant the like `versionsCollection.total` + 1. For example, if `versionsCollection` is empty it means it is the first version

gabiayako (approved):

apps/gp2-server/src/data-providers/output.data-provider.ts
peterstarling: nah, they didn't ask for this field at all nor are they aware of our internal versioning mechanisms (i think). They asked for the publish date to be able to poll and I suggested we return our system publish version for them to track if they have the most recent resource version just in case

gabiayako: ah ok! nice!

PR #4252, opened 2024-04-25
ASAP-447 Create L&M replica indexes

gabiayako:
This PR changes the algolia `set-settings` to create all the replica indexes necessary to sort the Leadership & Membership data. 

Screenshot 2024-04-26 at 07 35 37 (2)

### DIFF
--- a/apps/asap-cli/src/cli.ts
+++ b/apps/asap-cli/src/cli.ts
@@ -17,124 +17,126 @@ import {
   setAlgoliaSettings,
 } from './scripts/algolia';
 
+const stringType = 'string' as 'string';
+const trueType = true as true;
+
+const appIdOption = {
+  alias: 'a',
+  type: stringType,
+  description: 'The App ID',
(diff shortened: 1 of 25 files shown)


apps/asap-cli/src/scripts/algolia/set-settings.ts
peterstarling, line 1: do you think we should continue to use a single script for both indexes and then have a massive if-else inside? maybe best to split these script and reuse bits? just an idea, im not not super fond of recognising analytics index by the presence of the word 'analytics'

peterstarling (commented):
do you think

apps/asap-cli/src/cli.ts
gabiayako, line 1: I've done a refactor because it was giving a copy paste error (https://github.com/yldio/asap-hub/actions/runs/8849569706/job/24301870376#step:3:5)

copy-paste

peterstarling (approved):

PR #4256, opened 2024-05-01
feat: ASAP-414 Add user public endpoint

peterstarling:
* add missing output fetch-by-id endpoint
* add outputs to GP2 user response
* add user public endpoints 
* changes 404 error messages from both app servers


apps/gp2-server/src/routes/public/output.route.ts
```diff
@@ -27,6 +27,17 @@ export const outputRouteFactory = (
     },
   );
 
+  outputRoutes.get(
+    '/outputs/:outputId',
+    async (req, res: Response<gp2Model.PublicOutputResponse>) => {
+      const { outputId } = req.params;
+
+      const output = await outputController.fetchById(outputId);
```
lctrt, line 35: does this mean that if we know the id of a non public output we can access it with that api?

lctrt (approved):

apps/gp2-server/src/routes/public/output.route.ts
peterstarling: Good catch!! We shouldn't but yes it probably means that yeah. I'll work on the fix now

PR #4257, opened 2024-05-02
Asap 403

lctrt:


packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -157,3 +157,72 @@ export const FETCH_TEAM_PRODUCTIVITY = gql`
...
+              }
+              displayName
+              inactiveSince
+            }
+            role
+            inactiveSinceDate
+          }
+        }
+        linkedFrom {
+          researchOutputsCollection(limit: 20) {
```
gabiayako, line 194: I know there's another query where we fetch at most 20 research outputs, but I think we might need to make this number higher in both queries. I was taking a look at prod, it seems there are users with more than 100 ROs, these ones for example:

https://hub.asap.science/network/users/ca867aa8-d40a-4a79-9c84-90cf8a64f013/outputs

https://hub.asap.science/network/users/e8cace05-f2cf-4e46-9de7-a7c9a537b627/outputs

gabiayako (commented):
I'll submit some comments as I review :)

packages/contentful/src/crn/queries/analytics.queries.ts
lctrt: might be worth doing that work in a separate ticket for all analytics at once? This way we can focus on that problem separately and make sure we do it in a robust way.

peterstarling: yeah we can't leave it that way but this is behind a FF anyway so why don't you create a ticket for it and lets add it to that same epic - we will make sure that this feature doesn't get released until its addressed

apps/crn-server/src/utils/analytics/collaboration.ts
```diff
@@ -0,0 +1,145 @@
...
+export const getUserCoproductionItems = (
+  userCollection: FetchUserCoproductionQuery['usersCollection'],
+): UserCollaborationDataObject[] =>
+  cleanArray(userCollection?.items).map((user) => {
+    const teams = cleanArray(user?.teamsCollection?.items).map((team) => {
+      const analyticData =
+        user.linkedFrom?.researchOutputsCollection?.items.map((output) => {
+          const authorList = cleanArray(
+            cleanArray(output?.authorsCollection?.items).map((author) => {
+              if (author.__typename === 'Users') {
```
gabiayako, line 104: Not all research outputs linked to a user is linked because they are authors 😅 

This one for example, returns a total of 84 outputs 

Screenshot 2024-05-02 at 11 56 59

But in the hub https://hub.asap.science/network/users/1fhU4Tx6CedIzQX0js9ajD/research we see that there's 0 outputs

Screenshot 2024-05-02 at 12 00 18

she's linked to the outputs because she might be the one who updated it not necessarily the author. So it's necessary to check if the author is the user we are analyzing

apps/crn-server/src/utils/analytics/collaboration.ts
```diff
@@ -0,0 +1,145 @@
...
+  id: string;
+  teams: EntityWithId[];
+  labs: EntityWithId[];
+};
+
+export const checkDifferentTeams = (
+  referenceTeam: EntityWithId,
+  testTeams: EntityWithId[],
+): boolean =>
+  !testTeams.find((testTeam) => testTeam.sys.id === referenceTeam.sys.id);
```
gabiayako, line 21: small change, but I think it's more readable
```suggestion
  testTeams.every((testTeam) => testTeam.sys.id !== referenceTeam.sys.id)
```

apps/crn-server/test/utils/analytics/collaboration.test.ts
```diff
@@ -0,0 +1,219 @@
...
+  it('sums each collaboration count properly', () => {
+    const data = [flags(false, false), flags(true, false), flags(false, true)];
+    expect(getCollaborationCounts(data)).toEqual({
+      acrossTeamCount: 1,
+      withinTeamCount: 1,
+    });
+  });
+});
+
+describe('getUserCoproductionItems ', () => {
```
gabiayako, line 185: I think it would be nice to add the test cases from the JIRA tickets if possible. Here or in apps/crn-server/test/data-providers/contentful/analytics/coproduction.analytics.data-provider.test.ts :)

apps/crn-server/src/utils/analytics/collaboration.ts
```diff
@@ -0,0 +1,145 @@
...
+        ? withinTeamCount + 1
+        : withinTeamCount,
+    }),
+    {
+      acrossTeamCount: 0,
+      withinTeamCount: 0,
+    },
+  );
+
+export const getUserCoproductionItems = (
```
gabiayako, line 95:
```suggestion
export const getUserCoProductionItems = (
```

packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -157,3 +157,72 @@ export const FETCH_TEAM_PRODUCTIVITY = gql`
...
+            items {
+              addedDate
+              sharingStatus
+              authorsCollection(limit: 10) {
```
gabiayako, line 198: it's possible to add `where: { sys: { id_exists: true }}` so it returns only users not external users
```suggestion
              authorsCollection(limit: 10, where: { sys: { id_exists: true } }) {
```

apps/crn-server/src/utils/analytics/collaboration.ts
```diff
@@ -0,0 +1,160 @@
...
+        : withinTeamCount,
+    }),
+    {
+      acrossTeamCount: 0,
+      withinTeamCount: 0,
+    },
+  );
+
+export const getUserCoProductionItems = (
+  userCollection: FetchUserCoproductionQuery['usersCollection'],
```
gabiayako, line 108:
```suggestion
  userCollection: FetchUserCoProductionQuery['usersCollection'],
```

apps/crn-server/test/data-providers/contentful/analytics/coproduction.analytics.data-provider.test.ts
```diff
@@ -0,0 +1,86 @@
+import { FETCH_USER_COPRODUCTION } from '@asap-hub/contentful';
+import { AnalyticsContentfulDataProvider } from '../../../../src/data-providers/contentful/analytics.data-provider';
+import { getUserCoproductionQuery } from '../../../fixtures/analytics.fixtures';
+// import { getContentfulGraphqlUser } from "../../../fixtures/users.fixtures";
```
gabiayako, line 4:
```suggestion

```

apps/crn-server/test/data-providers/contentful/analytics/coproduction.analytics.data-provider.test.ts
gabiayako, line 1: I still think it would be nice to add the test cases from the JIRA tickets :)

lctrt: wording is different but I did :)

apps/crn-server/src/utils/analytics/collaboration.ts
lctrt: this is coming from autogen code, don't think I can rename it without making it worse. but I don't like the name to be honest, I think I'll rename to collaboration and make it consistent everywhere.

apps/crn-server/src/utils/analytics/collaboration.ts
```diff
@@ -0,0 +1,160 @@
...
+        differentTeamFlag = true;
+      }
+      if (
+        !sameTeamDifferentLabFlag &&
+        checkSameTeamDifferentLab(referenceTeam, referenceLabs, author)
+      ) {
+        sameTeamDifferentLabFlag = true;
+      }
+    } else if (author.id === referenceId) {
+      isAuthor = true;
```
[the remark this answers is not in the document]
lctrt: we still need the user to be part of the author list, this is used to check that the user is an actual author.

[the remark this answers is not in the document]
gabiayako: But if we add 

```
author.sys.id === user.sys.id
```

before the loop we still checking if the user is an author 🤔

packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -157,3 +157,75 @@ export const FETCH_TEAM_PRODUCTIVITY = gql`
...
+          }
+        }
+        linkedFrom {
+          researchOutputsCollection(limit: 20) {
+            items {
+              addedDate
+              sharingStatus
+              authorsCollection(
+                limit: 10
+                where: { sys: { id_exists: true } }
```
gabiayako, line 200: I know I suggested this, but I think we'll have to remove this

```
where: { sys: { id_exists: true } }
```
for now because it increased complexity 😢 
query-complexity

gabiayako (approved):

PR #4266, opened 2024-05-08
feat: ASAP-415 Add Working-Group Public API endpoint

peterstarling:
- **add new public endpoints**
- **add publish date and version to the responses**
## Notes
* we dont need pagination here really as the amount of working groups is so insignificant it was never implemented in our endpoints either - I think we can send it to GMMB as is but let me know what you think

### DIFF
--- a/apps/gp2-server/src/routes/public/working-group.route.ts
+++ b/apps/gp2-server/src/routes/public/working-group.route.ts
@@ -44,4 +44,6 @@ const mapWorkingGroupToPublicWorkingGroup = (
   title: workingGroup.title,
   primaryEmail: workingGroup.primaryEmail,
   secondaryEmail: workingGroup.secondaryEmail,
+  publishDate: workingGroup.publishDate,
+  systemPublishedVersion: workingGroup.systemPublishedVersion,
 });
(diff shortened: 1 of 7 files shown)


apps/gp2-server/src/routes/public/working-group.route.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  return workingGroupRoutes;
+};
+
+const mapWorkingGroupToPublicWorkingGroup = (
+  workingGroup: gp2Model.WorkingGroupResponse,
+): gp2Model.PublicWorkingGroupResponse => ({
+  id: workingGroup.id,
+  description: workingGroup.description,
+  members: workingGroup.members,
```
gabiayako, line 42: Hey, I think we need more data on members, don't we?

It's listed in ticket

- degree
- role
- dept
- institution
- city
- country

and in the website (https://gp2.org/working-group-directory/) passed as reference we have

Screenshot 2024-05-09 at 07 18 43

so these info seems to be used

gabiayako (approved):

apps/gp2-server/src/routes/public/working-group.route.ts
peterstarling: as per our discussion - we decided to leave these out and GMMB don't seem to need any extra information embedded for the nested resources

PR #4278, opened 2024-05-16
feat: ASAP-416 Add cache to Public API

peterstarling:


lctrt (approved):

AkosuaA (approved):
Looks good
Nit: Maybe a test for calling the controller with the same parameters after an hour

PR #4279, opened 2024-05-16
ASAP-470 Add teams reference and display manuscript title in the frontend

gabiayako:
This PR

- Adds the team reference to the `manuscripts` content model
Screenshot 2024-05-17 at 12 18 03

- Displays the manuscript in the frontend
Screenshot 2024-05-17 at 12 17 04


packages/contentful/src/crn/queries/manuscript.queries.ts
```diff
@@ -8,6 +8,13 @@ export const manuscriptContentQueryFragment = gql`
       id
     }
     title
+    teamsCollection(limit: 10) {
```
AkosuaA, line 11: Would we be having multiple teams assigned to a manuscript?

gabiayako: Yeah, we're going to have this field in the form in the future :)
Screenshot 2024-05-20 at 09 34 32

AkosuaA: Ah yes, I remember the discussion on this. Thanks! The teamId in the model confused me but it makes sense to add that when it becomes relevant. Looks good!

gabiayako: thanks 😄

PR #4286, opened 2024-05-23
ASAP-376 Process Productivity Metrics

gabiayako:


apps/asap-cli/src/scripts/algolia/tests/process-productivity-metrics.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    const numbers = [1, 2, 3, 4, 5];
+    const mean =
+      numbers.reduce((sum, value) => sum + value, 0) / numbers.length;
+    expect(getStandardDeviation(numbers, mean)).toBeCloseTo(1.414, 3);
+  });
+});
+
+describe('getBellCurveMetrics', () => {
+  it('should return correct bell curve metrics', () => {
+    const data = [1, 2, 3, 4, 5];
```
peterstarling, line 29: would be nice to have a data set where below and above average max and min define a range, not just one number

apps/asap-cli/src/scripts/algolia/tests/process-productivity-metrics.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    await deletePreviousObjects(mockIndex, 'user-productivity');
+    expect(mockIndex.search).toHaveBeenCalledWith('', {
+      filters: '__meta.type:"user-productivity-metrics"',
+    });
+    expect(mockIndex.deleteObjects).toHaveBeenCalledWith(['1', '2']);
+  });
+});
+
+describe('processUserProductivityMetrics', () => {
+  it('should process user productivity metrics', async () => {
```
peterstarling, line 80: what do you mean exactly, that it should save new records yes? I think you are also deleting before saving and this is not asserted so it's potentially open to regression

apps/asap-cli/src/scripts/algolia/tests/process-productivity-metrics.test.ts
peterstarling, line 1: there is a lot going on in all these exported functions, are you sure there aren't any tests missing?

peterstarling (changes requested):
the approach is correct, I guess it's probably the best way to process metrics from algolia once they are saved in the previous step. 

The coverage seems a little thin but maybe it's because with previous metrics there were so many edge cases and variations and here it's pretty straight forward, it either works or it doesn't? I would make sure you cover all the different cases though

apps/asap-cli/src/scripts/algolia/process-productivity-metrics.ts
```diff
@@ -0,0 +1,230 @@
...
+      },
+      {} as Record<string, PerformanceMetrics>,
+    );
+
+    await index.saveObject(
+      {
+        ...userMetrics,
+        __meta: {
+          range,
+          type: `${type}-metrics`,
```
peterstarling, line 155: yeah... are you sure about that `-metrics` suffix? It took a me a while to understand what you were saving here. Maybe this ball curve thing is a better way as thats essentially what these records are, whereas technically speaking everything we have in Algolia for the analytics is a metric, no?

peterstarling (changes requested):

.github/workflows/reusable-crn-analytics-algolia-sync.yml
```diff
@@ -118,3 +118,11 @@ jobs:
           ALGOLIA_APP_ID: ${{ steps.algolia-keys.outputs.algolia-app-id }}
           ALGOLIA_INDEX: ${{ steps.setup.outputs.crn-analytics-algolia-index }}
           ALGOLIA_INDEX_TEMP: '${{ steps.setup.outputs.crn-analytics-algolia-index }}_${{ github.run_id }}_temp'
+      - name: Process Productivity Performance
+        if: ${{ inputs.metric == 'all' || inputs.metric == 'team-productivity' || inputs.metric == 'user-productivity' }}
```
gabiayako, line 122: When running only for team-leadership https://github.com/yldio/asap-hub/actions/runs/9224439267/job/25379705095#step:13:11

Screenshot 2024-05-24 at 10 12 03

When running only for team-productivity https://github.com/yldio/asap-hub/actions/runs/9224927942/job/25381269805#step:15:8

Screenshot 2024-05-24 at 10 55 26

peterstarling (approved):

PR #4289, opened 2024-05-29
Asap 469

lctrt:


packages/react-components/src/templates/AnalyticsLeadershipPageBody.tsx
```diff
@@ -119,11 +127,18 @@ const LeadershipPageBody: React.FC<LeadershipAndMembershipAnalyticsProps> = ({
     </div>
     <div css={searchContainerStyles}>
       <Subtitle>Teams:</Subtitle>
-      <span css={searchStyles}>
-        <SearchField
-          value={searchQuery}
+      <span role="search" css={searchStyles}>
+        <MultiSelect
```
gabiayako, line 131: There's a 15px marginTop in this component that makes "Teams" to seems unaligned. Maybe we could add noMargin props as there is in another components :)
Screenshot 2024-05-29 at 09 24 51

gabiayako (approved):
Looks good! ✅

PR #4290, opened 2024-05-29
[ASAP-373]- analytics productivity sort

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-373
This pr covers points 1 & 2 in the requirements (Display sort icons on both team and user productivity pages as per designs & Enable sort by the selected column)

Indices have been added for some of the user productivity & team productivity columns.

On the Team & Role columns (User Productivity table), we have the 'Multiple Teams'/'Multiple Roles'  & 'No team'/'No role' values which we don't store in Algolia. It makes the results a bit weird as the actual team name or lack of is used for sorting and not what is seen in the table (see images below). I've updated the export script to add team & role fields with similar data as what we display in the table and use them to sort those columns.
Screenshot 2024-05-30 at 12 21 20

Screenshot 2024-05-30 at 12 22 00

The ascending & descending icons for the numerical columns are different from the designs. I'll confirm with João when he's back but I assume the icons in the design are outdated.


gabiayako (approved):
nice! 🎉

PR #4297, opened 2024-06-04
feat: ASAP-475 Add missing fields to GP2 Public Api

peterstarling:
- **add system publish version to users**
- **add title and institution to users**
- **add last modified date to users**
- **add last modified date to wgs**
- **add last modified date to outputs**
- **add final publish date to outputs**
- **add preprint publish date to outputs**

### DIFF
--- a/apps/gp2-server/src/routes/public/user.route.ts
+++ b/apps/gp2-server/src/routes/public/user.route.ts
@@ -48,6 +48,7 @@ const mapUserToPublicUser = (
   firstName: user.firstName,
   id: user.id,
   lastName: user.lastName,
+  lastModifiedDate: user.lastModifiedDate,
   middleName: user.middleName,
   institution: user.positions[0]?.institution,
   title: user.positions[0]?.role,
--- a/apps/gp2-server/src/routes/public/user.route.ts
+++ b/apps/gp2-server/src/routes/public/user.route.ts
@@ -39,16 +39,15 @@ export const userRouteFactory = (userController: UserController): Router => {
 const mapUserToPublicUser = (
   user: gp2Model.UserResponse,
 ): gp2Model.PublicUserResponse => ({
-  id: user.id,
-  displayName: user.displayName,
-  country: user.country,
-  degrees: user.degrees,
-  firstName: user.firstName,
-  lastName: user.lastName,
-  publishDate: user.createdDate,
   avatarUrl: user.avatarUrl,
   biography: user.biography,
   city: user.city,
+  country: user.country,
+  degrees: user.degrees,
+  displayName: user.displayName,
+  firstName: user.firstName,
+  id: user.id,
+  lastName: user.lastName,
   middleName: user.middleName,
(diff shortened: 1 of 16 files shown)


apps/gp2-server/src/routes/public/user.route.ts
```diff
@@ -39,24 +39,28 @@ export const userRouteFactory = (userController: UserController): Router => {
...
   firstName: user.firstName,
+  id: user.id,
   lastName: user.lastName,
-  publishDate: user.createdDate,
-  avatarUrl: user.avatarUrl,
-  biography: user.biography,
-  city: user.city,
+  lastModifiedDate: user.lastModifiedDate,
   middleName: user.middleName,
+  institution: user.positions[0]?.institution,
```
gabiayako, line 53: Looking at the ticket, it seems we should display all institutions, right?
Screenshot 2024-06-04 at 07 37 30

apps/gp2-server/src/routes/public/user.route.ts
```diff
@@ -39,24 +39,28 @@ export const userRouteFactory = (userController: UserController): Router => {
...
+  title: user.positions[0]?.role,
```
gabiayako, line 54: And title is confusing because in one part of the ticket says it should be the role in CMS and in other it says it should be Degree, but we're not returning either 😅 
 
Screenshot 2024-06-04 at 07 38 27


Screenshot 2024-06-04 at 07 38 33

apps/gp2-server/src/routes/public/user.route.ts
```diff
@@ -39,24 +39,28 @@ export const userRouteFactory = (userController: UserController): Router => {
...
   degrees: user.degrees,
+  displayName: user.displayName,
   firstName: user.firstName,
+  id: user.id,
   lastName: user.lastName,
-  publishDate: user.createdDate,
-  avatarUrl: user.avatarUrl,
-  biography: user.biography,
-  city: user.city,
+  lastModifiedDate: user.lastModifiedDate,
```
gabiayako, line 51: Shouldn't it be publishDate?

Screenshot 2024-06-04 at 07 42 52

gabiayako (changes requested):

apps/gp2-server/src/routes/public/user.route.ts
peterstarling: it is https://github.com/yldio/asap-hub/blob/master/apps/gp2-server/src/data-providers/user.data-provider.ts#L310 - in the case of users the last modified date was already present so I figured this would be easier than copying the publish date again

apps/gp2-server/src/routes/public/user.route.ts
peterstarling: yeah this is strange cause degrees are actually already present. I added a title and referenced a role there, I think I am just going to leave it and if they decide it was the degree they were after then they can just do that.

gabiayako (approved):

PR #4302, opened 2024-06-14
fix: Short Description empty validation error when generating with AI

peterstarling:
update role-modal test

force the blur to fix short desc empty error msg bug

add cleanup to shareoutput


gabiayako:
Notes: Dependent on react update

PR #4310, opened 2024-06-24
feat: ASAP-538 Add output type filter to Team Productivity Analytics page

peterstarling:
- **fix the config and add missing env vars for vite**
- **add type selection to team productivity**


apps/crn-server/scripts/export-analytics.ts
```diff
@@ -155,6 +169,7 @@ const transformRecords = (
       type,
```
AkosuaA, line 169: The payload's objectID should have the outputType included since objectIDs are unique

AkosuaA (changes requested):

packages/react-components/src/molecules/AnalyticsControls.tsx
```diff
@@ -123,9 +132,12 @@ const generateLink = (
...
   documentCategory?: string,
+  outputType?: string,
 ) =>
   `${href}?range=${timeRange}${
     documentCategory ? `&documentCategory=${documentCategory}` : ''
+  }${
+    outputType ? `&outputType=${outputType}` : ''
   }&currentPage=${currentPage}${tagsQueryString}`;
 
 const updateSearchParams = (): URLSearchParams => {
```
AkosuaA, line 143: here the outputType searchParam should be deleted otherwise switching between output types cause multiple params in the url
Screenshot 2024-06-27 at 12 11 03

AkosuaA (approved):

PR #4314, opened 2024-06-25
[ASAP-522] - add document category filter to user productivity table

AkosuaA:
- updates process user productivity performance to calculate performance metrics per range & documentCategory combination.

- updates algolia schema sync to forward `attributesForFaceting` setting updates to replicas. Without this, existing replicas' `attributesForFaceting` are not updated. eg: we start filtering by 'documentCategory' on the main index schema. The sorting replicas need to be configured to allow filtering by 'documentCategory' as well.

- FE & BE updates to allow filtering by document category


apps/crn-server/scripts/export-analytics.ts
```diff
@@ -32,27 +33,44 @@ export const exportAnalyticsData = async (
...
+      for (let j = 0; j < documentCategories.length; j += 1) {
+        await exportData(metric, file, {
+          timeRange: timeRanges[i],
+          documentCategory: documentCategories[j],
+        });
+        if (j != documentCategories.length - 1) {
+          await file.write(',');
+        }
+      }
+      if (i != timeRanges.length - 1) {
```
peterstarling, line 56: looks to me like you could get double `,` if both `j` and `i` reach the lengths of their corresponding arrays, not that it should matter (it would just be seen as an empty row but im sure algolia wouldn't index an empty row)

This is quite a complicated logic, would be nice to test this file at some point (not now)

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
peterstarling, line 1: it's great that the new filter is being tested in isolation as part of the `common..test.ts` test suite but what I have an issue with is that this isn't being tested as part of the data-provider tests, so technically we wouldn't know what the desired filtering behaviour is or even if these filters are being applied at all. I'm pretty sure if you remove the filter lines it will still all be green. 

I have added some tests to address that here:
https://github.com/yldio/asap-hub/blob/8f7210b80d3918adbdd7deb971ffe677c470ad7f/apps/crn-server/test/data-providers/contentful/analytics/productivity.analytics.data-provider.test.ts#L408

TECHNICALLY we could mock out the filtering functions and just say that these filters are being applied but I think that's silly, so imho it's best to test the whole filtering logic at that level again.

peterstarling (approved):

packages/model/src/analytics.ts
```diff
@@ -76,10 +76,25 @@ export const timeRanges = [
...
+
 export type TimeRangeOption = (typeof timeRanges)[number];
+export type DocumentCategoryOption = (typeof documentCategories)[number];
+
+export type FilterAnalyticsOptions = {
+  timeRange?: TimeRangeOption;
+  documentCategory?: DocumentCategoryOption;
+};
 
 export type FetchAnalyticsOptions = FetchPaginationOptions & {
```
peterstarling, line 96: this are generally used in various places and in many of those the `documentCategory` filter does not apply. Leave it though as I can address it in my PR

apps/crn-server/scripts/export-analytics.ts
AkosuaA: I don't think that's the case as we don't include the `,` when the `j` and `i` reach the end of their arrays? That is what the condition above is for. I do agree about testing the file at some point

apps/crn-server/src/data-providers/contentful/analytics.data-provider.ts
AkosuaA: I agree, I'll add a test for this

packages/model/src/analytics.ts
AkosuaA: Why would this be an issue since it can be undefined?

PR #4325, opened 2024-07-04
ASAP-510 Fetch engagement data

gabiayako:
Jira Link: https://asaphub.atlassian.net/browse/ASAP-510
Figma Link: https://www.figma.com/design/qM73Om8hMDi8o15nj09ECw/CRN-%2F-Designs-%2F-Other-Pages?node-id=24392-92025&t=POJC35qxvi0JnaLI-0
Functional Specification: https://docs.google.com/document/d/1QzXXFpUUIvSGEHwoq3vHw5iVMlBmKBKsOG7qQVl7D-w/edit


packages/contentful/src/crn/queries/analytics.queries.ts
```diff
@@ -269,3 +269,55 @@ export const FETCH_TEAM_COLLABORATION = gql`
...
+      total
+      items {
+        sys {
+          id
+        }
+        displayName
+        inactiveSince
+        linkedFrom {
+          teamMembershipCollection(limit: 1) {
+            total
```
AkosuaA, line 285: Can a user have multiple roles in a team? From the cms that can happen but I don't know if that's a bug. We can add the same team & role multiple times to a user as well (I'm sure that's a bug though). In the first scenario, if it's not a bug wouldn't we have a wrong count?

gabiayako: Nice observation, I'll ask on Slack, but I think we should not allow multiple roles 🤔

gabiayako:
I'll merge this one and I'll open another PR after #4327 is merged to plug frontend and backend then we can move it to PO review :)

PR #4333, opened 2024-07-18
ASAP-426 (Collaboration) Add CSV export

gabiayako:
JIRA ticket: https://asaphub.atlassian.net/browse/ASAP-426

---

Note: the function `algoliaResultsToStream` that was inside `/analytics/leadership/export.ts` was moved to `/analytics/export.ts`.

### DIFF
--- a/apps/crn-frontend/src/analytics/__tests__/Routes.test.tsx
+++ b/apps/crn-frontend/src/analytics/__tests__/Routes.test.tsx
@@ -2,6 +2,8 @@ import { mockConsoleError } from '@asap-hub/dom-test-utils';
 import {
   performanceByDocumentType,
   userProductivityPerformance,
+  userCollaborationPerformance,
+  teamCollaborationPerformance,
 } from '@asap-hub/fixtures';
 import { analytics } from '@asap-hub/routing';
 import {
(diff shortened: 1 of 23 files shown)


AkosuaA (approved):
Looks good! 🎉🎉

With the csv headers there are a few difference when compared to the sample csv. Most make sense though so I have no problem approving.
eg: For team collaboration, in the sample csv, within team uses ASAP Article Output and across team uses ASAP Output Article. I assume we've agreed to go with one (as implemented). 
I wasn't sure if when the Type selected is ASAP Public Output the heading should also have Public included (as in the sample csv) or not.

Anyway, if they are issues it'll be raised in acceptance testing so happy to approve

gabiayako:
Thanks for the review @AkosuaA ! 😄 
I've made the suggested changes in the last two commits:
- [standardize field header for consistency](https://github.com/yldio/asap-hub/pull/4333/commits/74544210a6017149ee6f77b03039452beaf7ba0c)
- [add filter names to header rows](https://github.com/yldio/asap-hub/pull/4333/commits/907a8b3d75ab7cc492973cefdb5be3e281d11ec5)

PR #4342, opened 2024-07-31
[ASAP-424] -  allow sorting on analytics collaboration pages

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-424

`UserCollaborationDataObject` now includes `totalUniqueOutputsCoAuthoredAcrossTeams` & `totalUniqueOutputsCoAuthoredWithinTeam` which will be used in algolia when sorting User Collaboration data by 'Outputs Co-Authored'. It's the number of distinct research outputs a user has co-authored within/across all their teams. 

eg: In the past 30 days my user in the hub has the following data under 'Outputs Co-Authored' within team.
Screenshot 2024-08-01 at 14 53 13

Although there is 1 under Team Test 1 and 1 under A-Team Sprint 16, these refer to the same output so my `totalUniqueOutputsCoAuthoredWithinTeam` is 1.

The user_role & user_team algolia schemas have been updated to fix sorting by the team & role columns on the User Productivity & Collaboration pages. When sorting by the Team/Role columns, the User column should be ascending for users with the same Team/Role

Previously
Screenshot 2024-08-01 at 12 55 21

Currently
Screenshot 2024-08-01 at 14 47 24


apps/crn-frontend/src/analytics/collaboration/Collaboration.tsx
```diff
@@ -45,14 +63,38 @@ const Collaboration = () => {
...
+      setTeamSortingDirection(teamCollaborationInitialSortingDirection);
+    }
     history.push(
       analytics({})
         .collaboration({})
         .collaborationPath({ metric, type: newType }).$,
     );
+  };
+
+  const { client: userClient } = useAnalyticsUserCollaboration({
```
gabiayako, line 81: I was wondering if we could get this client by a function like

```
  const getUserClient = (sort: SortUserCollaboration) => {

    const indexName =
    sort === 'user_asc'
    ? ANALYTICS_ALGOLIA_INDEX
    : `${ANALYTICS_ALGOLIA_INDEX}_user_${sort.replace('user_', '')}`;
    
    return useAnalyticsAlgolia(indexName);
  } 
```

That would be called here and inside `useAnalyticsUserCollaboration` instead of making
`useAnalyticsUserCollaboration` to return `userClient`. Because the way it is I believe it's doing more calls than it need to.

Here on https://4342.hub.asap.science/analytics/collaboration/user/within-team after cleaning the network and refreshing the page, we have 14 requests

Screenshot 2024-08-02 at 07 48 52

and in my local without calling `useAnalyticsUserCollaboration` and `useAnalyticsTeamCollaboration` on `apps/crn-frontend/src/analytics/collaboration/Collaboration.tsx` we have 7 requests
Screenshot 2024-08-02 at 07 49 23

gabiayako (approved):

PR #4351, opened 2024-08-14
ASAP-627 CRN Public API - Research Output

gabiayako:
JIRA ticket: https://asaphub.atlassian.net/browse/ASAP-627

---

It's possible to test it here: https://api-4351.hub.asap.science/public/research-outputs


AkosuaA (approved):
Nice 🎉

There are a few naming differences when compared with the spreadsheet eg: resourceType & dateCreated but I'm not sure if it's supposed to be exactly as it is in the sheet

PR #4356, opened 2024-08-22
update standard deviation to quartiles

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-609
I tried a few scenarios on this sheet https://docs.google.com/spreadsheets/d/1Z-zWlTwiJeiftr5HhKqpAmRfX_wAQoCzdLWdj7tTtNk/edit?gid=868813298#gid=868813298 to come up with a 'formula'. 
I set duplicated values to -1 to make it easier on the frontend (i.e if the range is negative then display ---


gabiayako (approved):
really nice! 👏

PR #4359, opened 2024-08-27
ASAP-513 Engagement Custom Timeframe

gabiayako:
JIRA: https://asaphub.atlassian.net/browse/ASAP-513

---

Screenshot 2024-08-29 at 07 06 38

### DIFF
--- a/apps/crn-frontend/src/analytics/engagement/state.ts
+++ b/apps/crn-frontend/src/analytics/engagement/state.ts
@@ -5,13 +5,11 @@ import {
   ListEngagementAlgoliaResponse,
   SortEngagement,
 } from '@asap-hub/model';
-import { useEffect } from 'react';
 import {
   atomFamily,
   DefaultValue,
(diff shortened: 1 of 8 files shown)


apps/crn-frontend/src/analytics/engagement/state.ts
```diff
@@ -81,6 +83,15 @@ export const useAnalyticsEngagement = (
   const [engagement, setEngagement] = useRecoilState(
     analyticsEngagementState(options),
   );
+
+  const resetEngagement = useResetRecoilState(
```
AkosuaA, line 87: We can use the objectID field instead of id since those are unique. That way we wouldn't need to force refetch. Example from user productivity
Screenshot 2024-08-29 at 12 00 52

gabiayako: nice, I'll do that! Thanks!

PR #4360, opened 2024-08-28
Frontend works

lctrt:


packages/react-components/src/templates/AnalyticsEngagementPageBody.tsx
gabiayako, line 1: The icons beside the values on the table are missing. The text in the caption card is missing as well
Screenshot 2024-08-28 at 11 58 50

gabiayako (changes requested):

apps/crn-frontend/src/analytics/engagement/Engagement.tsx
```diff
@@ -39,6 +39,8 @@ const Engagement = () => {
     timeRange,
   });
 
+  const performance = useEngagementPerformance({ timeRange: '30d' });
```
gabiayako, line 42:
```suggestion
  const performance = useEngagementPerformance({ timeRange });
```

apps/crn-frontend/src/analytics/engagement/Engagement.tsx
```diff
@@ -39,6 +39,8 @@ const Engagement = () => {
...
+
   const { numberOfPages, renderPageHref } = usePagination(total, pageSize);
 
   const exportResults = () =>
```
gabiayako, line 46: Hey, I think we should also update export results since it should show the performance https://docs.google.com/spreadsheets/d/1ito19ZHryOSu0JxE1YQMltYkI7S_3S8PmwZgbHJb72Y/edit?gid=1516875906#gid=1516875906

Screenshot 2024-09-06 at 05 49 28

gabiayako (changes requested):

apps/crn-frontend/src/analytics/engagement/export.ts
```diff
@@ -1,14 +1,34 @@
...
+  (performance: EngagementPerformance) => (data: EngagementDataObject) => ({
+    Team: data.name,
+    'Team Status': data.inactiveSince ? 'Inactive' : 'Active',
+    'Inactive Since': data.inactiveSince
+      ? data.inactiveSince.split('T')[0]
+      : '',
+    Members: data.memberCount,
+    'Presentations: Value': data.eventCount,
+    'Presentations: Average': utils.getPerformanceText(
+      data.eventCount,
```
gabiayako, line 14:
```suggestion
    'Events: Value': data.eventCount,
    'Events: Average': utils.getPerformanceText(
      data.eventCount,
```

I believe we could still call this 'Events' since we're calling it this way on the table. I believe it started as 'Presentation' and then it became 'Events' but I think maybe it wasn't updated in all places ?

Screenshot 2024-09-06 at 07 54 37

packages/react-components/src/templates/AnalyticsEngagementPageBody.tsx
```diff
@@ -70,8 +89,28 @@ const AnalyticsEngagementPageBody: React.FC<
       href={analytics({}).engagement({}).$}
       exportResults={exportResults}
     />
+    <CaptionCard>
+      <>
+        <CaptionItem label="Presentations" {...performance.events} />
```
gabiayako, line 94:
```suggestion
        <CaptionItem label="Events" {...performance.events} />
```

I think we could call 'Events' here as well

Screenshot 2024-09-06 at 07 54 37

gabiayako (approved):
nice 🎉

PR #4361, opened 2024-08-28
ASAP-607 - filter out outputs that are not asap funded

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-607


gabiayako (approved):
nice 🎉

PR #4364, opened 2024-08-30
[ASAP-651] - fix alignment on performance card

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-651


dianaguimaraes (commented):
We still need to address this. Probably not the correct breakpoint.
image

dianaguimaraes (approved):

PR #4365, opened 2024-09-03
[ASAP-530] - display analytics last updated date

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-530

The ticket says 'The data will be refreshed everyday at 7:00 am EDT (New York time zone)'. I took a look at the workflow file and it's 6am UTC so that's what I used.
Screenshot 2024-09-04 at 10 14 45

### DIFF
--- a/packages/react-components/src/molecules/AnalyticsControls.tsx
+++ b/packages/react-components/src/molecules/AnalyticsControls.tsx
@@ -156,16 +156,16 @@ const updateSearchParams = (): URLSearchParams => {
   return searchParams;
 };
 
-const getLastUpdated = () => {
-  const lastUpdated = new Date();
-  const currentHour = lastUpdated.getUTCHours();
+const getLastUpdate = () => {
+  const lastUpdate = new Date();
+  const currentUTCHour = lastUpdate.getUTCHours();
 
-  if (currentHour <= 6) lastUpdated.setDate(lastUpdated.getDate() - 1);
+  if (currentUTCHour <= 6) lastUpdate.setDate(lastUpdate.getDate() - 1);
 
-  lastUpdated.setUTCHours(6, 0, 0);
+  lastUpdate.setUTCHours(6, 0, 0);
 
   return formatDateToTimezone(
-    lastUpdated.toISOString(),
+    lastUpdate.toISOString(),
     'do MMMM y, h:mm aaa (z)',
   );
 };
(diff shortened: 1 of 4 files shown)


packages/react-components/src/molecules/AnalyticsControls.tsx
```diff
@@ -150,6 +156,20 @@ const updateSearchParams = (): URLSearchParams => {
   return searchParams;
 };
 
+const getLastUpdated = () => {
+  const lastUpdated = new Date();
+  const currentHour = lastUpdated.getUTCHours();
+
+  if (currentHour <= 6) lastUpdated.setDate(lastUpdated.getDate() - 1);
```
AkosuaA, line 163: Although the sync runs at 6am UTC, since it runs for about an hour,  till 7am UTC the last update would be the previous day's run.

Screenshot 2024-09-04 at 10 13 34

packages/react-components/src/molecules/AnalyticsControls.tsx
```diff
@@ -150,6 +156,21 @@ const updateSearchParams = (): URLSearchParams => {
   return searchParams;
 };
 
+const getLastUpdate = () => {
+  const lastUpdate = new Date();
+  lastUpdate.setUTCHours(6, 0, 0);
+
+  const now = new Date();
+
+  if (lastUpdate.getUTCHours() >= now.getUTCHours())
```
dianaguimaraes, line 165: I think you can remove the `.getUTCHours()` from both `lastUpdate` and `now` as we just need to check that we're not setting `lastUpdate` to a future date, not just the time.

dianaguimaraes (approved):

packages/react-components/src/molecules/AnalyticsControls.tsx
AkosuaA: If its currently 6:15am UTC we would still want to set the previous days sync time even though lastUpdated is not a future date since the sync wouldn't be done running and the data seen in the hub would be from the previous day.

dianaguimaraes: Hmmm...then I think you need to represent some offset to consider the time it takes to update rather than comparing the hours directly, no?

AkosuaA: I'm not sure I understand why not to compare the hours directly. The sync takes about an hour (I added this comment earlier but it's outdated since I pushed another commit https://github.com/yldio/asap-hub/pull/4365#discussion_r1743402161). Since last update is 6am, if the current hour now is 6 or less then set date to previous day.

PR #4370, opened 2024-09-06
ASAP-608 change default time range filter

lctrt:


gabiayako (changes requested):
There are some places where 30d is a fallback, I think it would be nice to change that

- [ ] https://github.com/yldio/asap-hub/blob/ASAP-608/apps/crn-frontend/src/analytics/engagement/api.ts#L20
- [ ] https://github.com/yldio/asap-hub/blob/ASAP-608/apps/crn-server/src/utils/analytics/common.ts#L48
- [ ] https://github.com/yldio/asap-hub/blob/ASAP-608/packages/algolia/src/analytics/api.ts#L35
- [ ] https://github.com/yldio/asap-hub/blob/ASAP-608/packages/algolia/src/analytics/api.ts#L62

gabiayako (approved):

PR #4372, opened 2024-09-09
[ASAP-629] - add users public api

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-629

The api spec indicates returning the name and role for the users' interest groups. Currently I return only the interest group name since the user could be associated via a team and also be a leader on the interest group and it's not clear how to present the role in that scenario. I discussed with Padmini and she's agreed to leave that out till we hear back from gmmb.

Endpoints: 
List: https://api-4372.hub.asap.science/public/users
Single: https://api-4372.hub.asap.science/public/users/00c8ad49-f0ff-4f3b-a529-ba94353dc6df


apps/crn-server/src/data-providers/contentful/user.data-provider.ts
```diff
@@ -523,6 +575,33 @@ export const parseTeamsCollection = (
...
+      const isAuthor = researchOutput?.authorsCollection?.items.some(
+        (author) => author?.__typename === 'Users' && author.sys.id === userId,
+      );
+
+      if (isAuthor) {
+        return [
+          ...userResearchOutputs,
+          {
+            id: researchOutput?.sys.id || '',
+          },
```
dianaguimaraes, line 596: I'm thinking we probably don't need this to be an object with just the id and we can probably just return an array of ids. What do you think?

apps/crn-server/src/routes/public/user.route.ts
```diff
@@ -0,0 +1,81 @@
...
+  interestGroups: user.interestGroups.map((ig) => ({
+    name: ig.name,
+  })),
+  labs: user.labs,
+  lastName: user.lastName,
+  lastModifiedDate: user.lastModifiedDate,
+  orcid: user.orcid,
+  researchTheme: user.researchTheme,
+  researchOutputs: user.researchOutputs || [],
+  social: user.social,
```
dianaguimaraes, line 69: We can probably just return each of the social at the root level and remove the object wrap. If you notice we're currently even having orcid in 2 different places.

packages/contentful/src/crn/queries/users.queries.ts
```diff
@@ -153,10 +153,44 @@ export const usersContentQueryFragment = gql`
           }
         }
       }
+      researchOutputsCollection {
```
dianaguimaraes, line 156: Maybe we can make this piece composable so we don't make the query more complex, since this information is not used by the hub, just the public api.

dianaguimaraes (changes requested):

dianaguimaraes (approved):

PR #4377, opened 2024-09-12
add fields to gp2 output export

lctrt:


apps/gp2-frontend/src/outputs/export.ts
```diff
@@ -55,8 +83,40 @@ export const outputToCSV = ({
     sorted(workingGroups?.map((workingGroup) => workingGroup.title)) || '',
   projects: sorted(projects?.map((project) => project.title)) || '',
   date: formatDate(new Date(addedDate)),
+  description,
+  shortDescription,
+  relatedEvents: sorted(relatedEvents.map((event) => event.title)) || '',
+  relatedResearch: sorted(relatedOutputs.map((event) => event.title)) || '',
+  gp2Supported: gp2Supported ? 'yes' : 'no',
```
gabiayako, line 90: I believe `gp2Supported` comes as a string
Screenshot 2024-09-12 at 11 35 18

so maybe we could return just

```suggestion
  gp2Supported,
```

because it can also be "Don't Know" I think

apps/gp2-frontend/src/outputs/export.ts
```diff
@@ -55,8 +83,40 @@ export const outputToCSV = ({
...
+  ...getVersionFields(versions),
 });
 
+export const getVersionFields = (versions?: gp2.OutputVersion[]) => {
+  if (versions && versions.length > 0) {
+    return {
+      firstVersionTitle: versions[0]?.title,
+      firstVersionType: versions[0]?.type,
+      firstVersionRRID: versions[0]?.rrid,
+      firstVersionAccession: versions[0]?.accessionNumber,
```
gabiayako, line 106: I believe we're not returning rrid and accessionNumber in versions https://github.com/yldio/asap-hub/blob/master/packages/contentful/src/gp2/queries/output.queries.ts#L141:L151 yet

I think these fields are in the model though because we are using them when we're creating a version https://github.com/yldio/asap-hub/pull/4349/files. So they exist in write mode and not read mode.

apps/gp2-frontend/src/outputs/export.ts
```diff
@@ -55,8 +83,40 @@ export const outputToCSV = ({
...
+  doi,
+  rrid,
+  accession: accessionNumber,
+  publishDate,
+  lastUpdatedPartial,
+  ...getVersionFields(versions),
 });
 
+export const getVersionFields = (versions?: gp2.OutputVersion[]) => {
+  if (versions && versions.length > 0) {
```
dianaguimaraes, line 101:
```suggestion
  if (versions?.length > 0) {
```

gabiayako (approved):

PR #4378, opened 2024-09-13
add short description and first version data to research outputs export and fix typo in news filter

dianaguimaraes:
Tickets:
https://asaphub.atlassian.net/browse/ASAP-578
https://asaphub.atlassian.net/browse/ASAP-659


apps/crn-frontend/src/shared-research/export.ts
```diff
@@ -5,22 +5,58 @@ import {
...
+const getFirstVersionData = (
+  versions: Array<ResearchOutputVersion>,
+): FirstVersionCSV => {
+  if (versions[0]) {
+    const { title, type, rrid, accession, link } = versions[0];
+    return {
+      firstVersionTitle: title,
+      firstVersionType: type,
+      firstVersionRrid: rrid,
+      firstVersionAccession: accession,
```
gabiayako, line 46: I believe we need to modify the query to return te rrid and accession in the version
https://github.com/yldio/asap-hub/blob/master/packages/contentful/src/crn/queries/research-outputs.queries.ts#L216:L225

gabiayako (changes requested):

apps/crn-frontend/src/shared-research/export.ts
dianaguimaraes: Nice catch! I didn't realise it cause the other fields were retrieved and I knew some were only recently added.

gabiayako: If I’m not mistaken, we also need to add those fields to the data provider to ensure they appear in the export
https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/contentful/research-output.data-provider.ts#L274 :)

gabiayako (approved):

PR #4402, opened 2024-10-02
(fix): env creation based on github labels

gabiayako:
This PR adds the yellow row from the table below:

Screenshot 2024-10-02 at 11 37 11

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -316,12 +316,17 @@ runs:
           echo "gp2-app-url=https://$PR_NUMBER.gp2.asap.science" >> $GITHUB_OUTPUT
           echo "gp2-api-url=https://api-$PR_NUMBER.gp2.asap.science" >> $GITHUB_OUTPUT
 
-          if [ "$CRN_CREATE_CONTENTFUL_PR_ENV" = true ]; then
+          if [[ "$CRN_CREATE_ANALYTICS_ALGOLIA_INDEX" = true && "$CRN_CREATE_CONTENTFUL_PR_ENV" = true ]]; then
+            echo "setting up dedicated Contentful PR env, Entity Algolia Index and Analytics Algolia Index"
+            echo "crn-contentful-env-id=crn-$PR_NUMBER" >> $GITHUB_OUTPUT
+            echo "crn-algolia-index=asap-hub_CI-$PR_NUMBER" >> $GITHUB_OUTPUT
+            echo "crn-analytics-algolia-index=crn-analytics_CI-$PR_NUMBER" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/workflows/on-remove.yml
```diff
@@ -214,15 +214,15 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-default-region: ${{ steps.setup.outputs.aws-default-region }}
       - name: Delete the entity index
-        if: ${{ steps.setup.outputs.crn-contentful-on-branch-env == 'true' && (steps.setup.outputs.crn-algolia-index != 'asap-hub_dev' || steps.setup.outputs.crn-algolia-index != 'asap-hub_prod')}}
```
gabiayako, line 217: I believe here we should have `&&` instead of `||` because, for example, if we have `crn-algolia-index` = `'asap-hub_dev'`

`steps.setup.outputs.crn-algolia-index != 'asap-hub_dev'` would be `false`
`steps.setup.outputs.crn-algolia-index != 'asap-hub_prod'` would be `true`

so `(steps.setup.outputs.crn-algolia-index != 'asap-hub_dev' || steps.setup.outputs.crn-algolia-index != 'asap-hub_prod')` would be `false || true` -> `true` and it could delete `asap-hub_dev` eventually

.github/actions/setup-environment/action.yml
```diff
@@ -316,12 +316,17 @@ runs:
           echo "gp2-app-url=https://$PR_NUMBER.gp2.asap.science" >> $GITHUB_OUTPUT
           echo "gp2-api-url=https://api-$PR_NUMBER.gp2.asap.science" >> $GITHUB_OUTPUT
 
-          if [ "$CRN_CREATE_CONTENTFUL_PR_ENV" = true ]; then
+          if [[ "$CRN_CREATE_ANALYTICS_ALGOLIA_INDEX" = true && "$CRN_CREATE_CONTENTFUL_PR_ENV" = true ]]; then
```
gabiayako, line 319: I was unsure about the syntax but it works (I mean, it doesn't give an error and it reaches the content inside the `if` statement) as we can see here https://github.com/yldio/asap-hub/actions/runs/11146299276/job/30977928232#step:3:98
Screenshot 2024-10-02 at 12 01 02

I've cancelled the run with the two labels (`crn-create-environment` and `crn-create-analytics-algolia-index`) so it won't create a new analytics algolia index

.github/workflows/on-remove.yml
dianaguimaraes: To be fair it should never be `crn-contentful-on-branch-env` and `crn-algolia-index`.

dianaguimaraes (approved):

.github/workflows/on-remove.yml
gabiayako: yeah, but I think it's safer to leave this condition here 😅

PR #4407, opened 2024-10-04
[ASAP-542] - submit compliance report

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-542


apps/crn-frontend/src/network/teams/ManuscriptToastProvider.tsx
```diff
@@ -15,16 +18,24 @@ export const ManuscriptToastProvider = ({
...
+  };
 
   return (
-    <ManuscriptToastContext.Provider value={{ setShowSuccessBanner }}>
+    <ManuscriptToastContext.Provider
+      value={{ setShowSuccessBanner, setFormType }}
+    >
       <>
-        {showSuccessBanner && (
+        {showSuccessBanner && !!formType && (
```
gabiayako, line 33: maybe we could remove this `showSuccessBanner` and check if `formType` is truthy?

packages/react-components/src/templates/ComplianceReportForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+                enabled={!isSubmitting}
+                onClick={() => history.goBack()}
+              >
+                Cancel
+              </Button>
+              <Button
+                primary
+                noMargin
+                submit
+                enabled={!isSubmitting}
```
gabiayako, line 175: It seems the button starts as disabled when all the form is empty https://www.figma.com/design/zY4fHBAsXCiyW6gGrnS4Yy/CRN-%2F-Design-%2F-Network?node-id=24087-188526&t=KN9QcOYPgA7EeHyB-4 but maybe we could ask Padmini/João/Irene if it is supposed to be that way since it is not that consistent with other forms behavior

gabiayako (approved):
nice! 🎉

packages/react-components/src/templates/ComplianceReportForm.tsx
AkosuaA: Thanks, great catches! I've raised this with Padmini and she says this is the ideal behaviour. I'm updating this form (as well as the other changes raised) but there will be a subsequent ticket to update other forms.

PR #4409, opened 2024-10-07
Asap 615

lctrt:


packages/react-components/src/organisms/ResearchOutputFormSharingCard.tsx
```diff
@@ -203,14 +203,13 @@ const ResearchOutputFormSharingCard: React.FC<
...
+      <LabeledTextEditor
         title="Description"
         subtitle="(required)"
         tip="Add an abstract or a summary that describes this work."
         onChange={onChangeDescription}
         getValidationMessage={() => 'Please enter a description'}
         required
-        enabled={!isSaving}
         value={descriptionMD}
         info={
```
gabiayako, line 214: I don't think we need this info anymore 😄 
Screenshot 2024-10-14 at 06 36 16

packages/react-components/src/organisms/ResearchOutputFormSharingCard.tsx
```diff
@@ -203,14 +203,13 @@ const ResearchOutputFormSharingCard: React.FC<
...
       )}
-      <LabeledTextArea
+      <LabeledTextEditor
         title="Description"
         subtitle="(required)"
         tip="Add an abstract or a summary that describes this work."
         onChange={onChangeDescription}
         getValidationMessage={() => 'Please enter a description'}
         required
-        enabled={!isSaving}
```
gabiayako, line 213: maybe we can use this 

```
editor.setEditable(true/false)
```

to set to disabled when is saving (https://lexical.dev/docs/concepts/read-only)

AkosuaA (approved):
Looks great! 🎉🎉

Noticed this on the UI, minor and not very noticeable. May not even be an issue and can be addressed later if it is. 
The list portion of the description has a different colour from the paragraph portion
Screenshot 2024-10-15 at 17 27 31

PR #4420, opened 2024-10-18
[ASAP-534] - update quick check details fields

AkosuaA:
This pr updates the quick check details fields on the Manuscript Version content model in the cms to reference a Discussion. In order not to break the app, I've also updated the manuscript data provider to work with the cms changes. The next pr will add the replies functionality and update how we fetch the quick check detail's created by user.


apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
```diff
@@ -285,3 +307,72 @@ const parseComplianceReport = (
...
+    ? getLinkEntity(quickCheckDetails.protocolsDepositedDetails)
+    : null,
+});
+
+const createQuickCheckDiscussions = async (
+  environment: Environment,
+  quickCheckDetails: QuickCheckDetailsObject,
+  userId: string,
+) => {
+  const generatedDiscussions = {} as QuickCheckDetailsObject;
```
gabiayako, line 347: I wonder if we could remove `parseQuickCheckDiscussions` doing some modifications like this one:

```suggestion
  const generatedDiscussions = {} as Record<
    QuickCheckDetails,
    Link<'Entry'> | null
  >;
```

and the one in the comment below, what do you think?

apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
```diff
@@ -285,3 +307,72 @@ const parseComplianceReport = (
...
+            fields: addLocaleToFields({
+              message: getLinkEntity(messageId),
+            }),
+          });
+
+          await discussionEntry.publish();
+
+          generatedDiscussions[quickCheckDetail as QuickCheckDetails] =
+            discussionEntry.sys.id;
+        }
```
gabiayako, line 373:
```suggestion
          generatedDiscussions[quickCheckDetail as QuickCheckDetails] =
            getLinkEntity(discussionEntry.sys.id);
        } else {
          generatedDiscussions[quickCheckDetail as QuickCheckDetails] = null;
        }
```

gabiayako (approved):
nice

apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
AkosuaA: Yea, it simplifies the steps. Thanks! 👌🏾

PR #4421, opened 2024-10-18
Manuscript UID Display

lctrt:


packages/react-components/src/organisms/ManuscriptVersionCard.tsx
```diff
@@ -108,9 +113,52 @@ const hasAdditionalInfo = (version: ManuscriptVersion) =>
   version.requestingApcCoverage ||
   version.otherDetails;
 
-const ManuscriptVersionCard: React.FC<ManuscriptVersionCardProps> = (
-  version,
+export const getLifecycleCode = (
+  version: Pick<ManuscriptVersion, 'type' | 'lifecycle'>,
```
AkosuaA, line 117: minor suggestion, we can simplify since we only need the lifecycle here
```suggestion
  lifecycle: ManuscriptLifecycle,
```

AkosuaA (approved):
Looks good! 🎉 
I've just merged some changes which are breaking the pr environment so you'll need to rebase

PR #4426, opened 2024-10-24
Update Engagement analytics logic

lctrt:
- **update engagement % counting to use members instead of total speakers**
- **engagement performance uses percentage data**
- **update engagement table to use % for performance comparison**

### DIFF
--- a/apps/asap-cli/src/scripts/algolia/performance/engagement.ts
+++ b/apps/asap-cli/src/scripts/algolia/performance/engagement.ts
@@ -12,8 +12,8 @@ import {
 type EngagementHit = Hit & {
   eventCount: number;
   totalSpeakerCount: number;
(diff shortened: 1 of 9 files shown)


AkosuaA:
This wasn't mentioned in the ticket but we'd need to update the description on the performance card
Screenshot 2024-10-24 at 12 44 54

Will acceptance testing be done in production? Asking since although the logic looks good, we haven't updated algolia data so I couldn't confirm from the frontend. Happy to approve if we'll be doing acceptance in prod

gabiayako (approved):
Although it's not written down in the ticket I remember we discussed about making some adjustments to the Caption



We could take a look with Padmini how it should change though, but I think at least we agreed at the time that we would have `%` in the `Unique Speakers All Roles` and `Unique Speakers: Key Personnel`

PR #4428, opened 2024-10-28
Fix caption card ui for engagement analytics

lctrt:


AkosuaA (commented):
All the performance cards are now single column even when the screen can contain two columns (compared to dev version). Is this deliberate?

https://github.com/user-attachments/assets/e22faae8-4f26-4a68-86f6-56c872c1089d

AkosuaA (approved):

PR #4429, opened 2024-10-30
[ASAP-534] - submit quick check replies

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-534


apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
```diff
@@ -236,36 +237,41 @@ export const parseGraphqlManuscriptVersion = (
       submissionDate: version?.submissionDate,
       otherDetails: version?.otherDetails,
       acknowledgedGrantNumberDetails:
-        version?.acknowledgedGrantNumber === 'No'
-          ? version?.acknowledgedGrantNumberDetails?.message?.text
+        version?.acknowledgedGrantNumber === 'No' &&
```
lctrt, line 240: these fields handling could be using a function to reduce duplication. Something like
```
const maybeParseDiscussion = (field, details) => {
  if (field === 'No' && details) {
    return parseGraphQLDiscussion(details)
  }
  return undefined
};
```

then in this you could have 
```
acknowledgedGrantNumberDetails: maybeParseDiscussion(version?.acknowledgedGrantNumber, version?.acknowledgedGrantNumberDetails) 
```

AkosuaA: updated here: https://github.com/yldio/asap-hub/pull/4429/commits/dedb1c63effefeb37e0a94c3cb7f1c993f72c52a

lctrt (approved):

PR #4434, opened 2024-11-06
[ASAP-590] - add count field to compliance reports

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-590


packages/react-components/src/organisms/ShareComplianceReportModal.tsx
dianaguimaraes, line 1: Is this modal different to the ConfirmModal component?

dianaguimaraes (approved):

PR #4436, opened 2024-11-08
ASAP-752 Cookie Modal

gabiayako:
How to test it

1. Access https://4436.hub.asap.science/
2. Activate cookie modal feature flag
3. Refresh the page and check that cookie modal appears
4. Verify that cookie is saved on the browser when toggle is off for analytics
5. Verify that if cookie exists you can refresh several times and the cookie modal won’t appear until you delete the cookie
6. Delete the cookie, refresh the page and it’s possible to check that the modal appears again
7. Switch the analytics toggle and click save. Then it’s possible to check that the cookie is saved with analytics: true
8. The same flow applies to GP2 https://4436.gp2.asap.science/, check the logo and colors are correct

https://github.com/user-attachments/assets/988ff1fa-cbda-43f3-945f-2254a8c534c0


packages/react-components/src/organisms/CookiesModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+type ThirdPartyCookieLinkProps = {
+  link: string;
+  label: string;
+};
+
+const ThirdPartyCookieLink = ({ link, label }: ThirdPartyCookieLinkProps) => (
+  <Link href={link}>
+    <Pill>
+      <span css={pillStyles}>
+        {label} {<ExternalLinkIcon size={16} color={colors.lead} />}
```
dianaguimaraes, line 97: I don't think we need to wrap the ExternalLinkIcon with curly braces

dianaguimaraes (approved):

PR #4438, opened 2024-11-11
ASAP-756 Save Cookie Preferences in DynamoDB

gabiayako:


packages/server-common/test/handlers/cookie-preferences/save-cookie-preferences-handler-factory.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    jest.clearAllMocks();
+  });
+
+  describe('Validations', () => {
+    test('returns 400 when cookieId is not defined', async () => {
+      const request = getLambdaRequest({ preferences: {} }, {});
+      expect(handler(request)).rejects.toThrow('Validation error');
+    });
+
+    test('returns 400 when preferences is not defined', async () => {
```
dianaguimaraes, line 28: It looks like we're using both `test` and `it`. Can we pick one and stick with it please?

PR #4446, opened 2024-11-20
Asap 757 crn gp2 enable users view their cookie preferences

AimeurAmin:
[ASAP-757](https://asaphub.atlassian.net/browse/ASAP-757)

[ASAP-757]: https://asaphub.atlassian.net/browse/ASAP-757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-components/src/organisms/CookiesModal.tsx
dianaguimaraes, line 1: I'm not sure about the naming of the component.
It feels to me it should be more inclusive since we added the button as well.

packages/react-components/src/atoms/CookieButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    bottom: '7em',
+  },
+});
+
+const CookieButton: React.FC<CookieButtonProps> = ({
+  toggleCookieModal,
+  isOnboardable,
+}) => (
+  <span
+    className={isOnboardable ? 'is-not-onboarded' : ''}
```
dianaguimaraes, line 33: This behaviour is specific to an app making the component less independent.
My suggestion would be to detach the positioning from the button itself.

packages/react-components/src/organisms/CookiesModal.tsx
AimeurAmin: @dianaguimaraes as agreed; the component name will be kept as is (for now at least)

dianaguimaraes (approved):

PR #4451, opened 2024-12-02
add rules to compliance review visibility within team workspace depending on role and team

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-745

### DIFF
--- a/apps/crn-frontend/src/network/teams/TeamProfile.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamProfile.tsx
@@ -84,7 +84,7 @@ const TeamProfile: FC<TeamProfileProps> = ({ currentTime }) => {
       .then(loadOutputs)
       .then(loadTeamOutput)
       .then(loadEventsList);
-  }, [team]);
+  }, [team, isStaff]);
 
   const canShareResearchOutput = useCanShareResearchOutput(
     'teams',
--- a/apps/crn-frontend/src/network/teams/TeamProfile.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamProfile.tsx
@@ -4,7 +4,7 @@ import { v4 as uuid } from 'uuid';
 
 import { Frame } from '@asap-hub/frontend-utils';
 import { NotFoundPage, TeamProfilePage } from '@asap-hub/react-components';
-import { ResearchOutputPermissionsContext } from '@asap-hub/react-context';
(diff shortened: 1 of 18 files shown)


apps/crn-server/src/controllers/team.controller.ts
```diff
@@ -60,6 +60,7 @@ export default class TeamController {
       throw new NotFoundError(undefined, `team with id ${teamId} not found`);
     }
 
+    console.log(team);
```
AimeurAmin: forgotten dev log

AimeurAmin (changes requested):
Looks good, code wise, left a small comment below

apps/crn-server/src/controllers/team.controller.ts
dianaguimaraes: I believe you were seeing an old version

apps/crn-frontend/src/network/teams/TeamProfile.tsx
```diff
@@ -71,15 +74,17 @@ const TeamProfile: FC<TeamProfileProps> = ({ currentTime }) => {
...
   const { teamId } = useRouteParams(route);
   const team = useTeamById(teamId);
+  const user = useCurrentUserCRN();
+  const isStaff = user?.role === 'Staff';
 
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
     loadAbout()
-      .then(team?.tools ? loadWorkspace : undefined)
+      .then(team?.tools || isStaff ? loadWorkspace : undefined)
```
gabiayako, line 83: Is this right? In the backend it seems we show tools to everyone that belongs to the team https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/routes/team.route.ts#L39-L41

If that has changed for sure, I think it would be nice to keep the rule in the backend as it was before :)

gabiayako (changes requested):

apps/crn-frontend/src/network/teams/TeamProfile.tsx
gabiayako: oh sorry, I don't know why I skipped the `||` in the change above 😅 
But I still think it would be nice to have all rules in the same place (here https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/routes/team.route.ts#L39-L41)

AimeurAmin (approved):

gabiayako (approved):

PR #4454, opened 2024-12-05
ASAP-791 Add Lab Membership + data migration

gabiayako:
Ticket: https://asaphub.atlassian.net/browse/ASAP-791

---

This PR adds the content model `labMembership`, renames `labs` to `labs_old` in `users` (this will be deleted in a following PR) and adds a new field `labs` to `users` that can only contain `labMembership` items.

The `deriveEntryForLocale` used in the schema migration apparently only can attach one derived entry in the new field (I've tried to return an array with links and it didn't work), so I've also created a script to migrate the remaining data.

I've tested this script to migrate the rest of the data here in this action:
https://github.com/yldio/asap-hub/actions/runs/12198007677/job/34028835210#step:5:1

Finally, I ran Algolia Sync to make sure that the data we are seeing in the PR hub is indeed coming from the new field `labs`
Screenshot 2024-12-06 at 09 44 00


dianaguimaraes (approved):
Looks good to me!

PR #4471, opened 2024-12-12
[ASAP-749] - add compliance rules

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-749


packages/react-components/src/organisms/ManuscriptCard.tsx
```diff
@@ -283,38 +287,40 @@ const ManuscriptCard: React.FC<ManuscriptCardProps> = ({
...
-                    enabled={!!currentManuscriptVersion?.complianceReport}
-                  >
-                    <span css={buttonStyles}>
-                      {resubmitManuscriptIcon} Submit Revised Manuscript
-                    </span>
-                  </Button>
-                </span>
-              )}
-            </span>
+            {isActiveManuscript && (
```
gabiayako, line 290: There's some stale state when we change manually to compliant / closed. We need to refresh to see the button disappear:
Screenshot 2024-12-13 at 12 27 14

AkosuaA: I've just pushed an update to compute `isActiveManuscript` based on `selectedStatus` and not status

gabiayako (approved):
Really nice 👏👏👏

PR #4475, opened 2024-12-13
ASAP-813 Allow only lab membership creation on the CMS

gabiayako:
Adding an existing linked entry was causing confusion

Screenshot 2024-12-13 at 07 39 54

so we're going to change it to always create a new lab-membership entry
Screenshot 2024-12-13 at 08 54 50

### DIFF
--- a/packages/contentful/migrations/crn/users/20241213103301-remove-labs-link-existing-entries.js
+++ b/packages/contentful/migrations/crn/users/20241213103301-remove-labs-link-existing-entries.js
@@ -1,0 +1,19 @@
+module.exports.description = 'Allow only creation of lab membership';
+
+module.exports.up = (migration) => {
+  const users = migration.editContentType('users');
+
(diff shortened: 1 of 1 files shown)


packages/contentful/migrations/crn/users/20241213103301-remove-labs-link-existing-entries.js
```diff
@@ -0,0 +1,19 @@
...
+    showLinkEntityAction: false,
+    showCreateEntityAction: true,
+  });
+};
+
+module.exports.down = (migration) => {
+  const users = migration.editContentType('users');
+
+  users.changeFieldControl('labs', 'builtin', 'entryLinksEditor', {
+    showLinkEntityAction: false,
```
gabiayako, line 16: actually this should be true 😅

AkosuaA (approved):
🎉

PR #4485, opened 2024-12-18
ASAP-800 Teams and versions updated correctly

gabiayako:
This PR fixes the issue of teams not being updated in the `Manuscript` content type and the edit updating the last version, but the first one.

Jira ticket: https://asaphub.atlassian.net/browse/ASAP-800

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
@@ -206,7 +206,7 @@ export class ManuscriptContentfulDataProvider
       fields: addLocaleToFields({
         ...plainFields,
         count: currentCount + 1,
-        teams: getLinkEntities([teamId]),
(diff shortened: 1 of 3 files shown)


dianaguimaraes (commented):
Now we can probably simplify the following function:
https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/contentful/team.data-provider.ts#L338

AimeurAmin (approved):
Clear and straight to the point 👏

gabiayako:
> Now we can probably simplify the following function: https://github.com/yldio/asap-hub/blob/master/apps/crn-server/src/data-providers/contentful/team.data-provider.ts#L338

Yeah, it would be nice to simplify it, but we don't retrieve the teams from the manuscript yet, just from manuscript versions, we would need to add `teamsCollection` to the manuscript query. I'll do that in another PR if that's ok 😄 
Screenshot 2024-12-19 at 06 57 45

PR #4498, opened 2025-01-06
[ASAP-807] - add script to migrate preprint versions from csv

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-807

`preprint_data.csv` is a copy of https://docs.google.com/spreadsheets/d/1NNST5GsEp2ZTBPYglUTqVlReu7gwcPTnC3qLF7zRAGE/edit?gid=0#gid=0. 
The `packages/contentful/scripts/migrate-research-output-versions.ts` script reads the csv file and for each research output in the file, creates a preprint version if one doesn't exist. 

I tested the script locally with a dummy csv file [
[preprint_data.csv](https://github.com/user-attachments/files/18322555/preprint_data.csv)] in the dev environment. 

No existing preprint version
Before
Screenshot 2025-01-06 at 17 19 50

After
Screenshot 2025-01-06 at 17 23 50

Existing preprint version
Before
Screenshot 2025-01-06 at 17 20 37

After
no new version was created
Screenshot 2025-01-06 at 17 24 09

I limited the workflow to the prod environment since we don't need to replicate this in dev.


gabiayako (approved):
Nice!! 👏

PR #4499, opened 2025-01-10
ASAP-793 Manuscript Reminders: Created, Resubmitted  and Status Changed

gabiayako:


AimeurAmin (approved):
Looks great 👌🏻

apps/crn-server/src/data-providers/contentful/reminder.data-provider.ts
```diff
@@ -236,9 +259,32 @@ export const getSortDate = (reminder: ReminderDataObject): DateTime => {
     return DateTime.fromISO(reminder.data.notesUpdatedAt);
   }
 
+  if (reminder.type === 'Manuscript Created') {
```
AimeurAmin, line 262: just an opinion but I always find it more appealing to have a record instead of multiple `if`s in this kind of cases, something like this:

```
const typeToDateField: Record<string, keyof typeof reminder.data> = {
  'Manuscript Created': 'publishedAt',
  'Manuscript Status Updated': 'updatedAt',
  'Manuscript Resubmitted': 'resubmittedAt',
};

const dateField = typeToDateField[reminder.type] || 'endDate';
return DateTime.fromISO(reminder.data[dateField]);
```

but again It's just a preference and maybe it's more readable with the ifs in this case.

PR #4502, opened 2025-01-14
ASAP-845 Fix GP2 filter event by project

gabiayako:
Ticket: https://asaphub.atlassian.net/browse/ASAP-845

This bug happens because the filter `project.id` was missing in the query sent to algolia and in the facet configuration.

---

It's possible to see below that in dev the past events tab is showing events from another project (left side of screenshot) in the PR env it doesn't.

Screenshot 2025-01-14 at 14 44 33

### DIFF
--- a/packages/algolia/src/filters.ts
+++ b/packages/algolia/src/filters.ts
@@ -1,4 +1,9 @@
-import { EventConstraint } from '@asap-hub/model';
+import {
+  EventConstraint as CRNEventConstraint,
+  gp2 as gp2Model,
+} from '@asap-hub/model';
+
+type EventConstraint = CRNEventConstraint | gp2Model.EventConstraint;
 
 type Filters = {
   before?: string;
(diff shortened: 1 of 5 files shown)


packages/model/src/event.ts
```diff
@@ -82,6 +82,7 @@ export type EventConstraint = {
   userId?: string;
   teamId?: string;
   notStatus?: string;
+  projectId?: string;
```
dianaguimaraes, line 85: I believe this should belong to [`gp2/EventConstraint`](https://github.com/yldio/asap-hub/blob/master/packages/model/src/gp2/event.ts#L40) type instead

gabiayako: Without this, there's this TS error
Screenshot 2025-01-14 at 15 01 53

gabiayako: if we separate the types, we get a bunch of errors
Screenshot 2025-01-14 at 15 02 11

gabiayako: before I try to fix these bunch of errors, just wanted to confirm if your suggestion is to have 

```
constraint?: EventConstraint | gp2Model.EventConstraint
```

dianaguimaraes: Yes, something along those lines.
But maybe an intersection instead of an union will make it easier if we're checking constraints one by one anyway.

gabiayako: Thanks, I ended up checking each property in the object [trying to fix the type](https://github.com/yldio/asap-hub/pull/4502/commits/2ad903cb6fea27a3ba7e02b1f6908149068d8b11)

packages/algolia/src/filters.ts
```diff
@@ -11,20 +16,39 @@ const timeFilter = (time: string, symbol: string) => {
...
+  }
+  if ('notStatus' in constraint && constraint.notStatus) {
+    constraintFiltersList.push(`NOT status:${constraint.notStatus}`);
+  }
+
+  if ('projectId' in constraint && constraint.projectId) {
+    constraintFiltersList.push(`project.id: "${constraint.projectId}"`);
+  }
+
+  const constraintFilters = constraintFiltersList.filter(Boolean).join(' AND ');
```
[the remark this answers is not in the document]
gabiayako: Yeah, you're right! Thanks! I've removed it here [remove filter(Boolean)](https://github.com/yldio/asap-hub/pull/4502/commits/e0f16b3484490934a20c9081beed7c07b34d30a3)

dianaguimaraes (approved):

PR #4504, opened 2025-01-16
[ASAP-568] - remove getManuscriptVersionUID from frontend

AkosuaA:
This pr removes the `getManuscriptVersionUID` function from the `ManuscriptVersionCard` component and passes the uid from the backend.


apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
gabiayako, line 1: Hey, isn't it necessary to change it here https://github.com/yldio/asap-hub/blob/ASAP-568-refactor-manuscript-version-uid/apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts#L379 as well?

AkosuaA: I left it because we currently don't use the version uid when fetching a specific manuscript but it could be better to just have it. Updating now

gabiayako (approved):
nice 🎉

PR #4507, opened 2025-01-17
ASAP-847 Upgrade actions artifact from v3 to v4

gabiayako:


.github/actions/test-unit/action.yml
gabiayako, line 1: This doesn't seem to used anywhere

.github/workflows/reusable-test.yml
```diff
@@ -103,19 +103,23 @@ jobs:
           ## get package directory
           mkdir -p coverage
           yarn workspace ${{ matrix.package }} exec mv coverage/coverage-final.json ../../coverage/$PACKAGE_NAME.json
-      - uses: actions/upload-artifact@v3
+      - name: Sanitize package name
+        id: sanitize
+        run: echo "MATRIX_PACKAGE_NAME=${{ matrix.package }}" | sed 's/\//-/g' >> $GITHUB_ENV
```
gabiayako, line 108: forward slash is not allowed, since the `matrix.package` is something like `@asap-hub/algolia` we were having these errors:
 
forward slash

so this part replaces `/` by `-`

AimeurAmin (approved):

PR #4509, opened 2025-01-20
[ASAP-753] - add discussion rules

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-753


packages/react-components/src/organisms/ManuscriptVersionCard.tsx
```diff
@@ -403,7 +408,7 @@ const ManuscriptVersionCard: React.FC<ManuscriptVersionCardProps> = ({
                       <Suspense fallback={<Loading />}>
                         <Discussion
                           modalTitle="Reply to quick check"
-                          canReply={isActiveManuscript && isTeamMember}
+                          canReply={isActiveVersion && canUpdateDiscussion}
```
gabiayako, line 411: `canUpdateDiscussion` already takes `isActiveVersion` into consideration :)
```suggestion
                          canReply={canUpdateDiscussion}
```

gabiayako (approved):
Nice! 👏

packages/react-components/src/organisms/ManuscriptVersionCard.tsx
AkosuaA: Good catch, thanks! 👌🏾

PR #4510, opened 2025-01-21
use manuscript version count on the compliance report count

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-848


packages/contentful/migrations/crn/complianceReports/20250120161725-remove-count-field.js
```diff
@@ -0,0 +1,21 @@
+module.exports.description = 'Add count field';
```
gabiayako, line 1:
```suggestion
module.exports.description = 'Delete count field';
```

PR #4516, opened 2025-01-30
[ASAP-853] Enable Hub Admin assign user role to a Lab in CMS

AimeurAmin:
[ASAP-853](https://asaphub.atlassian.net/browse/ASAP-853)

[ASAP-853]: https://asaphub.atlassian.net/browse/ASAP-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contentful/scripts/assign-users-as-lab-pi.ts
```diff
@@ -0,0 +1,196 @@
...
+        content_type: 'users',
+        'fields.oldLabs.sys.id': lab.sys.id,
+      } as QueryOptions);
+
+      if (!users.items.length) {
+        console.log(`No users found for lab: ${labId}`);
+        continue;
+      }
+
+      const firstUser = users.items[0]; // Pick the first user - maybe the last user would be better? - also maybe checking if the user is published would be a good idea?
```
gabiayako, line 156: I believe we should check the role in here. But I'll ask in the engineers channel as well

gabiayako (changes requested):

.github/workflows/on-demand-migrate-user-labs.yml
gabiayako, line 1: Just make sure to revert these changes before merging it 😉

packages/contentful/migrations/crn/labs/20250202113054-rollback-lab-membership-and-adds-pi-field.js
```diff
@@ -0,0 +1,47 @@
+module.exports.description =
+  'Renames lab membership reference (labs to oldLabs) and creates new labs field (references labs) adds PI field';
+
+module.exports.up = (migration) => {
+  const labs = migration.editContentType('labs');
+
+  labs
+    .createField('labPI')
+    .name('Lab PI')
```
gabiayako, line 9: Maybe the display name could be

```suggestion
    .name('Principle Investigator')
```

as it appears in the ticket 
Screenshot 2025-02-11 at 09 32 56

packages/contentful/scripts/assign-users-as-lab-pi.ts
gabiayako: as per the conversation on slack, we should check if the lab memberships has role "Lead PI"  before assigning the user to the `labEntry.fields.labPI`

packages/contentful/scripts/assign-users-as-lab-pi.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      // Update the labs field
+      userEntry.fields.labs = userEntry.fields.labs || {};
+      userEntry.fields.labs['en-US'] = labIds.map((labId) => ({
+        sys: {
+          type: 'Link',
+          linkType: 'Entry',
+          id: labId,
+        },
+      })) as Link<'Entry'>[];
```
gabiayako, line 107: here it seems we are fetching the `labMemberships` just to get the `labIds`, right? but the `labIds` can be found on `labMemberships`, no? So couldn't we do something like:

```
const labMemberships: Entry[] = user.fields.oldLabs?.['en-US'] || [];

if (!labMemberships.length) {
  console.log('No labs, skipping - 2');
  continue;
}

const isPublished = user.isPublished();

const userEntry = await environment.getEntry(user.sys.id);

// Update the labs field
userEntry.fields.labs = userEntry.fields.labs || {};
userEntry.fields.labs['en-US'] = labMemberships.map((labMembership) => ({
  sys: {
    type: 'Link',
    linkType: 'Entry',
    id: labMembership.lab.sys.id,
  },
})) as Link<'Entry'>[];
```

packages/contentful/scripts/assign-users-as-lab-pi.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    const labMembershipIds = users.items.flatMap(
+      (user) =>
+        user.fields.oldLabs?.['en-US'].map(
+          (membershipRef: any) => membershipRef.sys.id,
+        ) || [],
+    );
+
+    const labMemberships = await environment.getEntries({
+      content_type: 'labMembership',
```
gabiayako, line 142: Here we're 
- fetching all users who have some labs
- then fetching the labMemberships and 
- checking if the role is `Lead PI`
 - assigning the user to the field labPI in Labs
 
Imagining we have 1000 users and only 5  have lab membership with role Lead PI we were doing a lot of unnecessary requests, since the majority of users wouldn't have Lead PI as a role.

Maybe we could go the other way:
- Fetch all labMemberships that has role === `Lead PI`
- Fetch the user that has this labMembership
- Assign the user to the field labPI in Labs

In order to fetch all labMemberships that has role === `Lead PI` we could do something like

```
const leadPILabMemberships = await environment.getEntries({
  content_type: 'labMembership',
  'fields.role': 'Lead PI',
} as QueryOptions);
```

and fetch the user and lab that has this labMembership:

```
const labIdsByLabMembershipId = leadPILabMemberships.items.reduce(
  (acc, labMembership) => {
    acc[labMembership.sys.id] = labMembership.fields.lab['en-US'].sys.id;
    return acc;
  },
  {} as Record<string, string>,
);

for (const [labMembershipId, labId] of Object.entries(
  labIdsByLabMembershipId,
)) {
  const users = await environment.getEntries({
    content_type: 'users',
    'fields.labs.sys.id[in]': labMembershipId,
    include: 2,
    select: 'fields.email,sys',
  } as QueryOptions);

  const userEntry = users.items[0];
  const labEntry = await environment.getEntry(labId);
  (...)
}
```

This way we would do less requests :)

AimeurAmin: but if we're assigning all labs to the `user`, no? he can be a member of a `lab` but not a `Lead PI`, right?

```
const leadPILabMemberships = await environment.getEntries({
  content_type: 'labMembership',
  'fields.role': 'Lead PI',
} as QueryOptions);
```

this will only fetch labMemberships that has the role `Lead PI`

gabiayako: Yeah, but you're already assigning the labs to the user in the `rollbackLabsMembership`, right? and in `assignLabPI` the goal is to search for lab memberships with role Lead PI

AimeurAmin: correct, sorry I got confused thought you were talking about `rollbackLabsMembership `

PR #4533, opened 2025-02-24
[ASAP-842] - Add RTF capability on the compliance report form

AimeurAmin:
[ASAP-842](https://asaphub.atlassian.net/browse/ASAP-842)

[ASAP-842]: https://asaphub.atlassian.net/browse/ASAP-842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-components/src/molecules/ExpandableText.tsx
+++ b/packages/react-components/src/molecules/ExpandableText.tsx
@@ -32,7 +32,8 @@ const expandableTextStyles = css({
   } 100%)`,
   WebkitBackgroundClip: 'text',
   backgroundClip: 'text',
-  transition: 'max-height 300ms linear',
+  maskImage: 'linear-gradient(to bottom, rgba(0,0,0,1), rgba(0,0,0,0))',
+  WebkitMaskImage: 'linear-gradient(to bottom, rgba(0,0,0,1), rgba(0,0,0,0))',
 });
 
 const buttonContainerStyles = css({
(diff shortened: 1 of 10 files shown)


AkosuaA (approved):
Looks good! 🎉🎉

I noticed on forms with the LabeledTextEditor field when the form is opened those fields are automatically active. It's not as noticeable on the other forms as it is on the ComplianceReportForm since that's the only one using the onBlur prop. So basically if you try to enter the url before the description you get an error to add the description. We could add it to the ticket to fix form errors if it's a lot to fix now.

packages/react-components/src/molecules/ExpandableText.tsx
```diff
@@ -31,9 +31,7 @@ const expandableTextStyles = css({
     colorWithTransparency(lead, 0).rgba
   } 100%)`,
   WebkitBackgroundClip: 'text',
-  WebkitTextFillColor: 'transparent',
```
AkosuaA, line 34: removing this takes out the fade out effect
Screenshot 2025-02-26 at 15 21 10

**Currently**
Screenshot 2025-02-26 at 15 21 53

AkosuaA: Cool! Looks good to me

PR #4535, opened 2025-03-03
[ASAP-901] - compliance table update

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-901

### DIFF
--- a/packages/react-components/src/organisms/ComplianceTable.tsx
+++ b/packages/react-components/src/organisms/ComplianceTable.tsx
@@ -18,7 +18,6 @@ import ComplianceTableRow from './ComplianceTableRow';
 import ConfirmStatusChangeModal from './ConfirmStatusChangeModal';
 
 const container = css({
-  paddingTop: rem(8),
   overflowX: 'auto',
   table: {
     borderSpacing: 0,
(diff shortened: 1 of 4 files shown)


packages/react-components/src/organisms/ComplianceTable.tsx
```diff
@@ -1,67 +1,65 @@
...
-  display: 'grid',
-  padding: `${rem(20)} ${rem(24)} 0`,
-  borderBottom: `1px solid ${steel.rgb}`,
-  ':first-of-type': {
-    borderBottom: 'none',
+  paddingTop: rem(8),
+  overflowX: 'auto',
+  table: {
+    borderSpacing: 0,
+    width: 'max-content',
```
gabiayako, line 25: I think we could have something like
```suggestion
    width: '100%',
    minWidth: 'max-content',
```

so in large screens the width is expanded. It looks like this in my big monitor


To see how it looks like, it's always possible to use this "device toolbar"


my monitor for example is 2560 x 1440

gabiayako (approved):
really nice! 🎉

packages/react-components/src/organisms/ComplianceTable.tsx
AkosuaA: Thanks!

PR #4540, opened 2025-03-11
[ASAP-827] - send compliance emails

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-827

- Added two flags `SEND_COMPLIANCE_NOTIFICATIONS` and `COMPLIANCE_NOTIFICATION_LIST`.
The `SEND_COMPLIANCE_NOTIFICATIONS` flag, if enabled, sends emails to all expected recipients. However, it is disabled by default. 
While the feature is disabled, you can still test sending emails by setting the  `COMPLIANCE_NOTIFICATION_LIST` flag to a string containing accepted emails, 
eg: `document.cookie = 'ASAP_COMPLIANCE_NOTIFICATION_LIST=deborah.attuah@yld.io';`. 
The notes on the ticket says to send emails only to email addresses in the yld.io domain, however, this approach should make it easier for us and ASAP testers. 

- Previously, the flags implementation only accepted boolean values but has been expanded to include strings for flags with the _LIST suffix.


gabiayako (approved):
Really nice and smart solution! 👏

PR #4547, opened 2025-03-24
[ASAP-785] - Enable users to set Manuscript Status within Share Compliance Report form

AimeurAmin:
[ASAP-785](https://asaphub.atlassian.net/browse/ASAP-785)

[ASAP-785]: https://asaphub.atlassian.net/browse/ASAP-785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/crn-server/src/validation/compliance-report.validation.ts
+++ b/apps/crn-server/src/validation/compliance-report.validation.ts
@@ -15,12 +15,12 @@ const complianceReportPostRequestValidationSchema: JSONSchemaType<ComplianceRepo
       manuscriptVersionId: { type: 'string' },
       discussionId: { type: 'string', nullable: true },
       versionId: { type: 'string', nullable: true },
-      manuscriptId: { type: 'string', nullable: true },
-      status: { enum: manuscriptStatus, type: 'string', nullable: false },
+      manuscriptId: { type: 'string' },
+      status: { enum: manuscriptStatus, type: 'string' },
       sendNotifications: { type: 'boolean', nullable: true },
       notificationList: { type: 'string', nullable: true },
     },
-    required: ['description', 'url', 'manuscriptVersionId', 'status'],
+    required: ['description', 'url', 'manuscriptVersionId', 'status', 'manuscriptId'],
     additionalProperties: false,
   };
 
--- a/apps/crn-server/src/validation/compliance-report.validation.ts
+++ b/apps/crn-server/src/validation/compliance-report.validation.ts
@@ -1,4 +1,4 @@
-import { ComplianceReportPostRequest } from '@asap-hub/model';
+import { ComplianceReportPostRequest, manuscriptStatus } from '@asap-hub/model';
 import { validateInput } from '@asap-hub/server-common';
 import { urlExpression } from '@asap-hub/validation';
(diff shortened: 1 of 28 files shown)


packages/react-components/src/templates/ComplianceReportForm.tsx
```diff
@@ -68,31 +76,58 @@ type ComplianceReportFormProps = {
...
+  'Waiting for Report',
+  'Review Compliance Report',
+  'Manuscript Resubmitted',
+  'Submit Final Publication',
+  'Addendum Required',
+  'Compliant',
+  'Closed (other)',
+] as const;
+
+type ManuscriptStatus = (typeof manuscriptStatus)[number];
```
gabiayako, line 91: I believe we can import this from `@asap-hub/model`

packages/react-components/src/templates/ComplianceReportForm.tsx
```diff
@@ -68,31 +76,58 @@ type ComplianceReportFormProps = {
...
+      content:
+        'If you elect to share the compliance report, all associated team members (First Author(s), PM, PIs, Corresponding Author and Additional Authors) will receive a reminder on the CRN Hub and an email to notify them that this report is now available.',
+      confirmButtonText: 'Share Compliance Report',
+      confirmButtonStyle: 'primary' as const,
+    },
+    confirmClosedStatus: {
+      title:
+        'Share compliance report and set status to <compliant/closed (other)>?',
+      content:
+        'After you update the status to <compliant/closed (other)>, this change will be permanent and cannot be altered. If you need to make changes in the future, please reach out to the CMS admin. Additionally, by sharing the compliance report, all associated team members (First Author(s), PM, PIs, Corresponding Author and Additional Authors) will receive a reminder on the CRN Hub and an email to notify them that this report is now available.',
```
gabiayako, line 119: Screenshot 2025-03-27 at 10 25 18

It's better to confirm with Padmini/João, but I believe this `<compliant/closed (other)>`should be replaced by `compliant` or `closed` depending on the value selected, like it is when we try to change the status in this page below

Screenshot 2025-03-27 at 10 26 38

apps/crn-server/src/validation/compliance-report.validation.ts
```diff
@@ -16,8 +16,11 @@ const complianceReportPostRequestValidationSchema: JSONSchemaType<ComplianceRepo
       discussionId: { type: 'string', nullable: true },
       versionId: { type: 'string', nullable: true },
       manuscriptId: { type: 'string', nullable: true },
```
gabiayako, line 18: Hey, I was wondering if this `manuscriptId` is ever nullable, is it? if not maybe we can have `nullable: false` and then in the route file we could have

```
    const manuscript = await manuscriptController.fetchById(manuscriptId);
```

instead of 

```
    const manuscript =
      manuscriptId && (await manuscriptController.fetchById(manuscriptId));
```

gabiayako (changes requested):

gabiayako (approved):

PR #4556, opened 2025-04-09
fix avatarUrl not being retrived in CRN's users public API

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-966


AkosuaA (approved):
🎉

PR #4577, opened 2025-05-08
ASAP-979 Contributors Validation in Manuscript Form

gabiayako:
Jira Ticket: https://asaphub.atlassian.net/browse/ASAP-979

---

validation


AkosuaA (commented):
Nice work! 🎉

I noticed a few things:
Screenshot 2025-05-12 at 10 34 17
In the screenshot above Robert's user doesn't have a team. This throws an error for the corresponding author field but not the first author or additional author. I'm not sure what the expected behaviour should be but I assume it should be the same for all the author fields.

Also noticed this behaviour 
- Add two users with teams not listed above to the first author field
- Add a lab
- Remove one of the users added in the first author field
- Error changes to please select at least one author
Screen Rec added below
https://github.com/user-attachments/assets/c9a84a21-1dfa-4e86-a3aa-c944956a33f3
This also happens with the Additional Authors field

Also want to confirm that having the pill in red is out of scope?
Screenshot 2025-05-12 at 11 20 33

gabiayako:
Thanks for the review! 😄 

> Also want to confirm that having the pill in red is out of scope?
Yes, this pill in red became a separate ticket https://asaphub.atlassian.net/browse/ASAP-980

PR #4580, opened 2025-05-13
[ASAP-905] - add link to manuscript

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-905

Note: The scroll behaviour will have to be updated as it doesn't always work (eg: on slow page loads). Padmini agreed to leave as is and come back to it later.


gabiayako (approved):
nice! 👏

PR #4582, opened 2025-05-14
ASAP-985 Fixes form validation triggers and disables Submit button until all required fields are filled 

AimeurAmin:
[ASAP-985](https://asaphub.atlassian.net/browse/ASAP-985)

[ASAP-985]: https://asaphub.atlassian.net/browse/ASAP-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-components/src/templates/ManuscriptForm.tsx
```diff
@@ -1012,13 +1037,18 @@ const ManuscriptForm: React.FC<ManuscriptFormProps> = ({
...
                     fieldState: { error },
                   }) => (
                     <LabeledTextField
                       title="Please provide details"
                       subtitle={'(required)'}
-                      onChange={onChange}
+                      onChange={async (e) => {
+                        onChange(e);
+                        await trigger('versions.0.otherDetails');
+                        onBlur();
```
gabiayako, line 1049: hey, I didn't quite understand why we're calling `onBlur` inside `onChange`. Can you explain, please? Is it only to validate the field?

packages/react-components/src/templates/ManuscriptForm.tsx
```diff
@@ -883,7 +877,15 @@ const ManuscriptForm: React.FC<ManuscriptFormProps> = ({
                     value: option,
                     label: option,
                   }))}
-                  onChange={onChange}
+                  onChange={(e) => {
+                    onChange(e);
+                    // Clear lifecycle when type changes to prevent mismatched states
+                    if (e !== watchType) {
```
gabiayako, line 883: just out of curiosity, when does this happen? I mean, when the value `e` can be different than `watchType`?

gabiayako (changes requested):
Hey Amin, I noticed the validation message is not showing below corresponding author anymore. This video below was recorded on https://4582.hub.asap.science
https://github.com/user-attachments/assets/a8570083-787c-44f9-9ee5-f79004e881b5

In dev with the same input we have
Screenshot 2025-05-19 at 06 15 03

packages/react-components/src/templates/ManuscriptForm.tsx
AimeurAmin: must have forgot to remove that one, the trigger does the validation on its own just fine

packages/react-components/src/templates/ManuscriptForm.tsx
AimeurAmin: the idea is that `watchType` which is basically the `versions.0.type` if it changes then `versions.0.lifecycle` should be reset, on the other hand, now looking back at it, the if check isn't needed, cause if the `onChange` gets triggered then obviously there has been a change

https://github.com/user-attachments/assets/c53316f1-a354-4c5a-b898-ee805921b4a6

packages/react-components/src/templates/ManuscriptForm.tsx
gabiayako: yeah, but also, if we want to validate `onChange` instead of `onBlur` we just need to change here 

https://github.com/yldio/asap-hub/blob/ASAP-1012-update-manuscript-and-version-counts-logic/packages/react-components/src/templates/ManuscriptForm.tsx#L350

from `mode: 'onBlur'` to `mode: 'onChange'` but I think it's a product decision as it affects the experience. We can always change to `onChange`. It would be nice to ask Padimini and the designers first

AimeurAmin: I believe I already confirmed that with Padmini, I showed both behaviors and she went for the current one.

gabiayako: The current one you mean on change?

PR #4583, opened 2025-05-15
update manuscript count and manuscript version count generation logic

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-1012


apps/crn-server/test/data-providers/contentful/manuscript.data-provider.test.ts
```diff
@@ -1482,6 +1482,8 @@ describe('Manuscripts Contentful Data Provider', () => {
         userId: 'user-id-0',
       });
 
+      //console.log(JSON.stringify(environmentMock.createEntry));
```
AimeurAmin, line 1485: commented log missed 👀

AimeurAmin (approved):
Looks all good to me as long as the commented log statement and the conflict are out of the way ✌🏻

packages/contentful/src/crn/queries/manuscript.queries.ts
```diff
@@ -484,7 +484,7 @@ export const FETCH_MANUSCRIPTS_BY_TEAM_ID = gql`
   query FetchManuscriptsByTeamId($id: String!) {
     teams(id: $id) {
       linkedFrom {
-        manuscriptsCollection(limit: 500) {
+        manuscriptsCollection(limit: 500, order: count_DESC) {
```
gabiayako, line 487: I was taking a look at this query, we grab 500 manuscripts, then check if the team is inside teamsCollection of the manuscript to grab the count.

I think we could have something like this 

```
  manuscriptsCollection(
    where: { teams: { sys: { id: $id } } }
    limit: 1,
    order: count_DESC
  ) {
    items {
      count
    }
  }
```

instead, right?

gabiayako: I see, it makes sense. I think we can at least add a `limit: 1` here :)

```    
teams(id: $id) {
  linkedFrom {
    manuscriptsCollection(limit: 500, order: count_DESC) {
      items {
        teamsCollection(limit: 1) {    <---
          items {
            sys {
              id
            }
          }
        }
        count
      }
    }
  }
}
```

gabiayako (approved):

PR #4584, opened 2025-05-16
 ASAP-913 Adds OS Checklist link on the manuscript form

AimeurAmin:
[ASAP-913](https://asaphub.atlassian.net/browse/ASAP-913)

### DIFF
--- a/packages/react-components/src/icons/external-link.tsx
+++ b/packages/react-components/src/icons/external-link.tsx
@@ -6,13 +6,16 @@ import type { OpaqueColor } from '../colors';
 type ExternalLinkIconProps = {
   size?: number;
   color?: OpaqueColor;
+  style?: React.CSSProperties;
 };
 
 const ExternalLinkIcon: React.FC<ExternalLinkIconProps> = ({
   size = 24,
   color = charcoal,
+  style,
 }) => (
   <svg
+    style={style}
     width={size}
     height={size}
     viewBox="0 0 24 24"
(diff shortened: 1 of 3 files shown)


packages/react-components/src/molecules/FormCard.tsx
```diff
@@ -25,14 +29,38 @@ const FormCard: React.FC<FormCardProps> = ({
...
   <Card padding={false} overrideStyles={cardStyles} title={title}>
     <div role="presentation" css={[paddingStyles]}>
       <Headline3 noMargin>{title}</Headline3>
     </div>
     {!!description && (
       <div css={[paddingStyles, descriptionStyles]}>
-        <Paragraph accent="lead">{description}</Paragraph>
+        <Paragraph accent="lead">
+          {description}
+          {link && (
```
AkosuaA, line 43: Rather than add the link & label as separate props, I think it will be better to modify the description prop to accept a  react element, eg: what we have for the description prop in `ConfirmModal`. Since we reuse the FormCard it's possible that we'll have to add another description with a link and the link could be anywhere within the description.
Screenshot 2025-05-19 at 10 55 37

packages/react-components/src/icons/external-link.tsx
```diff
@@ -6,13 +6,16 @@ import type { OpaqueColor } from '../colors';
 type ExternalLinkIconProps = {
   size?: number;
   color?: OpaqueColor;
+  style?: React.CSSProperties;
```
dianaguimaraes, line 9: Can we move this outside the icon and have a container with the styles wrapping the icon instead, please?

AkosuaA (approved):
Nice 🎉

PR #4592, opened 2025-05-21
ASAP-1017 APC Coverage buttons in Compliance Table

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-1017

----
### Notes

- As discussed on slack, the APC filter dropdown was hidden regardless of the feature flag
- To see the add/edit button is necessary  to activate the feature flag using
```
document.cookie = 'ASAP_DISPLAY_NEW_APC_COVERAGE = true'
```
- In order to see the edit button when changing the `apcRequested` to true on the CMS is necessary to run Algolia Sync


packages/react-components/src/organisms/ComplianceTableRow.tsx
```diff
@@ -123,24 +125,87 @@ type ComplianceTableRowProps = {
...
+
+  return (
+    <div
+      data-testid="apc-coverage"
+      css={css({ display: 'flex', alignItems: 'center' })}
+    >
+      {/* TODO: fix this once the apc coverage label is properly implemented */}
+      <p css={apcCoverageStyles}>{''}</p>
+      {isComplianceReviewer ? (
+        apcRequested ? (
```
AkosuaA, line 172: I think when apcRequested is false it should show the Edit button, right?  Not sure 😅 but from this 
Screenshot 2025-05-23 at 16 03 35

PR #4599, opened 2025-05-28
Asap 213 multiple invites and possible race condition nth

AimeurAmin:


gabiayako (approved):
Approved, but I would like @dianaguimaraes or @AkosuaA to review it too 😄

apps/crn-server/src/handlers/user/forward-invite-events.ts
gabiayako, line 1: I believe since 

- `apps/crn-server/src/handlers/user/forward-invite-events.ts`
- `apps/gp2-server/src/handlers/user/forward-invite-events.ts`

are pretty much the same, it's possible to create a factory and move the code to `packages/server-common` the same way it was done to `invite.handler.ts` 😄

gabiayako (approved):

PR #4600, opened 2025-05-29
ASAP-1000 Add url and changelog to manuscript and research outputs

gabiayako:


apps/crn-server/src/validation/research-output.validation.ts
```diff
@@ -213,6 +214,7 @@ const researchOutputPutRequestValidationSchema: JSONSchemaType<ResearchOutputPut
       description: { type: 'string', nullable: true },
       descriptionMD: { type: 'string' },
       shortDescription: { type: 'string', nullable: true, maxLength: 250 },
+      changelog: { type: 'string', nullable: true },
```
AkosuaA, line 217:
```suggestion
      changelog: { type: 'string', nullable: true, maxLength: 250 },
```

packages/react-components/src/templates/ManuscriptForm.tsx
```diff
@@ -878,6 +892,36 @@ const ManuscriptForm: React.FC<ManuscriptFormProps> = ({
               )}
             />
 
+            <Controller
+              name="url"
+              control={control}
+              rules={{
+                required: isURLRequired ? 'Please enter a URL.' : false,
```
AkosuaA, line 899: Just nitpicking here since this works fine. On the preprintDoi field we use the selected lifecycle to determine whether it should be required or not. Just wondering if doing the same here will work or not, so we don't have additional state.

AkosuaA (approved):
Looks good! 🎉

Approving because the comments left are minor suggestions.

packages/react-components/src/templates/__tests__/ManuscriptForm.other.test.tsx
```diff
@@ -390,6 +393,8 @@ describe('renders the necessary fields', () => {
 
     description: 'Please provide a description',
     shortDescription: 'Add a short description',
+    changelog:
+      'Briefly explain what’s new or changed in this version in comparison to the prior version of the manuscript.',
```
AkosuaA, line 397: sorry, I missed this. we won't need this anymore I think

PR #4624, opened 2025-06-18
ASAP-1064 enable os members to edit manuscript form

AimeurAmin:
[ASAP-1064](https://asaphub.atlassian.net/browse/ASAP-1064)


apps/crn-frontend/src/network/teams/TeamManuscript.tsx
```diff
@@ -128,6 +130,7 @@ const TeamManuscript: React.FC<TeamManuscriptProps> = ({
           onUpdate={updateManuscript}
           onResubmit={handleResubmitManuscript}
           teamId={teamId}
+          user={user}
```
gabiayako, line 133: Not a big deal in this case, but passing only the `user?.openScienceTeamMember` keeps the child component focused on what it actually needs and avoids unnecessary re-renders if the user object changes. I believe in our case the user object only changes when the user logs in, but still, I think it's a good practice to pass only the necessary field if we're using only one field in the object 😄

PR #4625, opened 2025-06-18
ASAP-1035 Enable users to submit Impact and Category details on a Manuscript

AimeurAmin:
[ASAP-1035](https://asaphub.atlassian.net/browse/ASAP-1035)


apps/crn-server/src/data-providers/contentful/manuscript.data-provider.ts
```diff
@@ -528,6 +534,27 @@ const parseGraphQLManuscript = (
       teamId,
       count,
     ),
+    impact:
+      manuscript.impact &&
+      typeof manuscript.impact.name === 'string' &&
```
AkosuaA, line 539: checking the type of impact name might not be necessary, same with category name. just nitpicking here so not necessary to change

AkosuaA (approved):
Looks good

PR #4630, opened 2025-06-27
add os member replied to discussion email notification

AkosuaA:


dianaguimaraes (approved):
LGTM. If anything I would prefer not to have OS email hardcoded everywhere, but I'm aware it wasn't introduced only in this PR

PR #4637, opened 2025-07-11
ASAP-948 Remove old teams field

gabiayako:
This PR renames the new teams field inserted in PR https://github.com/yldio/asap-hub/pull/4636 from `teams_new` to `teams` and removes the old teams field. It was also necessary to modify some queries and data providers.


AkosuaA (approved):
Looks good! 🎉

PR #4638, opened 2025-07-11
ASAP-984 Download compliance data

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-984


apps/crn-frontend/src/network/teams/api.ts
```diff
@@ -576,44 +576,15 @@ export const uploadManuscriptFileViaPresignedUrl = async (
...
-  if (!resp.ok) {
-    throw new BackendError(
-      `Failed to generate presigned URL. Expected status 200. Received status ${resp.status}: ${resp.statusText}`,
-      response,
-      resp.status,
-    );
-  }
+): Promise<string> => {
+  const { presignedUrl: downloadUrl } = await getPresignedUrl(
+    'ComplianceFullDataset.csv',
```
gabiayako, line 583: I think we can send the filename without the `.csv`. When we download the file, it's coming with a `.csv` in the middle and in the end

Image

```suggestion
    'ComplianceFullDataset',
```

packages/react-components/src/molecules/ExportButton.tsx
```diff
@@ -34,31 +35,73 @@ const exportButtonStyles = css({
...
-      </Button>
+      <strong>
+        Download:
+        {info ? (
+          <TooltipInfo
+            overrideWrapperStyles={infoWrapperStyle}
+            overrideTooltipStyles={tooltipStyle}
+          >
+            {info}
+          </TooltipInfo>
```
gabiayako, line 83: very minor, but I wonder if we couldn't use the `Tooltip` instead of `TooltipInfo` and have the tooltip appearing on hover instead of when we click on the tooltip, like in here

https://github.com/yldio/asap-hub/blob/master/packages/react-components/src/organisms/ManuscriptCard.tsx#L429:L440

gabiayako (approved):

PR #4644, opened 2025-07-22
[ASAP-988] - research output manuscript search

AkosuaA:
ref: https://asaphub.atlassian.net/jira/software/c/projects/ASAP/boards/13?selectedIssue=ASAP-988

### Notes
- I use the term `valid version` for versions in the specified lifecycles, i.e `Preprint`, `Publication`, `Publication with addendum or corrigendum`

The version sync has been implemented differently from the other entity syncs. Rather than store all versions, it stores only the latest valid version of a manuscript. So rather than loop through all versions and upload or loop through all versions and check whether it's the latest valid version of a manuscript, it goes through each manuscript to find the latest version. 
Also, rather than use the version's id in the cms as the objectID in algolia it uses a modified version of the related manuscript's id `(mv-{manuscriptid})`.  I thought this will be easier to keep track of when deleting / updating so we won't have multiple records for the same manuscript in algolia.

Also, I couldn't think of a good naming for versions that are in the valid lifecycles so if you think of more intuitive naming please suggest 😅

I tried to breakdown the various flows here (screenshots below):
Link to excalidraw: https://excalidraw.com/#json=SPPkiZWfEfy7LdCWoLR7d,ZmurraA1BLBD6Z6N9mPKAg

Screenshot 2025-07-25 at 13 38 51
<br />
Screenshot 2025-07-25 at 13 39 07
Screenshot 2025-07-25 at 13 39 37


apps/crn-server/src/data-providers/contentful/manuscript-version.data-provider.ts
```diff
@@ -0,0 +1,147 @@
...
+    const { manuscriptVersions } = await this.contentfulClient.request<
+      FetchManuscriptVersionByIdQuery,
+      FetchManuscriptVersionByIdQueryVariables
+    >(FETCH_MANUSCRIPT_VERSION_BY_ID, {
+      id,
+    });
+
+    if (!manuscriptVersions) {
+      return {
+        versionFound: false,
```
AkosuaA, line 104: I added the versionFound found field to differentiate between when the version isn't found at all which should throw an error verses it was found but not linked to a manuscript or the linked manuscript doesn't have a recent version. It feels weird though so do let me know if there's a cleaner way to do this

gabiayako: It looks good to me the way it is :)

apps/crn-server/scripts/export-entity.ts
```diff
@@ -37,35 +39,42 @@ export const exportEntity = async (
...
+          2,
+        ).slice(1, -1),
+      );
 
-    page++;
-    recordCount += records.items.length;
-  } while (total > recordCount);
+      page++;
+      recordCount += records.items.length;
+    } while (total > recordCount);
```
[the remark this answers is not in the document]
AkosuaA: yea, the only difference is the while condition but no need to redo the whole loop just for that :sweat_smile: Will update

gabiayako (approved):
really nice! 👏👏👏

PR #4672, opened 2025-09-02
ASAP-990 Create article version from manuscript

dianaguimaraes:
Ticket: https://asaphub.atlassian.net/browse/ASAP-990


packages/react-components/src/organisms/SharedResearchOutputButtons.tsx
```diff
@@ -121,7 +127,22 @@ const SharedResearchOutputButtons: React.FC<
...
+              noMargin
+              small
+              primary
+              onClick={checkForNewerManuscriptVersion}
+            >
+              <VersionIcon /> Import Manuscript Version
+            </Button>
+          </div>
+        )}
+      {canVersionResearchOutput && published && !hasRelatedManuscript && (
```
gabiayako, line 145: maybe we can add `!isManuscriptOutputFlagEnabled` here? because otherwise the `Add version` button just disappears when the flag is not enabled

PR #4687, opened 2025-09-18
[ASAP-1040] - Collaboration-  Display  'Preliminary Findings' data

AimeurAmin:
[ASAP-1040](https://asaphub.atlassian.net/browse/ASAP-1040)


apps/crn-frontend/src/analytics/collaboration/api.ts
```diff
@@ -47,3 +50,68 @@ export const getUserCollaborationPerformance =
...
+      {
+        'teamName.keyword': {
+          order: 'asc',
+        },
+      },
+    ],
+  };
+
+  const response =
+    await opensearchClient.request<
```
AkosuaA, line 108: Rather than use the request method directly, there's a search method where you pass the search options and it builds the query, makes the request and returns the parsed response, see [here](https://github.com/yldio/asap-hub/blob/master/apps/crn-frontend/src/analytics/leadership/api.ts#L40). 
The only thing is at the time it was added the time range bit hadn't been added. I've just added that in my [pr](https://github.com/yldio/asap-hub/pull/4686/files#diff-205ffc206c35c6c67ed6c19abcf581c575007a38ac21b4034f02dd0cc4eec804)

packages/react-components/src/organisms/SharingPrelimFindingsTable.tsx
```diff
@@ -131,7 +140,17 @@ const SharingPrelimFindingsTable: React.FC<SharingPrelimFindingsTableProps> = ({
                   </p>
                 </td>
                 <td>
-                  <p>{row.teamPercentShared}%</p>
+                  <p css={iconStyles}>
+                    <span css={valueStyles}>
+                      {!row.teamPercentShared || row.limitedData
```
gabiayako, line 145: I believe we should remove this `!row.teamPercentShared`. Because the data can come with `teamPercentShared: 0` and  `limitedData: true` which can cause

Image

apps/crn-frontend/src/analytics/collaboration/api.ts
gabiayako: I forgot this `buildOpensearchQuery` existed. I'm not against reusing because so far the queries are very similar. What do you thin @dianaguimaraes ?

AimeurAmin: I tried using the `buildOpensearchQuery`, unfortunately looks like the data mapping is different between `os-champion` & `preliminary-data-sharing` when it comes to `timeRange`, please check the images down below:

image

image


having `timeRange` as `text` means this part of the query won't work:

image

AkosuaA: Yea, the os-champion timeRange was text previously but I changed that in this pr to keyword based on https://docs.opensearch.org/latest/field-types/supported-field-types/text/.

dianaguimaraes: Cool, if we make it consistent, which should be the case, I'm all in for reusability.

gabiayako (approved):

PR #4699, opened 2025-10-01
[ASAP-1078] - update FormCard component to match designs

damianmr:
Please check https://asaphub.atlassian.net/browse/ASAP-1078 to get a list of impacted areas/URLs.

The way I'd approach this PR is by first having a look at the `atoms` and `molecules` folder:

image

In these folders, I'm adding support for the `noPadding` property across multiple components that didn't have support for this attribute. I've done it in a similar way to other existing components that do have support for this attribute (comments are welcome). Most of the changes in these folders are backwards compatible, i.e. the changes won't affect existing use-cases. The only exceptions are the molecules that use a `Paragraph` to display a title/subtitle. The default `Paragraph` component has a default margin on the top/bottom which contributed to the overall problem of the aggregated white-spacing.

image

Before/after
Screenshot 2025-10-01 at 12 57 16
Screenshot 2025-10-01 at 12 59 40

This tiny difference may cause existing forms using these components to render slightly differently after the removal of the top margin. I don't think it will cause a lot of trouble and it can be solved on a case by case basis, but please let me know if this is an issue.

After checking those folders, the changes in organisms and templates aim to make use of the `noPadding` attribute in all the fields rendered within the `FormCard` (and derivatives) and also set the gaps between those components and the save buttons.

image

FormCard Before/After

before

after

### DIFF
--- a/packages/gp2-components/src/templates/OutputForm.tsx
+++ b/packages/gp2-components/src/templates/OutputForm.tsx
@@ -61,6 +61,9 @@ const DOC_TYPES_COHORTS_NOT_REQUIRED: gp2Model.OutputDocumentType[] = [
   'Code/Software',
 ];
 
+/** Auxiliary empty space to match designs */
+const EmptyGap = () => <div />;
+
 export const getRelatedOutputs = (
   relatedOutputs: gp2Model.OutputResponse['relatedOutputs'],
 ) =>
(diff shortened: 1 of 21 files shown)


damianmr:
Closing as we discussed have a different approach about the `noPadding` entirely

PR #4966, opened 2026-03-20
[ASAP-1412] - Compliance Live Sheet Sync

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1412

To test the on demand sync run the [Migrate Impact and Categories](https://github.com/yldio/asap-hub/actions/workflows/on-demand-migrate-impact-categories.yml) workflow with this pr's branch. eg: https://github.com/yldio/asap-hub/actions/runs/25064301658

For testing, this is the Sheet to use: https://docs.google.com/spreadsheets/d/1YsA9nnlolG092XDn5IXt_G1Mx_aln4k8co_caPcSpiA/edit?gid=0#gid=0 

Before merging I will need to do the ff:
- update the serverless file to only create the added queues and handlers only for the prod environment
- revert changes to .github/workflows/on-demand-migrate-impact-categories.yml
- uncomment the notify_failure job in .github/workflows/reusable-compliance-sheet-sync.yml
- update `CONTENTFUL_ENV_ID`, `ENVIRONMENT` and `GOOGLE_API_CREDENTIALS_SECRET_ID` vars in .github/workflows/reusable-compliance-sheet-sync.yml
- change references to 'Sheet1' to appropriate sheet name


apps/crn-server/scripts/sync-compliance-sheet.ts
```diff
@@ -0,0 +1,111 @@
...
+const writeInChunks = async (
+  sheets: Awaited<SheetsClient>,
+  rows: string[][],
+) => {
+  for (let i = 0; i < rows.length; i += CHUNK_SIZE) {
+    const chunk = rows.slice(i, i + CHUNK_SIZE);
+
+    await sheets.spreadsheets.values.update({
+      spreadsheetId,
+      range: 'Sheet1!A2',
```
damianmr: I suppose that if you're writing in chunks, you also need to account for the chunk's size when updating values in the sheet. Otherwise, you'll end up replacing the whole `A` column for every chunk.

For example:

`range: `Sheet1!A${i + 2}`

apps/crn-server/test/handlers/compliance-sheet/compliance-spreadsheet-sync-handler.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    } as any;
+
+    await handler(event);
+
+    expect(sheetsMock.spreadsheets.values.batchUpdate).toHaveBeenCalledWith(
+      expect.objectContaining({
+        requestBody: expect.objectContaining({
+          data: expect.arrayContaining([
+            expect.objectContaining({
+              values: [Array(48).fill('')],
```
damianmr, line 174: Just want to confirm: so, if the manuscript gets deleted or unpublished, we essentially blank out the row? i.e. the row isn't deleted, just filled out with empty cells.

apps/crn-server/serverless.ts
```diff
@@ -1282,6 +1310,60 @@ const serverlessConfig: AWS = {
...
+      environment: {
+        SENTRY_DSN: sentryDsnHandlers,
+        COMPLIANCE_DOC_SYNC_QUEUE_URL: { Ref: 'ComplianceDocSyncQueue' },
+      },
+    },
+    complianceSpreadsheetSyncHandler: {
+      handler:
+        './src/handlers/compliance-sheet/compliance-spreadsheet-sync-handler.handler',
+      timeout: 40,
+      reservedConcurrency: 2,
```
damianmr: Question: For this handler, don't you think it may be beter to reduce concurrency to 1 to prevent race conditions when multiple updates are made at once? I'm thinking about bulk actions 🤔 Same for `maximumConcurrency` below.

apps/crn-server/serverless.ts
```diff
@@ -1282,6 +1310,60 @@ const serverlessConfig: AWS = {
...
+    complianceSpreadsheetEntryHandler: {
+      handler:
+        './src/handlers/compliance-sheet/compliance-entry-handler.handler',
+      events: [
+        {
+          eventBridge: {
+            eventBus: 'asap-events-${self:provider.stage}',
+            pattern: {
+              source: [eventBusSourceContentful],
+              'detail-type': [
```
damianmr: Does it make sense to add these?

```
LabsPublished,
LabsUnpublished,
ExternalAuthorsPublished,
ExternalAuthorsUnpublished
```

apps/crn-server/src/data-providers/contentful/manuscript-version.data-provider.ts
```diff
@@ -116,6 +133,114 @@ export class ManuscriptVersionContentfulDataProvider
...
+
+    const searchQuery: ManuscriptVersionsFilter = {
+      sys: {
+        ...(ids.length ? { id_in: ids } : {}),
+      },
+    };
+
+    const variables: FetchComplianceManuscriptVersionsQueryVariables = {
+      where: searchQuery,
+      ...(ids.length ? {} : { limit: take, skip }),
```
damianmr, line 227: I think using the `limit: take` here would make the code in
apps/crn-server/src/handlers/compliance-sheet/compliance-spreadsheet-sync-handler.ts

Image

Return always 8 results. You could make the property `limit` use the ids passed in the `filters` property:

`limit: ids.length` 🤔

damianmr: Please verify if what I'm saying is correct 😁

apps/crn-server/test/handlers/compliance-sheet/compliance-spreadsheet-sync-handler.test.ts
AkosuaA: yep

apps/crn-server/src/data-providers/contentful/manuscript-version.data-provider.ts
AkosuaA: we only use the limit when there are no ids `...(ids.length ? {} : { limit: take, skip }),`

apps/crn-server/serverless.ts
AkosuaA: I didn't add those as well as the impact and categories because I don't expect them to change much but it can happen so will add them

apps/crn-server/serverless.ts
AkosuaA: Makes sense to reduce to 1

damianmr (approved):
Went through the changes in a huddle, all good now! 🚀 

@AkosuaA just needs to update the settings to run this in production

PR #4986, opened 2026-04-10
[ASAP-1455] - Sync Aims and Milestones data in OpenSearch

AimeurAmin:
[ASAP-1455](https://asaphub.atlassian.net/browse/ASAP-1455)

## Summary

This PR implements event-based synchronisation of Aims and Milestones data to OpenSearch. Previously, these indices were only updated by a scheduled full reindex. Now, any change to an Aim, Milestone, Project, or Supplement Grant in Contentful is immediately reflected in OpenSearch.

### Event pipeline

Two new Lambda handlers listen to EventBridge events:

**`opensearchIndexAims`** handles:

- `AimsPublished` — rebuilds the aim document and its linked milestone documents
- `AimsUnpublished` — deletes the aim document and its linked milestone documents
- `ProjectsPublished` — rebuilds all aims and milestones for the project
- `ProjectsUnpublished` — deletes all aims and milestones for the project
- `SupplementGrantPublished` / `SupplementGrantUnpublished` — finds the parent project, then rebuilds all its aims and milestones

**`opensearchIndexMilestones`** handles:

- `MilestonesPublished` — rebuilds the milestone document and all aims linked to it (because aim status and article counts derive from milestones)
- `MilestonesUnpublished` — rebuilds linked aims, then deletes the milestone document

### New files

| File                                                              | Purpose                                                                           |
| ----------------------------------------------------------------- | --------------------------------------------------------------------------------- |
| `src/handlers/aim/opensearch-index-aim-handler.ts`                | Lambda handler for aim/project/supplement grant events                            |
| `src/handlers/milestone/opensearch-index-milestone-handler.ts`    | Lambda handler for milestone events                                               |
| `src/handlers/opensearch/aims-milestones-reindex.ts`              | Core reindex logic — builds documents and calls OpenSearch                        |
| `src/data-providers/contentful/aims-milestones.data-provider.ts`  | Contentful queries for targeted lookups (by ID, reverse lookups via `linkedFrom`) |
| `src/data-providers/types/aims-milestones.data-provider.types.ts` | TypeScript interface for the data provider                                        |
| `packages/contentful/src/crn/queries/aims-milestones.queries.ts`  | GraphQL query definitions                                                         |

### How the reindex works

Each reindex function follows a **delete-then-reinsert** pattern:

1. Fetch the current state from Contentful
2. Delete existing documents from OpenSearch that match the affected IDs
3. Build fresh documents from the Contentful data
4. Upsert the new documents into OpenSearch

This pattern avoids duplicates and stale data. The delete step uses `deleteByQuery` matching on the document's `id` field (not the OpenSearch `_id`), which is critical because the full reindex script uses auto-generated `_id` values while event-based sync uses `_id: doc.id`. Deleting by document `id` works regardless of which indexing strategy created the original document.

### Aim ordering

Added an `aimOrder` field to aim documents. This is a 1-based integer representing the aim's position within its collection in Contentful (original grant aims and supplement grant aims are ordered independently). The OpenSearch query now sorts by `aimOrder` instead of `createdDate`, so the frontend displays aims in the same order as Contentful.

The sort includes `missing: '_last'` to handle the transition period before the first full reindex populates the field on all existing documents.

### Webhook configuration

Added `aims`, `milestones`, and `supplementGrant` to `ContentfulWebhookPayloadType` so that Contentful publishes webhook events for these content types. The event flow is:

```
Contentful publish → Webhook → SQS → Poller → EventBridge → Lambda handler
```

### OpenSearch utilities

Added two new functions to `@asap-hub/server-common`:

- `upsertOpensearchDocuments(client, index, docs)` — bulk upsert using `_id: doc.id`
- `deleteByDocumentIds(client, index, ids)` — delete by querying the `id` field in the document body, works regardless of `_id` strategy

### What each event triggers

| Event                                  | Aims index                                        | Milestones index                                        |
| -------------------------------------- | ------------------------------------------------- | ------------------------------------------------------- |
| Aim published                          | Rebuild this aim                                  | Delete + reinsert milestones linked to this aim         |
| Aim unpublished                        | Delete this aim                                   | Delete milestones linked to this aim                    |
| Milestone published                    | Rebuild aims linked to this milestone             | Rebuild this milestone                                  |
| Milestone unpublished                  | Rebuild aims linked to this milestone             | Delete this milestone                                   |
| Project published                      | Delete + reinsert all aims for this project       | Delete + reinsert all milestones for this project       |
| Project unpublished                    | Delete all aims for this project                  | Delete all milestones for this project                  |
| Supplement grant published/unpublished | Delete + reinsert all aims for the parent project | Delete + reinsert all milestones for the parent project |

### Changes to full reindex script

- Added `aimOrder` field to `ProjectAimsDataObject` type and OpenSearch mapping
- `exportAimsData` now populates `aimOrder` based on the aim's position in the Contentful collection

### 1. Unpublish paths cannot fetch the unpublished entity

When Contentful unpublishes an entity, the CDA (Content Delivery API) no longer returns it. This means any handler triggered by an `*Unpublished` event cannot fetch the entity that was just unpublished. `deleteByProjectId` solves this by querying OpenSearch directly (using `deleteByQuery` on the `projectId` field) instead of fetching aim/milestone IDs from Contentful. For `AimsUnpublished`, `deleteMilestonesByAimId` gracefully returns early if the aim is gone, deferring milestone cleanup to the next full reindex.

### 2. AimsPublished reindexes the entire project, not just the aim

When an aim is published, milestones may have been added or removed from it. Milestones removed from the aim still exist in Contentful but their `aimNumbers` need updating. Rather than trying to diff what changed, `AimsPublished` resolves the parent project via `fetchProjectWithAimsDetailByAimId` and calls `reindexByProjectId`, which deletes and rebuilds all aims and milestones for the project. This ensures removed milestone links are cleaned up.

### 3. Shared milestones are reindexed, not deleted, on aim unpublish

Milestones can be linked to multiple aims. When an aim is unpublished, `deleteMilestonesByAimId` checks each linked milestone: if it's still referenced by other aims, it reindexes the milestone (updating its `aimNumbers` to exclude the unpublished aim); if no other aims reference it, it deletes the milestone from OpenSearch. This prevents over-deletion of milestone documents that are still valid.

[ASAP-1455]: https://asaphub.atlassian.net/browse/ASAP-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/server-common/src/utils/opensearch.ts
```diff
@@ -234,6 +234,106 @@ const updateAliasAndCleanup = async (
...
+    const errors = response.body.items
+      .filter(
+        (item: { index?: { error?: { reason?: string } } }) =>
+          item.index?.error,
+      )
+      .map(
+        (item: { index?: { _id?: string; error?: { reason?: string } } }) =>
+          `${item.index?._id}: ${item.index?.error?.reason}`,
+      );
+    console.error(`Upsert errors: ${errors.join(', ')}`);
```
damianmr: @AimeurAmin I'm just checking: is it worth it to throw an error instead of just the logging? i.e. 

```ts
throw new Error(`Upsert errors: ${errors.join(', ')}`
```

This function (and the one below) are called by other event handlers which are all surrounded by the Sentry wrapper. So, throwing here would allow us to see the error in Sentry. 

Maybe you considered this and thought it was not necessary, but since I'm seeing error processing code I assume it's also worth seeing this in Sentry in order to track it properly.

packages/server-common/src/utils/opensearch.ts
```diff
@@ -234,6 +234,106 @@ const updateAliasAndCleanup = async (
...
+    const errors = response.body.items
+      .filter(
+        (item: { delete?: { error?: { reason?: string } } }) =>
+          item.delete?.error,
+      )
+      .map(
+        (item: { delete?: { _id?: string; error?: { reason?: string } } }) =>
+          `${item.delete?._id}: ${item.delete?.error?.reason}`,
+      );
+    console.error(`Delete errors: ${errors.join(', ')}`);
```
damianmr: Same as https://github.com/yldio/asap-hub/pull/4986/changes#r3078652360

packages/server-common/src/utils/opensearch.ts
```diff
@@ -234,6 +234,106 @@ const updateAliasAndCleanup = async (
...
+ * Useful for cleaning up related documents when the parent entity
+ * is no longer available in the source API (e.g. after unpublish).
+ */
+export const deleteByFieldValue = async (
+  client: Client,
+  indexAlias: string,
+  field: string,
+  value: string,
+): Promise<void> => {
+  await client.deleteByQuery({
```
damianmr: Here and in `deleteByDocumentIds` we could have the error handling strategy we used for `deleteOpensearchDocuments` (and maybe extract all of that into a common function, if possible).

I'm not super picky on this, if you think is not necessary then it's fine but at least a simple check for errors with a `throw` will give us some visibility in Sentry (and Slack)

apps/crn-server/serverless.ts
```diff
@@ -1424,6 +1424,61 @@ const serverlessConfig: AWS = {
...
+          eventBridge: {
+            eventBus: 'asap-events-${self:provider.stage}',
+            pattern: {
+              source: [eventBusSourceContentful],
+              'detail-type': [
+                'AimsPublished',
+                'AimsUnpublished',
+                'ProjectsPublished',
+                'ProjectsUnpublished',
+                'ProjectMembershipPublished',
```
damianmr: `ProjectMembershipPublished` and `ProjectMembershipUnpublished` are not being handled at all in `./src/handlers/aim/opensearch-index-aim-handler.handler`. Do we need to receive events from these sources?

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
```diff
@@ -0,0 +1,478 @@
...
+        ) ??
+        project.supplementGrant?.aimsCollection?.items?.find(
+          (a) => a?.sys?.id === aimId,
+        );
+
+      if (aimDetail) {
+        const msIds =
+          aimWithMilestones.milestonesCollection?.items
+            ?.filter((m): m is NonNullable<typeof m> => m !== null)
+            .map((m) => m.sys.id) ?? [];
```
damianmr: This fragment repeats multiple times with minor differences, let's move it into a private utility function in this file

damianmr:
```ts
const getMilestoneIds = (aim: AimWithMilestonesDataObject | null): string[] =>
  aim?.milestonesCollection?.items
    ?.filter((m): m is NonNullable<typeof m> => m !== null)
    .map((m) => m.sys.id) ?? [];
```

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
```diff
@@ -0,0 +1,478 @@
...
+          (m): m is MilestoneDataObject => m !== null,
+        );
+
+        const doc = buildAimDocument(
+          aimDetail,
+          project,
+          grantType,
+          validMilestones,
+          msIds,
+        );
```
damianmr: This `doc` construction can also be extracted out to a private function and then reused in `buildAndUpsertAim`

Image

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
```diff
@@ -0,0 +1,478 @@
...
+
+  const aimIds = allAimDetails.map((a) => a.sys.id);
+
+  const client = await getOpensearchClient();
+
+  // Delete ALL existing aims and milestones for this project by projectId.
+  // Using projectId field (not document IDs) ensures we also remove aims/milestones
+  // that were previously linked but have since been unlinked from the project.
+  await deleteByFieldValue(client, AIMS_INDEX, 'projectId', projectId);
+  await deleteByFieldValue(client, MILESTONES_INDEX, 'projectId', projectId);
```
damianmr: This could probably just call `deleteByProjectId`, right? i.e. left over from refactor.

damianmr (changes requested):
I still need to give thorough testing to the plumbing of everything but I noted a few issues in the code that require either a 
1. Confirmation of the intentions
2. or small refactors

I'll keep testing this in the afternoon, thanks!

packages/server-common/src/utils/opensearch.ts
AimeurAmin: Good call 👌🏻 the Sentry wrapper will pick it up and we get proper alerting instead of silent `console.error`. Will change both `upsertOpensearchDocuments` and `deleteOpensearchDocuments` to throw.

packages/server-common/src/utils/opensearch.ts
AimeurAmin: check previous comment: https://github.com/yldio/asap-hub/pull/4986#discussion_r3079372763

packages/server-common/src/utils/opensearch.ts
AimeurAmin: Added throwIfDeleteByQueryHasFailures helper. Both deleteByDocumentIds and deleteByFieldValue now throw when the response contains failures[] entries (partial-failure case that deleteByQuery doesn't surface as a client error).

apps/crn-server/serverless.ts
AimeurAmin: we don't actually need them. Team name changes on aims/milestones are low-priority and the scheduled full reindex catches them. Removed `ProjectMembershipPublished`/`ProjectMembershipUnpublished` from the event pattern in `serverless.ts` so we don't receive events we ignore.

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
AimeurAmin: Extracted into `getMilestoneIdsFromAim(aim)` helper.

apps/crn-server/src/data-providers/contentful/aims-milestones.data-provider.ts
```diff
@@ -168,4 +174,106 @@ export class AimsMilestonesContentfulDataProvider
...
+        linkedFrom?: {
+          projectsCollection?: {
+            items: Array<{ sys: { id: string } } | null>;
+          } | null;
+        } | null;
+      } | null;
+    }>(FETCH_PROJECT_ID_BY_SUPPLEMENT_GRANT_ID, { supplementGrantId });
+
+    return (
+      supplementGrant?.linkedFrom?.projectsCollection?.items?.find(Boolean)?.sys
```
diegonvs: nit: Hey, when I first read this find(Boolean) It took me a while and I needed to search how it works. Maybe we can separate this logic into a new function, something like:

```typescript
const getFirstValid = <T>(items?: (T | null)[] | null): T | null =>
  items?.find(Boolean) ?? null;
```

and then reuse it.

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
AimeurAmin: the two inline deleteByFieldValue calls are functionally identical to what `deleteByProjectId` already does, so calling it directly is cleaner. For context: this wasn't leftover, I'd switched from `deleteByDocumentIds(currentAimIds)` to `deleteByProjectId` in the previous commit to fix a stale-aims bug (if aims were unlinked from a project in `Contentful`, their OpenSearch docs were never cleaned up because they didn't appear in the current aimIds list). But yes, it's the same operation as `deleteByProjectId` — consolidated.

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
AimeurAmin: Done — extracted the doc construction out of `buildAndUpsertAim` into a separate private `buildAimDocumentForAim` function.  `buildAndUpsertAim` now just calls it and upserts the returned doc. 
Also applied the same split on the milestone side for symmetry (`buildMilestoneDocumentForMilestone` + `buildAndUpsertMilestone`).

apps/crn-server/src/data-providers/contentful/aims-milestones.data-provider.ts
AimeurAmin: extracted it to `@asap-hub/server-common` as `getFirstValid` so we can reuse it across data providers 👍

Went with a slightly different implementation than `find(Boolean)` for better typing:                    
```
  export const getFirstValid = <T>(
    items?: ReadonlyArray<T | null | undefined> | null,
  ): T | null => items?.find((item): item is T => item != null) ?? null;
 ```

Two reasons:                                                                                           
1. `Boolean` isn't typed as a type predicate in TS, so find(Boolean) returns `T | null | undefined`; callers would still need `?? null` or `!` to satisfy the compiler, which defeats the point of the helper.  The explicit type guard narrows the return to `T | null` honestly, no cast needed.                     
2. `item != null` only excludes `null/undefined`, so falsy-but-valid values (`0`, `''`, `false`) are preserved if `T` ever includes them; Boolean would wrongly drop those.

**Review by damianmr — APPROVED** [review]
All good, my comments were addressed. I verified the sync by going to Contentful, updating an aim/milestone and then checking we would see the data as expected in the hub. The quick verifications I ran in Opensearch worked as expected.

LGTM!

**Review by diegonvs — APPROVED** [review]

File: apps/crn-server/src/handlers/milestone/opensearch-index-milestone-handler.ts [hunk 9]
```diff
@@ -0,0 +1,34 @@
...
+  async (event) => {
+    const eventType = event['detail-type'];
+    const milestoneId = event.detail.resourceId;
+    logger.debug(`Event ${eventType} for milestone ${milestoneId}`);
+
+    if (eventType === 'MilestonesPublished') {
+      await reindexMilestoneById(provider, milestoneId);
+      await reindexAimsByMilestoneId(provider, milestoneId);
+    } else {
+      await reindexAimsByMilestoneId(provider, milestoneId);
```
AkosuaA: for MilestonesUnpublished since no aims will be returned maybe there's no need to call this

apps/crn-server/src/handlers/opensearch/aims-milestones-reindex.ts
```diff
@@ -0,0 +1,499 @@
...
+ *
+ * Note: if the aim is already gone from the delivery API, we cannot resolve
+ * its milestone IDs and rely on the scheduled full reindex.
+ */
+export const deleteMilestonesByAimId = async (
+  provider: AimsMilestonesDataProvider,
+  aimId: string,
+): Promise<void> => {
+  const aim = await provider.fetchAimWithMilestonesById(aimId);
+  if (!aim) {
```
AkosuaA: we only call this when an aim is unpublished. wouldn't aim then always be null?

AkosuaA (commented):
This works, especially since the daily sync resolves any inconsistencies from unpublish operations.

One concern is that publishing an aim currently reindexes all aims and milestones for the project, which feels quite heavy. If we stored related entry IDs in OpenSearch (e.g. `milestoneIds` on aims and `aimIds` on milestones), then on publish, we could take a union of linked IDs from both Contentful and OpenSearch and only reindex those. It would also help with unpublish, as in, we could use OpenSearch to look up previously linked entries and resync. 

However, the current approach is simpler, unpublishing should be rare and I guess if we're expecting max 10 aims per project and around 50 milestones max (avg 20; based on the team milestones sheet) it's ok. We will have to manage the client's expectation with the Hub when creating milestones but that's more a separate UX conversation. 

Overall I’m fine with keeping the current approach given the expected scale.

PR #4987, opened 2026-04-13
[ASAP-1422] - create milestones

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1422

Notes:
- To test the entire flow, for now, you'd have to run the app locally with the `crn-4986` environment since that has the updates sync enabled. Once that pr is merged, testing in the pr environment should be fine.
- Added milestone-created to the existing ManuscriptToast since the behaviour is identical and we can reuse the same provider. This does make the naming a bit misleading, so I’ll follow up with a separate PR to generalise/rename it.

- There is currently a setTimeout delay after creating a milestone before invalidating the milestones list. I added this as a workaround since the newly created milestone is not immediately returned by the API but it's not reliable. Raising this so we discuss alternatives.


apps/crn-frontend/src/projects/ProjectMilestones.tsx
```diff
@@ -84,12 +98,36 @@ const ProjectMilestones: React.FC<{
...
+        type: 'milestone-created',
+        accent: 'successLarge',
+      });
+    } catch (e) {
+      setFormType({
+        type: 'default-error',
+        accent: 'error',
+      });
+    } finally {
+      window.scrollTo({ top: 0, behavior: 'smooth' });
```
AkosuaA: This is currently not behaving as expected after milestone creation. I'm not sure why 🤔

damianmr: Don't you need to defer it? My guess is that the scroll animation is trying to happen as the component is being remounted.

```ts
setTimeout(() => window.scrollTo({ top: 0, behavior: 'smooth'}), 0); // try with 100ms as well.
```

AimeurAmin: `setTimeout` should do the trick but I'd rather have a small custom hook that triggers the scroll, it can also be used every time we need that

**useScrollToTop.tsx**
```
export const useScrollToTop = () => {
  const [trigger, setTrigger] = useState(false);

  useEffect(() => {
    if (trigger) {
      window.scrollTo({ top: 0, behavior: 'smooth' });
      setTrigger(false);
    }
  }, [trigger]);

  return () => setTrigger(true);
}

```

**Usage:**
```
const scrollToTop = useScrollToTop();
```

**in finally:**
```
...

} finally {
  scrollToTop();
}
```

apps/crn-frontend/src/projects/api.ts
```diff
@@ -190,7 +191,7 @@ export const getProjectMilestones = async (
   const url = createListApiUrl(`projects/${projectId}/milestones`, {
     ...searchOptions,
   });
-  if (grantType) {
+  if (grantType && grantType !== 'all') {
```
diegonvs: The `'all'` sentinel seems unnecessary. selectedGrantType is always `'original'` or `'supplement'`, so `'all'` is never used at runtime.

If the goal is deterministic Recoil keys, `grantType: GrantType` is enough. The `'all'` branch becomes dead code.

Is this for a future "show all" UI tab?

apps/crn-frontend/src/projects/ProjectDetail.tsx
```diff
@@ -38,6 +37,14 @@ const ProjectDetail: FC<Props> = ({ config }) => {
   const isStaff = user?.role === 'Staff';
   const { hash: targetManuscript } = useLocation();
 
+  const teamId =
+    projectDetail && 'teamId' in projectDetail && projectDetail.teamId
+      ? projectDetail.teamId
+      : undefined;
```
diegonvs, line 43:
```suggestion
const teamId = projectDetail && 'teamId' in projectDetail ? projectDetail.teamId : undefined;
```

I think the second verification is unnecessary, does makes sense?

AimeurAmin: @diegonvs 
it is necessary if we want to have a truthy value on teamId.
`'teamId' in projectDetail` only ensures that the property is present.

apps/crn-server/src/routes/project.route.ts
```diff
@@ -98,6 +99,38 @@ export const projectRouteFactory = (
...
 
+  projectRoutes.post<{ projectId: string }>(
+    '/projects/:projectId/milestones',
+    async (req, res) => {
+      const { projectId } = validateProjectParameters(req.params);
+      const data = validateProjectMilestoneCreateRequest(req.body);
+
+      if (!req.loggedInUser) {
+        throw Boom.unauthorized();
+      }
```
AimeurAmin: it would be better to check the use auth first

```suggestion
      if (!req.loggedInUser) { 
        throw Boom.unauthorized();
      }
      
      const { projectId } = validateProjectParameters(req.params);
      const data = validateProjectMilestoneCreateRequest(req.body);
```

packages/react-components/src/templates/CreateMilestoneModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+  grantType,
+  onCancel,
+  onSubmit,
+  getArticleSuggestions,
+}) => {
+  const articleSelectComponents =
+    createArticleSelectComponents<ResearchOutputOption>({
+      getIcon: () => articleIcon,
+      showArticlePill: (data) => !!data.type,
+    });
```
AimeurAmin: Creating a component dynamically inside another one is bad practice in react, it surely will rebuild `articleSelectComponents` on every render (state change, props change)

The easiest way to avoid issues in this case is to wrap  it with useMemo to avoid  `react-select` re-mount issues (loss of focus, flicker)

```suggestion
  const articleSelectComponents = useMemo(
    createArticleSelectComponents<ResearchOutputOption>({
      getIcon: () => articleIcon,
      showArticlePill: (data) => !!data.type,
    }),
  []);
```

AimeurAmin: PS: if articleSelectComponents is not dependent on any of the props of the component maybe declare it outside of the component?

packages/react-components/src/atoms/PillSelector.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { colors, noop } from '..';
+import { lead, paper, steel, tin } from '../colors';
+import { borderWidth } from '../form';
+import { rem } from '../pixels';
+
+const containerStyles = css({
+  padding: rem(6),
+  display: 'flex',
+  flexWrap: 'wrap',
+  gap: 8,
```
AimeurAmin:
```suggestion
  gap: rem(8),
```

AimeurAmin (approved):
My biggest concern is the hard-coded 5-second wait after milestone creation — race with OpenSearch indexing

`apps/crn-frontend/src/projects/state.ts:316-323`
```
  await new Promise<void>((resolve) => { setTimeout(resolve, 5000); });
  invalidateProjectMilestonesIndex();
```

- Every successful create blocks the UI 5s regardless of whether indexing is ready; 5s is ALSO not guaranteed to be enough.
So basically this is a race disguised as a fix and will cause flaky "my milestone didn't appear" bug reports from client. AC03 ("Then the new Milestone … are displayed in the table") is not reliably satisfied — it's hoped.

- Proper fix (for later): return the created milestone payload from the server, optimistically upsert it into projectMilestonesListItemState/index, and then invalidate. Or poll OpenSearch until visible. 

- At minimum (please add on this PR): document the workaround with a TODO and a linked ticket.

apps/crn-frontend/src/projects/api.ts
diegonvs: I can't see it as a blocker. I'll resolve it and later we can work on this.

diegonvs (approved):

PR #4990, opened 2026-04-15
[ASAP-1373] Allow fuzzy search of milestones

damianmr:
https://asaphub.atlassian.net/browse/ASAP-1373

Can be reviewed commit by commit!


apps/crn-frontend/src/projects/ProjectMilestones.tsx
```diff
@@ -68,14 +88,26 @@ const ProjectMilestones: React.FC<{
...
-  );
+  const selectedGrantType: GrantType = (() => {
+    if (!hasSupplementGrant) {
+      return 'original';
+    }
+
+    const grantType = new URLSearchParams(search).get('grantType');
+
+    return grantType === 'original' ? 'original' : 'supplement';
+  })();
```
diegonvs:
```suggestion
const selectedGrantType: GrantType =                                                                          
    hasSupplementGrant && new URLSearchParams(search).get('grantType') !== 'original'                           
      ? 'supplement'                                                                                            
      : 'original';
```

You can use a ternary.

An IIFE forces the interpreter to parse the implementation before understanding what's being computed, while providing no actual benefit, inside a React component you already have function scope, so there's nothing to isolate.

apps/crn-server/src/data-providers/contentful/project.data-provider.ts
```diff
@@ -885,10 +886,16 @@ export class ProjectContentfulDataProvider implements ProjectDataProvider {
...
+    const normalizedSearch = search?.trim();
+    const statusFilters = (filter ?? []).filter(
+      (value): value is MilestoneStatus =>
+        milestoneStatuses.includes(value as MilestoneStatus),
+    );
     const filters = [
       { term: { projectId: id } },
       ...(grantType ? [{ term: { grantType } }] : []),
+      ...(statusFilters.length ? [{ terms: { status: statusFilters } }] : []),
     ];
```
diegonvs:
```suggestion
    const { take = 10, skip = 0, grantType, search, filter } = options;
    const normalizedSearch = search?.trim();
    const statusFilters = (filter ?? []).filter(
      (value): value is MilestoneStatus =>
        milestoneStatuses.includes(value as MilestoneStatus),
      );
     const filters = cleanArray([
        { term: { projectId: id } },
        grantType ? { term: { grantType } } : undefined,
        statusFilters.length ? { terms: { status: statusFilters } } : undefined,
  ]);
```

We can use clearArray function

damianmr:
@diegonvs ready to have another look, you can see the last 3 commits one by one. 

- First commit addresses the IIIFE

- The second commit addresses an unrelated refactor that:
  - adds support for using `undefined` as part of the arguments of cleanArray (so that it filters out `undefined` args as well...)
  - moves the utility into `server-commons` which is a more predictable place for it to be and updates references to that function across multiple places.

- Third commit rewrites the data provider with your suggestion

diegonvs (approved):
Image

PR #4992, opened 2026-04-17
[ASAP-1423] - Associate Articles to Milestones

diegonvs:
This PR depends on https://github.com/yldio/asap-hub/pull/4986 and https://github.com/yldio/asap-hub/pull/4987.

To test it, use the crn-4986 environment in Contentful and follow these steps:

1. Enable the ASAP_PROJECT_AIMS_AND_MILESTONES feature flag.
2. Go to Projects.
3. Search for “Project Banteng”.
4. Open the project milestones.
5. Edit a milestone article.

https://github.com/user-attachments/assets/ea07e0eb-8384-45ea-8465-6f0943068ed7


apps/crn-server/src/routes/milestone.route.ts
```diff
@@ -17,5 +17,18 @@ export const milestoneRouteFactory = (
     res.json(articles);
   });
 
+  milestoneRoutes.put(
+    '/milestones/:milestoneId/articles',
+    async (req, res) => {
+      if (!req.loggedInUser) throw Boom.forbidden();
+
+      const { milestoneId } = req.params;
+      const { articleIds } = req.body as { articleIds: string[] };
```
damianmr: We need to check whether the current user is a project lead, right?

You can do so here with function `isProjectLead`.

See apps/crn-server/src/routes/project.route.ts

Image

So the resulting code could look something like this:

```ts
     const data = validateMilestoneArticleUpdateRequest(req.body);
     const milestone = await milestoneController.fetchById(milestoneId);
     const project = await projectController.fetchById(milestone.projectId);

     if (!isProjectLead(req.loggedInUser.id, req.loggedInUser.teams, project)) {
       throw Boom.forbidden();
     }
```

You'll have to refactor the `milestoneRouteFactory` to receive the project controller as well.

packages/react-components/src/organisms/ProjectMilestonesTable.tsx
```diff
@@ -131,6 +131,10 @@ type ProjectMilestonesProps = {
   readonly fetchLinkedArticles: (
     milestoneId: string,
   ) => Promise<ReadonlyArray<ArticleItem>>;
+  readonly onSaveArticles: (
```
damianmr, line 134: I think this left the Storybook's stories for this component broken. My compiler says this property is missing in `apps/storybook/src/ProjectMilestonesTable.stories.tsx`

damianmr (changes requested):
Requested a few changes, thanks!

apps/crn-frontend/src/projects/articles-state.ts
```diff
@@ -56,3 +60,17 @@ export const useFetchMilestoneArticles = (): ((
...
+  articles: ReadonlyArray<ArticleItem>,
+) => Promise<void>) =>
+  useRecoilCallback(({ set, snapshot }) => async (milestoneId, articles) => {
+    const authorization = await snapshot.getPromise(authorizationState);
+    await putMilestoneArticles(
+      milestoneId,
+      articles.map((a) => a.id),
+      authorization,
+    );
+    set(milestoneArticlesState(milestoneId), articles);
```
damianmr, line 76: We need to update the cache for the row of the affected milestone, this would work but we need to update tests:

```ts
// after the last line set(milestoneArticlesState(milestoneId), articles);
    const cachedMilestone = snapshot
      .getLoadable(projectMilestonesListItemState(milestoneId))
      .valueMaybe();

    if (cachedMilestone) {
      set(projectMilestonesListItemState(milestoneId), {
        ...cachedMilestone,
        articleCount: articles.length,
      });
    }
```

See what's happening now (the code above fixes this):

https://github.com/user-attachments/assets/d5018b98-6b59-439d-a231-710e0b64a894

packages/react-components/src/organisms/Milestone.tsx
```diff
@@ -241,9 +244,16 @@ const Milestone: FC<MilestoneProps> = ({
       </div>
       {isModalOpen && (
         <MilestoneArticlesModal
-          articles={articles}
+          articles={articles ?? []}
           onClose={() => setIsModalOpen(false)}
-          onConfirm={noop}
+          onConfirm={(updated) => {
```
damianmr, line 249: Let's update this code to make the property `onConfirm` in `MilestoneArticlesModal` be a function that returns a promise and make the call to `onConfirm` in that modal disable the button and display a `<Loader/>` component while the async action is running.

E.g.:
In `MilestonesArticleModal`:

Image

change the onClick for something that keeps a `isLoading` state so that we have a nice "in progress" indicator for the user adding the article.

damianmr (changes requested):
A few extra changes requested, thanks!

packages/react-components/src/organisms/Milestone.tsx
diegonvs: > change the onClick for something that keeps a isLoading state so that we have a nice "in progress" indicator for the user adding the article.

I'll do it in this PR but I had conversations with Elena about showing a skeleton loading while the data isn't updated yet. But I'll work on this in another ticket.

diegonvs: We had a private discussion about the risk of leaving the Confirm button active after the first click. Since we’ll be addressing the skeleton state soon, we can simply close the modal after the action. After a short delay, the updated list of articles will be displayed.

Error states and the skeleton implementation will be handled in a separate ticket.

diegonvs: See [7afb59c](https://github.com/yldio/asap-hub/pull/4992/commits/7afb59cd715323b126957802f901dc04ccdb2884)

apps/crn-frontend/src/projects/articles-state.ts
diegonvs: See [e59aca3](https://github.com/yldio/asap-hub/pull/4992/commits/e59aca3c526d5393cb11f8dd1e896b4568c8369a). I simplified a bit with just receiving current and then apply the change on the articleCount.

damianmr (approved):
Excellent, let's see if the rebase against master goes as expected

damianmr (approved):

PR #4994, opened 2026-04-20
[ASAP-1383] - User based project workspace display manuscripts and related data

AimeurAmin:
[ASAP-1383](https://asaphub.atlassian.net/browse/ASAP-1383)

[ASAP-1383]: https://asaphub.atlassian.net/browse/ASAP-1383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-server/src/validation/manuscript.validation.ts
```diff
@@ -28,6 +32,28 @@ export const validateManuscriptParameters = validateInput(
...
 );
 
+const manuscriptBatchRequestValidationSchema: JSONSchemaType<ManuscriptBatchRequest> =
+  {
+    type: 'object',
+    properties: {
+      ids: {
+        type: 'array',
+        items: { type: 'string' },
+        minItems: 1,
```
diegonvs, line 42: What about adding a maxItems as well? Just to set a limit in case someone uses this endpoint to send thousands of ids.

apps/crn-server/src/data-providers/contentful/project.data-provider.ts
```diff
@@ -486,22 +499,43 @@ export const parseContentfulProjectDetail = (
...
         originalGrantProposalId,
         supplementGrant,
         members: userMembers.length > 0 ? userMembers : undefined,
+        manuscripts,
       } as ResourceProjectDetail;
     }
 
     case 'Trainee Project': {
       const allMembers = processTraineeProjectMembers(members);
+      const manuscripts = sortProjectManuscripts(
```
diegonvs: Hey, the way we're calling sortProjectManuscripts seems duplicated. Can we centralize it in a helper?

```ts
const getManuscriptIdsFromMembers = (members: ProjectMembershipItem[]) =>
  sortProjectManuscripts(
    deduplicateProjectManuscripts(
      members.flatMap((member) =>
        member.projectMember?.__typename === 'Users'
          ? getManuscriptItemsFromProjectMember(member.projectMember)
          : [],
      ),
    ),
  ).map((m) => m.sys.id);
```

diegonvs: Also, deduplicateProjectManuscripts runs a filter and inside it, we are running a flatMap, so, we're quadratic.

Maybe, we can implement the deduplicateProjectManuscripts like this(AI helped me with this implementation, need to verify 😂 )?

```ts
const deduplicateProjectManuscripts = (items: ProjectManuscriptItem[]) => {
  const seen = new Set<string>();
  return items.filter(({ sys }) => {
    if (seen.has(sys.id)) return false;
    seen.add(sys.id);
    return true;
  });
};
```

apps/crn-server/src/validation/manuscript.validation.ts
AimeurAmin: No one should be able to do so, the api is not public; not an urgent thing I'd say, and not part of the ticket requirements, I added this as an improvement, I'm all in for creating a ticket for this though.

apps/crn-server/src/data-providers/contentful/project.data-provider.ts
AimeurAmin: good catch on both, thanks ✌🏻

diegonvs (approved):

packages/react-components/src/templates/ProjectProfileWorkspace.tsx
```diff
@@ -329,29 +329,39 @@ const ProjectProfileWorkspace: React.FC<ProjectProfileWorkspaceProps> = ({
...
+                    />
+                  </div>
+                ))
+              ) : (
+                <Paragraph
+                  noMargin
+                  accent="lead"
+                  data-testid="project-manuscripts-empty-state"
+                >
+                  No manuscripts available.
```
AkosuaA: I believe this is the empty state design https://www.figma.com/design/dka9iyR2A1H5ElhGZEfNq8/CRN---Designs---Projects?node-id=483-560390&t=Mqrp5Ge6wwCr19mo-0

Image

apps/crn-server/src/data-providers/contentful/project.data-provider.ts
```diff
@@ -486,22 +515,27 @@ export const parseContentfulProjectDetail = (
         .filter((m) => m.projectMember?.__typename === 'Users')
         .map((m) => parseProjectUserMember(m));
 
+      const manuscripts = getManuscriptIdsFromMembers(members);
```
AkosuaA: I believe the manuscripts for a user-based project are the manuscripts on which the project is directly linked

AkosuaA (changes requested):

apps/crn-server/src/data-providers/contentful/project.data-provider.ts
AimeurAmin: nice catch Debi, I had a talk with Damian, he mentioned the same thing, fixing that... thanks

packages/contentful/src/crn/queries/projects.queries.ts
```diff
@@ -82,6 +82,23 @@ export const projectsContentQueryFragment = gql`
...
       }
     }
+    linkedFrom {
+      manuscriptsCollection(limit: 100, order: sys_firstPublishedAt_DESC) {
+        items {
+          sys {
+            id
+          }
+          status
+          teamsCollection(limit: 1) {
```
AkosuaA, line 92: nit: I guess we won't need this anymore? can be removed

packages/contentful/src/crn/queries/projects.queries.ts
```diff
@@ -104,6 +121,26 @@ export const projectsContentQueryFragment = gql`
             avatar {
               url
             }
+            linkedFrom {
```
AkosuaA, line 124: same here

AkosuaA (approved):
LGTM! 🎉 Just left a minor nit

packages/contentful/src/crn/queries/projects.queries.ts
AkosuaA: ah, never mind. we use this for team based manuscripts to determine whether it's a collaborating manuscript or not 💡

diegonvs (approved):

PR #4996, opened 2026-04-21
[ASAP-1471] - Debounce MultiSelect loadOptions

diegonvs:
All async selects now batch rapid keystrokes into a single Algolia query via a 300ms debounce (configurable via loadOptionsDebounceMs).

The loadOptions function is kept in a ref so react-select always receives the same stable reference, preventing internal state resets on parent re-renders. Also memoizes the custom Control component when leftIndicator is provided, and exposes defaultOptions as a configurable prop.

AnalyticsControls opts out of the internal debounce (loadOptionsDebounceMs=0) since it already debounces loadOptions externally.

**Before:**

https://github.com/user-attachments/assets/465ab094-4266-494f-b2bb-41b613a07202

**After:**

https://github.com/user-attachments/assets/0be2777a-5559-447d-a265-de39ab8ddf2b

### DIFF
--- a/packages/react-components/src/atoms/MultiSelect.tsx
+++ b/packages/react-components/src/atoms/MultiSelect.tsx
@@ -157,7 +157,7 @@ const MultiSelect = <
       debounceTimerRef.current = setTimeout(() => {
         const result = loadOptionsRef.current?.(inputValue, callback);
         if (result instanceof Promise) {
-          result.then(callback);
+          void result.then(callback);
         }
       }, loadOptionsDebounceMs);
     },
--- a/packages/react-components/src/atoms/MultiSelect.tsx
+++ b/packages/react-components/src/atoms/MultiSelect.tsx
@@ -2,8 +2,9 @@ import { css, useTheme } from '@emotion/react';
 import {
   ReactElement,
   ReactNode,
+  createContext,
+  useContext,
   useEffect,
-  useMemo,
   useRef,
   useState,
 } from 'react';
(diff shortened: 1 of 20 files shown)


packages/react-components/src/atoms/__tests__/MultiSelect.test.tsx
```diff
@@ -217,6 +217,53 @@ describe('invalidity', () => {
   });
 });
 
+describe('Async debounce', () => {
+  it('does not call loadOptions on mount when defaultOptions is false (default)', () => {
+    const loadOptions = jest.fn().mockResolvedValue([]);
+    render(<MultiSelect loadOptions={loadOptions} />);
+    expect(loadOptions).not.toHaveBeenCalled();
```
AimeurAmin, line 232:
```suggestion
    jest.useFakeTimers();
    const loadOptions = jest.fn().mockResolvedValue([]);
    render(
      <MultiSelect
        loadOptions={loadOptions}
        loadOptionsDebounceMs={300}
        defaultOptions={false}
      />,
    );
    jest.advanceTimersByTime(350); // wait a bit more than 300ms to make sure  the test is not passing because the 300ms didn't pass yet
    expect(loadOptions).not.toHaveBeenCalled();
    jest.useRealTimers();
```

AimeurAmin:
Careful with the case when typing fast and removing chars fast; sometimes causes sending wrong payload, the following video reproduces that, the first **two** attempts were ok, but third and fourth were not: 

- Third attempt: I typed "**amin-testing**" but the payload had "**amin-testin**" => missing a "**g**"
- Fourth attempt (more critical cause it can cause misleading empty array): typed "**amin-tesin**" then removed "**in**" but the payload kept "**amin-tesin**"

https://github.com/user-attachments/assets/f4c3a72a-996f-4340-b581-e514d2414721

packages/react-components/src/atoms/MultiSelect.tsx
```diff
@@ -109,6 +120,8 @@ const MultiSelect = <
 >({
   customValidationMessage = '',
   loadOptions,
+  loadOptionsDebounceMs = 300,
```
AimeurAmin, line 138: I'd suggest to raise the delay to **500ms** as I'm expecting the users of the hub to type in a slow fashion and we won't benefit a lot from this delay. (to be checked with the team)

diegonvs:
Rebased with current master and force pushed.

packages/react-components/src/atoms/MultiSelect.tsx
diegonvs: We decided to go with 300ms after some conversations on slack. https://yld.slack.com/archives/G01A6J33G8H/p1777559663913579?thread_ts=1777301803.483709&cid=G01A6J33G8H

AimeurAmin (approved):
Looks good 🚀

PR #4997, opened 2026-04-22
[ASAP-1459] - Add 'Project' to Team ASAP Compliance table

damianmr:
This updates the compliance manuscripts table in the ASAP Team Workspace to show the project name, icon, and link when project data is available.

The main backend change is how we resolve a manuscript's project. We now use the direct `manuscript.project` relation first, which is the source of truth for user-based projects. If that field is empty, we fall back to the first project linked through the manuscript's first team, which keeps the existing team-based behaviour working. The same precedence is used in both `fetch` and `fetchById`, so the project shown in the table, the team visibility, and the computed manuscript ID all stay consistent.

On the frontend, the compliance table now renders a `Project` column behind the feature flag (`ASAP_PROJECT_WORKSPACE`), shows the correct project icon and route, and hides the team name for user-based project rows. I also regenerated the Contentful GraphQL types and added coverage around the new resolution logic and table rendering.

Commits can be reviewed individually to reduce diff noise.

Team ASAP Compliance (in env 4997)
https://4997.hub.asap.science/network/teams/8c88a42b-9fc7-4ec1-8606-204be75d252c/compliance

We don't sync manuscripts with Algolia daily, so if you want to verify the "Project inference" logic, you need to open the Manuscript entry in Contentful and republish the entity to make the Algolia index for this environment get in sync with the latest data. 

To find the manuscript entry in Contentful from the compliance table, click the manuscript ID pill in the first column. 
image

The link goes to the team workspace and adds the Contentful manuscript entry id as the URL hash, for example:
`/network/teams/<team-id>/workspace#2IHoW77gfBzGSJJgBGJvCr`
The part after `#` is the manuscript entry id. You can open it directly in Contentful with:
`https://app.contentful.com/spaces/5v6w5j61tndm/environments/<CONTENTFUL_ENV_ID>/entries/<entry-id>`

For example, if the hash is `2IHoW77gfBzGSJJgBGJvCr` and the environment is `crn-4997`, the Contentful URL is:
https://app.contentful.com/spaces/5v6w5j61tndm/environments/crn-4997/entries/2IHoW77gfBzGSJJgBGJvCr


packages/react-components/src/organisms/ComplianceTableRow.tsx
```diff
@@ -250,6 +321,14 @@ const ComplianceTableRow: React.FC<ComplianceTableRowProps> = ({
 
   const canEditAssignedUsers =
     !completeStatuses.includes(status ?? '') && isComplianceReviewer;
+  const manuscriptHref = team.id
+    ? `${network({}).teams({}).team({ teamId: team.id }).workspace({}).$}#${id}`
+    : undefined;
+  const teamHref = team.id
+    ? network({}).teams({}).team({ teamId: team.id }).workspace({}).$
```
diegonvs: The logic of both urls are almost the same. Can we centralize this?

packages/react-components/src/organisms/ComplianceTable.tsx
```diff
@@ -204,17 +235,19 @@ const ComplianceTable: React.FC<ComplianceTableProps> = ({
...
             <tr>
-              <th css={titleStyles} className={'sticky'}>
+              <th css={[titleStyles, idColumnStyles]} className={'sticky'}>
                 ID
               </th>
-              <th css={titleStyles}>Team</th>
-              <th css={titleStyles}>Last Updated</th>
-              <th css={titleStyles}>Status</th>
-              <th css={titleStyles}>
-                APC <br />
```
diegonvs, line 214: We removed that <br /> and the styles will change from two lines to just one line. Is this the intent?

packages/react-components/src/organisms/ComplianceTableRow.tsx
```diff
@@ -139,6 +172,42 @@ type ComplianceTableRowProps = {
...
+      return <DiscoveryProjectIcon />;
+    case 'Resource Project':
+      return <ResourceProjectIcon />;
+    case 'Trainee Project':
+      return <TraineeProjectIcon />;
+    default:
+      return null;
+  }
+};
+
```
diegonvs: nit: We can centralize it in something like getProjectConfig that holds the route and the icon

packages/react-components/src/organisms/ComplianceTable.tsx
damianmr: Yep, designs show it in one line

diegonvs (approved):
Image

PR #4998, opened 2026-04-24
[ASAP-1473] - Improve calendar link handling for Outlook

damianmr:
I could not find any official Microsoft documentation for this specific behavior, but after testing it locally the issue seems to be in new Outlook for Windows.

The problem is that new Outlook does not handle standard `webcal://...` subscription links correctly. When the user clicks a link like:

`webcal://path.to.domain/etc/calendar.ics`

and chooses Outlook from the Windows app picker, Outlook opens the "Add calendar" dialog with the `webcal:` scheme stripped out. That leaves only the path portion of the URL, which Outlook then treats as invalid.

This is what happens today if the user clicks "Add to Default System Calendar" and selects Outlook from the app picker:

image

image

image

The workaround is to generate a deliberately non-standard URL for the Outlook-specific action:

`webcal:https://calendar.google.com/calendar/ical/2457a371511efcaca32732b476834906a3c57116f04b1afb5c94e74b4d20b132%40group.calendar.google.com/public/basic.ics`

This looks wrong, but it works with new Outlook. Outlook strips the leading `webcal:` and keeps the `https://...` portion, so the "Add calendar" dialog is prefilled with a valid URL.

With this change, clicking "Add to Outlook" behaves as expected:

image

The odd part is that this appears to be a regression in the new Outlook client. Outlook (classic) handles the original `webcal://...` URL correctly, so I kept a separate "Add to Outlook (classic)" link that still uses the standard `webcal://...` format.

I also changed the manual setup action to copy the plain `https://...ics` URL, since that is the format Outlook accepts when users add the subscription manually.

Relevant links:
https://techcommunity.microsoft.com/discussions/outlookgeneral/webcal-links-not-opening-correctly-in-outlook/4462816
https://office-watch.com/2022/yes-you-can-add-an-online-calendar-to-outlook-for-mac/

Relevant screenshot from that thread:

image
image

Here's a few guides that explain how to add a Shared calendar to Outlook, they may be useful in case we want to write one of our own:

https://it.queens.ox.ac.uk/help/calendars-internet-subscription/
https://office-watch.com/2022/yes-you-can-add-an-online-calendar-to-outlook-for-mac/


packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -38,15 +40,27 @@ const CalendarLink: React.FC<CalendarLinkProps> = ({
...
       }}
+      {{
+        item: <>{outlookIcon} Add to Outlook</>,
+        // this URL is terribly malformed (`webcal:https://route.to.calendar/etc`) but it's the
+        // only known hack to make the browser open the app picker and that works as expected
+        // when the selected app is Outlook.
+        // See https://asaphub.atlassian.net/browse/ASAP-1473
+        href: `webcal:${httpCal}`,
+      }}
+      {{
```
damianmr: You may notice the link used for the `Outlook (classic)` app is the same as the `Add to Default System Calendar`. This is to prevent users that still use the classic Outlook click on the (new) Outlook link (which is above this one). It's just a UX enhancement to prevent confusion.

packages/react-components/src/molecules/CalendarLink.tsx
```diff
@@ -38,15 +40,27 @@ const CalendarLink: React.FC<CalendarLinkProps> = ({
         item: <>{googleCalendarIcon} Add to Google Calendar</>,
         href: url.toString(),
       }}
+      {{
+        item: <>{outlookIcon} Add to Outlook</>,
+        // this URL is terribly malformed (`webcal:https://route.to.calendar/etc`) but it's the
+        // only known hack to make the browser open the app picker and that works as expected
+        // when the selected app is Outlook.
+        // See https://asaphub.atlassian.net/browse/ASAP-1473
```
diegonvs, line 45: nit: https://techcommunity.microsoft.com/discussions/outlookgeneral/webcal-links-not-opening-correctly-in-outlook/4462816 can we reference that url in the comment too?

diegonvs (approved):

PR #5004, opened 2026-04-28
[ASAP-1479] New Milestones Fields

diegonvs:
### Summary

Adds five new fields to the `milestones` Contentful content model and exposes them in the backend read path:

- `bulkImported` (Boolean, hidden in editor) — flag for milestones created via bulk import; reminder generation will skip these
- `statusUpdatedAt` (Date) and `statusUpdatedBy` (Link → users) — audit trail for status changes
- `outputsLinkedAt` (Date) and `outputsLinkedBy` (Link → users) — audit trail for when research outputs are linked via `relatedArticles`

Audit fields follow the same pattern as `manuscripts.statusUpdatedBy/At` (editable in the CMS, app logic populates them — Contentful has no native read-only widget). `bulkImported` uses `disabled(true)` so it stays hidden from the entry editor.

### Testing

I configured a milestone with the new fields, as you can see at:

Screenshot 2026-04-28 at 16 45 30

When testing using:

```bash
curl -s "https://preview.contentful.com/spaces/5v6w5j61tndm/environments/Development/entries/3HLf8Kmi8KJE0aMwdHIwF1?access_token=ACCESS_TOKEN_CHANGE_HERE" | jq '.fields | keys'
```

It responds:

```json
[
  "description",
  "outputsLinkedAt",
  "outputsLinkedBy",
  "relatedArticles",
  "status",
  "statusUpdatedAt",
  "statusUpdatedBy"
]
```


packages/contentful/src/crn/queries/aims-milestones.queries.ts
```diff
@@ -284,6 +284,23 @@ export const FETCH_MILESTONE_BY_ID = gql`
       }
       description
       status
+      bulkImported
```
AkosuaA: nit: we probably don’t need to include them in the milestone queries or data provider since they’re not being used on the frontend. we'll use them for reminders but that's a different query

AkosuaA (approved):
Approved with a minor comment 🎉

packages/contentful/src/crn/queries/aims-milestones.queries.ts
diegonvs: makes sense, we can use when needed, yagni. [7ea43ea](https://github.com/yldio/asap-hub/pull/5004/commits/7ea43eaf869eca5ad4d04c5d4d83427fca0e539c)

PR #5005, opened 2026-04-29
[ASAP-1485] - Expand roles inline instead of broken scroll-to-element on network/users listing page

diegonvs:
**Context**

The links used document.getElementById('user-teams-tabbed-card').scrollIntoView(), but that element only exists on the user profile page. On the network/users listing (PeopleCard), clicking did nothing because the scroll target was absent.

tldr; When the teams card exists on the page (user profile), scroll to it as before and when it doesn't exist (network/users listing), expand the truncated list inline.

**Test plan**

- On network/users listing, click "View all roles" on a user with 3+ teams — should
expand inline and hide the link
- On network/users listing, click "View all former roles" on a user with 3+ inactive
teams — same behavior
- On user profile page, click "View all roles" — should scroll to the Teams and Roles
card (no regression)
- Verify "View all roles" does not appear when user has 2 or fewer teams

**Before:**

https://github.com/user-attachments/assets/a4488739-ef4a-4af4-869a-d3063b2223e7

**After:**

https://github.com/user-attachments/assets/4f6690f1-a71f-49f0-8d43-97ace06e5bdc

### DIFF
--- a/packages/react-components/src/templates/UserProfileHeader.tsx
+++ b/packages/react-components/src/templates/UserProfileHeader.tsx
@@ -268,7 +268,6 @@ const UserProfileHeader: React.FC<UserProfileHeaderProps> = ({
                     jobTitle={jobTitle}
                     teams={teams}
                     labs={labs}
-                    userActiveTeamsRoute={tabRoutes.research({}).$}
                     isAlumni={!!alumniSinceDate}
                   />
                 </div>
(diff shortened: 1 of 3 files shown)


packages/react-components/src/templates/UserProfileHeader.tsx
```diff
@@ -268,7 +268,6 @@ const UserProfileHeader: React.FC<UserProfileHeaderProps> = ({
                     jobTitle={jobTitle}
                     teams={teams}
                     labs={labs}
-                    userActiveTeamsRoute={tabRoutes.research({}).$}
```
diegonvs, line 271: This prop was unused, so I removed it.

AkosuaA (approved):
Nice! 🎉

PR #5006, opened 2026-05-01
[ASAP-1481] - Remove property 'text' in Tutorial model from the sync with Algolia

damianmr:
https://asaphub.atlassian.net/browse/ASAP-1481

After adding the Tutorial page to the Development environment, the sync with Algolia won't work anymore. The workaround is that, whenever there's a deploy to master, we need to unpublish this tutorial https://app.contentful.com/spaces/5v6w5j61tndm/environments/Development/entries/4EOkCzukXz6qNReA3PFz7T

So, the change in this PR makes the sync work no matter the size of the `text` property in the `tutorial` model.

This is to prevent errors like
https://github.com/yldio/asap-hub/actions/runs/25174201905/job/73811760054 when there's a large tutorial in the dev environment, which uses a cheaper plan in Algolia with lower limits for record size.

Error:
image

`could not send intermediate batch: Algolia API error [400] Record at the position 1 objectID=16JlmykX0R2tH5NGwPSDDq is too big size=10787/10000 bytes. Please have a look at
https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/in-depth/index-and-records-size-and-usage-limitations/#record-size-limits`

### DIFF
--- a/apps/crn-server/scripts/export-entity.ts
+++ b/apps/crn-server/scripts/export-entity.ts
@@ -236,10 +236,14 @@ const transformRecords = (
 
   // type 'tutorial'
   if ('usedInPublication' in record && 'shortText' in record) {
-    return {
-      ...payload,
-      _tags: record.tags,
+    const { text: _text, ...payloadWithoutText } = payload as typeof payload & {
+      text?: string;
     };
+
+    return {
+      ...payloadWithoutText,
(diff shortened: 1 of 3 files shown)


AkosuaA (approved):
Looks good! 🎉

PR #5007, opened 2026-05-04
[ASAP-1476] - Sort Milestones by Aim Number

diegonvs:
Adds asc/desc sort toggle to the Project Milestones table. Clicking the **Aim #** header flips order; pagination resets on toggle.

## Acceptance ordering

Given the input set:

`(1), (1,2), (1,2), (1,2,4,5), (1,2,5), (1,3), (1,3,4,5), (2), (2,3), (2,3,4), (3), (3,4), (3,4), (5)`

- **Asc**: `(1), (1,2), (1,2), (1,2,4,5), (1,2,5), (1,3), (1,3,4,5), (2), (2,3), (2,3,4), (3), (3,4), (3,4), (5)`
- **Desc**: `(5), (5,4,3,1), (5,4,2,1), (5,2,1), (4,3), (4,3), (4,3,2), (3), (3,2), (3,1), (2), (2,1), (2,1), (1)`

### Sort rules

**Asc** (single rule): walk aims left-to-right (smallest first), compare ascending. Shorter wins on prefix tie.

**Desc** (single rule): walk aims right-to-left (largest first), compare descending. Shorter wins on prefix tie.

`(4,3)` beats `(4,3,2)` in desc because after matching `4` and `3`, `(4,3)` ends — shorter prefix-equal wins.

## Approach

Sort happens at query time in OpenSearch via two `_script` sorts with Painless, one per direction.

- **Asc**: zero-pad each aim and join with `,` (e.g. `[1,2,5]` → `"001,002,005"`). Lex asc gives the AC order.
- **Desc**: walk aims right-to-left, encode each as `999 - n` zero-padded, join with `|` (e.g. `[1,2,5]` → `"994|997|998"`). Lex asc on encoded output gives the AC order. Inverse-encoding flips per-aim compare; shorter strings naturally sort before longer ones with the same prefix, so `(4,3)` beats `(4,3,2)`.

Both clauses use `order: 'asc'`; the desc rule is baked into the encoding.

## Why query-time

The first iteration ([7121e6d](https://github.com/yldio/asap-hub/pull/5007/commits/7121e6d)) used a single indexed `aimNumbersDesc` field — it failed AC because lexical compare puts `"3"` before `"3,2"`. A second iteration added `aimMax` + `aimCount` indexed fields for multi-key sort. That worked but persisted three redundant fields per document for a small index whose only purpose was sort ordering.

Final iteration drops all sort-helper fields and keeps only `aimNumbers` (the canonical source) on the index. The sort keys are computed per query in Painless. Trade-off: ~5–20ms of script CPU per query vs. ~16B/doc storage + indexing cost. For this index size, query-time wins.

`runtime_mappings` was tried first but rejected by OpenSearch (`Unknown key for a START_OBJECT in [runtime_mappings]`); switched to `_script` sort, which is supported on every version and runs the same Painless logic inline on the sort clause.

## Reindex

The OpenSearch sync renames `aimNumbersAsc` → `aimNumbers` and drops `aimNumbersDesc`, `aimMax`, `aimCount` from the mapping. Already run on the `crn-5007` preview. Production reindex required before merge.

## API response shape

The `/projects/:id/milestones` payload exposes a single `aims` field per milestone (comma-separated aim numbers, e.g. `"1,2"`). This is unchanged externally — the data-provider maps the new internal `aimNumbers` source field to `aims` when building the response. The `aimNumbersAsc` and `aimNumbersDesc` fields no longer exist in the OpenSearch document or in the API payload; they were write-only sort helpers with no external consumers.

## Deploy order

Run **Opensearch Analytics Sync** against production before (or immediately after) merging. The query reads `doc['aimNumbers']`; while the production index still has `aimNumbersAsc`, the Milestones tab returns 400.


AkosuaA (approved):
Really nice! 🎉🎉

apps/crn-server/src/utils/opensearch/aim-numbers-sort-scripts.ts
```diff
@@ -0,0 +1,26 @@
+export const aimNumbersAscSortScript = `
```
AkosuaA, line 1: really minor but would this file fit better under `apps/crn-server/src/utils` or maybe `apps/crn-server/src/data-providers/transformers` since it's like a helper?

diegonvs: Makes sense. I created a folder called opensearch under utils and then moved the code there. [eca61e7](https://github.com/yldio/asap-hub/pull/5007/commits/eca61e7c645cbde6220e6b16fbf22150edbcc713)

PR #5008, opened 2026-05-04
[ASAP-1480] - Update information when output is linked to milestone

AimeurAmin:
[ASAP-1480](https://asaphub.atlassian.net/browse/ASAP-1480)

[ASAP-1480]: https://asaphub.atlassian.net/browse/ASAP-1480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/project.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/project.data-provider.ts
@@ -1098,14 +1098,17 @@ export class ProjectContentfulDataProvider implements ProjectDataProvider {
 
     const articleIds = data.relatedArticleIds ?? [];
     const hasArticles = articleIds.length > 0;
+    const now = new Date();
 
     const milestoneEntry = await environment.createEntry('milestones', {
       fields: addLocaleToFields({
         description: data.description,
         status: data.status,
         relatedArticles: articleIds.map((id: string) => getLinkEntity(id)),
+        statusUpdatedAt: now,
+        statusUpdatedBy: getLinkEntity(userId),
         ...(hasArticles && {
-          outputsLinkedAt: new Date(),
+          outputsLinkedAt: now,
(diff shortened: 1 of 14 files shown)


apps/crn-server/src/data-providers/contentful/project.data-provider.ts
```diff
@@ -1086,20 +1086,28 @@ export class ProjectContentfulDataProvider implements ProjectDataProvider {
...
+
     const milestoneEntry = await environment.createEntry('milestones', {
       fields: addLocaleToFields({
         description: data.description,
         status: data.status,
-        relatedArticles: data.relatedArticleIds?.length
-          ? data.relatedArticleIds.map((id: string) => getLinkEntity(id))
-          : [],
+        relatedArticles: articleIds.map((id: string) => getLinkEntity(id)),
+        ...(hasArticles && {
```
diegonvs, line 1110: Hey, here we’re checking whether we have articles, but in updateArticlesForMilestone (in aims-milestones.data-provider.ts), we still update the milestones even when articleIds is an empty array. Should we add a guard there as well?

AimeurAmin: @diegonvs  this is intentional to follow AC#2 and also to keep track of all updates, let's say me as a lead I add articles, then someone else (another lead) removes those articles and saves (which is allowed), if we don't do it this way, it will seem like I never added articles, but like this we know the latest person who made the update must have removed the articles.

diegonvs: Makes sense. Solving

diegonvs (approved):

PR #5017, opened 2026-05-12
[ASAP-1494] - Implement loading spinner pattern in add/edit article modal

diegonvs:
### Changes:

- MilestoneArticlesModal now stays open during save with all inputs disabled and a spinner in the confirm button; closes only on success, shows inline error on failure.

- Extracted the duplicated button spinner into a shared loading prop on Button, replacing identical implementations in ConfirmModal, ToolModal, and MilestoneArticlesModal.

### How it's looking like:

Normal flow:

https://github.com/user-attachments/assets/95278d37-0931-45c8-b068-02912e4214c2

When there is an error:
_when we have an error when trying to persisting data to the server (e.g when we're offline)_

**Before:**

https://github.com/user-attachments/assets/7b1fd7fe-3950-45e8-89d1-5bc118017c35

**After:**

https://github.com/user-attachments/assets/ad04ff68-ba46-4b4b-8f3b-674a22f711de

### DIFF
--- a/packages/react-components/src/atoms/Button.tsx
+++ b/packages/react-components/src/atoms/Button.tsx
@@ -2,7 +2,12 @@
 import { css, SerializedStyles } from '@emotion/react';
 import { InputHTMLAttributes } from 'react';
 
-import { getButtonChildren, getButtonStyles } from '../button';
+import {
+  buttonLoadingContentStyles,
+  buttonSpinnerStyles,
+  getButtonChildren,
+  getButtonStyles,
+} from '../button';
 import { defaultThemeVariant, ThemeVariant } from '../theme';
 import { noop } from '../utils';
 import { getLinkColors, styles as linkStyles } from './Link';
@@ -47,6 +52,7 @@ interface LinkStyleButtonProps {
 type ButtonProps = (NormalButtonProps | LinkStyleButtonProps) & {
   readonly preventDefault?: boolean;
   readonly submit?: boolean;
+  readonly loading?: boolean;
   readonly children?: React.ReactNode;
   readonly overrideStyles?: SerializedStyles;
   readonly onClick?: () => void;
@@ -63,6 +69,7 @@ const Button: React.FC<ButtonProps> = ({
   noMargin,
   theme = defaultThemeVariant,
   submit = primary,
+  loading = false,
   id,
   children,
   onClick = noop,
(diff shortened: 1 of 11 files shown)


packages/react-components/src/organisms/MilestoneArticlesModal.tsx
```diff
@@ -111,20 +116,45 @@ const MilestoneArticlesModal: React.FC<MilestoneArticlesModalProps> = ({
...
+    setHasError(false);
+    setIsRequestInProgress(true);
+    try {
+      await onConfirm(optionsToArticles(selectedOptions));
+      onClose();
+    } catch {
+      setIsRequestInProgress(false);
+      setHasError(true);
+    }
+  };
```
gabiayako, line 136: It doesn't make a difference here because we're closing the modal anyway, but I usually use `finally` to the request in progress state in these kinds of try/catch blocks:

```suggestion
  const handleConfirm = async () => {
    if (isRequestInProgress) return;
    setHasError(false);
    setIsRequestInProgress(true);
    try {
      await onConfirm(optionsToArticles(selectedOptions));
      onClose();
    } catch { 
      setHasError(true);
    } finally {
       setIsRequestInProgress(false);
    }
  };
```

packages/react-components/src/atoms/Button.tsx
```diff
@@ -97,7 +104,14 @@ const Button: React.FC<ButtonProps> = ({
...
   >
-    {getButtonChildren(children)}
+    {loading ? (
+      <span css={buttonLoadingContentStyles}>
+        <div css={buttonSpinnerStyles} />
+        {getButtonChildren(children)}
+      </span>
+    ) : (
+      getButtonChildren(children)
+    )}
```
gabiayako, line 114: nice 🎉

gabiayako (approved):
really nice! approved ✅

packages/react-components/src/organisms/MilestoneArticlesModal.tsx
diegonvs: using `finally` makes it much clearer now. thanks for the suggestion! [316fcff](https://github.com/yldio/asap-hub/pull/5017/commits/316fcffec48f9414dbd9a9a8f763e5f46642c32d)

PR #5020, opened 2026-05-14
[ASAP-1498] - Enable users to filter resource projects by research theme

AimeurAmin:
[ASAP-1498](https://asaphub.atlassian.net/browse/ASAP-1498)

[ASAP-1498]: https://asaphub.atlassian.net/browse/ASAP-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-frontend/src/projects/ResourceProjects.tsx
```diff
@@ -99,23 +110,33 @@ const ResourceProjects: FC<ResourceProjectsProps> = ({
 }) => {
   const { currentPage, pageSize } = usePaginationParams();
   const resourceTypes = useResourceTypes();
+  const researchThemes = useResearchThemes(RESOURCE_THEME_TYPES);
   const statusFilters = useMemo(() => toStatusFilters(filters), [filters]);
   const resourceTypeFilters = useMemo(
     () => toResourceTypeFilters(filters, resourceTypes),
     [filters, resourceTypes],
   );
+  const themeFilters = useMemo(
```
gabiayako, line 128: I’d rather use `filterNames` in `useSearch` here (like we do in Network) than split filters in the component with `toResourceTypeFilters` / `toResearchThemeFilters`.

Right now everything goes into one filter param and we figure out what’s what by matching against loaded types/themes. We’re inferring categories from a flat set instead of reading them directly from the URL. Not sure if that can happen with real data but if different filters end up sharing the same string value, we can run into issues because the parsing relies on string matching.

If we use separate params (status, resourceType, researchTheme), we get a filtersMap we can pass straight through, same as TeamList 😄

gabiayako (changes requested):

apps/crn-frontend/src/projects/ResourceProjects.tsx
AimeurAmin: Done: moved to per-category filtersMap pattern, same as TeamList. Filters now read from `filtersMap.researchTheme` / `filtersMap.resourceType` / `filtersMap.status` directly. The `toResourceTypeFilters` / `toResearchThemeFilters` helpers are gone.

apps/crn-frontend/src/projects/__tests__/TraineeProjects.test.tsx
```diff
@@ -90,7 +91,7 @@ beforeEach(() => {
 
 const renderTraineeProjects = (
   searchQuery: string = '',
-  filters?: Set<string>,
+  filtersMap: import('@asap-hub/model').FetchTeamsFilter = {},
```
gabiayako: we can move the import to the top 😄

packages/model/src/research-theme.ts
```diff
@@ -8,6 +8,11 @@ export const DISCOVERY_THEME_TYPES = ['Discovery' as const];
 
 export const RESOURCE_THEME_TYPES = ['Resource' as const];
 
+export const isResearchThemeType = (
+  value: unknown,
+): value is ResearchThemeType =>
+  (researchThemeTypes as readonly string[]).includes(value as string);
+
```
gabiayako: I believe this is not being used anymore

apps/crn-frontend/src/projects/__tests__/DiscoveryProjects.test.tsx
```diff
@@ -5,7 +5,7 @@ import { ComponentProps, Suspense } from 'react';
 import { RecoilRoot } from 'recoil';
 import { createCsvFileStream } from '@asap-hub/frontend-utils';
 import { EMPTY_ALGOLIA_RESPONSE } from '@asap-hub/algolia';
-import { DiscoveryProject } from '@asap-hub/model';
+import { DiscoveryProject, FetchTeamsFilter } from '@asap-hub/model';
```
gabiayako: Maybe we can rename this `FetchTeamsFilter` to something more generic :)

gabiayako (approved):
nice 🎉

PR #5023, opened 2026-05-15
[ASAP-1376] - Implement rules for project workspace

AimeurAmin:
[ASAP-1376](https://asaphub.atlassian.net/browse/ASAP-1376)

[ASAP-1376]: https://asaphub.atlassian.net/browse/ASAP-1376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/crn-frontend/src/projects/ProjectDetail.tsx
+++ b/apps/crn-frontend/src/projects/ProjectDetail.tsx
@@ -1,7 +1,7 @@
 import { FC, lazy } from 'react';
 import { Navigate, Route, Routes, useLocation, useParams } from 'react-router';
 import { Frame } from '@asap-hub/frontend-utils';
-import { isProjectLead } from '@asap-hub/model';
+import { isProjectLead, isProjectMember } from '@asap-hub/model';
 import {
   ProjectDetailPage,
   ProjectDetailAbout,
@@ -39,7 +39,8 @@ const ProjectDetail: FC<Props> = ({ config }) => {
   const fetchArticles = useFetchAimArticles();
   const { isEnabled } = useFlags();
   const user = useCurrentUserCRN();
-  const isStaff = user?.role === 'Staff';
+  const isOpenScienceMember =
+    user?.role === 'Staff' && !!user?.openScienceTeamMember;
   const { hash: targetManuscript } = useLocation();
 
   const teamId =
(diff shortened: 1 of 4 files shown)


apps/crn-frontend/src/projects/ProjectDetail.tsx
```diff
@@ -60,11 +61,16 @@ const ProjectDetail: FC<Props> = ({ config }) => {
...
-  const isProjectMember = !!user?.projects.find(({ id }) => id === projectId);
+  const isMember =
+    !!user && isProjectMember(user.id, user.teams ?? [], projectDetail);
   const isLead =
     !!user && isProjectLead(user.id, user.teams ?? [], projectDetail);
   const showWorkspace =
-    isEnabled('PROJECT_WORKSPACE') && (isProjectMember || isStaff);
+    isEnabled('PROJECT_WORKSPACE') && (isMember || isOpenScienceMember);
+  const canSubmitManuscript = showWorkspace && isMember;
+  const canEditOrResubmitManuscript =
```
diegonvs, line 71: Hey, on the ticket it says:

Image

but here we're adding `isOpenScienceMember` in the conditional, is intended to have OS members edit/submit a manuscript?

packages/model/src/project.ts
```diff
@@ -279,6 +279,39 @@ export const isProjectLead = (
...
+ * resolved from `teamId` or, when only the detail payload is available,
+ * from `fundedTeam.id`.
+ *
+ * For individual-based projects (Trainee, user-based Resource): the user is a
+ * member if they appear in the project's members list.
+ */
+export const isProjectMember = (
+  userId: string,
+  userTeams: ReadonlyArray<{ id: string }>,
+  project: Pick<Project, 'projectType'> &
```
diegonvs: Is this `Pick<Project, 'projectType'>` really necessary?

If we drop this Pick and just use the Partial I think it might work

apps/crn-frontend/src/projects/__tests__/ProjectDetail.test.tsx
```diff
@@ -300,6 +300,29 @@ type TestVariant = {
...
+  teams: [{ id: 'team-1', role: 'Project Manager' }],
+  role: 'Grantee',
+};
+
+const resourceMemberUser = {
+  id: 'user-team',
+  projects: [],
+  teams: [{ id: 'team-1', role: 'Project Manager' }],
+  role: 'Grantee',
+};
```
diegonvs: `discoveryMemberUser` and `resourceMemberUser` are the same. Can we centralize it?

apps/crn-frontend/src/projects/ProjectDetail.tsx
AimeurAmin: I talked to Padmini about this, we need a kinda big refactor to have these so for now we're keeping the same logic we have in Network => teams

image

packages/model/src/project.ts
AimeurAmin: Partial is not specific on fields, and I don't think we need it here at all, I'll change it to `project: Project `since we are passing the full project anyway.

diegonvs (approved):

PR #5024, opened 2026-05-15
[ASAP-1502] - Prevent description text from triggering file picker in LabeledFileField

diegonvs:
Move description outside <label> into forContent so clicking it no longer activates the associated button. Also, added aria-describedby to preserve screen reader association.

**Before:**

https://github.com/user-attachments/assets/bfc3f0ba-1184-493c-a7ec-ae02a977705b

**After:**

https://github.com/user-attachments/assets/7516cbcb-dd09-4a3c-a662-ffa3704f245b


gabiayako (approved):
nice 🎉

PR #5026, opened 2026-05-18
[ASAP-1432] - Create Project Outputs tab

diegonvs:
## Summary                                               

Adds Outputs and Draft Outputs tabs to the Project Detail page (all project   
types), behind the `PROJECT_OUTPUTS` feature flag. Both tabs render an empty state
page. List components come in the next ticket.

https://github.com/user-attachments/assets/6d1b51e5-ab37-4e68-b968-ff60cab50829

https://github.com/user-attachments/assets/6b1b3973-7ae5-4620-a31f-3a58c9ff6f12

### DIFF
--- a/apps/crn-frontend/src/projects/ProjectDetail.tsx
+++ b/apps/crn-frontend/src/projects/ProjectDetail.tsx
@@ -6,6 +6,7 @@ import {
   ProjectDetailPage,
   ProjectDetailAbout,
   NotFoundPage,
+  NoOutputsPage,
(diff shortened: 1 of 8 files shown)


AkosuaA (approved):
LGTM! 🎉 
Even though the count is in the designs I think since this ticket is just to have the tab, it's ok to add that bit when wire in the data.

diegonvs:
> add that bit when wire in the data

Yes, that's what I was thinking

PR #5030, opened 2026-05-20
[ASAP-1501] - update fields in live compliance sheet

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1501

The instant live sheet process currently works only in Production. We’ll create a follow-up ticket to support a dedicated test environment.

To test this I created a Google Sheet and linked it to my Google service account. I then updated the newly added fields in the CMS and ran the sync script locally.
Since the setup process is fairly involved, I’ve linked a video so there's no need to configure everything locally in order to verify the changes.

https://drive.google.com/file/d/1QYhLy2rW0XELAK3xuvRJYzmXUHpZAF07/view?usp=sharing


AimeurAmin (approved):
nice 🚀

PR #5034, opened 2026-05-22
[ASAP-1460] - Display collaborating teams on team based project detail page

AimeurAmin:
[ASAP-1460](https://asaphub.atlassian.net/browse/ASAP-1460)

[ASAP-1460]: https://asaphub.atlassian.net/browse/ASAP-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-server/src/data-providers/contentful/project.data-provider.ts
```diff
@@ -412,6 +419,102 @@ const parseProjectManuscripts = (
...
+        }
+      });
+    });
+
+  return Array.from(teamsById.values())
+    .sort((a, b) => a.displayName.localeCompare(b.displayName))
+    .map((team) => ({
+      ...team,
+      articles: team.articles,
+    }));
```
diegonvs: I believe this .map is unnecessary. Can we remove it?

diegonvs:
Maybe in the future we could structure the ProjectContributors component like this:

```jsx
<ProjectContributors>
  <ProjectContributors.Tab label="Funded Team">
    <FundedTeamPanel team={fundedTeam} />
  </ProjectContributors.Tab>
  <ProjectContributors.Tab label={`Collaborators (${count})`}>
    <CollaboratorsPanel teams={collaboratingTeams} />
  </ProjectContributors.Tab>
</ProjectContributors>
```

It doesn’t make much sense right now since we only have a few tabs, but if the number grows, this approach could be useful.

AimeurAmin:
> Maybe in the future we could structure the ProjectContributors component like this:
> 
> ```js
> <ProjectContributors>
>   <ProjectContributors.Tab label="Funded Team">
>     <FundedTeamPanel team={fundedTeam} />
>   </ProjectContributors.Tab>
>   <ProjectContributors.Tab label={`Collaborators (${count})`}>
>     <CollaboratorsPanel teams={collaboratingTeams} />
>   </ProjectContributors.Tab>
> </ProjectContributors>
> ```
> 
> It doesn’t make much sense right now since we only have a few tabs, but if the number grows, this approach could be useful.

compound components are not used anywhere in the app right now, it would be nice to have them when needed, but for this specific case I believe it will only make sense if we the area grows.
if we decide to move that way it would be nice to have a dedicated ticket for it.

diegonvs (approved):
Image

PR #5039, opened 2026-05-27
ASAP-1308 Display Project Research Outputs for User and Team based Projects

gabiayako:
JIRA tickets:
- https://asaphub.atlassian.net/browse/ASAP-1308
- https://asaphub.atlassian.net/browse/ASAP-1434

---

### Description

This PR:
- Adds `projects` to `researchOutputs` content model
- Replaces mock data for project research outputs with real API data fetched from Contentful
- Adds a new API endpoint (GET /projects/:projectId/research-outputs) with filtering support for draft/published outputs
- Fetches published research outputs from Algolia and draft research outputs from Contentful like it was done for Teams


AkosuaA (commented):
A research output should only be linked to a single project in the CMS, similar to how manuscripts work today.

My understanding is that:

* For a team-based project, the outputs should include all outputs linked to the funded team, regardless of whether a project is linked.
* For a user-based project, the outputs should include only those outputs directly linked to the project.

Currently, the frontend does not enforce that when an author is added to an output, their team must also be added (assuming they are not already a member of the project) [designs](https://www.figma.com/design/JgcEOEbozf8CzVjA0SLsaq/CRN---Designs---Shared-Research?node-id=27447-160004&t=q9MTKAJFXCTVLsbu-0). I guess this is why we need to fetch outputs where the authors are members of the funded team. However, it's also possible to add the funded team to an output without adding any authors from that team 🤔 in which case I'm guessing that output should still show in the team.

I'll ask the Padmini on this ticket's chat for context and clarification.

On the project card, the first project is the main project (i.e the project that created the output) and the count is the collaborating projects (based on the collaborating teams). It's confusing and I don't think I'm explaining it properly. Let me ask both Padmini and Elena to explain 😅

packages/contentful/migrations/crn/researchOutputs/20260527113953-add-projects.js
```diff
@@ -0,0 +1,33 @@
+module.exports.description = 'Add projects to research outputs';
+
+module.exports.up = (migration) => {
+  const researchOutputs = migration.editContentType('researchOutputs');
+
+  researchOutputs
+    .createField('projects')
+    .name('Projects')
+    .type('Array')
```
AkosuaA: Hey Gabi, I've been thinking about this a bit more because it's different from the initial approach, but I think this could work too.

Originally, the idea was, a research output could only be linked to a single project, and that would only happen when the submitting project was a user-based project (similar to how we handle manuscripts and working group outputs). That's why the ticket references a `project` field.  Image 
In that model:
* If the `project` field is set, it's a user-based project output.
* If the `workingGroup` field is set, it's a working group output.
* If neither is set, it's a team-based output.

With this approach (linking multiple projects), we'd store the selected teams in the `teams` field and their linked projects in the `projects` field when creating a research output. To determine the output type, we'd then look at the members of the project(s): if they're teams, it's team-based; if they're users, it's user-based.

We'd also need to backfill existing outputs so the `projects` field is populated.

What do you think?

packages/react-components/src/molecules/ProjectOutputBody.tsx
```diff
@@ -359,11 +358,22 @@ const ProjectOutputBody: React.FC<ProjectOutputBodyProps> = ({
   isInReview,
   keywords,
   showTags = true,
-  projects,
+  project,
 }) => {
-  const primaryProject = projects?.[0];
+  const primaryProject = project;
```
AkosuaA: For team-based project outputs (where there is no `project`), `primaryProject` is the project from the first team, and `otherProjects` contains the projects from the remaining teams. https://www.figma.com/design/dka9iyR2A1H5ElhGZEfNq8/CRN---Designs---Projects?node-id=361-529018&t=l7OCjwHhUmvYGxzu-0

AkosuaA (approved):
Really nice!! 🎉🎉

There's one minor change I left a comment on 🙂

PR #5043, opened 2026-06-01
[ASAP-1517] -  user based manuscript fix

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1517

### Changes Made
* Introduced separate identifiers for project manuscripts:

  * `projectId` is now used for user-based project manuscripts.
  * `teamId` is now used for team-based project manuscripts.
  * Manuscripts created from the Teams section of the Hub should be unaffected and continue to behave like team-based project manuscripts (i.e use teamId). 

* Updated manuscript retrieval logic for teams and projects:
  * When fetching `manuscripts` and `collaborationManuscripts`, the manuscript's `project` field in the CMS is now taken into account.
  * For teams and team-based projects, a manuscript is considered a collaboration manuscript if its `project` field is populated, even if the team appears as the first entry in the manuscript's `teams` field.

* Updated manuscript count logic when creating team-based project manuscripts and user-based project manuscripts:
  * For team-based projects, the manuscript count should be determined only by manuscript's linked to the funded team which do not have a project linked.
  * For user-based projects, the manuscript count should be determined by manuscripts with the project linked.

* For user-based project manuscripts, on the Hub UI no team is fixed, allowing all associated teams to be edited.

### Not Covered (pending confirmation from Padmini)

* When a user-based project manuscript is opened from a team-based project's workspace (for example, when the project's funded team is a collaborator and the current user is an author), it should display the user-based project UI. Currently, it follows the team-based project UI.

* The validation logic in the Contributors section has to be reviewed, particularly for edit and resubmission workflows:
  * if `User A` doesn't belong to a user-based project (Project B) and they are added to Project B's manuscript and so their team (Team A) is added, then when editing if User A is removed should Team A be automatically removed? What if another Team is added ie Team A and Team C. Should both be removed?
  * When editing, the validation in the Contributors section doesn't work always work correctly (both on the teams and projects manuscripts). I have an idea of what may be the cause but since it's something that has to be tested more thoroughly and the form is currently being used I'd like to do that in a [separate ticket](https://asaphub.atlassian.net/browse/ASAP-1523).


packages/contentful/src/crn/queries/manuscript.queries.ts
```diff
@@ -612,6 +612,25 @@ export const FETCH_MANUSCRIPTS_BY_TEAM_ID = gql`
...
+      }
+    }
+  }
+`;
+
+export const FETCH_MANUSCRIPTS_BY_PROJECT_ID = gql`
+  query FetchManuscriptsByProjectId($id: String!) {
+    projects(id: $id) {
+      linkedFrom {
+        manuscriptsCollection(limit: 500, order: count_DESC) {
```
gabiayako: I wonder if we could make this change

```suggestion
        manuscriptsCollection(limit: 1, order: count_DESC) {
```

since we're using only the first item where we consume the query

```
projects?.linkedFrom?.manuscriptsCollection?.items[0]?.count || 0;
```

packages/contentful/src/crn/queries/manuscript.queries.ts
```diff
@@ -612,6 +612,25 @@ export const FETCH_MANUSCRIPTS_BY_TEAM_ID = gql`
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+export const FETCH_MANUSCRIPTS_BY_PROJECT_ID = gql`
```
gabiayako, line 627: Debbie, about this count, should it start from `1` for manuscripts only linked to the project without any team associated? If so, I believe it's not working that way with this query because even for user based project manuscripts, there could be teams associated, so it would take the count from team's manuscript to form the id to the project manuscript
Image

packages/contentful/src/crn/queries/manuscript.queries.ts
AkosuaA: yep, nice catch. Will update this ~and the manuscripts by team id query~.

packages/contentful/src/crn/queries/manuscript.queries.ts
AkosuaA: Yea, this one is a bit confusing because the data in dev isn't correct. Because we were using the team counts to set the count for user-based manuscript before, the counts for user-based project manuscripts in dev are not correct and don't start from 1. Testing with a user-based project that isn't linked to any manuscripts yet should start from one and ignore the count from the teams linked. I'll update the cms counts for the manuscripts linked to this project which should give a clearer picture

AkosuaA: I've fixed the counts for this project but you can also test with `Diego Trainee Project`, [here](https://5043.hub.asap.science/projects/trainee/5cOKEp6xST91VdGvGn9UTe/workspace)

gabiayako: Ohh nice, in my head it would be required to add some conditions in the query, but it works perfectly on `Diego Trainee Project`. Thanks 😄

gabiayako (approved):
nice 👏👏👏👏👏

PR #5048, opened 2026-06-05
[ASAP-1489] - Upgrade Algolia API Client

AimeurAmin:
[ASAP-1489](https://asaphub.atlassian.net/browse/ASAP-1489)

[ASAP-1489]: https://asaphub.atlassian.net/browse/ASAP-1489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/algolia/src/no-token-client.ts
```diff
@@ -33,25 +30,29 @@ export const EMPTY_ALGOLIA_FACET_HITS = {
   exhaustiveFacetsCount: true,
 };
 
+export type CRN = 'crn';
```
gabiayako: I believe this is not being used

apps/asap-cli/src/scripts/algolia/set-settings.ts
```diff
@@ -14,21 +13,19 @@ export type SetAlgoliaSettings = {
 export const getIndexSchema = async ({
   algoliaAppId,
   algoliaCiApiKey,
-  indexName,
+  indexName: _indexName,
```
gabiayako: if it's not being used, I believe we can remove it from here :)

gabiayako (approved):
nice 🎉

PR #5050, opened 2026-06-05
[ASAP-1512] - Adds new optional fields to GP2 Contentful

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1512

Depends on #5049 — alumniSinceDate and alumniLocation are created there.

#### Users Content Model

- `alumniLastUpdated` (Date, read-only) — **"Alumni Last Updated"**
  - Auto-stamped when either `alumniSinceDate` or `alumniLocation` changes.
  - Placed after `alumniLocation`.

> `alumniSinceDate` and `alumniLocation` are handled by #5049. This migration depends on #5049 landing first; its timestamp (`20260606120000`) is set to run after #5049's (`20260605000000`).

#### Working Group Membership Content Model

- `inactiveSinceDate` (Date) — **"Inactive Since Date"**

#### Project Membership Content Model

- `inactiveSinceDate` (Date) — **"Inactive Since Date"**

#### `field-as-updated-at` Extension

The existing `field-as-updated-at` Contentful App (ID `mqvX9KU5AthRTlnIRhNNh`) only supported a single `observedField`.

The requirement is that `alumniLastUpdated` re-stamps when either `alumniSinceDate` **or** `alumniLocation` changes, so the extension was extended to parse `observedField` as a comma-separated list.

## Backward Compatibility

Existing single-field configurations continue to work unchanged:

- Events
  - `notesUpdatedAt`
  - `videoRecordingUpdatedAt`
  - `presentationUpdatedAt`

- Outputs
  - `lastUpdatedPartial`


packages/contentful-app-extensions/field-as-updated-at/src/locations/Field.tsx
```diff
@@ -8,28 +8,36 @@ const Field = () => {
...
     if (
       currentPublishedVersion &&
       initialPublishedVersion &&
       currentPublishedVersion > initialPublishedVersion &&
-      currentValue !== initialValue
+      anyChanged
     ) {
-      setField(currentValue ? sys.publishedAt : undefined);
+      setField(anyPresent ? sys.publishedAt : undefined);
       await sdk.field.setValue(sys.publishedAt);
```
AimeurAmin: if `anyPresent` is falsy you'll have undefined passed to `setField` whereas a real value is passed in `sdk.field.setValue`

```suggestion
      const newValue = anyPresent ? sys.publishedAt : undefined;
      setField(newValue);
      await sdk.field.setValue(newValue);
```

packages/contentful-app-extensions/field-as-updated-at/src/test/locations/Field.test.tsx
```diff
@@ -140,4 +160,71 @@ describe('Field component', () => {
...
+          .fn()
+          .mockReturnValueOnce(makeDoc('something'))
+          .mockReturnValueOnce(undefined),
+      },
+    });
+    (useSDK as jest.Mock).mockReturnValue(sdk);
+
+    render(<Field />);
+
+    expect(sdk.field.setValue).toHaveBeenCalledWith('2026-06-03T10:00:00.000Z');
```
AimeurAmin: this is happening because of the issue I pointed above, the title is "clears the timestamp" yet the test expects `sdk.field.setValue` to receive a value, after fixing the above it should be fine to expect it to be called with `undefined`

```suggestion
    expect(sdk.field.setValue).toHaveBeenCalledWith(undefined);
```

AimeurAmin (changes requested):
Looks good, just one small issue to be fixed

packages/contentful-app-extensions/field-as-updated-at/src/test/locations/Field.test.tsx
diegonvs: Makes sense. Pushed [a2ceec8](https://github.com/yldio/asap-hub/pull/5050/commits/a2ceec83e9ec6fe0505cb6c534e538d883ff624a)

PR #5052, opened 2026-06-08
[ASAP-1447] - move inactive teams to bottom of teams list

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1447

### DIFF
--- a/packages/model/src/team.ts
+++ b/packages/model/src/team.ts
@@ -42,6 +42,11 @@ export const teamStatus = ['Active', 'Inactive'] as const;
 
 export type TeamStatus = (typeof teamStatus)[number];
 
+export const TeamStatusRank: Record<TeamStatus, number> = {
+  Active: 1,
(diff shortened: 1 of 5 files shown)


packages/model/src/team.ts
```diff
@@ -42,6 +42,11 @@ export const teamStatus = ['Active', 'Inactive'] as const;
 
 export type TeamStatus = (typeof teamStatus)[number];
 
+export const TeamStatusRank: Record<TeamStatus, number> = {
```
gabiayako:
```suggestion
export const teamStatusRank: Record<TeamStatus, number> = {
```

I wouldn't use uppercase just because it seems it is a type with that

gabiayako (approved):

packages/model/src/team.ts
AkosuaA: makes sense, will update 👍🏾

PR #5053, opened 2026-06-08
ASAP-1465 Display Collaborating Members on User based Project Detail page

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-1465

---

This PR adds a "Collaborating Members" tab to user based project detail pages (Resource individual and Trainee projects). Shows members who are not part of the project and belong to the research output authors.

Diego Trainee Project _ ASAP Hub · 9 34am · 06-09

It's also possible to see the component here: https://5053.hub.asap.science/.storybook/?path=/story/molecules-collaborating-members--normal


packages/react-components/src/templates/ProjectDetailAbout.tsx
AkosuaA: should we add the collaboratingMembers here as well?

AkosuaA (approved):
Looks good! 🎉

I noticed two minor things from the designs (subtitle and project member count) which were probably missing from a previous pr but maybe we could add now. It wasn't in the scope of this ticket though so happy to approve 🙂

Image

gabiayako:
nice catch @AkosuaA! thanks! I'll fix it soon

packages/react-components/src/templates/ProjectDetailAbout.tsx
gabiayako: Thanks! Nice catches! I've added the missing subtitle and count from the designs and added collaborators to user based Resource Projects as well 😄 

Screenshot 2026-06-10 at 10 03 01

PR #5055, opened 2026-06-09
[ASAP-1477] - add script to import aims and milestones for projects

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1477

To test I cloned the environment `crn-5055` from Production and run the script `yarn import:aims-milestones <aims-csv-path> <milestones-csv-path>` (takes about 10 minutes).
I've also synced the aims and milestones in opensearch with the created aims and milestones.

So to cross-check/confirm you can either view directly in the cms or use this pr's hub, eg: https://5055.hub.asap.science/projects/discovery/6pZpD6OLR2sMquJdAsVfwr/milestones

### DIFF
--- a/apps/crn-server/scripts/import-aims-and-milestones.ts
+++ b/apps/crn-server/scripts/import-aims-and-milestones.ts
@@ -128,37 +128,39 @@ const app = async () => {
   }
 
   for (const row of milestoneRows) {
-    const projectId = cell(row, milestoneColumns.projectId);
-    const milestoneKey = cell(row, milestoneColumns.milestoneDescription);
-    const status = cell(row, milestoneColumns.status);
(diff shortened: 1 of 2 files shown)


apps/crn-server/scripts/import-aims-and-milestones.ts
```diff
@@ -0,0 +1,298 @@
...
+
+      project.aims[aimNumber] = {
+        aimNumber,
+        description: aimDescription,
+        milestoneKeys: [],
+      };
+    }
+  }
+
+  for (const row of milestoneRows) {
```
gabiayako, line 130: maybe we could add

```
if (!isEmptyRow(row)) {
```

the same way it was made for aims :)

gabiayako (approved):

PR #5058, opened 2026-06-10
ASAP-1530 Display User-Based Manuscript when Team is Collaborator

gabiayako:
Jira: https://asaphub.atlassian.net/browse/ASAP-1530

--- 

### Problem
When a user-based project manuscript was displayed through a team-based workspace's collaboration section, it was being opened as if it were a team-based manuscript. This caused the "Teams" field to appear as required in the contributors section, even though that field is not required for user-based projects.
The root cause was how the URL path was being calculated inside ManuscriptCard. When the card was rendered in a team-based workspace collaboration section, it used the surrounding team's project ID to build the route, instead of the manuscript's own projectId and projectType. So the manuscript form opened in the wrong project context.

### Solution
Fixed the URL path calculation in ManuscriptCard so that when a manuscript has its own projectId and projectType, those take precedence over the hrefs injected by the surrounding workspace. This applies to the edit, resubmit, and compliance report routes.
As part of this, the route-building logic was extracted into a projectRouteByType map in @asap-hub/routing, which removed the repeated if-chains and also simplified the getRoute type in projectDetailConfig.
I'm not a fan of putting this kind of logic inside a component, but moving it up to ProjectDetail would require fetching extra data upfront and duplicating the fix across both the project and team workspaces, so keeping it in ManuscriptCard felt better in this case.

### DIFF
--- a/apps/crn-frontend/src/projects/__tests__/ProjectDetail.test.tsx
+++ b/apps/crn-frontend/src/projects/__tests__/ProjectDetail.test.tsx
@@ -314,6 +314,29 @@ const renderProjectDetail = async (
   });
 };
 
+const getWorkspaceRoute = (routeKeyword: string, projectId: string) => {
+  const base = projects({});
+  switch (routeKeyword) {
+    case 'discovery':
+      return base
+        .discoveryProjects({})
+        .discoveryProject({ projectId })
+        .workspace({});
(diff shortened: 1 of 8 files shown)


AimeurAmin (approved):
nice 🚀

PR #5059, opened 2026-06-11
[ASAP-1512] - add alumniLastUpdated and membership inactiveSinceDate

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1512

Depends on #5049 — `alumniSinceDate` and `alumniLocation` are created there.

#### Users Content Model

- `alumniLastUpdated` (Date) — **"Alumni Last Updated"**
  - Auto-stamped via the `field-as-updated-at` Contentful App whenever `alumniSinceDate` or `alumniLocation` changes and the entry is published.
  - Also stamped server-side from `gp2-server`'s user data-provider whenever a `PATCH /users/:id` includes `alumniSinceDate` or `alumniLocation`. Covers product-UI flows where the CMS editor never opens.
  - Placed after `alumniLocation`.

> The migration's timestamp (`20260611120000`) runs after #5049's (`20260605000000`).

#### Working Group Membership Content Model

- `inactiveSinceDate` (Date) — **"Inactive Since Date"** — editor-writable via `datePicker`.

#### Project Membership Content Model

- `inactiveSinceDate` (Date) — **"Inactive Since Date"** — editor-writable via `datePicker`.

Both fields are also exposed on the `Member<T>` model type so future consumers can read them through the existing membership selections.

#### `field-as-updated-at` Extension

The existing Contentful App (ID `mqvX9KU5AthRTlnIRhNNh`) was extended in several ways:

- **Multi-field observed list.** `observedField` is now parsed as a comma-separated list. `alumniLastUpdated`'s migration installs the app with `observedField: 'alumniSinceDate,alumniLocation'`, so any change to either field re-stamps it on publish.
- **Type-aware value reads.** The extension piped every observed field through `documentToHtmlString` regardless of type. For Date/Symbol fields (e.g. `alumniSinceDate`, `orcid`) that returned `""`, so the diff `currentValue !== initialValue` was always `"" !== ""` — the stamp never fired. The extension now reads each field's type from `sdk.contentType.fields` and only applies `documentToHtmlString` to `RichText`, returning the raw value otherwise.
- **Effect lifecycle.** The subscription used to register in the render body; `useSDK` re-renders on every entry save, so subscriptions accumulated. Stale ones fired `setField + setValue + publish` on each keystroke (observed as 409 PATCH conflicts). The subscription is now inside `useEffect` with cleanup, so only one listener is alive at a time.
- **Rolling baseline.** The callback used to unsubscribe after the first stamp, which meant a later publish that cleared the observed fields never reached the listener and the stamp remained. The effect now tracks `baselineValues` / `baselinePublishedVersion` and updates them after each stamp instead of unsubscribing.
- **Proper clearing.** When all observed fields are empty the extension now calls `sdk.field.removeValue()`; `sdk.field.setValue(undefined)` does not reliably clear Date fields.
- **Cancellation on unmount.** A `cancelled` flag flipped in the cleanup is checked at the top of every callback so callbacks that fire after unmount bail out before touching state or the SDK.

## See it in action

https://github.com/user-attachments/assets/28bddb96-1b1f-4260-9b62-c7e1d278ce86

## Backward Compatibility

Existing single-field configurations continue to work unchanged — a one-item list behaves the same as the old single-field path:

- Events
  - `notesUpdatedAt`
  - `videoRecordingUpdatedAt`
  - `presentationUpdatedAt`
- Outputs
  - `lastUpdatedPartial`

For RichText observed fields (the original use case) the value read still goes through `documentToHtmlString`, so the diff behaviour is byte-for-byte identical.

## Notes

The `field-as-updated-at` bundle is deployed org-wide via `on-push-app-extensions.yml` on push to `master`; the new bundle behaviour activates in dev/prod after merge.


AimeurAmin (approved):
looks good 👏🏻

PR #5060, opened 2026-06-11
[ASAP-1542] - [Fix] Discussion link always pointing project workspace

AimeurAmin:
[ASAP-1542](https://asaphub.atlassian.net/browse/ASAP-1542)

The fix includes:

- If the user replied from Network team workspace the link received in the email should taken them to the discussion in the network team workspace

- If the user replied from Project workspace then the link received in the email should take them to the discussion in the project workspace

https://github.com/user-attachments/assets/383392bb-92e9-468d-b541-8e8fdd100e6e

[ASAP-1542]: https://asaphub.atlassian.net/browse/ASAP-1542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-frontend/src/network/teams/useDiscussionHandlers.ts
```diff
@@ -19,12 +19,14 @@ const useDiscussionHandlers = () => {
     message: string,
     files?: ManuscriptFileResponse[],
   ): Promise<string | undefined> => {
+    const workspaceLink = `${window.location.origin}${window.location.pathname}`;
```
gabiayako: I believe this falls into de same issue from [ASAP-1530 Display User-Based Manuscript when Team is Collaborator](https://github.com/yldio/asap-hub/pull/5058): if we are in the collaborators section, the manuscript is not from the current team/project page we're in. For example, in the print below, I'm in the `Ant-Discovery` Project Workspace, but I opened a discussion on a manuscript from `Diego's Trainee Project` and the `workspaceLink` sent is from `Ant-Discovery`. So the url doesn't work in this case.

Image

decode-jwt.sh
gabiayako, line 1: I believe this is committed by mistake 😉

gabiayako (changes requested):
Requesting changes mainly because of the file committed by mistake 😉 
But I also wanted to understand if this issue https://github.com/yldio/asap-hub/pull/5060/changes#r3397836328 was intentionally left unresolved because it’s not a very common user flow

decode-jwt.sh
AimeurAmin: oh that totally shouldn't have been there 😭 it's leftover from the auth0 roles sharing ticket, the token was  invalidated before so no concerns, I'll just remove the file, nice catch, thanks Gabi ✌🏻

gabiayako (approved):
Nice!! 👏 👏 👏 👏 👏

PR #5061, opened 2026-06-11
[ASAP-1515] - Display Active and Past members for Working Group

AimeurAmin:
[ASAP-1515](https://asaphub.atlassian.net/browse/ASAP-1515)

[ASAP-1515]: https://asaphub.atlassian.net/browse/ASAP-1515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/model/src/working-group.ts
```diff
@@ -11,7 +11,12 @@ export const deliverableStatus = [
 ] as const;
 export type DeliverableStatus = (typeof deliverableStatus)[number];
 
-export const workingGroupRole = ['Chair', 'Project Manager'] as const;
+export const workingGroupRole = [
+  'Chair',
+  'Project Manager',
+  'Lead',
+  'Co-Lead',
```
diegonvs: Hey, while searching for occurrences of "Co-lead," I found that this is the only case where it's written as "Co-Lead" with a capital L. Even in Contentful(https://github.com/yldio/asap-hub/blob/064a1cc22f5e0dec6b41489b71de3d642c03f96f/packages/contentful/migrations/gp2/workingGroupMembership/20230504104909-redefine-enums.js#L10), the naming convention is "Co-lead." Could we use "Co-lead" with a lowercase l in this model for consistency?

packages/gp2-components/src/templates/WorkingGroupOverview.tsx
```diff
@@ -55,6 +58,34 @@ const columnStyles = css({
   },
 });
 
+const memberListStyles = css({
+  paddingTop: rem(32),
+});
+
+const tabDescriptionStyles = css({
+  fontWeight: 'bold',
+  margin: '8px 0',
```
diegonvs:
```suggestion
  margin: `${rem(8)} 0`,
```

packages/validation/src/permissions/research-output.ts
```diff
@@ -35,12 +35,20 @@ export const isActiveAndBelongsToAssociation = (
   return belongsToAssociation && isActive;
 };
 
+const ELEVATED_WORKING_GROUP_ROLES = new Set([
```
diegonvs, line 38: Does makes sense to add `Chair` as a elevated as well?

diegonvs (commented):
PR looks good but left some minor questions and nits 😅

packages/validation/src/permissions/research-output.ts
AimeurAmin: I believe this is a business decision and from what I know `Chair` should not be added

diegonvs (approved):

PR #5062, opened 2026-06-12
[ASAP-1524] - outputs refactor add flow ids

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1524

I considered splitting the article flows into separate ids and wasn't sure if it was necessary. It might be but I think once we get to the point of using the flow ids with the actions in the form it'll be clearer so I left that out for now.

The update to impact suggestions fixes the issue shown in the attached screen recording, where impact requests were triggered on every keystroke in the Title and URL fields.
https://github.com/user-attachments/assets/06c85b24-d0ee-4ac5-8e24-dd92675cdb1a


apps/crn-frontend/src/shared-research/util.ts
```diff
@@ -52,3 +54,59 @@ export const mapManuscriptVersionToResearchOutput = (
   doi: manuscriptVersion.doi,
   publishDate: manuscriptVersion.firstPublicDate,
 });
+
+export type ResolveFlowIdParams = {
+  entityType: 'project' | 'team' | 'working-group';
```
AimeurAmin: `entityType` accepts `project` but there are no project flow ids and the function only checks `isTeam`.
in `resolveResearchOutputFlowId` when entityType is `project`, it silently returns the WORKING_GROUP flow ids

packages/model/src/research-output.ts
```diff
@@ -499,3 +499,22 @@ export const convertBooleanToDecision = (bool?: boolean): DecisionOption => {
...
+  TEAM_ADD_VERSION: 'team-add-version',
+  TEAM_ADD_VERSION_FROM_MANUSCRIPT: 'team-add-version-from-manuscript',
+  TEAM_DUPLICATE: 'team-duplicate',
+
+  WORKING_GROUP_CREATE: 'working-group-create',
+  WORKING_GROUP_EDIT_DRAFT: 'working-group-edit-draft',
+  WORKING_GROUP_EDIT_PUBLISHED: 'working-group-edit-published',
+  WORKING_GROUP_ADD_VERSION: 'working-group-add-version',
+  WORKING_GROUP_DUPLICATE: 'working-group-duplicate',
+} as const;
```
AimeurAmin, line 517: these are CRN-only but they live in @asap-hub/model, GP2 has its own output flows that aren't covered here. if you check the dev note it mentions that the export should go in CRN-frontend shared state, not model

packages/react-components/src/organisms/ResearchOutputFormSharingCard.tsx
```diff
@@ -147,20 +147,16 @@ const ResearchOutputFormSharingCard: React.FC<
...
   useEffect(() => {
-    const fetchImpactOptions = async () => {
-      if (documentType === 'Article') {
-        await getImpactOptions();
-      }
+    if (documentType !== 'Article') return;
+
+    const loadImpactOptions = async () => {
+      const options = await getImpactSuggestions('');
+      setImpactOptions(options || []);
```
AimeurAmin: nit:
I believe `getImpactSuggestions` will always return a promise of an array, so no need to do `options || []`

apps/crn-frontend/src/shared-research/util.ts
AkosuaA: `project` was added in anticipation of future project research output flows but since the flows aren't implemented yet, it makes sense to remove project from the type

AimeurAmin (approved):
Looks good to me, just left few comments

packages/model/src/research-output.ts
AkosuaA: I think it's better to keep the flow IDs here and have the resolver in CRN frontend. There are some things under `model/src/research-output` that are shared and used by GP2 as well but where CRN and GP2 differ, we keep the GP2-specific versions under `model/src/gp2/output` so I don't think that should be a problem if we need GP2-specific flow IDs in the future.

The reason I added the IDs and `ResearchOutputFlowId` type here is that the type can be reused by `ResearchOutputForm` in the `react-components` package, which already imports types from `@asap-hub/model`.

gabiayako (approved):
nice 🎉

PR #5065, opened 2026-06-15
[ASAP-1513] - Filter people by membership status

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1513

Adds a **Type of Users** filter (GP2 Member / Alumni Member) to the GP2 People List, mirroring CRN's `UserMembershipStatus`. Membership is derived from `alumniSinceDate` at write time, both in the EventBridge indexer and in the bulk `export-entity.ts` script, so both write paths stay consistent.

Algolia gains a `membershipStatus` facet. The Algolia filter builder now wraps each group in parentheses so multi-value groups combine correctly. The REST `/users` endpoint translates `filter[membershipStatus]` into `OR { alumniSinceDate_exists }` for the CSV export.

The new section is rendered with `LabeledCheckbox` in a flex row — 2 columns on desktop, stacked with 16 px gap on mobile.

## Deploy

After merge, trigger **Sync GP2 Algolia** (`entity: users`) so existing records get the `membershipStatus` field; without this the filter returns empty values.

## See it in action

https://github.com/user-attachments/assets/d4e6f43e-2bf9-4356-bb60-45660bb64fe6

### DIFF
--- a/packages/algolia/schema/gp2/algolia-schema.json
+++ b/packages/algolia/schema/gp2/algolia-schema.json
@@ -44,6 +44,7 @@
     "documentType",
     "eventTypes",
     "id",
+    "membershipStatus",
     "project.id",
     "projects.id",
     "projectIds",
--- a/apps/gp2-frontend/src/users/UserDirectory.tsx
+++ b/apps/gp2-frontend/src/users/UserDirectory.tsx
@@ -1,6 +1,6 @@
 import { createCsvFileStream } from '@asap-hub/frontend-utils';
 import { UsersPageList } from '@asap-hub/gp2-components';
-import { useCurrentUserGP2 } from '@asap-hub/react-context';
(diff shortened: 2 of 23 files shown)


packages/algolia/schema/gp2/algolia-schema.json
gabiayako, line 1: Although we don't call the reverse timestamp index when searching users, I believe it's a good practice to add the same facet here `packages/algolia/schema/gp2/algolia-reverse-timestamp-schema.json` just in case we need to use it :)

gabiayako (approved):

packages/algolia/schema/gp2/algolia-schema.json
diegonvs: See [c108617](https://github.com/yldio/asap-hub/pull/5065/commits/c108617d4eaf26bffd38a3cbb71310476c22753d)

gabiayako (approved):

PR #5068, opened 2026-06-17
ASAP-1516 Display Active and Past members for Projects

gabiayako:
Ticket: https://asaphub.atlassian.net/browse/ASAP-1516

---

#### Before / After

Screenshot 2026-06-17 at 09 39 52


packages/gp2-components/src/templates/ProjectOverview.tsx
```diff
@@ -54,6 +58,43 @@ const columnStyles = css({
...
+  'Project lead',
+  'Project manager',
+]);
+
+const isActive = (status: gp2.ProjectStatus) => (member: gp2.ProjectMember) =>
+  status !== 'Completed' &&
+  !member.alumniSinceDate &&
+  !member.inactiveSinceDate;
+
+const splitActivePast = <T,>(
```
diegonvs: We have a `splitListBy` which can be used like:

```
const [activeLeaders, pastLeaders] = splitListBy(leaders, isActive(status));
const [activeMembers, pastMembers] = splitListBy(regularMembers, isActive(status));
```

So, this splitActivePast is not needed.

packages/gp2-components/src/templates/ProjectOverview.tsx
```diff
@@ -54,6 +58,43 @@ const columnStyles = css({
...
+  predicate: (item: T) => boolean,
+): [T[], T[]] => {
+  const active = [];
+  const past = [];
+  for (const elmnt of arr)
+    predicate(elmnt) ? active.push(elmnt) : past.push(elmnt);
+  return [active, past];
+};
+
 const ProjectOverview: React.FC<ProjectOverviewProps> = ({
```
diegonvs, line 65: I believe it not worth to be worked on this PR, since it extrapolates a lot the scope of the ticket, but GroupLeadersTabbedCard inside packages/react-components/src/organisms/WorkingGroupMembers.tsx does mostly the same as this ProjectOverview component. What it differs is the shape of the data. Maybe we can create a ticket for it.

diegonvs (commented):
I tested and it's looking okay. Just some suggestions/ideas

diegonvs (approved):

PR #5071, opened 2026-06-18
[ASAP-1525] - resolve flow ids in containers

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1525


apps/crn-frontend/src/shared-research/util.ts
```diff
@@ -59,25 +59,23 @@ export type ResolveFlowIdParams = {
...
-  isImportedFromManuscript,
-  isDuplicate,
+  isImportedFromManuscript = false,
+  isDuplicate = false,
   hasResearchOutputId,
 }: ResolveFlowIdParams): ResearchOutputFlowId => {
   const isTeam = entityType === 'team';
 
   if (isDuplicate) {
-    return isTeam
```
AkosuaA, line 78: I realized for working group we don't currently pass the isDuplicate prop when duplicating a working group output so I removed that flow (it looked to be similar to the creation flow just with pre-populated fields). 

Image

However taking another look I realized with duplicating outputs, if the user doesn't update the description we show an alert and that is controlled by the `descriptionUnchangedWarning` prop 

Image

so maybe we maintain the duplicate flow just for that case and use that flow to determine whether to show the warning or not (and remove the `descriptionUnchangedWarning` prop once we start using the flow ids). 
Let me know what you think 🙂

gabiayako: Hey, I believe we only use `isDuplicate` to resolve this `showManuscriptOutputFlow`

```
  const [showManuscriptOutputFlow, setShowManuscriptOutputFlow] = useState(
    !isManuscriptVersion &&
      !isDuplicate &&
      documentType === 'Article' &&
      !updatedOutput?.id,
  );
```

I don't think it actually prevents the duplication when is not passed as true

AkosuaA: Yea, we only use it to resolve `showManuscriptOutputFlow` which is something we only do for team outputs and not working group outputs. That's why it's not passed for working groups. So I wasn't sure if we need a duplicate flow for working groups or not.

Can we use the `showManuscriptOutputFlow` and the`descriptionUnchangedWarning` modals to help me understand? 😅 
So we have the flow TEAM_DUPLICATE. Then in the flow registry for TEAM_DUPLICATE we'd have as one of the fields `supportsManuscriptOutputFlow: false` or something like that.
Screenshot 2026-06-18 at 1 00 46 PM

Then in the actions resolver 
Screenshot 2026-06-18 at 12 53 22 PM
we'd have an action called `showManuscriptOutputFlow` which would be computed using the flow's `supportsManuscriptOutputFlow` field and some other conditions. And so for team duplicates since `supportsManuscriptOutputFlow: false` then showManuscriptOutputFlow will be false. And this action `showManuscriptOutputFlow` will be used to set the state value 😅😅 This is how I'm thinking of it but not sure if I understand. I wanted to check this first before bringing up the `descriptionUnchangedWarning`

gabiayako: Yeah, I think that’s right 😊
Sorry, I thought you were saying earlier that we don’t have duplicates in working groups. I misunderstood your message 😅
Even though `WORKING_GROUP_DUPLICATE` doesn’t seem to be special in terms of having different rules of its own, I think it would still be nice to keep it as a flow so it’s clear that it’s an existing flow. Do you agree?

AkosuaA: Yea, I'll put it back in. The `descriptionUnchangedWarning` is present only for duplicates (both working group and teams) so that would be the difference between working group create and duplicate. Thanks 😊 Making the update

gabiayako: thanks 😄

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
```diff
@@ -105,6 +106,14 @@ const WorkingGroupOutput: React.FC<WorkingGroupOutputProps> = ({
     published,
   );
 
+  const flowId = resolveResearchOutputFlowId({
+    entityType: 'working-group',
+    versionAction,
+    published,
+    hasResearchOutputId: !!researchOutputData?.id,
+    isDuplicate: !!descriptionUnchangedWarning,
```
gabiayako, line 114: I get what you mean now 😄 it seems messy, but I think it's a good thing to have this prop called `isDuplicate` in `resolveResearchOutputFlowId` so it's clearer

apps/crn-frontend/src/shared-research/__tests__/util.test.ts
```diff
@@ -81,16 +81,6 @@ describe('resolveResearchOutputFlowId', () => {
   });
 
   describe('working group flows', () => {
-    it('returns WORKING_GROUP_DUPLICATE when duplicating', () => {
```
gabiayako, line 84: I believe this could come back :)

gabiayako (approved):
nice! 👏

apps/crn-frontend/src/network/working-groups/WorkingGroupOutput.tsx
AkosuaA: Yep, agreed!

PR #5074, opened 2026-06-19
ASAP-1538 Add BlueSky on User Profile

gabiayako:
Screenshot 2026-06-19 at 09 27 55
Screenshot 2026-06-19 at 10 58 34
Screenshot 2026-06-19 at 10 58 45


packages/react-components/src/templates/ContactInfoModal.tsx
```diff
@@ -87,6 +90,7 @@ const ContactInfoModal: React.FC<ContactInfoModalProps> = ({
           contactEmail: newEmail || undefined,
           social: {
             twitter: formatUserSocial(newTwitter, 'twitter') || undefined,
+            blueSky: newBlueSky || undefined,
```
diegonvs: Hey, why we don't used formatUserSocial here? 

Since we're not doing it, the url for bluesky if the user give the full link of his profile will be broken. 

https://github.com/user-attachments/assets/7aef2ddb-5fd7-44b5-a356-034953d29921

diegonvs: It will look like this: 

Screenshot 2026-06-19 at 17 28 14

gabiayako: ops 🙈 I didn't notice that. I'll fix it now

gabiayako: thank you 😄

gabiayako:
I also deleted `packages/contentful/migrations/crn/users/20260619132627-move-bluesky-field.js.` I had added it because, in the CMS Development environment, the BlueSky field was showing up much later in the form than I expected. I hadn’t noticed that we already had `users.moveField('blueSky').afterField('researchGate')` in the original migration.
The idea was to move the BlueSky field closer to the other social fields, but it didn’t seem to have any effect. After taking a closer look in Contentful, I realized the field order wasn’t what I expected because of the field Group configuration. 
Screenshot 2026-06-19 at 14 10 12

That doesn’t seem to be something we can configure through migrations.

diegonvs (approved):

PR #5078, opened 2026-06-23
[ASAP-818] - Enable users to manage their profile photo

AimeurAmin:
[ASAP-818](https://asaphub.atlassian.net/browse/ASAP-818)

[ASAP-818]: https://asaphub.atlassian.net/browse/ASAP-818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/crn-server/src/data-providers/contentful/user.data-provider.ts
+++ b/apps/crn-server/src/data-providers/contentful/user.data-provider.ts
@@ -351,11 +351,18 @@ export class UserContentfulDataProvider implements UserDataProvider {
       ...(data.tagIds ? { researchTags: getLinkEntities(data.tagIds) } : {}),
     });
 
+    // only delete the old asset once the entry update actually applied,
+    // otherwise a suppressed conflict would leave the entry linking a
+    // deleted asset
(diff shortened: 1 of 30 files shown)


apps/crn-server/src/data-providers/contentful/user.data-provider.ts
```diff
@@ -334,18 +335,53 @@ export class UserContentfulDataProvider implements UserDataProvider {
...
-    if (!result || !polling) {
+
+    // only delete the old asset once the entry update actually applied,
+    // otherwise a suppressed conflict would leave the entry linking a
+    // deleted asset
+    if (!result) {
+      return;
+    }
+
+    if (removedAvatarId) {
```
diegonvs: Here we are checking for fields.avatar === null to delete an asset. But what If the user replaces an existing avatar? This condition will not be true and the deleteAsset is never called, right?

diegonvs (approved):

PR #5082, opened 2026-06-25
fix: discussion links

gabiayako:
Discussion links were directing users to a team or project workspace that they weren't allowed to view in some cases. For example, a team-based manuscript where the user isn't part of the submitting team.

To fix that, discussion emails now send the user to a new route, `/compliance/manuscripts/:id?tab=discussions`. When the user access this route, it makes a request to the backend to decide which URL the user should be sent to, so we can check the teams and projects the user currently belongs. It also makes it easier to manually test different scenarios without having to send emails everytime.

This PR also removes the old approach that used `workspaceType`.

I've made this diagram to organize the rules https://whimsical.com/my-workspace939368/manuscripts-MESnSeTvN93yPgaRU2udBV although in the end I made a refactor to start with "is feature flag enabled?" or not so it is easier to remove the branch where the answer is no.


AimeurAmin (approved):
very nice 🚀👏🏻

PR #5089, opened 2026-07-03
[ASAP-1572] - Require Labs and lock Date made Public on the Output form

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1572

Makes the **Labs** field required on the CRN research Output form (create, edit and new-version). Saving or publishing with Labs empty is now blocked and shows "Please add at least one lab." 

https://github.com/user-attachments/assets/4894ff76-6b9f-4c7b-a944-eb6fb087afe4

Disables the **"Public Repository Published Date"** field when editing an existing output or creating a new version, but only when it already has a value. Additional: outputs that were published without a date can still have one entered.

https://github.com/user-attachments/assets/e8034d30-6a91-4e2d-ac0a-861db2a562c8

https://github.com/user-attachments/assets/e5e6e686-7c6c-4897-a0ac-b46b0f180505


packages/react-components/src/organisms/ResearchOutputPublishingCard.tsx
```diff
@@ -115,6 +115,7 @@ const ResearchOutputFormSharingCard: React.FC<
         description={
           'This should be the date your output was shared publicly on its repository.'
         }
+        enabled={!researchOutputData?.publishDate}
```
gabiayako: I believe the same issue from the other PR happens here where if we change from Public to CRN Only we keep sending the publishDate

https://github.com/user-attachments/assets/99f1e691-7cb0-4a20-a645-784121d51119

gabiayako (approved):

PR #5090, opened 2026-07-03
ASAP-1563 URL Sign In Redirect

gabiayako:
There was an issue where, after logging in, users would remain on the original URL they had tried to access, but the content displayed did not match that URL. This happened because:
* replaceState only updates the browser’s history.
* React Router maintains its own internal location state.
* Updating window.history directly does not guarantee that React Router’s internal location state will be updated.

The solution was to use React Router’s navigation instead of calling `window.history.replaceState` directly.

Additionally, since `apps/crn-frontend/src/auth/react-auth0-spa.tsx` and `apps/gp2-frontend/src/auth/react-auth0-spa.tsx` were very similar, they were consolidated into a single shared component at `packages/react-context/src/Auth0SpaProvider.tsx`.

Another thing is that I needed to regenerate `.pnp.cjs`. Since Yarn 4 enables `enableGlobalCache: true` by default, running yarn install locally rewrote `.pnp.cjs` to reference my machine’s global cache, which would break the Zero-Install CI jobs. To avoid that, I set `enableGlobalCache: false` in `.yarnrc.yml` before regenerating `.pnp.cjs`, so it references the committed `.yarn/cache` instead.


AimeurAmin (approved):
very nice ✌🏻

PR #5093, opened 2026-07-06
[ASAP-1526] - add save drafts project flow

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1526

- Adds the `Share an Output` button to Discovery Projects only when the `PROJECT_OUTPUTS` flag is on, since those are team-based. 
Screenshot 2026-07-08 at 3 14 14 PM

The idea here is the Project Outputs flow (which is behind a flag) will use the available actions on the ResearchOutputForm while the other areas (Teams and Working Groups) use the existing implementation. Once all the available actions have been implemented and tested we can safely switch Team and Working Groups to use available actions as well.

- For the `SharedResearchOutputButtons` component, I realized the buttons were not resulting from flows but rather triggers for the flows, eg: viewing the Edit button on a Research Output is not a flow but clicking the Edit button opens the Output Form in either the TEAM_EDIT_DRAFT flow, TEAM_EDIT_PUBLISHED flow or the working group equivalents. It made sense to still centralize the conditions for the buttons, which I've tried to do with the `getVisibleResearchOutputActions` function. I did not put this behind a flag because it seemed straightforward. It should be the same conditions that were previously in the buttons component, just moved into one function. Do let me know what you think. If it's necessary or can be implemented in a different way.

I struggled with figuring out which parts were available actions or not, so combining [the info from the spike](https://app.notion.com/p/Outputs-Form-Refactor-FlowRegistry-AvailableActions-3870dae759ea801687e5f4187fa0264b?v=972c280e3f7a422dbedf67dea107742a) I put together t[his sheet](https://docs.google.com/spreadsheets/d/1yZ0TLaM1o6QmuIH7_eInXZkqC10GMVaTKgudBP-sBdo/edit?gid=0#gid=0) to clear my understanding. I'm still building it out but basically it should help determine the conditions for a field/button/action to be visible/occur and also clarify which fields should go into the flow registry and how the combination of permissions, flow registry and output data should determine an available action. It's based on point 5 (especially 5.2) in the spike document linked and aside the first (Save Draft) action, I'm not 100% sure of the rest :sweat_smile:. Feel free to comment/fix/update etc


packages/react-components/src/templates/SharedResearchOutput.tsx
AkosuaA, line 196: I was wondering whether we should move this check into `getVisibleResearchOutputActions` and if all actions are false the `SharedResearchOutputButtons` component returns nothing. 🤔 Let me know what you think.

packages/react-components/src/organisms/__tests__/SharedResearchOutputButtons.test.tsx
```diff
@@ -12,506 +20,36 @@ const props: ComponentProps<typeof SharedResearchOutputButtons> = {
...
-    </ResearchOutputPermissionsContext.Provider>,
+it.each([
+  ['canEdit', 'link', /edit/i],
+  ['canDuplicate', 'link', /duplicate/i],
+  ['canRequestReview', 'button', /ready for pm review/i],
+  ['canImportManuscriptVersion', 'button', /import manuscript version/i],
+  ['canAddVersion', 'link', /add version/i],
+  ['canSwitchToDraft', 'button', /switch to draft/i],
+  ['canPublish', 'button', /publish/i],
+] as const)('renders %s when %s is true', (action, role, name) => {
```
gabiayako: Very minor, but the title of the tests are like this now

Image

maybe it would be better to use a table like

```
it.each`
  title                       | action                      | role        | name
  ${'Edit'}                   | ${'canEdit '}               | ${'link'}   | ${/edit/i}
(...)
`('renders $title when $action is true', ({ action, role, name }) => {
```

packages/react-context/src/permissions/research-output.ts
```diff
@@ -21,3 +22,64 @@ export const ResearchOutputPermissionsContext =
...
+    !state.hasRelatedManuscript,
+  canImportManuscriptVersion:
+    !!permissions.canVersionResearchOutput &&
+    state.published &&
+    state.hasRelatedManuscript,
+  canSwitchToDraft:
+    !state.published &&
+    state.isInReview &&
+    !!permissions.canPublishResearchOutput,
+  canPublish: !state.published && !!permissions.canPublishResearchOutput,
```
gabiayako: I was wondering if we should call the actions "show":

```
showEdit,
showDuplicate,
...
```

because "can" seems it's a permission

packages/react-components/src/templates/SharedResearchOutput.tsx
gabiayako: Yeah, I believe it makes sense 😄

apps/crn-frontend/src/projects/ProjectDetail.tsx
```diff
@@ -143,6 +169,7 @@ const ProjectDetail: FC<Props> = ({ config }) => {
 
   const workspaceHref = showWorkspace ? route.workspace({}).$ : undefined;
   const isProjectOutputsEnabled = isEnabled('PROJECT_OUTPUTS');
+  const isDiscoveryProject = config.projectType === 'Discovery Project';
```
gabiayako: should we also add the route for Resource Projects that are team based?

gabiayako (approved):
really nice ✅

apps/crn-frontend/src/projects/ProjectDetail.tsx
AkosuaA: Thanks! I've made the suggested improvements in [this commit](https://github.com/yldio/asap-hub/pull/5093/changes/d4c21f60a08530b1821cf63dca586f1864971725). Could you take a look? Should be quick 🙂

gabiayako (approved):
It looks nice!

PR #5096, opened 2026-07-08
ASAP-1575 Fix scroll to manuscript card

gabiayako:
Sometimes scrolling to the Manuscript Card didn’t work properly because the element’s position changed while the page was still loading. In order to fix that, this PR reuses `useScrollToHash` mounted in Layout and refactors the hook so that:
* it keeps checking every animation frame until the element stops changing position before scrolling
* if more than 1 second passes, it scrolls anyway
* after scrolling, it keeps checking in case the layout shifts again and the element moves out of view
* if more than 10 seconds pass, it gives up

### DIFF
--- a/apps/crn-frontend/src/compliance/ManuscriptWorkspaceRedirect.tsx
+++ b/apps/crn-frontend/src/compliance/ManuscriptWorkspaceRedirect.tsx
@@ -20,7 +20,9 @@ const ManuscriptWorkspaceRedirect: FC = () => {
     return <ManuscriptWorkspaceUnavailablePage />;
   }
 
-  return <Navigate to={workspaceUrl.url} replace />;
+  return (
+    <Navigate to={workspaceUrl.url} replace state={{ scrollToHash: true }} />
+  );
 };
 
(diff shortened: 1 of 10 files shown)


gabiayako:
@AimeurAmin Thanks a lot for pointing the issue with the scroll. I noticed it was happening on production as well

https://github.com/user-attachments/assets/ee7d8462-4a5d-46d0-92bb-f90606803356

I think it was introduced here https://github.com/yldio/asap-hub/pull/5090. Since the auth0 value in Auth0SpaProvider wasn't memoized, every URL change created a new auth0 object. That new reference re-triggered the auth selector, which made the data-dependent components suspend for a moment and show a loading fallback. So when we expanded cards, the URL sync was making the page jump up unnecessarily.

The other issue was with ProfileSwitch, where we were passing Workspace as an inline function, so React saw a new component type every render and re-mounted the whole subtree instead of just re-rendering it. That remount showed the same loading fallback and caused the same jump.

AimeurAmin (approved):
very nice and clean 🚀

PR #5097, opened 2026-07-08
[ASAP-1204] - Remove react-lottie, add centralized CSS Spinner atom

diegonvs:
`react-lottie` was only used for three submit/loading spinners. It pulls a heavy runtime into the bundle, which forced two components to be `lazy()`-loaded purely to code-split it away. PR #4706 already introduced a pure-CSS spinner and let's reuse them.

 **Example**
 
Before:

https://github.com/user-attachments/assets/cea1ffaf-12a2-48c0-bd3f-3e820735cab9

After: 

https://github.com/user-attachments/assets/90687582-774d-4926-b08f-c1283a3c1c28


apps/storybook/src/Spinner.stories.tsx
AimeurAmin, line 1: not sure the spinner is worth a story 😅

apps/storybook/src/CookiesModal.stories.tsx
AimeurAmin, line 1: how is cookies modal related?

AimeurAmin (approved):
looks good to me, approved with two minor comments 👌🏻🚀👏🏻

apps/storybook/src/CookiesModal.stories.tsx
diegonvs: oops, I was using it to test the spinner with the button. I'll remove it 🙈

apps/storybook/src/Spinner.stories.tsx
diegonvs: I can remove it

PR #5101, opened 2026-07-10
[ASAP-1282] - Replace recoil with tanstack react query

AimeurAmin:
[ASAP-1282](https://asaphub.atlassian.net/browse/ASAP-1282)
[ASAP-1410](https://asaphub.atlassian.net/browse/ASAP-1410)

# Replace Recoil with TanStack Query

This PR moves both frontends from Recoil to **TanStack Query v5**. Recoil is completely gone — no imports, no dependency, nothing in the lockfile. Recoil states were all server data cached on the client, which is exactly what TanStack Query is made for (and it supports Suspense, which our loading UX depends on).

## How writes work — `useMutation` + `setQueryData`

Every server write (24 across both apps) is a `useMutation`, so you can see them all in the devtools Mutations tab. But the hooks still return a plain async function (`mutateAsync`), so **forms work exactly like before and no component was changed**.

```ts
const { mutateAsync } = useMutation({
  mutationFn: async (patch) => patchUser(id, patch, await getAuthorization()),
  onSuccess: (user) => queryClient.setQueryData(userQueryKeys.detail(id), user),
});
return mutateAsync;
```

**Why we write to the cache instead of refetching:** Contentful is slow to return fresh data right after a write. If we refetched, we would often get the *old* data back and the user's edit would look like it was lost. So we take the write's response and put it straight into the cache — in the code we call these entries **"overlays"** (see `refetchAllButOverlays`). Rule: **overlay entries are never refetched, and invalidation never touches them.** When a write changes more than its response tells us (e.g. creating an output), we mark the lists stale with `refetchType: 'none'` — they refetch on the next visit, after the search index has caught up.

Cache updates always live in the hook (`onSuccess` / `onMutate`), never in per-call callbacks (those are skipped if the component unmounts). Some things are on purpose **not** mutations because they are reads: the batch manuscript fetch (a POST, but a read), the AI content generator, and the suggestion fetchers.

## Cache behavior

The defaults are set so the app behaves the same as it did with Recoil: data never refetches by itself (`staleTime`/`gcTime: Infinity`), failed requests are not retried (`retry: false`), and there is no refetch on tab focus or reconnect. The QueryClient is created inside the logged-in part of the app, so logging out throws away the whole cache — same as the old RecoilRoot. Turning on smarter refetching is a follow-up, not this PR.

## Structure

Components never import `@tanstack/react-query`. All query and mutation logic lives in each module's `state.ts` behind custom hooks, with one key factory per module (e.g. `userQueryKeys.detail(id)`). Call sites are basically unchanged.

## Behavior changes (intentional)

- Some GP2 "refresh counters" were broken silently on `master`, due to recoil behavior — writes silently refreshed nothing. They now really refresh the lists (safely, after the index catches up).
- Profile edits show up again in other cached views (team lists etc.), without the old side effect where the whole app unmounted.
- If the Auth0 token refresh fails after a successful save, we log a warning instead of wrongly telling the user the save failed. (On localhost the silent token refresh always fails — Auth0 cannot skip the consent check there — so this happened on every profile save.)

## Bugs found and fixed along the way

- **Infinite refetch loop on GP2 startup**: the QueryClient provider never finished mounting when a child suspended, so React created a new client (and refetched) over and over. Fixed with a Suspense boundary inside the provider (both apps) + a regression test.
- **False "unable to save" error** on profile edits (see above).
- **Dead code removed**: `usePatchTeamById` and its `patchTeam` API function (unused since ASAP-1278), `usePresignedUrl`, `TestQueryClientWrapper`, leftover jest recoil setup.
- From the post-migration review: the OpenSearch token was frozen for the whole session (analytics would get 401s after it expired), project tools could be overwritten by a refetch right after saving, and empty manuscript lists showed a loading state for no reason.

## Extras

The `QUERY_DEVTOOLS` feature flag shows the TanStack devtools icon on Development and PR environments only — never on Production.

# **A small demo of the benifits of saying goodbye to Recoil:**
This video is a comparison between current Development env version and this PR env, not only updating the profile data causes a refresh of the page, it also keeps data stale in some parts of the application until next refresh, the second part of the video shows how Tanstack Query handles that gracefully with the current implementation and spreads new data across the app:

https://github.com/user-attachments/assets/c52f6bfe-932c-4bc6-a3c0-cbc2d463785a

# Final Dev note:
- Tanstack Query Devtools are available on all PR envs and Development env through the flag `document.cookie="ASAP_QUERY_DEVTOOLS=true"`.
- Devtools are not available on Production.

[ASAP-1410]: https://asaphub.atlassian.net/browse/ASAP-1410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ASAP-1282]: https://asaphub.atlassian.net/browse/ASAP-1282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-frontend/src/news/state.ts
```diff
@@ -1,97 +1,35 @@
...
-    },
-});
+export const newsQueryKeys = {
+  all: ['news'] as const,
+  lists: () => [...newsQueryKeys.all, 'list'] as const,
+  list: (options: GetListOptions) =>
+    [...newsQueryKeys.lists(), normalizeListOptions(options)] as const,
+  details: () => [...newsQueryKeys.all, 'detail'] as const,
+  detail: (id: string) => [...newsQueryKeys.details(), id] as const,
+};
```
gabiayako: I was wondering if we could do something like

```
export const createQueryKeys = <O extends object = GetListOptions>(
  scope: string,
) => ({
  all: [scope] as const,
  lists: () => [scope, 'list'] as const,
  list: (options: O) => [scope, 'list', normalizeListOptions(options)] as const,
  details: () => [scope, 'detail'] as const,
  detail: (id: string) => [scope, 'detail', id] as const,
});
```

and reuse across the files

apps/crn-frontend/src/network/interest-groups/state.ts
```diff
@@ -1,113 +1,55 @@
...
-  string
->({
-  key: 'fetchInterestGroup',
-  get:
-    (id) =>
-    async ({ get }) => {
-      get(refreshInterestGroupState(id));
-      const authorization = get(authorizationState);
-      return getInterestGroup(id, authorization);
+export const useInterestGroups = (
```
gabiayako, line 21: I noticed this pattern is repeated across some `state` files

Image

I wonder if it makes sense to extract a helper function to keep these functions consistent

apps/crn-frontend/src/network/interest-groups/state.ts
```diff
@@ -1,113 +1,55 @@
...
-  if (interestGroups instanceof Error) {
-    throw interestGroups;
-  }
-  return interestGroups;
+  }).data;
 };
 
-export const useInterestGroupById = (id: string) =>
-  useRecoilValue(interestGroupState(id));
+export const useInterestGroupById = (
```
gabiayako, line 35: The same thing here, I wonder if we could extract a helper function

Image

apps/crn-frontend/src/news/state.ts
AimeurAmin: Great call, it's done with two deviations from your sketch worth knowing:
- It's spreadable, because a few modules extend the standard shape (`teams` adds `batch(ids)`, gp2 `working-groups` adds `network()`): `{ ...createQueryKeys('manuscripts'), batch: ... }.`
- There's also a `createListQueryKeys` variant for list-only modules with no detail queries (project milestones, gp2 news/tag-search), `createQueryKeys` composes on top of it, so there's one source of truth and no dead methods.

The analytics factories deliberately don't adopt it, they key by specific option-Picks rather than `normalizeListOptions`, and forcing them through the helper would change their serialized keys and orphan existing cache entries.

apps/crn-frontend/src/network/interest-groups/state.ts
AimeurAmin: I extracted the two queryFn patterns repeated across the state files:
 - `withEmptyListFallback(fetch, empty)`: an Error re-throws to the error boundary, a non-Error rejection becomes an empty list.
 - `nullOnUndefined(fetch)`: the "a queryFn must not return undefined, cache null instead" rule for detail hooks that 404 as undefined.
 
Each module still owns its `useSuspenseQuery` call, only the shared semantics moved. All existing tests pass unchanged, which is the proof behavior didn't move.

apps/crn-frontend/src/network/interest-groups/state.ts
AimeurAmin: covered in previous comment ✌🏻

apps/crn-frontend/src/news/state.ts
gabiayako: Oh, I didn't notice there was already these helpers in place 🙈 I scanned the code to get a general idea of the refactor and ended up noticing the places that weren’t using the helper function.

gabiayako (approved):

diegonvs (approved):

PR #5104, opened 2026-07-13
[ASAP-1570] - Event Attendance component

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1570

Check Storybook on Organisms > Events > Attendance

Figma references: https://www.figma.com/design/qM73Om8hMDi8o15nj09ECw/CRN---Designs---Other-Pages?node-id=31964-80269&t=K9b4t02Hl5LTTRHR-4


AimeurAmin (commented):
I believe the empty state from the designs is missing: 

Image

Image

also it would be nice to have this as a story as well (case when percentage is low - orangish color):

Image

packages/react-components/src/organisms/EventAttendance.tsx
```diff
@@ -0,0 +1,314 @@
...
+  const attendancePercentage =
+    teamsTotal > 0 ? Math.round((teamsAttended / teamsTotal) * 100) : 0;
+
+  return (
+    <Card padding={false}>
+      <div css={[contentStyles, showFooter && contentWithFooterStyles]}>
+        <div css={headerStyles}>
+          <Headline3 noMargin>Attendance</Headline3>
+          <div css={actionsStyles}>
+            <button
```
AimeurAmin: I believe we should use the component `Button` in here instead of the raw `<button>`

packages/react-components/src/organisms/EventAttendance.tsx
```diff
@@ -0,0 +1,314 @@
...
+              type="button"
+              css={iconButtonStyles}
+              onClick={onExport}
+              aria-label="Download attendance"
+            >
+              {ExportIcon}
+            </button>
+            <button
```
AimeurAmin: same as previous

diegonvs:
> also it would be nice to have this as a story as well (case when percentage is low - orangish color):

You can use the controls on the story 

Screenshot 2026-07-15 at 10 34 14

packages/react-components/src/organisms/EventAttendance.tsx
diegonvs: <Button> is built for text buttons, for a compact icon square it forces min-width: 100% on mobile, stretches the SVG to 24px, and applies the secondary border/shadow + 14px padding. Matching the design would mean overriding nearly all of its styles, so we'd be using <Button> as an empty shell. A native <button> is cleaner and it's already the pattern here (for example `MeetingRepAttendanceTable` sort buttons).

diegonvs:
Added new stories for the empty state. Thanks for catching that, Amin! I completely forgot about it. 🤦🏼

AimeurAmin:
> > also it would be nice to have this as a story as well (case when percentage is low - orangish color):
> 
> You can use the controls on the story
> 
> Screenshot 2026-07-15 at 10 34 14

I know I can 😅 I suggested that because it's clearer and easier to spot, not cause it's impossible to do

AimeurAmin (approved):

diegonvs:
Pushed commit [`033ae8d` (this PR)](https://github.com/yldio/asap-hub/pull/5104/commits/033ae8de084d7100dacc4f641f2686bbcd8c5ca6)  with the changes requested by Padmini. Following yesterday's meeting, we changed the displayed value from a percentage to a count.

PR #5111, opened 2026-07-15
ASAP-1527 Refactoring Modals inside Research Output Form

gabiayako:
This PR adds 3 new conditions to `getResearchOutputFlowBehavior`:
* `requiresAddVersionConfirm`
* `requiresPublishConfirm`
* `requiresSameDescriptionConfirm`

These conditions control whether the confirmation modals are shown when the user saves or publishes an output. They are derived from the flow descriptor, since in this PR `FLOW_REGISTRY` was reworked into `FLOW_DEFINITIONS`, where each flow is described by its entity, action, origin and startsPublished. This makes the flow registry the single source of truth for this behavior.

In this PR the modals were also moved to a separate component. The two states that tracked whether a modal had already been dismissed were grouped into a single `alreadySeenModals` state, and the states that controlled which modal is open were grouped into a single modal state.

I also ended up removing the `behaviorMode` ('legacy' | 'flow') flag, since I think it would be difficult to maintain both paths at once. As a result, flowId and availableActions are now required props and the descriptionUnchangedWarning prop was removed.

---

### Same description

Screenshot 2026-07-14 at 06 35 40

Only appears when duplicating an output. 
Variations:
- Button Copy: “Keep and publish” → when publishing
- Button Copy: “Keep and save” → when saving a draft

If user presses “Keep and (publish|save)” once, the modal won’t appear again.

### Publish new version

Screenshot 2026-07-14 at 06 35 19

Only appears when adding a new version.
If user presses “Publish new version” once, the modal won’t appear again.

### Publish output
Screenshot 2026-07-14 at 06 33 32

Appears when publishing an output that is not a version.
It does not have the same rules as the other modals that once seen it doesn’t appear again.

### DIFF
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -185,7 +185,11 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
     published,
     isImportedFromManuscript,
     isDuplicate,
-    hasResearchOutputId: !!researchOutputData?.id,
+    // When importing an already-published manuscript whose preprint was
+    // never shared as an output, the backend auto-creates the preprint
+    // output and updatedOutput gains its id at runtime, turning this
+    // import into an add-version flow.
+    hasResearchOutputId: !!(updatedOutput?.id || researchOutputData?.id),
   });
   const availableActions = resolveResearchOutputAvailableActions({
     flowId,
--- a/apps/crn-frontend/src/network/teams/TeamOutput.tsx
+++ b/apps/crn-frontend/src/network/teams/TeamOutput.tsx
@@ -18,20 +18,17 @@ import {
   Toast,
   usePrevious,
 } from '@asap-hub/react-components';
-import { InnerToastContext } from '@asap-hub/react-context';
+import {
+  InnerToastContext,
+  resolveResearchOutputAvailableActions,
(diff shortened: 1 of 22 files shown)


apps/crn-frontend/src/network/teams/__tests__/TeamOutput.test.tsx
```diff
@@ -1028,7 +1031,7 @@ describe('manuscript outputs flow', () => {
       await user.click(screen.getByRole('button', { name: /Publish/i }));
 
       const button = screen.getByRole('button', {
-        name: /Publish new version/i,
+        name: 'Publish Output',
```
AimeurAmin:
```suggestion
        name: /Publish new version/i,
```

apps/crn-frontend/src/network/teams/TeamOutput.tsx
```diff
@@ -193,7 +187,11 @@ const TeamOutput: React.FC<TeamOutputProps> = ({
     isDuplicate,
     hasResearchOutputId: !!researchOutputData?.id,
```
AimeurAmin:
```suggestion
    hasResearchOutputId: !!(updatedOutput?.id || researchOutputData?.id),
```

AimeurAmin (approved):
Nice ! 👌🏻🚀

PR #5113, opened 2026-07-16
ASAP-1527 Refactoring banners and buttons on Research Output Form

gabiayako:
This PR deals with the toasts we display on the Research Output detail page when a draft is created or the output is published.

## The way it worked before

When the output was published we redirected to a /publishedNow route. The ResearchOutput page checked if there was a match on this route and, if so, it displayed the published successfully toast.

When a draft was created for the first time we added ?draftCreated=true to the redirect url. The ResearchOutput page checked if this search param was present and, if so, it displayed the created successfully toast.

## How it works now

We pass the type of toast that should be displayed in location.state of react-router and ResearchOutput displays the toast accordingly. After reading it, the page clears the state (replace navigation), so the toast doesn't show up again on refresh or back navigation.

Since the toast info no longer lives in the url, the /publishedNow route was removed from @asap-hub/routing, along with its extra <Route> in the ResearchOutput page.

The publish modal on the detail page (PM publishing a draft) also uses this mechanism now — ConfirmModal accepts a successState prop that is passed to the navigation on success.

To decide whether saving should show the published toast, a new publishesOnSave flag was added to getResearchOutputFlowBehavior in the model package (e.g. editing an already published output doesn't show it, but publishing or adding a version does).

## Other changes

- Buttons from ResearchOutputForm were moved to a separate component ResearchOutputFormActions
- SharedResearchOutputBanners was renamed to SharedResearchOutputToasts, and the publishedNow/draftCreated props were consolidated into a single toast prop
- In ResearchOutputForm, the single save(draftSave) function was split into save and saveDraft

## Note

I tried to remove setRedirectOnSave from Form, but the navigation stopped working. I believe it's better to try to solve this when we change to react-hook-form, because the redirect is coupled to the form's status state — the same state that drives the unsaved-changes navigation warning. It's hard to touch that now.

---

The `Team|Working Group {documentType} published successfully` toast is shown whenever a non-published output is published. It isn’t shown when an already published output is edited and published again.

Screenshot 2026-07-15 at 07 29 49

The `Draft Team|Working Group {documentType} created successfully` toast appears whenever a draft is created. It isn’t shown when a draft output is edited and saved again.

Screenshot 2026-07-15 at 07 51 14

So the rule of thumb is: if the status of the output changed we display a toast when navigating to Research Output Detail page. We don’t display it if it goes from

- `draft` → `draft`
- `published` → `published`


AimeurAmin (approved):
Really nice work 👌🏻

PR #5114, opened 2026-07-17
[ASAP-1560] - display os badges on user memberships

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1560


apps/crn-server/src/data-providers/transformers/users.ts
```diff
@@ -52,3 +74,27 @@ export const parseOrcidWorkFromCMS = (orcidWork: OrcidWorkCMS): OrcidWork => ({
...
+      return [
+        ...awards,
+        {
+          name,
+          date: award.date,
+          iconUrl:
+            award.awardType?.icon?.url ??
+            award.awardType?.smallIcon?.url ??
+            undefined,
+        },
```
AimeurAmin, line 94: small icon is showing fine right now, but that's only because membership graphql query doesn't fetch the normal icon, if at some point we do that then this will show the normal icon everywhere, so I believe this should separate the two fields
```suggestion
        {
          name,
          date: award.date,
          iconUrl: award.awardType?.icon?.url,
          smallIconUrl: award.awardType?.smallIcon?.url
        },
```

packages/contentful/migrations/crn/awardType/20260716144340-add-small-icon-field.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports.up = (migration) => {
+  const awardType = migration.editContentType('awardType');
+
+  awardType
+    .createField('smallIcon')
+    .name('Small Icon')
+    .type('Link')
+    .localized(false)
+    .required(false)
```
AimeurAmin: according to requirements of the ticket, the smallIcon should be required, no?

```suggestion
    .required(true)
```

AkosuaA: The icon field isn't required so I thought it may have been a mistake on the ticket to make it required. Will check with Padmini

AimeurAmin (approved):
LGTM 🚀

PR #5115, opened 2026-07-17
[ASAP-1569] Event Speakers component

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1569

Stacked on top of #5104. I'll merge that one Monday, but this is ready for review.

This PR does two things:

1. Adds the EventSpeakers component — the speakers card for the redesigned CRN event detail page (header + metrics + per-team accordion table + External Users row + role/past-aware empty states).
2. Extracts the chrome shared with Attendance (#5104) into reusable modules, so both cards stay in sync instead of drifting:
  - shared-event-card-styles.ts — card/header/table-cell styling
  - shared-event-card.tsx — teamIcon, EventTeamType, defaultVisibleTeams
  - shared-metric-card-styles.ts — metric shell styling
  - Attendance (#5104) was migrated onto these in the same PR.
  
Design decision — preliminary-findings chart as a conic gradient

The designer's spec for the "Preliminary findings" wheel/bar is a multi-stop brand gradient (purple → blue → teal → green) that fills along the arc, not a solid colour.

You can test on storybook under Events > Speakers > Playground

PR #5125, opened 2026-07-22
[ASAP-1602] - Extract GradientProgressBar/ProgressWheel

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1602

This PR builds on top of https://github.com/yldio/asap-hub/pull/5115, as it depends on the gradient implementation introduced there.

This PR extracts the `GradientProgressBar` and `ProgressWheel` components so they can also be reused in the Attendance card. The goal is to avoid code duplication, promote component reuse, and update the styling of the Attendance card.


AkosuaA (approved):
Looks good to me! 🎉

PR #5126, opened 2026-07-22
ASAP-1609 Add showChangelogAndVersionHistory to AvailableActions

gabiayako:
This PR moves the changelog and version history visibility logic into resolveResearchOutputAvailableActions so it lives in one place instead of being recalculated in each output page.

- Add `showChangelogAndVersionHistory` to ResearchOutputAvailableActions: true on edit/add-version flows when there are versions. This issue mentioned [here](https://yld.slack.com/archives/G01A6J33G8H/p1784727425124949) is fixed.
- Add `isEditFlow` to the flow behavior so the resolver can tell edit from the other flows.
- Pass researchOutputData/versions into the resolver and drop the `displayChangelog` prop from ResearchOutputForm it now reads from availableActions.
- Fix copy on the sharing card to say "output" instead of "manuscript".


AkosuaA (approved):
Nice 🎉👌🏾

gabiayako:
Rebased to master

PR #5128, opened 2026-07-23
[ASAP-1579] - display attendance information for past events

AimeurAmin:
[ASAP-1579](https://asaphub.atlassian.net/browse/ASAP-1579)

[ASAP-1579]: https://asaphub.atlassian.net/browse/ASAP-1579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/crn-frontend/src/events/Event.tsx
```diff
@@ -32,6 +50,29 @@ const Event: React.FC = () => {
     const displayCalendar =
       event.interestGroup === undefined || event.interestGroup.active;
 
+    const teams = mapAttendanceTeams(event.attendance);
+    const teamsTotal = teams.length;
+    const teamsAttended = teams.filter(({ attended }) => attended).length;
+    const isEventProjectManager = !!user?.interestGroups.some(
+      (ig) =>
+        ig.id === event.interestGroup?.id && ig.role === 'Project Manager',
```
gabiayako: I believe we need to check if the role is active

Image

I was taking a look in the backend it seems that to compute `active` we're just checking the IG status not the membership yet. So maybe this should be done in another ticket?

apps/crn-frontend/src/events/Event.tsx
```diff
@@ -32,6 +50,29 @@ const Event: React.FC = () => {
...
+        teamsTotal={teamsTotal}
+        teams={teams}
+        sinceLastEvent={
+          event.previousEventAttendance && {
+            count: teamsAttended - event.previousEventAttendance.teamsAttended,
+            teamsAttended: event.previousEventAttendance.teamsAttended,
+            teamsTotal: event.previousEventAttendance.teamsTotal,
+          }
+        }
+        onAddAttendance={isEventProjectManager ? noop : undefined}
```
gabiayako, line 69: hey not sure if this belongs in this ticket or the one that added the component, but the download and edit buttons don't rely on `onEdit` and `onExport` being passed, so they don't depend on whether the user is a PM

AimeurAmin: nice catch, and I agree, and it predates this PR
`EventAttendance.tsx` is inconsistent:
- Empty state (line 230): `{onAddAttendance ? ... : ...} ` => properly gated
- Populated state (lines 255-272): both buttons render unconditionally with `onClick={onExport}` / `onClick={onEdit}`, which `Event.tsx` never passes

incoming fix..

apps/crn-frontend/src/events/Event.tsx
AimeurAmin: Good catch. Added the `ig.active` check to the call to action, and went one step further on the backend since the data was the real gap: the user query now fetches the leader's own `inactiveSinceDate` (it was never selected; same field working groups already fetch), so the CTA also excludes PMs whose leadership has lapsed, not just inactive groups.

While testing this I found a related bug: when a PM is also on a team linked to their own interest group, the two membership entries get deduplicated and the team-derived one (which has no role) won! so `role: 'Project Manager'` was silently dropped and the PM check failed for exactly those users. Fixed the dedup to keep the role, with tests for the collision, the lapsed leader, and both combined.

One thing deliberately left out of scope: profile surfaces still show the role for lapsed leaders (current behavior, unchanged). Whether a stepped-down PM should stop being displayed as "Project Manager" on profile cards is a product call and we can bring that later as a follow-up ticket if so. Also note the usual caveat: the new field rides the Auth0 token, so existing sessions see it after next login.

AkosuaA (approved):
LGTM! 🎉

PR #5130, opened 2026-07-24
ASAP-1614 Add disableUsedInPublication to AvailableActions

gabiayako:
This PR adds `disableUsedInPublication` to AvailableActions following https://github.com/yldio/asap-hub/pull/5129 also with the goal of centralizing the conditions that are currently spread across child components.

Screenshot 2026-07-24 at 07 00 13

The old test [ResearchOutputForm.5.test.tsx‎](https://github.com/yldio/asap-hub/pull/5130/changes#diff-c43c4c6f3a87dacdf516dd1738019e7e8d64b7d8c6be6f305ebd92bb2aff1abb) was renamed to [FundingandPublicationDetails.test.tsx‎](https://github.com/yldio/asap-hub/pull/5130/changes#diff-436430b9788dd811b4b52034ce3366e24628103c95e270cbe257ff5c4da1b522) as an effort to remove the `ResearchOutputForm.{n}.test.tsx` tests pattern.

PR #5131, opened 2026-07-24
ASAP-1615 Add disableNonPublicSharingStatus to AvailableActions

gabiayako:
This PR adds `disableNonPublicSharingStatus` to AvailableActions following https://github.com/yldio/asap-hub/pull/5130 also with the goal of centralizing the conditions that are currently spread across child components.

Screenshot 2026-07-24 at 07 00 13

Also, some tests were moved to [‎FundingAndPublicationDetails.test.tsx‎](https://github.com/yldio/asap-hub/pull/5131/changes#diff-436430b9788dd811b4b52034ce3366e24628103c95e270cbe257ff5c4da1b522) to keep all the tests related to this section together.

### DIFF
--- a/packages/react-components/src/templates/ResearchOutputForm.tsx
+++ b/packages/react-components/src/templates/ResearchOutputForm.tsx
@@ -542,8 +542,6 @@ const ResearchOutputForm: React.FC<ResearchOutputFormProps> = ({
                   typeDescription="Select the type that matches your output the best."
                 />
                 <ResearchOutputPublishingCard
-                  documentType={documentType}
-                  researchOutputData={researchOutputData}
                   asapFunded={asapFunded}
                   onChangeAsapFunded={setAsapFunded}
                   usedInPublication={usedInPublication}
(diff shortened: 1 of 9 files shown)


packages/react-components/src/templates/ResearchOutputForm.tsx
AimeurAmin:
```suggestion
      behavior.isCreateFlow ? documentType : undefined,
```

AimeurAmin:
```suggestion
      behavior.isCreateFlow ? documentType : undefined,
```

AimeurAmin: all these can be removed with the small two changes bellow, note that `behavior` is already created on top ✌🏻

AimeurAmin (commented):
another really nice one 👏🏻 

One really small gap: the disabling moved to availableActions, but the logic that sets the value didn't. `isCreatingOutput` still matches only the `team` and `working-group` `createOutput`  routes, so on `/projects/:projectId/create-output/article` the two disagree, `isCreateFlow` is true (disables "No" / "Not Sure"), but the route match is false, so the value initializes to "Not Sure". The field ends up with a disabled option selected. 
I left suggestions on code, I believe they should fix that little gap 😄

packages/react-components/src/templates/ResearchOutputForm.tsx
gabiayako: Thanks a lot!!! 💯  It looks so much better now!

AimeurAmin (approved):

gabiayako:
rebased to master

PR #5134, opened 2026-07-27
[ASAP-1565] - Add Secondary Email address for user

AimeurAmin:
[ASAP-1565](https://asaphub.atlassian.net/browse/ASAP-1565)

[ASAP-1565]: https://asaphub.atlassian.net/browse/ASAP-1565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-components/src/templates/ContactInfoModal.tsx
```diff
@@ -126,6 +130,17 @@ const ContactInfoModal: React.FC<ContactInfoModalProps> = ({
             }
             hint="Note: This will not affect the way you login into the Hub."
           />
+          <LabeledTextField
+            type="email"
```
diegonvs, line 135: This is also happening with other email fields across the platform, so I’m okay with leaving it unchanged for now and creating a ticket to address it later.

However, if I enter an invalid email address, the validation only happens on the backend, and the error message is very generic: “There was an error, and we were unable to save your changes.”

Could we add client-side email validation and display a more specific error message?

https://github.com/user-attachments/assets/9dcc7b39-c732-448b-ab3a-eee1ba5ba8c0

diegonvs (approved):

PR #5137, opened 2026-07-27
[ASAP-1571] - import users and labs scripts

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1571

Summary

This pr adds three scripts that imports labs and team members into Contentful, plus shared helpers in `import-utils.ts`. They're split into three phases and must be run in order because of a circular dependency: a lab's PI has to be a User entry, but that user might not exist yet, and a user's labs field references a Lab that also has to exist first.

1. import-labs.ts — creates/publishes labs entries from the Labs CSV (without a PI link yet). Reconciles the sheet's "Lab On Hub?" column against what's actually in Contentful and logs any mismatch, but always ends with the lab created/published either way. Safe to re-run.
2. import-team-users.ts — creates users entries from the combined team-members CSV, linking each to their team (via a teamMembership entry) and lab. This is the most involved script; see "Matching & update logic" below.
3. link-lab-pis.ts — re-reads the same Labs CSV and sets each lab's labPI link, using the PI name/team columns. Must run last, since it needs the users from step 2 to already exist.

Matching & update logic (import-team-users.ts)

Each row is matched against Contentful first by ORCID, then by email if there's no ORCID match (covers users created before ORCID was collected). Depending on what's found:

- No match → a new user is created, provided their Team resolves to an existing team (otherwise the row is skipped and no user is created).
- Matched, and up to date → skipped.
- Matched, but missing something the sheet has (a team membership/role, a lab link, an ORCID value) → that's added. Nothing existing is ever removed or overwritten by this path.
- Matched by ORCID, but the CMS email differs → treated as authoritative on ORCID; the email difference is only flagged in the warnings for manual review, not written.
- Matched by email, but the CMS ORCID differs (or is missing) → the CMS ORCID is updated to match the sheet, and it's flagged in the warnings.

Position supports multiple comma-separated roles for the same team (e.g. "Data Manager, Project Manager") — one teamMembership is created per role that's missing.

Heads up on logging

The scripts log a lot, by design. Because we're writing into production data with several edge cases (ambiguous PI name matches, ORCID/email mismatches, teams or labs not found), every one of those gets an explicit log line and gets pushed into a warnings array that's dumped at the end of each run alongside a summary count (created/skipped/backfilled/etc.). This is intentional — we need to be able to go through the logs after each run to see exactly what happened, confirm how many records were affected by each edge case, and report that back to the client rather than have it happen silently.

Sheets used:
labs csv: https://docs.google.com/spreadsheets/d/1hvbp94jEmzvG_3k6MNkRKXIFSoV-OZhNL8VzTI-GjF0/edit?usp=sharing
team members csv: https://docs.google.com/spreadsheets/d/1mFVCUp2vIMj3U63R56FpseOR08OboDuhI4SBoi3uUVU/edit?usp=sharing


diegonvs:
Hey, some findings:


I tested import:team-users with this csv:

```
Team,Last Name,First Name,Position,Lab,Email,ORCID
Ulusoy,Reprobug4,Egv7dup,Data Manager,,egv7@duke.edu,0000-0009-9999-9991
```

We already have egv7@duke.edu in Contentful, but there the first E is uppercase (Egv7@duke.edu). When I ran it with the email all lowercase, it created a new user instead of updating the existing one. So, we end up with a duplicate. Looks like the email match is case-sensitive, and there are ~147 users with uppercase letters in their email, so any of those would get duplicated.

-----

Then I ran this csv (note I didn't set the ORCID):

```
Team,Last Name,First Name,Position,Lab,Email,ORCID
Ulusoy,Reprobug2,Zzztest,Data Manager,,zzztest.reprobug2@example.com,
```

Zzztest Reprobug2 doesn't exist yet, so I expected it to create the user. Instead it matched an unrelated existing user and added the membership to them:
```

EMAIL MISMATCH: existing user Zzztest Reprobug2 (3XrQ2TM7MwxVUSByta1b23) matched by ORCID (), but email differs (CMS: x@y.com, CSV: zzztest.reprobug2@example.com) - email not updated, please review.

Created and published team membership for existing user Zzztest Reprobug2: 3XO9yoVA10M7bN9dCn3YtP (Data Manager)
Linked 1 new membership(s) to existing user Zzztest Reprobug2 (3XrQ2TM7MwxVUSByta1b23).
```

At first I thought it matched because both had an empty ORCID, but that's not it, the matched user (x@y.com) actually has a real ORCID (0000-0001-0000-0001). When the ORCID is empty the lookup filter is basically ignored and returns every user, so it just grabs the first one. That means every row without an ORCID gets attached to the same random user, and the real person is never created. Same thing happens with an empty email.

AkosuaA:
> Hey, some findings:
> 
> I tested import:team-users with this csv:
> 
> ```
> Team,Last Name,First Name,Position,Lab,Email,ORCID
> Ulusoy,Reprobug4,Egv7dup,Data Manager,,egv7@duke.edu,0000-0009-9999-9991
> ```

I was wondering why I didn't see duplicates and realised for (at least some) instances where the email was in different cases the orcids were the same so they matched but you're right that there could be a case where the orcid doesn't match as well so I've updated.
> 
> Then I ran this csv (note I didn't set the ORCID):
> 
> ```
> Team,Last Name,First Name,Position,Lab,Email,ORCID
> Ulusoy,Reprobug2,Zzztest,Data Manager,,zzztest.reprobug2@example.com,
> ```
> 
> Zzztest Reprobug2 doesn't exist yet, so I expected it to create the user. Instead it matched an unrelated existing user and added the membership to them:
> 
> ```
The sheet only contains users with both email and orcid that's why. I've updated the script so if any row doesn't have either it is skipped. It shouldn't happen but it makes sense to have that explicitly written in the code as well.

Thanks!

diegonvs (approved):
I tested it, and everything looks good. Thanks, Debbie!

PR #5138, opened 2026-07-29
[ASAP-1627] - fix/increase team manuscript limits

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1627

- increases team manuscripts query limit to 100
- created fragments for reuse on similar queries
-  `/manuscripts/batch` now batches ids into chunks and fetches each chunk via `id_in` collection queries instead of N separate per-id requests.
- Expands `enrichLabPITeamIds` to handle multiple manuscripts at once


AimeurAmin (approved):
Nicely done 👌🏻🚀

PR #5139, opened 2026-07-29
ASAP-1597 Refactor Team Output

gabiayako:
The goal of this PR is to separate the manuscript import flow from the old TeamOutput, so it's easier to change each file without worrying too much about breaking things.

#### What was done

* `TeamOutput` is now just a re-export of `ProjectOutput`, so we don't need to duplicate new things in two files. It was confirmed in this thread https://yld.slack.com/archives/G01A6J33G8H/p1784551687161309 that they are supposed to be the same.
* `ManuscriptOutputSelectionScreen` was created to own the manuscript import flow
* The old `TeamOutput.test.tsx` and `ProjectOutput.test.tsx` were both deleted and replaced by 3 files, following the pattern we used for `ResearchOutputForm`. They live in `apps/crn-frontend/src/projects/__tests__/ProjectOutput/` 

#### Flows
Screenshot 2026-07-17 at 10 13 57 (2)


gabiayako:
rebased to master

AkosuaA (approved):
🤯 I can't imagine how much coffee went into this. Really nice job extracting the output selection and simplifying the *Output component. Looks good to me!

PR #5140, opened 2026-07-31
[ASAP-1635] - Improvement - manuscript collections and details decoupling

AimeurAmin:
[ASAP-1635](https://asaphub.atlassian.net/browse/ASAP-1635)

# Workspace manuscripts: show all manuscripts without caps

## Problem

Users could not see all manuscripts on a team workspace. Team Di Camilli's workspace only showed 20 manuscripts, so a manuscript with status "Addendum Required" was invisible to the team.

The first fix for this (#5138) raised the limit from 20 to 100 and preloaded the full detail of every manuscript through a `POST /manuscripts/batch` endpoint that fetches detail in chunked bulk GraphQL queries (30 manuscripts per chunk, 3 queries per chunk). That works well at current team sizes, but the design has two structural limits:

1. **100 is still a cap.** The team query still fetches `manuscriptsCollection(limit: 100)`, so a team with more than 100 manuscripts hits the same bug again.
2. **The cap cannot simply be raised within this design.** The workspace preloads the *full* detail (versions, discussions, labs) of every manuscript up front, whether or not the user ever expands it. Everything about that grows linearly with team size: the batch round-trips, the response payload, and the pressure on Contentful's query complexity limit (11k per query, which is why the bulk queries have to stay chunked at 30 in the first place). Lifting the cap would mean committing to unbounded up-front detail fetching inside a 16-second lambda budget.

So the cap and the preloading are coupled: to remove one for good, the other has to go too.

## The fix: light list + detail on demand

The workspace never needs the full detail of every manuscript up front. A collapsed manuscript card only shows the title, the status, and the manuscript ID pill.

So this PR splits the data in two:

- **One light list query.** A new `GET /manuscripts?teamId=...` (or `?projectId=...`) endpoint returns just `id`, `title`, `status`, and `versionUID` for every manuscript of a team or project. The server pages through Contentful 100 at a time until it has everything, so there is **no cap at any team size**. The query is tiny, so there is no complexity risk and no growth in payload as teams get bigger. For Alessi's 91 manuscripts this is a single request that returns in ~600ms (verified against the Development space).
- **Full detail only when a card is expanded.** When the user expands a manuscript card, a new `ManuscriptCardDetail` component loads (inside a `Suspense` boundary) and fetches that one manuscript's full detail. Deep-linked manuscripts start expanded, so they load their detail immediately.

The server keeps the existing behaviour for the list: team submissions are separated from collaborator submissions (first linked team + no project = team-owned), and Compliant / Closed (other) manuscripts sort to the bottom.

## What was removed

The entire batch path is gone, since nothing uses it anymore:

- `POST /manuscripts/batch` route, validation, and controller method
- `fetchByIds` / `fetchManuscriptsBatch` in the manuscript data provider, and the three bulk GraphQL queries behind them
- The `manuscriptsCollection` blocks inside the team and project queries, and the `manuscripts` / `collaborationManuscripts` fields on the team and project API responses (their only consumers were the two workspace pages, which now use the new list endpoint)
- The frontend batch hook (`useBatchManuscriptsByIds`) and API call

## Frontend cache behaviour

- Manuscript status/title changes patch the cached workspace list in place instead of refetching (Contentful has read-after-write lag, so an immediate refetch could return stale data).
- Creating a manuscript or compliance report invalidates the workspace list so the new manuscript appears.

[ASAP-1635]: https://asaphub.atlassian.net/browse/ASAP-1635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AkosuaA:
@AimeurAmin I'm curious about the second point and the timeout numbers. How did we measure the ~20–25s for Alessi’s 91 manuscripts? I tested the existing batch flow on Alessi and it seems to complete much quicker without timing out.

I agree with the overall approach of moving to a lightweight list + on-demand detail loading btw; just wondering if there’s a specific case that led to the timeout conclusion.

AimeurAmin:
> @AimeurAmin I'm curious about the second point and the timeout numbers. How did we measure the ~20–25s for Alessi’s 91 manuscripts? I tested the existing batch flow on Alessi and it seems to complete much quicker without timing out.
> 
> I agree with the overall approach of moving to a lightweight list + on-demand detail loading btw; just wondering if there’s a specific case that led to the timeout conclusion.

@AkosuaA 
totally correct calling out the wrong numbers up there, my bad, the calculation was on the pre-batch solution you introduced (`Promise.all(ids.map(fetchById))`  commit [6cc15a23e](https://github.com/yldio/asap-hub/commit/6cc15a23e4b956ee972426221fb12a4c9cf48d07)) => here are the numbers:
`fetchById` costs 3–4 Contentful GraphQL requests, all through the 15 req/s per-lambda limiter => For 91 manuscripts that's around 320 requests => resulting to 20 to 25 seconds minimum
my bad for running the numbers on that old commit, as after the commit you introduced [c0d5083](https://github.com/yldio/asap-hub/commit/c0d508367e59a70cc4a3f220f0369ddca07f49ae) which reworked the server fetching to bulk queries (chunks - `fetchByIds`) that version genuinely completes in a few seconds.

correcting the description up top 😉

AkosuaA (commented):
Looks good! 🎉
Just a few comments/questions

- Creating a manuscript throws an error.
`useManuscriptById(manuscriptId ?? '')` is called even when manuscriptId is missing, causing an invalid GET /manuscripts/ request that ends up hitting the added workspace manuscripts route and returning a 400. Maybe skip the fetch when there's no manuscriptId, similar to useWorkspaceManuscripts.

- Cache update is overwritten in one instance.
`usePutManuscript` patches the workspace list cache to avoid a full refetch because of Contentful's read-after-write lag. That holds up fine on the ASAP Compliance page and the status update button on the ManuscriptCard, since nothing else touches the cache afterwards there. On the ManuscriptForm's edit flow though, `usePutManuscript` is used but `TeamManuscript`/`ProjectManuscript`'s `onSuccess` calls `invalidateWorkspaceManuscripts` right after, which overwrites that same patch. Create and resubmit don't patch the workspace list at all, so they're relying on the invalidate + refetch. If there's an actual lag then changes wouldn't reflect, right?

- Since we've separated manuscripts from teams/projects do we need to invalidate teams/projects here?
Screenshot 2026-08-06 at 11 49 47

AkosuaA (approved):
Nice! 🎉🪄

PR #5142, opened 2026-08-03
[ASAP-1592] - Add Edit Speakers modal

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1592

Stacked on top of #5115 (Event Speakers card)

How to test (Storybook)

- **Organisms / Events / Speakers → Edit and Save**: end-to-end: open the pencil, add/remove speakers or toggle findings, Save, and watch the card refresh (no adapter between card and modal).
- **Organisms / Events / Edit Speakers Modal**: `Add Speakers` and `Edit Speakers`.

### DIFF
--- a/packages/react-components/src/organisms/EditEventSpeakersModal.tsx
+++ b/packages/react-components/src/organisms/EditEventSpeakersModal.tsx
@@ -59,7 +59,6 @@ type EditEventSpeakersModalProps = {
   ) => Promise<SpeakerSearchOption[]>;
   readonly onSave: (groups: SpeakerGroup[]) => void | Promise<void>;
   readonly onDismiss: () => void;
-  readonly disableAdding?: boolean;
 };
 
 const modalStyles = css({ width: '100%' });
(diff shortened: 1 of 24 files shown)


packages/react-components/src/organisms/EditEventSpeakersModal.tsx
```diff
@@ -0,0 +1,714 @@
...
+                    void handleSave();
+                  }}
+                >
+                  Save
+                </Button>
+              </div>
+            </>
+          )}
+        </div>
+      </footer>
```
AimeurAmin: I believe all this with the associated styles above are a duplicate of the component `ConfirmableModalFooter`

diegonvs: Yes! I missed that one while rebasing. Thanks for catching this!

diegonvs: Pushed 9902902 to reuse it. While doing that I noticed the padding bottom should be 32, not 24. UploadListModal was already overriding it locally to get there. So I changed the default in the component (all three usages now get 32) and dropped both the local override and the now-unused overrideStyles prop.

AimeurAmin (approved):
looks good 👌🏻

PR #5145, opened 2026-08-05
[ASAP-1625] Update Project Membership Roles

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1625

Renames two Project Membership roles and retires a third:

| Before | After |
| --- | --- |
| `Trainee Project - Lead` | `Independent Project - Lead` |
| `Trainee Project - Mentor` | `Independent Project - Mentor` |
| `Trainee Project - Key Personnel` | folded into `Independent Project - Mentor` |

The project type stays `Trainee Project`. Only membership roles change.

Project type stays `Trainee Project`. One person in production holds Key Personnel. Mapping him to Mentor is per Padmini's review. (https://asapparkinson-roadmap.slack.com/archives/C011NQ4U10B/p1786112333763369)

Members with no role were dropped from the members list entirely, pre-existing, not introduced here. They're now listed with an italic "No role assigned". Members carrying some other role stay hidden, as before.

`isProjectMember` now filters by role on Trainee Projects. Without it, listing roleless members would have granted them the workspace and manuscript submission. Only permissions change in the PR.

Contentful:

https://github.com/user-attachments/assets/f24114bd-1e89-4a0c-97b1-7d1d3d5ea575

Before:

Screenshot 2026-08-05 at 18 47 40

After(note a user without a role):

Screenshot 2026-08-11 at 11 51 53

**Key Personnel**

Before(Padmini was Key Personnel too):

Screenshot 2026-08-05 at 18 49 14

After:

Screenshot 2026-08-11 at 11 56 06


packages/model/src/project.ts
```diff
@@ -269,7 +269,28 @@ export const projectLeadMemberRoles = [
...
+
+export const traineeProjectMentorRoles = [
+  'Individual Project - Mentor',
+] as const;
+
+/**
+ * Splits Trainee Project members into the two rows the UI renders: trainees
+ * (leads) and mentors. Members holding any other role — including roles
+ * retired from the content model — are excluded from both lists.
+ */
```
gabiayako: Hey, I was wondering if this is right. The role is not required and if we're filtering members that don't have role, I can be listed on the CMS but not appear on the hub. It seems a little bit weird

Image

diegonvs: This is pre-existing behavior and wasn't introduced by this change. The migration clears the role on Key Personnel entries, so those members end up in the same state. I'll check with Padmini to confirm the expected behavior. Thanks a lot for spotting this, Gabi!

gabiayako: But this PR removes the Key Personnel role from existing entries. There’s currently one person in production with the Key Personnel role, and after this change, he won’t appear on the project anymore 😅

packages/react-components/src/templates/ProjectDetailHeader.tsx
gabiayako, line 403: I think it's weird that it appears on the members list but not at the top, maybe `unassigned` should  be grouped with "trainees"? But this is a another product related issue 😅 

Image
Image

gabiayako (approved):

packages/react-components/src/templates/ProjectDetailHeader.tsx
diegonvs: After today's daily, we decided to not update that section

diegonvs:
Waiting confirmation from https://asapparkinson-roadmap.slack.com/archives/C011NQ4U10B/p1786112333763369 before merging

PR #5148, opened 2026-08-07
[ASAP-1622] - Create Breadcrumbs components

AimeurAmin:
[ASAP-1622](https://asaphub.atlassian.net/browse/ASAP-1622)

[ASAP-1622]: https://asaphub.atlassian.net/browse/ASAP-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storybook/src/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+    homeHref: '/',
+    items: [
+      { label: 'Discovery Projects', href: '/projects' },
+      { label: 'Alpha-Synuclein Origins' },
+      { label: 'Submit Revised Manuscript' },
+    ],
+  },
+};
+
+export const DeepPath: Story = {
```
diegonvs, line 60: I was taking a look at this story and had a question: wouldn’t it make sense to show an ellipsis with a dropdown after a certain number of levels?

I’m raising this because titles like “Level (n)” look fine, but if we increase the amount of text, it could start to look a bit weird.

packages/react-components/src/icons/home.tsx
```diff
@@ -0,0 +1,19 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+    width={24}
+    height={24}
+    viewBox="0 0 24 24"
+    fill="none"
+  >
+    <title>Home</title>
+    <path
+      d="M5 19.0002V10.3082C5 10.0522 5.05733 9.8099 5.172 9.58123C5.28667 9.35257 5.44467 9.16423 5.646 9.01623L11.031 4.93823C11.313 4.7229 11.635 4.61523 11.997 4.61523C12.359 4.61523 12.683 4.7229 12.969 4.93823L18.354 9.01523C18.556 9.16323 18.714 9.3519 18.828 9.58123C18.9427 9.8099 19 10.0522 19 10.3082V19.0002C19 19.2682 18.9003 19.5019 18.701 19.7012C18.5017 19.9006 18.268 20.0002 18 20.0002H14.616C14.3867 20.0002 14.1947 19.9229 14.04 19.7682C13.8853 19.6129 13.808 19.4209 13.808 19.1922V14.4232C13.808 14.1946 13.7307 14.0029 13.576 13.8482C13.4207 13.6929 13.2287 13.6152 13 13.6152H11C10.7713 13.6152 10.5797 13.6929 10.425 13.8482C10.2697 14.0029 10.192 14.1946 10.192 14.4232V19.1932C10.192 19.4219 10.1147 19.6136 9.96 19.7682C9.80533 19.9229 9.61367 20.0002 9.385 20.0002H6C5.732 20.0002 5.49833 19.9006 5.299 19.7012C5.09967 19.5019 5 19.2682 5 19.0002Z"
+      fill="#34A270"
```
diegonvs, line 14: Can we use fill="currentColor"?

apps/storybook/src/Breadcrumbs.stories.tsx
AimeurAmin: we decided internally to not go with the ellipsis solution (at least for now and that's why there are no designs for it)

packages/react-components/src/icons/home.tsx
AimeurAmin: may you elaborate please? why would we want to do that? what we usually do is copy the svg directly from figma and we only change it if we need the icon to be used in different parts of the project, I don't see how changing the hex code of the svg to "currentColor" in this case can benefit us

diegonvs: This lives in the shared package and GP2's primary500 is blue. If the breadcrumb ever lands there, the icon stays CRN green next to blue links. We already use currentColor in some icons, but since the scope of this ticket is CRN-only, we can leave it. In the future, when we use it on GP2, we may need currentColor.

diegonvs (approved):

PR #5154, opened 2026-08-11
[ASAP-1599] - Display Speakers information for Upcoming and Past Events

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1599

Show the read-only Speakers card on the CRN event detail page for upcoming and past events. Requires NEW_EVENT_PAGE feature flag.

**Empty States**

**NON PM**
Screenshot 2026-08-11 at 20 53 15

**PM**
Screenshot 2026-08-11 at 20 58 58

**Past Event**

https://github.com/user-attachments/assets/e72eb02e-d0ef-4cce-9756-a9b5b49a2914

https://github.com/user-attachments/assets/286844f4-459a-4d46-8eab-87b0a0be807d

**Upcoming Event**

https://github.com/user-attachments/assets/b1919954-38f3-475b-b304-d7b9357b366d

### DIFF
--- a/apps/crn-frontend/src/events/state.ts
+++ b/apps/crn-frontend/src/events/state.ts
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { User } from '@asap-hub/auth';
 import {
   createQueryKeys,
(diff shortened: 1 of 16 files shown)


apps/crn-frontend/src/events/state.ts
```diff
@@ -25,6 +27,21 @@ export const useEventById = (id: string): EventResponse | undefined => {
   return data ?? undefined;
 };
 
+// Stable module-level reference so react-query memoizes the transform and
+// only re-runs it when the event data changes, not on every render.
+const selectSpeakerGroups = (data: EventResponse | null): SpeakerGroup[] =>
+  data ? mapSpeakersToGroups(data) : [];
+
+export const useEventSpeakerGroups = (id: string): SpeakerGroup[] => {
```
AkosuaA, line 34: we could do this instead
```
export const useEventSpeakerGroups = (id: string): SpeakerGroup[] => {
  const event = useEventById(id);
  return useMemo(() => (event ? mapSpeakersToGroups(event) : []), [event]);
};
```

suggesting this because it reuses the existing query instead of duplicating the queryKey/queryFn. Right now any change to how we fetch an event has to be made in two places.

AkosuaA (approved):
Looks good! 🎉

PR #5157, opened 2026-08-14
[ASAP-1656] - Group fields on Users Content model - part 1

AimeurAmin:
[ASAP-1656](https://asaphub.atlassian.net/browse/ASAP-1656)

- **First Part:** we create a new content-type called Socials which will hold our socials fields + a field holding the user that owns them - then we create a sript that creates a new content of type Socials for every user we have (it's a script that moves our data)
  
- **Second Part:** we remove the social fields from User content-type (at this point we can rely on the data we moved to Socials so we can remove those fields) - then we create another migration file that creates a new field called UserSocials - then we create a script that fills that field based on the Socials->User link we have

- **Third Part:** after confirming everything is working fine we remove the user link field from the content-type Socials

This PR covers the First Part, the action was run through act library simulating exactly what would happen if run through the github actions:

image

[ASAP-1656]: https://asaphub.atlassian.net/browse/ASAP-1656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/on-demand-migrate-user-socials.yml
+++ b/.github/workflows/on-demand-migrate-user-socials.yml
@@ -1,0 +1,54 @@
+name: Migrate user socials
+on:
+  workflow_dispatch:
+    inputs:
+      environment-name:
+        required: true
+        description: The environment name
+        type: choice
+        options:
+          - Branch
+          - Development
+          - Production
+      contentful-env:
+        required: true
+        type: string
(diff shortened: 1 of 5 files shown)


AkosuaA (approved):
Approved! 🎉

PR #5161, opened 2026-08-17
fix: unable to publish crn-report

gabiayako:
We were getting error 400 when trying to save/publish a CRN Report Output since d3c6191d1fc16f2f9cdce30bdde1b3a83312d84c because the `type` field in the payload we were sending to the backend was not `undefined` anymore, it was an empty string that don't belong to the allowed enums. So it was giving a validation error.


AkosuaA (approved):
👍🏾

PR #5174, opened 2026-08-21
[ASAP-1580] - manage event attendance

AkosuaA:
ref: https://asaphub.atlassian.net/browse/ASAP-1580

Updates to the event attendance component will be done in [this ticket](https://asaphub.atlassian.net/browse/ASAP-1675)


diegonvs:
Hey, If I block the PATCH request to save the attendance nothing happens. I believe we can ask Elena about how to improve the resiliency here. Maybe retrying the request and adding a toast on the modal can be good ones.

https://github.com/user-attachments/assets/0b872a81-c752-42ce-8b19-36cc0db18cc9

If I open two editors, remove a team in one, and then change the attendance of that already-removed team in the other, it results in a 500 error.

https://github.com/user-attachments/assets/aee82e74-3316-40bd-a5bc-54f241307393

AkosuaA:
@diegonvs I ignored the error state since we didn't have designs but it makes sense to have a toast here (similar to what we have on some of the other modals). I'll replicate here with the same message and design can update the copy if desired.
On the second comment, I'll update to skip attendance entries that no longer exist instead of erroring. What do you think?

diegonvs:
>  I'll update to skip attendance entries that no longer exist instead of erroring. What do you think?

Looks okay! 

About the first comment, we can create a ticket for it if you want

AkosuaA:
> About the first comment, we can create a ticket for it if you want

@diegonvs Thanks for the suggestions! I've updated the modal to display an error state as well as the other change discussed. Ready for another look. Thanks!
Screenshot 2026-08-24 at 13 59 35

diegonvs (approved):

PR #5182, opened 2026-08-26
ASAP-1654 Import GP2 Alumni

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-1654

This PR adds a script to import GP2 alumni data on the hub and also moves part of the import utils to `packages/server-common` to be used both by GP2 and CRN.

It also removes the feature flag `ASAP_STAGING_MODE` ONLY related to alumni part. It is still being used to display user award badges.


AkosuaA (approved):
Nice! 👌🏾👌🏾

PR #5185, opened 2026-08-27
[ASAP-1603] - Upload a List of Teams to Event Attendance

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1603

Lets Tech Support bulk add/update event attendance from CSV/XLSX files. Frontend-only, parsing and matching reuse 
the existing Algolia team index; no backend, model, or migration changes.

**In scope**

- Upload button + form; multiple files; 10 MB limit with error; cancel confirmation
- Parse Team + Attendance (Yes/No) columns
- Match names against the Hub; matched added with their status; unknown names shown as "not matched"
- Teams already on the table (including unsaved additions) updated in place, never duplicated
- Delete teams from the list

**Not in scope**

Saving/downloading the uploaded files (store the file, show + download source lists) → [ASAP-1605 Enable PMs to download Source files](https://asaphub.atlassian.net/browse/ASAP-1605)
Fuzzy “did you mean” suggestions for near-misses → [ASAP-1604 Enable PMs to manage Event Attendees ](https://asaphub.atlassian.net/browse/ASAP-1604)

**Happy path**

https://github.com/user-attachments/assets/4e644d47-25eb-46ca-a0b1-3fd37550b5d8

**Already added some teams**

https://github.com/user-attachments/assets/30c42fcc-db7a-4f19-878b-2f2426660f3b

**Too large**

https://github.com/user-attachments/assets/146ab41f-a2f5-4379-8fb8-73204f9dcf24

### DIFF
--- a/apps/crn-frontend/src/events/__tests__/Event.test.tsx
+++ b/apps/crn-frontend/src/events/__tests__/Event.test.tsx
@@ -8,6 +8,7 @@ import {
   createCalendarResponse,
   createEventResponse,
   createInterestGroupResponse,
+  createTeamListItemResponse,
 } from '@asap-hub/fixtures';
 import { events } from '@asap-hub/routing';
 import { disable, enable } from '@asap-hub/flags';
@@ -19,7 +20,7 @@ import {
   WhenReady,
 } from '@asap-hub/crn-frontend/src/auth/test-utils';
 import Event from '../Event';
-import { getEvent, patchEvent } from '../api';
+import { getEvent, getTeamsForMatching, patchEvent } from '../api';
 import { downloadEventSpeakers } from '../export';
 
 jest.mock('../api');
@@ -39,10 +40,14 @@ const id = '42';
 
 const mockGetEvent = getEvent as jest.MockedFunction<typeof getEvent>;
 const mockPatchEvent = patchEvent as jest.MockedFunction<typeof patchEvent>;
+const mockGetTeamsForMatching = getTeamsForMatching as jest.MockedFunction<
+  typeof getTeamsForMatching
+>;
 beforeEach(() => {
   disable('NEW_EVENT_PAGE');
   mockGetEvent.mockClear();
   mockPatchEvent.mockClear();
+  mockGetTeamsForMatching.mockClear();
   mockGetEvent.mockResolvedValue({
     ...createEventResponse(),
     id,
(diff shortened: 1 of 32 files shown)


apps/crn-frontend/src/events/parse-team-list.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const rows = XLSX.utils.sheet_to_json<unknown[]>(sheet, { header: 1 });
+  const { teamColumn, attendanceColumn, skipFirstRow } = findColumns(rows[0]);
+
+  return rows
+    .slice(skipFirstRow ? 1 : 0)
+    .map((row) => ({
+      name: toCellText(row[teamColumn]),
+      attended:
+        attendanceColumn === -1 ? false : parseAttended(row[attendanceColumn]),
```
gabiayako, line 98: Not sure if this can happen with real users, but the first format I've tried without reading the code was this one:
Image

So I didn't add headers, but I added Y or N as attendance and all became false because of this `attendanceColumn === -1  ? false`

apps/crn-frontend/src/events/__tests__/parse-team-list.test.ts
```diff
@@ -0,0 +1,255 @@
...
+  it('Should accept Team and Teams as header aliases', () => {
+    expect(names(parseSheet('Team\nAguzzi\n'))).toEqual(['Aguzzi']);
+    expect(names(parseSheet('  TEAMS  \nAguzzi\n'))).toEqual(['Aguzzi']);
+  });
+
+  it('Should use the headed column when it is not the first one', () => {
+    expect(
+      names(parseSheet('Email,Team Name\na@b.com,Aguzzi\nc@d.com,Alessi\n')),
+    ).toEqual(['Aguzzi', 'Alessi']);
+  });
```
gabiayako, line 60: nice!

apps/crn-frontend/src/events/__tests__/Event.test.tsx
```diff
@@ -386,6 +391,143 @@ describe('the NEW_EVENT_PAGE flag', () => {
...
+          },
+        ]);
+
+        const { container, findByText, findByRole, getByRole } =
+          await openUploadList();
+        await uploadCsv(container, 'Team Name,Attendance\nTeam One,Yes\n');
+
+        expect(
+          await findByText(
+            'All 1 Teams in this list are already in the attendance table.',
```
gabiayako: Hey, I think it would be nice to fix the singular/plural here.

I also noticed that we need to make the same adjustment in this part:
Image

packages/react-components/src/organisms/__tests__/UploadListModal.test.tsx
```diff
@@ -272,6 +286,91 @@ describe('UploadListModal', () => {
...
+        matched: [],
+        unmatched: [],
+      })),
+    });
+
+    await upload(makeFile('teams.csv'), container);
+
+    expect(
+      await screen.findByText(
+        'No Team names found. Check that the file has a Team column.',
```
gabiayako: I couldn't find this in Figma, but is "Team" supposed to be uppercase here?

```
No Team names found. Check that the file has a Team column
```

And here:

```
All 3 Teams in this list are already in the attendance table
```

It looks a little bit weird to me, so I just wanted to double-check.

packages/react-components/src/organisms/UploadListModal.tsx
```diff
@@ -583,148 +620,158 @@ const UploadListModal: React.FC<UploadListModalProps> = ({
...
+                              <Link
+                                href={
+                                  network({})
+                                    .teams({})
+                                    .team({ teamId: team.teamId }).$
+                                }
+                              >
+                                <span css={matchedTeamNameStyles}>
+                                  {team.teamName}
+                                </span>
```
gabiayako, line 714: I didn't notice this before, but we're not showing if the team is inactive here. Also, when we click the team name link, it navigates in the same window, so we lose our progress on the upload. Not sure if this comments are supposed to be addressed in another ticket though.

packages/react-components/src/organisms/UploadListModal.tsx
```diff
@@ -465,11 +494,18 @@ const UploadListModal: React.FC<UploadListModalProps> = ({
...
   const totalTeams = result
-    ? result.matched.length + result.alreadyInCount + result.unmatched.length
+    ? result.matched.length + result.unmatched.length
     : 0;
 
+  const hasSections = matchedTeams.length > 0 || remainingUnmatched.length > 0;
+  const emptyResultMessage =
+    totalTeams === 0
+      ? 'No Team names found. Check that the file has a Team column.'
+      : `All ${totalTeams} Teams in this list are already in the attendance table.`;
```
gabiayako: I believe there's a case where this message shouldn't appear as it does: if I delete all the matched teams that aren't already in the attendance table
https://github.com/user-attachments/assets/857a181c-e429-48db-9764-e759434ce4c1

apps/crn-frontend/src/events/parse-team-list.ts
diegonvs: thanks for trying a format I hadn't thought of. Fixed in c593eb354

apps/crn-frontend/src/events/__tests__/Event.test.tsx
diegonvs: Fixed. I pulled the pluralisation into a shared pluralizeTeams helper 271471854 and applied it in five places:

the counts line and the matched section header in the upload modal, the attendance card ("0 of 1 team"), the team search ("adds 1 team"), and the speakers card under Preliminary Findings, which also had it.

packages/react-components/src/organisms/__tests__/UploadListModal.test.tsx
diegonvs: I confirmed with Elena, the strings now use lowercase "teams" mid-sentence, matching the intro copy at the top of the modal, and keep "Team Name" capitalised only where it names the column in someone's file.

That state also changed shape on her feedback: a file we can't read any names from is now an error in the red line instead of a result with "0 Teams • 0 matched • 0 already in".

packages/react-components/src/organisms/UploadListModal.tsx
diegonvs: Both worth doing here, so I did. The badge is the same one the attendance table already uses, and the data was already reaching the row.

The link was more involved than it looked: Anchor decides on its own that internal links stay in the same tab and doesn't accept target at all. I added an opt-in openInNewTab prop to Link/Anchor 85773834f and used it only here. Default behaviour is unchanged. Elena confirmed a new tab is what she wants.

packages/react-components/src/organisms/UploadListModal.tsx
diegonvs: Reproduced and fixed. The message was describing the current view instead of the uploaded file, so deleting newly matched teams made it claim they'd been there all along. It now only appears when the file itself brought nothing new, and stays away once you delete the rows yourself.

gabiayako (approved):
nice!!! 👏

PR #5190, opened 2026-09-02
[ASAP-1692] - Fix false Manuscript Re-Submitted reminder on compliance report submission

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1692

**Problem:** After sharing a compliance report, a false **"… resubmitted a manuscript … 'Manuscript Re-Submitted'"** reminder shows on the dashboard, even though no resubmission happened. It's dated to the compliance-report day, not the actual resubmission.

### How to reproduce (before the fix)

1. Take a manuscript resubmitted **> 7 days ago** (≥ 2 versions).
2. Share a compliance report on it today (republishes the manuscript).
3. Open the dashboard as a PM / author (not the person who resubmitted) → the false **"Manuscript Re-Submitted"** reminder appears, dated today.

### Cause
Sharing a report calls `manuscriptController.update({ status })`, which `patchAndPublish`es the manuscript and bumps `sys.publishedAt`. The reminder fired on `versions > 1 && inLast7Days(sys.publishedAt)` — with no check that a resubmission actually happened.

### Fix
Key the reminder off the **last version's `sys.firstPublishedAt`** (the real resubmission time), which doesn't move when the manuscript is later republished. Adds a regression test.

**Before:**

I just changed the compliance status:

Screenshot 2026-09-03 at 08 51 14

**After:**

Note that the Resubmit notification is no longer displayed:

Screenshot 2026-09-03 at 08 39 50

### DIFF
--- a/packages/contentful/src/crn/queries/reminders.queries.ts
+++ b/packages/contentful/src/crn/queries/reminders.queries.ts
@@ -65,7 +65,7 @@ export const FETCH_REMINDERS = gql`
             }
           }
         }
-        versionsCollection(limit: 10) {
+        versionsCollection(limit: 10, order: count_ASC) {
           total
           items {
             count
--- a/packages/contentful/src/crn/queries/reminders.queries.ts
+++ b/packages/contentful/src/crn/queries/reminders.queries.ts
@@ -69,6 +69,9 @@ export const FETCH_REMINDERS = gql`
           total
           items {
             count
+            sys {
+              firstPublishedAt
+            }
             createdBy {
               sys {
                 id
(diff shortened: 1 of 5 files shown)


packages/contentful/src/crn/queries/reminders.queries.ts
gabiayako: just to make sure we're getting the last version correctly, I think it would be nice to add:


```suggestion
        versionsCollection(limit: 10,  order: sys_firstPublishedAt_DESC) {
```

gabiayako (approved):
nice!!! 👏👏👏

packages/contentful/src/crn/queries/reminders.queries.ts
diegonvs: Shouldn't be `sys_firstPublishedAt_ASC`? 

this code reads items[0] as the first version and items[length-1] as the last (used by both the resubmission trigger and the Manuscript Created reminder), so DESC would invert it

gabiayako: yes, that's right! I saw another query as reference and copied the order param 😅 but you're right

diegonvs: Pushed 0e4c6698ff741da95f288f226ff92d6aec654c25

gabiayako (approved):
nice! 👏

PR #5193, opened 2026-09-04
[ASAP-1693] - Show the manuscript project name in discussion reminders

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1693

If a user without a team comments on a discussion, you will see it

**Before:**

Screenshot 2026-09-04 at 16 08 41

**After:**

Screenshot 2026-09-04 at 16 35 46

### DIFF
--- a/apps/crn-server/src/controllers/reminder.controller.ts
+++ b/apps/crn-server/src/controllers/reminder.controller.ts
@@ -24,6 +24,9 @@ export const formattedMaterialByEventType = (
   }
 };
 
+const discussionTeamPrefix = (teams: string): string =>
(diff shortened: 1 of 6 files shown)


AkosuaA (approved):
Approving since this fixes the issue mentioned in the ticket. It's a bit confusing to verify that everything is correct  because I'm not sure now what the messages in the reminder should be. Padmini will confirm with Elena when she's back and if there are any changes we can revisit. eg: some of the reminders in the sheet say to list all the team names (should be project names I guess) mentioned in the Manuscript but it only shows the project the manuscript was created from.

PR #5214, opened 2026-09-09
[ASAP-1610] - Enable staff to manage event speaker details

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1610

Let Tech Support delete speakers and manage per-team "Preliminary Findings Shared" on upcoming and past events, mirroring the event attendance edit flow.

Preliminary findings is a per-team concept, so external speakers neither show nor persist a shared status. It will change in another ticket, where the preliminary findings will be a per-user concept but this ticket can add some code that the following tickets will require less code to be reviewed.


apps/crn-server/src/data-providers/contentful/event.data-provider.ts
```diff
@@ -416,18 +463,131 @@ export class EventContentfulDataProvider implements EventDataProvider {
...
+              team: createLink(teamId),
+              preliminaryDataShared: shared,
+            });
+            const updatedEntry = await existingEntry.update();
+            await updatedEntry.publish();
+          }
+          return null;
+        }
+
+        if (!shared) {
```
gabiayako, line 566: Hey, I know adding a speaker is going to be implemented in another ticket but I think this might cause confusion in the future. If we're adding a new team/speaker and shared is false, it would fall here, right? so it wouldn't create a new entry with preliminary data false

apps/crn-server/test/data-providers/contentful/event.data-provider.test.ts
```diff
@@ -1521,6 +1523,389 @@ describe('Events Contentful Data Provider', () => {
...
+
+      await eventDataProvider.updateEventDetails('123', {
+        preliminaryDataShared: [{ teamId: 'team-3', shared: false }],
+      });
+
+      expect(environmentMock.createEntry).not.toHaveBeenCalled();
+      expect(patchAndPublish).toHaveBeenCalledWith(eventEntry, {
+        preliminaryDataShared: [],
+      });
+    });
```
gabiayako: yeah, this test confirms my comment above about how this could be confusing for someone implementing the addition of team/speaker preliminary data in the future, because preliminaryData: true works, but false doesn’t

apps/crn-frontend/src/events/Event.tsx
```diff
@@ -119,6 +116,36 @@ const Event: React.FC = () => {
...
+          onExport={
+            isTechSupport
+              ? () => downloadEventSpeakers(event, speakerGroups)
+              : undefined
+          }
+          onAddSpeaker={isTechSupport ? openSpeakersEditor : undefined}
+          onEdit={isTechSupport ? openSpeakersEditor : undefined}
+        />
+        {isEditingSpeakers && (
+          <EditEventSpeakersModal
```
gabiayako, line 133: There's the problem with singular / plural here as well 😛 

Image

diegonvs: Awkward Season 4 GIF by The Office

fixed!

apps/crn-server/test/data-providers/contentful/event.data-provider.test.ts
```diff
@@ -1521,6 +1523,389 @@ describe('Events Contentful Data Provider', () => {
...
+        { id: '123' },
+      );
+      when(environmentMock.getEntry)
+        .calledWith('123')
+        .mockResolvedValue(eventEntry);
+      mockPollingConsistency();
+
+      await eventDataProvider.updateEventDetails('123', {
+        preliminaryDataShared: [{ teamId: 'team-3', shared: false }],
+      });
```
gabiayako: But I think we still need to create an entry, right? I know it’s not part of your ticket, but I think the current behavior could be misleading for the next person who works on this 😅

For example, imagine we add a **new** team through the frontend, "Team A", and set `preliminaryFindings` to `false`. The backend doesn’t create an entry for it. So, in Contentful, this event won’t have "Team A" among the preliminary findings entries.

Then, when we fetch the data, it’s not that we have "Team A" with preliminary findings as absent, we don’t have "Team A" at all

diegonvs: In the most recent changes on e5b0c2138 I'm already covering it

gabiayako: nice, thanks 🎉

gabiayako (approved):

PR #5218, opened 2026-09-14
[ASAP-1708] - Fix - GP2 working group members list capped at 50

AimeurAmin:
[ASAP-1708](https://asaphub.atlassian.net/browse/ASAP-1708)


apps/gp2-server/src/data-providers/working-group.data-provider.ts
```diff
@@ -42,14 +44,18 @@ export class WorkingGroupContentfulDataProvider
...
-      total: workingGroupsCollection.total,
-      items: workingGroupsCollection.items
+    const workingGroups = await Promise.all(
+      workingGroupsCollection.items
         .filter(
           (workingGroup): workingGroup is GraphQLWorkingGroup =>
             workingGroup !== null,
         )
-        .map(parseWorkingGroupToDataObject),
+        .map((workingGroup) => this.withAllMembers(workingGroup)),
```
gabiayako, line 53: hey, not sure if `withAllMembers` is really needed here in the `fetch` function, maybe only on `fetchById`?

gabiayako (approved):

apps/gp2-server/src/data-providers/working-group.data-provider.ts
AimeurAmin: I have it there too because the working group list card prints `members.length` as "N Members". With a truncated list, the Underrepresented Populations group shows 50 instead of 51. Also according to Madeline's report, the GP2 website is also showing just the first 50 members, and that data is coming from the public API list endpoint which returns members straight through.
I believe the cost of keeping this is negligible, the extra request only fires when a collection has more members than its first page, so today it is one request for one group. The users list memoizes page fetches per request, so ten users in that group cost one extra call, not ten. Removing it from fetch would save that single request and lose correctness on other surfaces.

PR #5220, opened 2026-09-14
[ASAP-1675] - Split Event Attendance into Interest Group and Additional Teams

diegonvs:
ref: https://asaphub.atlassian.net/browse/ASAP-1675

Round 2 of the design/product refinements to CRN Event Attendance (ASAP-1675). Attendance rows are split into a locked **From \<group\>** section and an editable **Additional teams** one, on the event detail card and in the edit modal. The card drops the "Since last event" tile and the progress wheel for a single percentage scoped to the hosting group, and always exposes Edit.

Behind `NEW_EVENT_PAGE` feature flag, which is `false` in every environment.

## How provenance works

The hosting group's teams come from the existing `GET /interest-groups/:id` and their ids are intersected with the attendance rows. Teams the group contributes with no attendance record yet are seeded as rows and get an entry created on save, which `updateEventDetails` already does for payload items carrying no `id`.

### Client-side only and temporary by design

The API carries no interest-group information with attendance: an entry is `{ team, attended }`, and a team the hosting group contributes has no entry at all until someone saves. Until Contentful carries both, the frontend stands in
for it.

**This code is written to be deleted.** When the backend ticket lands and
attendance arrives with its provenance, three things go:

| Goes away | What replaces it |
| --- | --- |
| `apps/crn-frontend/src/events/attendance-rows.ts` | `event.attendance` rendered as it arrives |
| `useInterestGroupTeams` + the `resolved` flag (`state.ts`) | nothing. No second request per event page |
| the `interestGroupResolved` gate on Edit (`Event.tsx`) | `isTechSupport` alone |


packages/react-components/src/molecules/EventAttendanceMetric.tsx
```diff
@@ -1,118 +1,81 @@
...
-      <div css={[metricContainerStyles, emptyStackStyles]}>
-        <p css={metricLabelStyles}>{props.label}</p>
-        <p css={metricLabelStyles}>{props.message}</p>
-      </div>
-    );
-  }
+const captionStyles = css({
+  margin: 0,
+  fontSize: rem(14),
+  lineHeight: rem(16),
```
gabiayako: I believe we need something like

```suggestion
  lineHeight: `${16 / 14}em`,
```

here because perRem is set to `17` on `packages/react-components/src/pixels.ts` so the lineHeight here becomes ~ 13px. The text feels a little too close together.

Image

packages/react-components/src/organisms/EventAttendance.tsx
```diff
@@ -111,198 +166,218 @@ const useHorizontalOverflow = () => {
...
               <col css={attendanceColStyles} />
             </colgroup>
             <thead>
               <tr>
-                <th css={headerCellStyles} scope="col">
+                <th css={teamColumnHeaderStyles} scope="col">
                   Teams
                 </th>
-                <th css={headerCellStyles} scope="col">
+                <th css={columnHeaderStyles} scope="col">
```
gabiayako, line 353: I think we need to align “Attendance” vertically with the icons
Image

packages/react-components/src/organisms/EditEventAttendanceModal.tsx
```diff
@@ -392,11 +369,155 @@ const placeholderStyles = css({
...
+            css={[
+              attendanceGridStyles,
+              (index < visibleTeams.length - 1 || canExpand) &&
+                rowDividerStyles,
+            ]}
+            role="listitem"
+          >
+            <span css={teamCellStyles}>
+              {teamIcon(team.teamType)}
+              <Link
```
gabiayako, line 477: I was thinking if it wouldn't be nice to add `openInNewTab` here as well since we're editing the teams

gabiayako (approved):
Nice! I really like the controls in the Storybook story! 👏

packages/react-components/src/organisms/EditEventAttendanceModal.tsx
diegonvs: yes, I've made the same changes to the speakers in the other PR after Elena flagged this as well. Thanks, Gabi!

gabiayako:
really nice! 👏

PR #5221, opened 2026-09-15
ASAP-1621 Split Event Detail Page into Two Tabs

gabiayako:
Jira ticket: https://asaphub.atlassian.net/browse/ASAP-1621

---

##### Note

You need to paste the following into the browser console:

```
document.cookie = 'ASAP_NEW_EVENT_PAGE=true;';
location.reload()
```


packages/react-components/src/organisms/RelatedResearchCard.tsx
```diff
@@ -133,102 +133,118 @@ const RelatedResearchCard = <
           ...(displayShowMoreButton ? [{ paddingBottom: 0 }] : []),
         ]}
       >
-        <Headline2 noMargin>{title}</Headline2>
+        <Headline3 noMargin>{title}</Headline3>
```
gabiayako, line 136: all these headings from "Related <something>" component should be H3

Image

AimeurAmin:
Looks really good just one small detail that I noticed for the on-going/up-coming events 

the card looks stuck to the devider:

image

here is how it looks like on dev:

image

AimeurAmin (approved):

PR #5222, opened 2026-09-15
[ASAP-1710] - Fix ci cache eviction

AimeurAmin:
[ASAP-1710](https://asaphub.atlassian.net/browse/ASAP-1710)

## Problem

All three master pipelines on 15 Sep failed in the deploy jobs (crn-sls-deploy, gp2-sls-deploy, auth0) with "Required unplugged package missing" and "ENOENT build/tenant.yaml". The code was fine.

The deploy jobs do not run yarn install. They restore `.yarn/unplugged` and the frontend/transpile build output from actions/cache entries keyed by commit sha, which the build jobs save ~90 minutes earlier in the run. The repo's Actions cache is at 10.9 GB against GitHub's 10 GB per-repo limit, so entries are evicted least-recently-used. This morning's PR burst (renovate security batch plus three feature PRs) cycled the whole cache in ~15 minutes, and the deploy-only entries were gone by the time deploy ran.

Not related to the move to the mjff-eng org: the limit is the same on every plan and the per-sha keys have been there since early 2024. It just needed enough concurrent PR runs to show up. "Re-run failed jobs" cannot recover it, since it restores the same evicted key.

## Fix Details

- `cache-unplugged` is keyed on the hash of yarn.lock, .yarnrc.yml and the CI Dockerfile instead of the sha and environment. One entry now serves every run and every environment (dev and prod unplugged content was identical, saved twice) and is refreshed on each hit. On a miss the action runs `yarn rebuild` itself instead of leaving the job to fail. Because the key no longer depends on the environment, the `environment-name` input is gone, which is why every `with:` block on this action is removed.
- The transpile and frontend build jobs upload their output as run artifacts (`transpile-output-<env>`, `frontend-output-<env>`, 1 day retention). The sls deploy action, auth0, ses and algolia-settings jobs download those instead of restoring caches. Artifacts are per run and never evicted, and a missing one fails the step instead of deploying an empty directory.

CI only, no application code. Verified with prettier and actionlint; this PR's own Branch deploy runs the same deploy action end to end.

### What happened

A pipeline run is a chain of jobs on separate machines. Early jobs install dependencies and build the apps. The deploy jobs run about 90 minutes later on fresh machines, and they need those installed dependencies and built files. To hand them over, the workflow used GitHub's cache: the build job puts the folder in the cache under a label that includes the commit id, and the deploy job asks for that exact label.

GitHub gives each repo 10 GB of cache. When it is full, GitHub deletes the entries nobody has asked for recently. Our repo was full, and this morning six PR pipelines ran at once, each adding about 1.3 GB. So the cache was being emptied and refilled every 15 minutes. Master's entries were written at 09:00 and nobody asked for them until deploy at 10:27. By then they had been deleted. The deploy job asked for the folder, got nothing, and serverless was simply not on disk.

Two things made master especially vulnerable. Its label included the commit id, so no other run ever asked for it and kept it alive. And the gap between build and deploy was long because tests and runner queueing sat in between.

### Why rerunning failed jobs cannot help

Rerunning failed jobs skips the build and goes straight to deploy, which asks for the same deleted label. Only a full rerun rebuilds the cache entry first.

### What the fix does

Two changes, both about not depending on a cache entry that can vanish.

1. The dependencies folder is now cached under a label based on the lockfile instead of the commit. Every run and every PR asks for the same entry, so it is always fresh and stays alive. If it is missing anyway, the job installs the dependencies itself in about 80 seconds instead of failing. The with: environment-name lines went away because the label no longer needs to know which environment it is for. Dev and prod were caching identical content twice.
2. The built files no longer go through the cache at all. The build job uploads them as a run artifact, which is a file attached to that specific run, and the deploy job downloads it. Artifacts are never evicted, so the 90 minute gap does not matter anymore.

### Why it did not happen before

The setup was always fragile. It needed enough PR pipelines running in the same hour to fill 10 GB faster than master could finish. Today was the first time that happened.

### Does this make our pipeline slower

No, if anything it is slightly faster in normal operation.

- Unplugged cache: same restore as before, about 215 MB, but now it hits far more often. Before, the rebuild job spent about 80 seconds running yarn rebuild on every run, because the sha-keyed entry never existed yet. Now that step only runs when the lockfile changes. The one-off cost is the first run after merge, where every job misses and rebuilds in parallel, so that run is about 80 seconds slower per job. After that it is a hit everywhere.
- Artifacts instead of cache for build output: upload and download are comparable in speed to cache save and restore for the same size, around 140 MB for the frontend output and 43 MB for the transpile output. The build jobs now do one extra upload each, a few seconds. The deploy jobs replaced a cache restore with an artifact download, roughly the same time. The existing cache saves are still there for the build-phase jobs that read them, so nothing there changed.

The only genuinely new work is the two uploads in the build jobs. That is seconds, not minutes.


.github/workflows/reusable-build.yml
```diff
@@ -256,6 +257,17 @@ jobs:
...
+      - name: Upload build output
+        uses: actions/upload-artifact@v4
+        with:
+          name: frontend-output-${{ inputs.environment-name }}
+          path: |
+            apps/**/build*
+            apps/**/dist*
+            packages/**/build*
+          include-hidden-files: true
+          retention-days: 1
```
diegonvs: Can we increase the retention-days to 3? Imagine I sent a PR on friday afternoon and it failed, if I rerun again on monday the artifact will be expired and I need to run a full re-run

diegonvs (approved):

.github/workflows/reusable-build.yml
AimeurAmin: Done, bumped to 3 days. I also raised the sls package artifacts from 1 to 3, which also had the same 1-day retention.

